PR #12, opened 2017-01-07
Add a makefile

geordieF:
just running make is so nice.

### DIFF
--- a/MAKEFILE
+++ b/MAKEFILE
@@ -1,0 +1,2 @@
+#!/bin/bash
+./setup.py build
--- a/MAKEFILE
+++ b/MAKEFILE
@@ -1,2 +1,3 @@
-#!/bin/bash
-./setup.py build
+SHELL := /bin/bash 
+all:
+  python3 `./setup.py build


kovidgoyal:
I committed a slightly simpler makefile :)

geordieF:
;)

geordieF:
Actually, if someone (like me) uses fish instead of bash, your doesn't work, but mine does! This is because fish needs context for this kind of stuff.

kovidgoyal:
Really, why does the shell matter? Make does not execute its build commands through a shell.

geordieF:
It seems that fish tries to do makefiles itself, which is faster somehow. This also helps in porting to mac as it expects a global path unless you yo tell it not to.

kovidgoyal:
You saying that invoking make in fish does not call /usr/bin/make? I tried running my Makefile under fish and it worked fine?

geordieF:
That's weird, mine dowsn't work. Must be some obscure setting that I didn't find. I just checked and Yosemite says that it cant fint the file.

nikrich (approved):

PR #22, opened 2017-01-08
fixed error when loading json due to binary mode

MaximilianKindshofer:
When I tested the new feature I got the error:

Failed to load cached value with error: the JSON object must be str not 'bytes'


kovidgoyal:
Thanks, I committed a slightly different fix.

Issue #5, opened 2017-01-07
OS X support


kovidgoyal:
Issue to track the status of porting kitty to OS X. 

Things needed to port:

0) ~~Fix any cross platform build issues when compiling the C code parts of kitty on OS X. Once it is building, setup continuous integration testing for OS X as well~~
1) ~~Replace use of xdpyinfo  to get screen DPI with native OS X API (CGDisplayBounds and CGDisplayScreenSize?)~~
2) ~~Replace use of freetype+fontconfig to rasterize fonts with native OS X APIs (this is low priority since it is possible to setup fontconfig+freetype on OSX as well)~~

Instructions for running kitty on OS X
---------------------------------------------------------

kitty now builds on OS X in Travis and all tests pass. A simple command to install the dependencies needed for kitty using brew:

```sh
brew update && brew install python3 && brew install glfw
```

Then simply follow the instructions in the README for building kitty from source.

toomim:
Here's a record of my OSX build attempt:

    brew install glew                             # install glew
    git clone https://github.com/glfw/glfw.git    # install glfw
    cd glfw
    cmake -DBUILD_SHARED_LIBS=ON .
    make install
    cd ..
    git clone https://github.com/kovidgoyal/kitty # install kitty
    cd kitty
    # Then delete ", '-l' + lib" on line 78 of setup.py... and run:
    python setup.py install

But I get this error:

    ld: symbol(s) not found for architecture x86_64
    clang: error: linker command failed with exit code 1 (use -v to see invocation)

I deleted the `, '-l' + lib` from setup.py because otherwise it errors with: `ld: library not found for -lPython.framework/Versions/3.5/Python`.

kovidgoyal:
I already committed code to handle that issue, see #2

toomim:
Ok, I'm getting further now. It turns out homebrew only has glew version 1.3, so I installed version 2 with:

```
wget https://sourceforge.net/projects/glew/files/glew/2.0.0/glew-2.0.0.tgz/download
mv download glew.tgz
tar xzvf glew.tgz
cd glew-2.0.0/
make
sudo make install
```

And now running `python3 setup.py build` produces this strange error:

```
glew >= 2.0.0 is required, found version: 2.0.0
```

kovidgoyal:
I guess `pkg-config --atleast-version=2 glew` is not working on OS X? I suppose it could be replaced by parsing the output of pkg-config --version manually on os x.

[...]

PR #40, opened 2017-01-18
Issue with MacOS build

cema-sp:
Hi, I had a few issue with MacOS build.
Having `python3` and `glfw` installed via **homebrew** and trying to build **kitty**, I faced the following error: https://gist.github.com/cema-sp/074b5422a0f6ce345f44b3163dad90a4

I played a little with _setup.py_ and found that "framework" python doesn't properly links on build.
And `LINKFORSHARED` has invalid value for brewed **python3**.

Anyone had similar issues? Any suggestions?

### DIFF
--- a/kitty/control-codes.h
+++ b/kitty/control-codes.h
@@ -63,6 +63,8 @@
 #define CSI 0x9b
 #define ST 0x9c
 #define OSC 0x9d
+#define PM 0x9e
+#define APC 0x9f
 
 // Sharp control codes
 // -------------------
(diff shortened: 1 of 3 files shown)


kovidgoyal:
What does

```
python3 -m sysconfig
```

output on your system?

cema-sp:
Here it is: https://gist.github.com/cema-sp/fbdf0271036618956bd2234807ebb01d

kovidgoyal:
According to that the path setup.py is using is correct, `-F/usr/local/Cellar/python3/3.6.0/Frameworks` so I'm not sure why linking is failing. Framework builds are the recommended way to link python on OS X, so I dont think the solution is to turn off frameworks.

kovidgoyal:
Also, many other users have built kitty against python from brew, see #5

Maybe your python install is corrupted, try re-installing it from brew.

cema-sp:
Closed. New pull-request: #41

PR #41, opened 2017-01-20
Link python shared lib

cema-sp:
I had n issue with MacOS build.
Having `python3` and `glfw` installed via **homebrew** and trying to build **kitty**, I faced the following error: https://gist.github.com/cema-sp/074b5422a0f6ce345f44b3163dad90a4

Here is `python3 -m sysconfig` output: https://gist.github.com/cema-sp/fbdf0271036618956bd2234807ebb01d

I found that linking python shared library could help.

If anyone faces the same issue, you could try setting **kitty** up with this fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,9 @@ matrix:
               packages:
                 - libfreetype6-dev
                 - libglew-dev
-                - libxi-dev 
-                - libxrandr-dev 
-                - libxinerama-dev 
+                - libxi-dev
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I committed a slightly different fix that makes use of the sysconfig LDLIBRARY variable, that seem to be correct as per your sysconfig output, try it and see.

kovidgoyal:
kitty now builds against brew python3 on Travis.

PR #43, opened 2017-01-23
Calculate pixel ratio in cursor_pos_callback

wezzynl:
On my Retina MacBook Pro when dragging to make a selection or when trying to click a link, the x and y coordinates were off. This calculates the ratio before calling the callback. Can't test on a non-Retina screen but it should work correctly.


kovidgoyal:
I fixed it slightly differently, to avoid needing to calculate the ratio on every callback

wezzynl:
Ah, there's a `ViewPortSize` object :) Perfect. Thanks!

binor:
Hey @wezzynl, for me this isn't fixed yet. I'm running kitty on a Macbook Pro (2016). Could you please verify?

Issue #141, opened 2017-10-12
Invisible text on macOS 10.11.6


Alok:
Using the latest build from `master`, all the text in the spawned terminal is invisible.

kovidgoyal:
Works for me. Run it with --debug-kitty-gl and see if it reports any errors.

Alok:
No errors get printed out when I run it with that. I notice that pressing `c-d` does not exit the spawned terminal, i have to type `exit` to do that, though control c works, judging by how the cursor jumps. The cursor is the only visible component.

kovidgoyal:
Hmm well without being able to reproduce it, it's hard for me to help. See if this commit makes a difference: https://github.com/kovidgoyal/kitty/commit/3f8b2184a6081c616378e34452c5dba9a47f2620

Alok:
It doesn't, but I'll try to reproduce the issue and report back.

kovidgoyal:
Here is the build I used that does not show this issue: https://download.calibre-ebook.com/betas/kitty-0.3.0.dmg

Alok:
That build displays the issue for me, so I'll have to continue trying to reproduce it.

kovidgoyal:
I should say I have tried it only on macOS 10.13 not 10.11.

kovidgoyal:
Another commit that might help, let me know if it works: https://github.com/kovidgoyal/kitty/commit/e679fa9a1c55c27345ae09524299571a22572b44

Alok:
didn't work

kovidgoyal:
I give up, I've just disabled graphics support on older macOS versions. That should take care of it.

kovidgoyal:
A better fix, that should allow graphics to works as well, though I have no way to test that. https://github.com/kovidgoyal/kitty/commit/66803e6873c9f8df1a75265aedc0670fce4411be

Alok:
Thank you. I can appreciate the headache of trying to fix a problem on software you don't even have.

Really enjoying the commit messages too.

eldentyrell:
Sorry about the slow response, I got dragged into a firefight on another project yesterday.

Simply disabling graphics on outdated OSes is probably the right approach.  Personally I'm trying to move off of them, and Kitty is a major help in doing this.

[...]

PR #145, opened 2017-10-16
fix for #141 as suggested by @kovidgoyal; I can confirm this checkout works on MacOS 10.9.5

eldentyrell:


kovidgoyal:
This is not an actual fix, it would break rendering of graphics. It just allows me to identify the bug in the mac opengl drivers I will have to workaround.

Alok:
This works for me.

PR #147, opened 2017-10-19
new command pipe_selection_to_new_tab (ALSO: enable --detach on macos; seems to work fine AFAICT)

eldentyrell:


eldentyrell:
> Given that kitty is typically launched via launch services (clicking the dock icon) which detach anyway, I dont see the need for this on OS X.

Right: it is needed in the situation where kitty is launched from the command line.

eldentyrell:
Well, my github-fu is weak, I cannot figure out how to create two separate pull requests for two separate groups of commits.  It keeps insisting on updating my existing pull request.

Anyways, there are two features here which ought to be evaluated separately: --detach and the new command pipe_selection_to_new_tab, which lets you bind a key such that when pressed the selection is piped to a config-file-programmable shell command taking the current selection as input on stdin.  Its uses are manifold.

kovidgoyal:
You need to split your commits into separate branches. Then push those
branches to github at which point you can use the web api to create the
PRs.

kovidgoyal:
I added the functionality to run arbitrary commands in new windows/tabs and pass botht he selection and the scrollback buffer ot the new commands. https://github.com/kovidgoyal/kitty/commit/9a1b85081e97e7e3564d1d7c8ced0111d40aa8e0

Issue #167, opened 2017-11-12
Strange mouse behavior in ncmpcpp


maksbotan:
Hi! Sorry for another strange issue from me :)

I use `ncmpcpp` MPD client as my music player. Here on the screenshot you can see progress bar of the current song in the bottom:

![screenshot-12-03](https://user-images.githubusercontent.com/275614/32697445-95b8b13a-c7a1-11e7-8342-0ee8b96f2bd4.png)

`ncmpcpp` allows to click with mouse on this bar in any position to scroll the song to that position -- very handy. But when I try to do this in kitty, I suddenly get to `ncmpcpp`'s search engine, as if I pressed `3` on the keyborad, and of course no scrolling hapens. At the same time, `ncmpcpp` in `xterm` reacts correctly to the click.

How can I debug the issue and provide you additional information? I'm running kitty from master commit 8431eef970afb84d10ca5a29052c08417d780f26 with `glfw` from master as well (to get clipboard without `xsel`).

kovidgoyal:
This will be a mismatch between mouse event  encoding in kitty and ncmcpp. The problem is that there are several existing mouse protocols. I know the oldest and newest work as they are used by htop and vim respectively and I use them myself in kitty. Probably ncmcpp is using one of the middle ones. Simplest way to debug is to first run kitty with --dump-commands and then look for the command that sets the mosue protocol version. Then write a small script that simply sets that protocol version and outputs any bytes received from the terminal so you can see the difference between xterm and kitty. And then finally fix the problem in encode_mouse_event() in mouse.c (or you can skip the --dump-commands part and just insert some printf() in encode_mouse_event() to see which protocol is being used).

maksbotan:
Adding this `printf`:

```c
printf("MOUSE %d\n", screen->modes.mouse_tracking_mode);
```

I see output "`MOUSE 1`". So if I understand the code correctly, `UTF8_PROTOCOL` is used.
By the way, when I click in `ncmpcpp`, i get __two__ "`MOUSE 1`" lines in the output, as if there were two mouse events. And in fact, before it shows search screen I see a fast flash of some other screen which I can't identify.

kovidgoyal:
There are all sorts of mouse events, move/press/release/drag. So the number should not matter. I have never tested the UTF8_PROTCOL, so it's quite possible there is a bug in there.

maksbotan:
Can you give me some directions on how to debug this? Description of protocol for example or how to write that simple script you mentioned.

[...]

PR #169, opened 2017-11-12
Fix URXVT and UTF8 mouse protocols implementations

maksbotan:
Fixes bug #167.

After experimenting I found a bug in UTF8 protocol -- kitty was not sending 'M' character after CSI.
After that I've found that I determined `ncmpcpp`'s protocol wrong and it uses URXVT one. So I checked it too and it turns out you have to add 32 to `Cb` value in that protocol as well. See:

http://cvs.schmorp.de/rxvt-unicode/src/command.C?revision=1.583&view=markup#l1353

Right now `ncmpcpp` works as it should with this patch. However, I don't know any UTF8 protocol program to check the fix with.


maksbotan:
Hm, I have no idea why GitHub linked some of your commit to this PR...

kovidgoyal:
My fault, I pushed a bad commit to master and had to rebase. I already committed the fix for the UTF*_PROTOCOL and I have added the fix for URXVT_PROTOL as well, thanks!

PR #183, opened 2017-11-25
Set default line thickness to level 0

Lokaltog:
This is likely a matter of personal preference, but the box drawing characters appear very thick on HiDPI monitors in kitty compared to other terminal emulators. In e.g. Konsole box drawing characters at normal font sizes are normally rendered as 1px regardless of DPI, this looks very nice and sharp on a HiDPI monitor.

This change makes the box drawing characters appear 1px wide by default. This also makes the contrast between thin and thick lines more noticeable. 

If this is not desired as the default appearance, then it would be great if this could be configurable since kitty overrides the box drawing characters.

Before/after:

2017-11-25-173233_318x358_scrot
2017-11-25-173238_318x358_scrot

### DIFF
--- a/kitty/fonts/box_drawing.py
+++ b/kitty/fonts/box_drawing.py
@@ -9,7 +9,7 @@ from itertools import repeat
 from kitty.utils import get_logical_dpi
 
 
-def thickness(level=1, horizontal=True):
+def thickness(level=0, horizontal=True):
     dpi = get_logical_dpi()[0 if horizontal else 1]
     pts = (0.001, 1, 1.5, 2)[level]
     return int(math.ceil(pts * dpi / 72.0))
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This is definitely a matter of personal preference. For example, on my 192 dpi monitor libvte based terminals and kitty draws 2px lines while konsole draws 1px lines. Personally I prefer 2px lines. The most common use of box drawing characters is to draw separator lines and box frames. 2px lines give a better visual indicator in these cases. However, I have no problem with making it configurable. Probably the best way to do it would be to make the point sizes used by the thickness() function a global variable and have kitty set that at startup based on the config.

Lokaltog:
Great stuff, the new config option works perfectly. Thanks for implementing this feature so quickly.

PR #240, opened 2017-12-19
Updated readme

FabioLolix:
Hello,

Added link for Arch Linux stable package

***

For building kitty in a clean chroot I had to add 'libxkbcommon' and 'libxkbcommon-x11' as dependency, otherwise it don't :)

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -119,7 +119,9 @@ python3 /path/to/kitty/folder
 
 === Linux packages
 
-* Arch linux: AUR git package https://aur.archlinux.org/packages/kitty-git/
+* Arch Linux: AUR git package https://aur.archlinux.org/packages/kitty-git/
+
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Just FYI, as of 0.6 glew, glfw, xsel, xdpyinfo are all no longer dependencies.

kovidgoyal:
And Imagemagick is an optional dependency for kitty icat (viewing images inside kitty)

FabioLolix:
I have started to working on this, I have encountered build fails without glew and need to add some others dependency, and have some questions

For you and anyone else which have issue and suggestion for Arch packaging, issues can be opened here: https://github.com/FabioLolix/PKGBUILD

kovidgoyal:
What build fail? glew is most definitely not used anywhere in kitty, as you can check by grepping for glew.

PR #250, opened 2017-12-23
Add remote functionality to kitty

grimpy:
Make it possible to send commands to a running instance of kitty

Signed-off-by: Jo De Boeck <deboeck.jo@gmail.com>

### DIFF
--- a/kitty/cli.py
+++ b/kitty/cli.py
@@ -47,6 +47,10 @@ Syntax: |_ name=value|. For example: |_ -o font_size=20|
 Run python code in the kitty context
 
 
+--remote -r
+Run command in running instance
+
+
 --directory -d
 default=.
 Change to the specified directory when launching
--- a/kitty/main.py
+++ b/kitty/main.py
@@ -131,13 +131,25 @@ def main():
         from kitty.client import main
         main(args.replay_commands)
         return
-    if args.single_instance:
+    if args.single_instance or args.remote:
         is_first = single_instance(args.instance_group)
         if not is_first:
             import json
-            data = {'cmd': 'new_instance', 'args': tuple(sys.argv), 'startup_id': os.environ.get('DESKTOP_STARTUP_ID')}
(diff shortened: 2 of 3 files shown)


kovidgoyal:
Thanks, I'll review this in detail when I have a little more time (probably after the holidays).

kitty/main.py
```diff
@@ -131,13 +131,25 @@ def main():
         from kitty.client import main
         main(args.replay_commands)
         return
-    if args.single_instance:
+    if args.single_instance or args.remote:
         is_first = single_instance(args.instance_group)
         if not is_first:
```
kovidgoyal: if args.remote and is_first, should raise SystemExit with appropriate error message.

kovidgoyal:
A concern I have is that this currently only works to address the active window in the remote instance. What is you want to send messages to some other window?

kitty/cli.py
```diff
@@ -47,6 +47,10 @@
 Run python code in the kitty context
 
 
+--remote -r
+Run command in running instance
```
kovidgoyal: The help message should be expanded to explain that it works via the --instance-group option that must be specified as well.

kovidgoyal:
Also, there should be a way to send messages to the current kitty instance, even if it does not use --single-instance and over ssh. i.e. if I run kitty send-msg inside a kitty instance, it should work transparently. That can be achieved by using a terminal escape code, probably from the APC family. 

Finally, the IPC code in kitty should probably be modified to allow responses to remote commands, so you can do things like listing windows/tabs/etc.

kovidgoyal:
I am in the process of implementing this, run

```
kitty @ -h
```

in a shell running in kitty to see how to send messages to any window/tab in that kitty process. 

I haven't bothered implementing messaging across different kitty processes, since it is better to just use --single-instance and have all your windows in a single process. But, if needed, it should be easy to add.

PR #271, opened 2018-01-07
split DRAG and MOVE events into correct indicators

### LINKED ISSUE #267: scrolling in Micro editor selects text
While using [Micro editor](https://github.com/zyedidia/micro), scrolling begins selecting text at the mouse cursor, as cursor begins to move, requiring an extra click to stop the selection and to be able to move cursor without selection. This only happens in this terminal.
Kitty:
peek 2018-01-06 15-04
xfce4-terminal:
peek 2018-01-06 15-08

### LINKED ISSUE #270: scrolling in Micro editor selects text: the cause
Since the developer has chosen to close and ignore the last issue (#267), blaming the editor, i'm opening another one to finish this.

Kitty is sending (or detecting) the wrong event type for mouse move (32) whereas other terminals have distinct events for mouse move (35) and mouse-down+move(32). The behaviour is shown here, using Micro's raw event inspector from the [nightly release](https://github.com/zyedidia/micro/releases/tag/nightly):
xfce4-terminal(left) - kitty(right)
peek 2018-01-07 16-23

I am confident this is the cause for the issue as all terminals which have the distinction do not have the bug.

evolbug:
fixes #267, fixes #270

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -19,7 +19,8 @@ typedef enum MouseActions { PRESS, RELEASE, DRAG, MOVE } MouseAction;
 #define SHIFT_INDICATOR  (1 << 2)
 #define ALT_INDICATOR (1 << 3)
 #define CONTROL_INDICATOR (1 << 4)
-#define MOTION_INDICATOR  (1 << 5)
+#define DRAG_INDICATOR  (1 << 5)
+#define MOVE_INDICATOR ((1 << 5) + 3)
 #define EXTRA_BUTTON_INDICATOR (1 << 6)
 
 static inline unsigned int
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I think a simpler fix is to simply remove the if statement on line 52. As far as I can tell from my testing that mimics the behavior of xterm in both SGR and UTF-8 modes with any mode mouse tracking. Let me know if it fixes your issue with micro as well.

evolbug:
removing the if statement does not fix the issue

evolbug:
the issue is the lack of differentiation between DRAG and MOVE, which i've solved.

evolbug:
unless you mean removing only the condition, seems to work

kovidgoyal:
Yes, I mean removing only the if condition, leaving behind the cb  = 3

If you want to update your PR let me know, and I will merge it, or I can just just commit the change myself.

evolbug:
done

PR #290, opened 2018-01-22
Vertical and horizontal layouts

mighty-pirate:
I've added very basic vertical and horizontal layout options

### DIFF
--- a/kitty/layout.py
+++ b/kitty/layout.py
@@ -241,5 +241,60 @@ class Grid(Tall):
         for i in range(ncols - 1):
             self.between_blank_rect(win_col_map[i][0], win_col_map[i + 1][0])
 
+class Vertical(Layout):
+
+    name = 'vertical'
+
+    def do_layout(self, windows, active_window_idx):
+        self.blank_rects = []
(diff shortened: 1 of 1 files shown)


kitty/layout.py
```diff
@@ -241,5 +241,60 @@ def do_layout(self, windows, active_window_idx):
...
+        ystart, ynum = next(ylayout)
+
+        for i in range(window_count):
+	        xstart, xnum = next(xlayout)
+	        windows[i].set_geometry(i, window_geometry(xstart, xnum, ystart, ynum))
+	        if i > 0 and i < window_count - 1:
+	        	# between blank rect
+		        self.between_blank_rect(windows[i - 1], windows[i])
+          	#bottom blank rect
+	        self.bottom_blank_rect(windows[i])
```
kovidgoyal: In a horizontal layout, all windows should have the same height, so you can just use a single blank rect for the bottom blank rect.

mighty-pirate: Ah, ok. I've applied that change in the latest commit.

Issue #190, opened 2017-11-27
Named key PLUS


zartstrom:
I would like to use `ctrl+shift+plus` as shortcut for increasing font size.
But this is obviously not supported:
```
Shortcut: ctrl+shift+plus    increase_font_size has an unknown key, ignoring
```
and `ctrl+shift++` gives an error.
In `kitty/config.py` there is a dict `named_key`. Would it be feasible to add a mapping PLUS -> '+'?

I guess there is a reason to why it's not in there .. if it's document, I'm sorry, I didn't find it yet.

kovidgoyal:
+ is not a key on most keyboards (leaving aside the numeric keyboard). The default shortcut is ctrl+shift+= 

The equal key typically has + as its shifted alternative. 

As for the numeric keyboard support for that requires https://github.com/glfw/glfw/issues/1140

zartstrom:
Thanks for the super fast answer! I see :)

PR #321, opened 2018-02-12
Add `+` to named_keys, #190 related

k1sul1:
I tried to change the hotkeys for changing font size to the ones used by every other application I use

```
# map ctrl+shift+equal    increase_font_size
# map ctrl+shift+minus    decrease_font_size
# map ctrl+shift+backspace restore_font_size
map ctrl+plus increase_font_size
map ctrl+minus decrease_font_size
map ctrl+0 restore_font_size
```

I get `Shortcut: ctrl+plus increase_font_size has an unknown key, ignoring` when starting. 

In issue #190 the following was said: 
> \+ is not a key on most keyboards (leaving aside the numeric keyboard). The default shortcut is ctrl+shift+=

Maybe it's not a key on most keyboards, but there's a "few" million keyboards out there where it is. Like mine. I don't have a dedicated <kbd>=</kbd>, but I do have a <kbd>+</kbd> key. 

Can we define our own named keys in the config so these kinds of PRs would be unnecessary? It wasn't apparent from the README or the 60 lines of source I "read". There are hundreds of keyboard layouts, it's not possible to have the right config for everyone.

### DIFF
--- a/kitty/config.py
+++ b/kitty/config.py
@@ -71,6 +71,7 @@ def parse_mods(parts):
 named_keys = {
     "'": 'APOSTROPHE',
     ',': 'COMMA',
+    '+': 'PLUS',
     '-': 'MINUS',
     '.': 'PERIOD',
     '/': 'SLASH',


kovidgoyal:
Sorry I dont understand what you are trying to say. There is no constant named GLFW_KEY_PLUS. http://www.glfw.org/docs/latest/group__keys.html So adding it to named keys is not going to do anything. What actual key code does your keyboard generate for this dedicated + key? Use the glfw test_events program from https://github.com/glfw/glfw/blob/master/tests/events.c to see that.

k1sul1:
Fair enough. If some underlying thing doesn't support it, then it doesn't. Doesn't make any sense to me why it doesn't but I don't really care either. 

Vim was unusable with this, and in general everything *felt* slower (didn't actually measure), so I went back to Konsole. 

After typing a HTML comment, this was displayed
before redraw

And this is what I wanted to see, which required a manual redraw:
after redraw

This isn't the only terminal that I've had a similar issue with, I just don't have the time or care enough to debug the cause.

kovidgoyal:
Presumably, you are running on a system without GPU acceleration. Since that is the only possible way kitty could be slower than konsole, which is by far the slowest of all terminal emulators. But anyway, given that you have neither the time nor interest to care, I dont have the time or interest to help you further either.

PR #327, opened 2018-02-15
Edge color matches background color

blueForestIcarus:
If only one window is visible, the edge color will be set to the current background color of the active window. If multiple windows are visible, the default background color is used.


kovidgoyal:
I'm traveling gor the enxt few days so I will look at this when I return.

kovidgoyal:
I had a little idle time so I took a look, that approach would cause a big performance regression. So I implemented it differently in the bg branch check it out.

PR #341, opened 2018-02-21
Add option for custom title bar colour on macOS.

splintah:
Hi,

This pull request adds the ability to customise the colour of the title bar on macOS.

**Example**
This:
Default macOS title bar colour.
Becomes this:
Customised colour.

This implementation uses three configuration variables:
- `macos_titlebar_custom_color` (bool)
  Apply a custom background color.
- `macos_titlebar_background` (colour)
  The title bar's background color. Only applied when `macos_titlebar_custom_color` equals `yes`.
- `macos_titlebar_light_foreground` (bool)
  Make the title bar's foreground light (by default it's dark).

I don't know whether this system is ideal, but it seemed the easiest to me.


kovidgoyal:
Hmm,I dont like having three config options for this. Instead I suggest, the following:

Have a single option name macos_custom_titlebar_color which can take a color value or the keyword system. Write a special parser for it that uses the lower 24bits for the color and the upper 8bits as a flag indicating whether the color should be used or not. 

As for the foreground boolean, it should be possible to set that automatically based on whether the  specified color is dark or light (convert to HSL or similar color space color and check the lightness value). Or, if you want to make it configurable, make the syntax of the option something like

```
<color name> <light|dark>
```

Have the parser put the boolean also into the upper eight bits, along with the used flag.

kovidgoyal:
Actually thinking about it, there is no need touse flags and bits at all. Dont go through state.c. Instead just write a python function in cocoa_window.m that sets a global with the parsed value of the option and write another function to apply the global to new windows and call that from glfw.c unconditionally. The function can do nothing when system is applied.

Also, in the name of DRYness, probably nice to have the parser recognize a special keyword (`background`) to set the color to be the same as the default background color.

splintah:
> I dont like having three config options for this.

I agree with that.

> As for the foreground boolean, it should be possible to set that automatically based on whether the specified color is dark or light (...).

I think that's a good idea. Going to take a look at it.

> (...) apply the global to new windows and call that from glfw.c unconditionally.

I don't think that will work, since there wouldn't be a possibility to use the default grey title bar.

> Also, in the name of DRYness, probably nice to have the parser recognize a special keyword (`background`) to set the color to be the same as the default background color.

Using the value of `background` is a good option, but it would give less customisability.

So concluding: I think the best way to keep the configuration simple, is to use one configuration variable, a boolean that configures whether or not the title bar uses the colour defined with `background` as a background. The text foreground should be based on the brightness of the background colour.
What do you think about that?

kovidgoyal:
Here is a commit that implements the option parsing code I was talking about: https://github.com/kovidgoyal/kitty/commit/e236539e3a9437be371b9ab3f0bdf3c0028bb782

Now you only need to implement the code to actually change the titlebar colors in cocoa_window.m

kovidgoyal:
You appear to have accidentally added cocoa_window.d to your commit

kitty/glfw.c
```diff
@@ -9,6 +9,7 @@
 #include "glfw-wrapper.h"
 extern bool cocoa_make_window_resizable(void *w);
 extern void cocoa_create_global_menu(void);
+void cocoa_set_titlebar_color(void *w);
```
kovidgoyal: should add extern here.

splintah:
I've now implemented the function that sets the background and foreground colours.
The foreground is based on the _lightness_ of the background colour, calculated as described [here](https://nl.wikipedia.org/wiki/HSL_(kleurruimte)#Omzetting_van_RGB_naar_HSL) (Dutch page, but summarised: `L = (max + min) / 2`). If it's less than `0.5` the foreground will be light, else dark. (The value `0.5` maybe not ideal)

kovidgoyal:
According to this https://en.wikipedia.org/wiki/HSL_and_HSV#Lightness it might be better to use Luma rather than lightness, but I'm OK either way, it's up to you.

splintah:
Okay, will change it to use Luma. I just picked the first I came across.

kovidgoyal:
Cool, could you also rebase them all into one commit, or if that's too much work I can do it for you.

splintah:
You can do it. I don't think I can do that successfully :).

andreypopp:
Is there an escape sequence available to change this dynamically while kitty is running? I'm working on a colorschema with dark/light variants and want to make it so I can toggle them while running.

splintah:
@andreypopp, I have not implemented escape sequences for the titlebar, so I guess they aren't available. Unless @kovidgoyal has implemented them, but I doubt that.

kovidgoyal:
No there are no escape sequences for this. But there should not be a need for an escape sequence. When set to background it should automatically track changes to the background color. I dont think it does that at the moment. @splintah if you want to add it, you will need to look at two places: 1) patch_colors() in boss.py which is triggered by the kitty @ set-colors --configured
2) default_bg_changed_for() in boss.py which is called when the default background is changed via an escape code or kitty @ set-colors

splintah:
I will take a look.

andreypopp:
That’d be even better! Thank you for working on kitty! It’s been my workhorse for a few months and I’ve never looked back to iterm2/terminal.app/alacritty.

splintah:
@kovidgoyal, how would I get a `NSWindow` from (I guess) the `class Window`? Is `os_window_id` the pointer?
And then, how would I call a C function (I think the `cocoa_set_titlebar_color`) in Python?

kovidgoyal:
Easier to show rather than tell, see https://github.com/kovidgoyal/kitty/commit/9cacfc0c2611bb7c512e6d9710e927af75e30935

I haven't actually tested it, so it may need some fine-tuning.

andreypopp:
This works! Thank you.

PR #372, opened 2018-03-10
Add options to ease linux packaging

oleastre:
I tried to create a fedora package for kitty, respecting the packaging guidelines (hope so)
I had a problem with the lib folder usage, as there are compiled files, it should be named "lib64" for 64bit architectures.
To solve that, I added libdir, bindir and datadir preferences to make it more flexible, and makes the linux-launcher use an absolute path to the libdir.

### DIFF
--- a/linux-launcher.c
+++ b/linux-launcher.c
@@ -94,8 +94,13 @@ int main(int argc, char *argv[]) {
 #ifdef FOR_LAUNCHER
     num = snprintf(lib, PATH_MAX, "%s%s", exe_dir, "/../Frameworks/kitty");
 #else
+#ifdef KITTY_LIB_DIR
+    num = snprintf(lib, PATH_MAX, KITTY_LIB_DIR);
+#else
     num = snprintf(lib, PATH_MAX, "%s%s", exe_dir, "/../lib/kitty");
 #endif
+#endif
+
 #endif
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Really? As far as I can tell from reading the FHS, you only use lib64/lib32 if there are both 32bit and 64bit libraries of a package. Is the situation different on Fedora?

I dont particularly mind adding the options, I'm just surprised it should be necessary.

oleastre:
What the Fedora packaging guidelines asks is that, for 64bit architectures, compiled dependencies are Just adding a --lib64 option could be enough, but I found it more flexible to simply add options for all the base folders (like a classic autoconf/automake project would have done).
The packaging is then simplified: I just need to pass the bindir, datadir and libdir options. 
setup.py creates correct folder structure under the linux-package and I just have to copy that in the package.
That means in my case bindir=/usr/bin, datadir=/usr/share and libdir=/usr/lib64/kitty

kovidgoyal:
I'm OK with adding the options, but, I dont like using absolute paths. Instead of --libdir, --datadir etc use --libdir-name and --datadir-name etc. and combine them with --prefix.

oleastre:
Thanks for this, I will adapt my packaging spec to reflect this change. And I will be able to maintain a copr repository with kitty.

kovidgoyal:
You're welcome :)

PR #379, opened 2018-03-12
Stop requiring git to run "make clean"

jamessan:
People building from the release tarball, e.g. packagers, shouldn't be expected
to be running from a git clone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,11 @@ test:
 	python3 setup.py $(VVAL) test
 
 clean:
-	python3 setup.py $(VVAL) clean
+	rm -f -r build
+	rm -f compile_commands.json
+	rm -f kitty/*.so kittens/unicode_input/*.so
+	rm -f glfw/*-protocol.h glfw/*-protocol.c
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm not really a fan of this. It means one has to now maintain a list of cleanable files in two places. I know I for one will forget to update the Makefile in the future. If you do wish to do it, I suggest make it conditional in setup.py. Look for a .git directory and use it if available, otherwise fallback to cleaning manually.

jamessan:
Thanks for the fix, @kovidgoyal.  This will help me finish the packaging for Debian.

Relatedly, is it necessary to embed glfw?  Are you just pulling intermediate states of upstream development to get fixes you need or are you actually patching the code?

kovidgoyal:
It's patched and will likely be patched even more in the future. https://github.com/glfw/glfw/issues/1140 and https://github.com/glfw/glfw/pull/1141 not to mention that thanks to embedding glfw a single kitty build can dynamically choose whether to use wayland or x11 at run time.

Issue #378, opened 2018-03-11
kitty hangs on exit if a disowned process has been run from it


josswright:
If I run a command in kitty and detach, such as "gvim &!"; or if I run a command directly and subsequently disown it, kitty doesn't close until the child process has also exited.

I found issue #229 relating to this from the specific pass command, and saw the discussion there. This does seem odd behaviour, though, as a general policy. I frequently run commands from the terminal that need to be detached, and keeping the parent terminal around forever doesn't seem desirable. It's certainly unexpected based on every other terminal I've used.

I saw in @kovidgoyal argue in #229 that because STDIN/STDOUT/STDERR are still directed to the terminal then it should remain open, but that does seem both non-standard and awkward for a range of everyday uses.

Could this at least be optional behaviour, if it's seen as desirable for the terminal to hang until all disowned child processes have closed? At the moment I'm having to forcibly quit terminal windows all over the place!

(Thanks for kitty, by the way. It's amazing!)

kovidgoyal:
kitty does not hang, you can close it by pressing ctrl+shift+w

And if you want to use the terminal to launch GUI programs that dont detach properly, create a launch alias and use it as described in #307

josswright:
I see the alternatives, but needing the interactive Ctrl+Shift+W is awkward for various situations.

The other alternatives, suggested in #307, all require you to have planned ahead. Quite often it's not unusual to end up in a situation where you want to Ctrl+Z a process, then bg and disown it.

I completely understand why your preferred behaviour here is to keep the terminal open for STDIN/STDOUT, but given the expected behaviour of other terminal emulators to close in this situation, please would you consider an option? Is there a good reason not to do it?

josswright:
Just as an additional argument here, I think it does make more logical sense for the terminal to close even when disowned processes are still outputting to it. The logic is that the controlling factor should be the shell's jobs list. The entire purpose of `disown` is to remove a job from the shell's jobs list. What's the purpose of `disown` existing if the terminal doesn't let go of the process after that?

I really hope you'll consider adding this option! Given that this is the normal behaviour for terminal emulators, there are a lot of old scripts and bits of code floating around out there that will rely on this behaviour, which isn't necessarily 'right' or 'wrong'. That's going to cause a lot of problems for people running those scripts in kitty.

I can see from your other threads on this that you don't see the point of this. I think I've given a few examples here:
[...]

PR #384, opened 2018-03-12
Adds close_on_child_death option for disowned processes.

josswright:
This adds the close_on_child_death option to kitty's configuration, with the default being the current behaviour. (close_on_child_death no).

You kindly refactored the code to make this easy as a result of issue #378, and pointed me in the right direction to alter it. I know you're not keen on this addition, but as I've got the code written I thought it would be polite to send a pull request. No offence if you don't want to merge; I'll just maintain the fork in sync.

Thanks again for kitty, and for taking the time to help with this.


mgoerlich90:
Just came here to say, that I'm really greatful for this option.
I developed the Habit to end my shell sessions with CTRL+D. 
But kitty doesn't handle this well sometimes, even if I don't work with ```disown``` or background jobs.

Enabling this option fixed it for me.
Thank you @josswright for adding this and thanks @kovidgoyal for merging it in despite actually opposing the idea.
 :+1:

Issue #358, opened 2018-03-04
Extending selection


vsajko:
Xterm an Urxvt supports extending selection with right mouse click. VTE terminals has similar behavior but with shift+click.
Is it possible to achieve similar behavior in Kitty?

kovidgoyal:
I tried it in gnome-terminal and as far as I can see, all it does is let you extend the current selection, not create disjoint selections. Doesn't seem very useful to me, why not just drag with the mouse to extend the selection? Maybe it's useful with touchpads where dragging is hard? 

Should be a simple patch to implement support for it via shift+click, so I'll leave this issue open in case anyone wants to contribute a patch.

vsajko:
It is simple patch to implement with right-click and I have this almost ready it but I don't like solution with shift+click or ctrl+shift+click.
This requires 2 hands and as such and don't see it it as usable.

Xterm/urxvt  will also distinguish what will be selected with right click based on initial click. If it was double-click, right click will select everything between first 'word' and right-click but right-click will behave like double-click. And this is really useful. 

I know that people from xterm and urxvt use that feature with right click and I don't know any VTE user to use shift+click. I might be wrong, I don't know everybody.

Should this be configurable (shift-click vs right-click), what do you think ?

kovidgoyal:
I'm OK with using right click instead of shift+click. But note that there is an additional complication. When a terminal program is running that has grabbed the mouse, then using shift+click will allow you to make selections using kitty rather than passing the click to the terminal program. So if you want to support right click to extend the seelction, you would also need to support shift+right click in this mode.

kovidgoyal:
Oh and there is also rectangle_select_modifiers to worry about.

PR #390, opened 2018-03-14
Extend selection with right click

vsajko:
This ads right mouse click extend selection in similar manner as it works in Xterm.
https://github.com/kovidgoyal/kitty/issues/358
Right mouse click respects type of first type of click (single, double, triple) and extends accordingly.
Previous functionality with drag, ctrl+alt drag, shift+click should work as before.
Please check if this can fit in Kitty.

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -22,6 +22,8 @@ typedef enum MouseActions { PRESS, RELEASE, DRAG, MOVE } MouseAction;
 #define MOTION_INDICATOR  (1 << 5)
 #define EXTRA_BUTTON_INDICATOR (1 << 6)
 
+int last_multi_clicks = 0;
+
 static inline unsigned int
 button_map(int button) {
     switch(button) {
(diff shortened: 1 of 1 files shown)


kitty/mouse.c
```diff
@@ -22,6 +22,8 @@ typedef enum MouseActions { PRESS, RELEASE, DRAG, MOVE } MouseAction;
 #define MOTION_INDICATOR  (1 << 5)
 #define EXTRA_BUTTON_INDICATOR (1 << 6)
 
+int last_multi_clicks = 0;
```
kovidgoyal: This should be static

kitty/mouse.c
```diff
@@ -246,9 +262,11 @@ HANDLER(add_click) {
     q->length++;
     // Now dispatch the multi-click if any
     if (q->length > 2 && N(1).at - N(3).at <= 2 * OPT(click_interval)) {
+        last_multi_clicks = 3;
         multi_click(w, 3);
         q->length = 0;
     } else if (q->length > 1 && N(1).at - N(2).at <= OPT(click_interval)) {
+        last_multi_clicks = 2;
         multi_click(w, 2);
     }
```
kovidgoyal: Set last_multi_clicks to zero here?

kovidgoyal: Never mind, I see you are doing it in handle_button_event()

PR #411, opened 2018-03-21
Mention tic as dependency

jubalh:
setup.py calls `tic` to compile the terminfo source files.


kovidgoyal:
tic is not a dependency for people trying to build and run kitty (the compiled terminfo files are checked into the source code). Only for linux packagers. As such I dont think it belongs in the list of dependencies. You can instead add a note about it to the linux packaging section if you like.

PR #427, opened 2018-03-31
Reduce icon set size to fit iconutil limits

justinnhli:
The existing iconset generation code leads to the following error and the lack of an icon for the OS X app:

> Mar 30 22:58:36 OXY4776M iconutil[63748] <Error>: ImageIO: _CGImageDestinationCreateWithWriter capacity parameter (12) is too large for this file format (max is 10)

`icon_64x64.png` and `icon_64x64@2x.png` is [unnecessary](https://developer.apple.com/library/content/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Optimizing/Optimizing.html). This PR prevents them from being created.


justinnhli:
Note: this may only be a problem on older OS X versions (I'm on 10.10.5), so feel free to reject.

PR #447, opened 2018-04-09
Describe another workaround for terminfo issues over SSH

maximbaz:
Useful if your only needs are opening vim via SSH and you can sacrifice things like True color.


kovidgoyal:
Better than changing the TERM variable for this would be to create a wrapper for ssh that copies over the terminfo file automatically. Something like

Create file ~/bin/myssh
```sh
#!/bin/sh
infocmp xterm-kitty | ssh $1 tic -x -o \~/.terminfo /dev/stdin && ssh $1

```

PR #468, opened 2018-04-19
setup.py: Support specifying CPPFLAGS distinctly from CFLAGS

jamessan:


setup.py
```diff
@@ -348,13 +359,13 @@ def compile_c_extension(kenv, module, incremental, compilation_database, all_key
...
-        cflgs = kenv.cflags[:]
+        cppflgs = kenv.cppflags[:]
         is_special = src in SPECIAL_SOURCES
         if is_special:
             src, defines = SPECIAL_SOURCES[src]
-            cflgs.extend(map(define, defines))
+            cppflgs.extend(map(define, defines))
 
-        cmd = [kenv.cc, '-MMD'] + cflgs
+        cmd = [kenv.cc, '-MMD'] + kenv.cflags + cppflgs
```
kovidgoyal: shouldn;t cppflags come before cflags?

jamessan: I don't know of a difference that it makes either way.  They're currently mixed up.

PR #474, opened 2018-04-20
Correct types for various PyArg_ParseTuple/Py_BuildValue calls

jamessan:
Debian's [mips] build failed one of the font tests:

    FAIL: test_sprite_map (kitty_tests.fonts.Rendering)
      File "/<<PKGBUILDDIR>>/kitty_tests/fonts.py", line 48, in test_sprite_map
        self.ae(test_sprite_position_for(0, 1), (0, 1, 1))
        self = <kitty_tests.fonts.Rendering testMethod=test_sprite_map>
    AssertionError: Tuples differ: (0, 0, 0) != (0, 1, 1)

    First differing element 1:
    0
    1

    - (0, 0, 0)
    ?     ^  ^

    + (0, 1, 1)
    ?     ^  ^

This was due to test_sprite_position_for() using the "I" code to parse
into an unsigned short.  Since mips is big endian, the "wrong" byte was
stored into extra_glyphs.data.

This commit changes all similar data type mismatches that were found
from auditing calls to PyArg_ParseTuple/Py_BuildValue.

[mips]: https://buildd.debian.org/status/fetch.php?pkg=kitty&arch=mips&ver=0.9.0-1&stamp=1524126606&raw=0

### DIFF
--- a/kitty/data-types.c
+++ b/kitty/data-types.c
@@ -194,7 +194,7 @@ PyInit_fast_data_types(void) {
         if (!init_fonts(m)) return NULL;
 
 #define OOF(n) #n, offsetof(Cell, n)
-        if (PyModule_AddObject(m, "CELL", Py_BuildValue("{sI sI sI sI sI sI sI sI sI}",
+        if (PyModule_AddObject(m, "CELL", Py_BuildValue("{sI sI sI sI sI sH sH sH sI}",
                     OOF(ch), OOF(fg), OOF(bg), OOF(decoration_fg), OOF(cc_idx), OOF(sprite_x), OOF(sprite_y), OOF(sprite_z), "size", sizeof(Cell))) != 0) return NULL;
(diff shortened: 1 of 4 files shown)


jamessan:
There was one other call that should be checked.

https://github.com/kovidgoyal/kitty/blob/dd18a1dfd838e3c55ce0d40edf9a246a4fe08a37/kitty/screen.c#L1636

`extend_mode` is an enum with 3 values, but it's being parse as a bool.

kovidgoyal:
I really need to find the time to write a gcc plugin to add compile time format string type checking for PyArg_ParseTuple. There used to be a patch for it floating around, but I cant find it anymore.

kovidgoyal:
Here it is: https://gcc-python-plugin.readthedocs.io/en/latest/cpychecker.html#format-string-checking

PR #489, opened 2018-04-24
Add macos_hide_from_tasks option

cfal:
Adds option for macOS to hide from tasks list. This is useful if you have kitty configured with a global hotkey and don't need to see it when alt tabbing.

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -228,10 +228,9 @@ macos_change_titlebar_color(PyObject *self UNUSED, PyObject *val) {
     Py_RETURN_NONE;
 }
 
-static PyObject*
-macos_set_hide_from_tasks(PyObject *self UNUSED, PyObject *val UNUSED) {
+void
+cocoa_set_hide_from_tasks(void) {
     [NSApp setActivationPolicy:NSApplicationActivationPolicyAccessory];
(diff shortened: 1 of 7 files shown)


kovidgoyal:
Some comments:

1) It needs to default to off
2) What happens if you have more than one kitty window?

kovidgoyal:
As for using a external tool to activate/minimize, doesn't https://www.hammerspoon.org/ work?

kovidgoyal:
You didn't answer (2)? You can open multiple top level windows in kitty (by pressing ctrl+shift+n) Currently your patch will only apply the taskbar skipping to the first window, not subsequent ones.

cfal:
Ah got it. It works correctly as-is because glfw only sets the activation policy once in initializeAppKit(), which is a static method called only when the first window is created:
https://github.com/kovidgoyal/kitty/blob/master/glfw/cocoa_window.m#L1205

cfal:
Done, please take a look. Thanks!

BIKA-C:
When this option is checked, and toggled to fullscreen, the titlebar disappears:

https://user-images.githubusercontent.com/50717946/227764879-2716dde9-62fa-411b-88aa-2b19cabcb887.mp4

PR #504, opened 2018-04-28
Add option to open URLs without modifiers. Fixes #371

### LINKED ISSUE #371: Open link just clicking
I hope this is the correct place for a question.

I know there is a `open_url_modifiers` option, but is possible to disable it? I want to open a link by just clicking on it, without pressing any modifier.

Limero:
This will allow urls to be clicked without holding a modifier, it is enabled by setting "open_url_modifiers" to "none".

The reason for the check for "!screen_has_selection(w->render_data.screen)" is because it would otherwise trigger opening the url when highlighting.

### DIFF
--- a/kitty/config.py
+++ b/kitty/config.py
@@ -60,7 +60,8 @@ def parse_mods(parts, sc):
         try:
             mods |= getattr(defines, 'GLFW_MOD_' + map_mod(m.upper()))
         except AttributeError:
-            log_error('Shortcut: {} has unknown modifier, ignoring'.format(sc))
+            if 'none' not in m.lower():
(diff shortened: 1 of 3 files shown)


kovidgoyal:
This breaks using the mouse in applications that want to handle the mouse themselves, such as vim. Click events will no longer be sent to vim and dragging will create a kitty selection instead of a vim selection. This can be fixed changing the handle_in_kitty condition, however there is a more fundamental problem. Using open_url_modifiers = 0 means that there is no way to both allow an application such as vim to handle the mouse and open URLs in that application with kitty, since now you have two consumers for the same click event. As such I don't think this is a good idea.

PR #591, opened 2018-05-31
ssh kitten: pass through extra params to ssh process

maximbaz:
The idea is to allow `kitty +kitten ssh server -v`, which should pass `-v` to `ssh` process.

### DIFF
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -42,7 +42,7 @@ def main(args):
     server = args[1]
     terminfo = subprocess.check_output(['infocmp']).decode('utf-8')
     sh_script = SHELL_SCRIPT.replace('TERMINFO', terminfo, 1)
-    cmd = ['ssh', '-t', server, sh_script]
+    cmd = ['ssh'] + args[2:] + ['-t', server, sh_script]
     p = subprocess.Popen(cmd)
     raise SystemExit(p.wait())
 


maximbaz:
You took parsing seriously, nicely done 😉 

> Note that currently specifying -N -f or similar will not work (it does not make sense anyway as you only use this kitten if you plan on an interactive ssh session).

Makes sense, but it is a bit annoying when you alias `ssh` to `kitty +kitten ssh`. Of course it's easy to use `\ssh` to bypass the alias, but thanks to muscle memory the workflow is usually (1) try `ssh -NfL ...`, (2) curse that it didn't work, (3) prepend with backslash and try again.

If it is complicated, then nevermind, but if it is possible to detect such options and make the kitten simply call ssh without doing terminfo magic, then it would be convenient.

kovidgoyal:
It should not be hard, a PR is welcome. You just have to be careful about the interactions of the various different options to make sure that you call ssh directly in all the correct situations.

maximbaz:
The difficult part is to know what else needs such special handling besides `-N` and `-f` 🙂 
Can I make a PR for just these two? These are most common to currently cause annoyance, I guess, and handling them is simple - if at least one of them is present, just call `ssh` directly.

kovidgoyal:
I had five minutes so I made the change but I have not tested it.

kovidgoyal:
The use case is passing arguments to the command to run on the server. All arguments found after the first non option are assumed to be arguments for the command to run. I suppose to be strictly correct the parser should only switch to command mode after two consecutive non option arguments

kovidgoyal:
This should take care of the combined options case: https://github.com/kovidgoyal/kitty/commit/d8f5dc958d15d2ff9fd4c721c688b508d082368d

kovidgoyal:
And at least according to the ssh man page, all options are supposed to be specified before the destination

maximbaz:
Now it is perfect, many thanks 👍 

As you probably realized, the trick seems to be indeed about two non-option arguments, except that they don't have to be consequent. In the example of `ssh host -v app -v` the first `-v` is a ssh param, but since `app` is the second non-option argument we've seen, the second `-v` that comes after `app` is already considered to be a server argument. 

Your code does exactly that, so everything is good.

PR #596, opened 2018-06-02
Fix build on FreeBSD 11

ricci:
This set of patches fixes the build on FreeBSD 11: it fixes errors resulting from the ordering of `#include`s, builds the correct (`null`) joystick driver for `glfw`, and deals with a `pthread` function naming issue.

I checked to make sure that these patches don't break the build in Linux (recent Arch)


ricci:
Hmm, I see that I've broken the build in other ways. I will update this PR.

PR #604, opened 2018-06-04
 Add support for sha1 hashes to the hint kitten 

ricci:
This is specifically intended for git commit hashes, so it allows for
hashes as short as 7 characters (git's default length for abbreviated
hashes) up to the full 40 characters


maximbaz:
Maybe name it more generically "hash", and allow length up to 128 (for `sha512`)?

ricci:
@maximbaz I'd be up for that. I was just trying to scratch one specific itch, but I agree that it will be trivial to support other use cases

PR #618, opened 2018-06-08
Super-hacky avoidance of spurious selection

dcolascione:
If we've previously reset the selection to EMPTY_SELECTION (say,
because we were marked dirty), the selection ends up being 0/0, so if
we click some random place, we suddenly get a non-empty selection
region and highlight random stuff even though all the user did was
left-click once somewhere.

This change ignores these unwanted region updates while (seemingly?) not breaking any actual selection features.

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -146,7 +146,10 @@ static inline void
 update_drag(bool from_button, Window *w, bool is_release, int modifiers) {
     Screen *screen = w->render_data.screen;
     if (from_button) {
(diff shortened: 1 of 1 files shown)


maximbaz:
So that's what happens! I also experienced random selections, but they happen very rarely and I could never realize what I did before they occur.

kovidgoyal:
This will break right click to extend selection. Rather than putting the check into screen_update_selection() it should go into the appropriate call site in mouse.c instead.

dcolascione:
Should be okay now. Tested right-click selection extension (which I didn't know existed) --- works fine.

PR #625, opened 2018-06-10
Autodpi x11

dcolascione:
Make kitty react to system DPI changes under X11


kovidgoyal:
Looks reasonable to me, but could you make a separate PR against my glfw fork for the changes to glfw. The glfw files in kitty are auto-generated by glfw.py from the glfw sources, so making patches against glfw in kitty will not work.

You could ideally send the PR to glfw upstream as well, though they are pretty dead at the moment.

kovidgoyal:
A new one, please.

PR #644, opened 2018-06-16
docs: improve generation of default config files

blueyed:
- config files are placed into `docs/generated/conf/`, using the
  expected name, and without any indentation
- `literalinclude::` is used instead of `include::`
- uses `:download:` instead of `:file:` to allow for downloading the
  generated files easily

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -500,35 +500,30 @@ def write_conf_docs(app, all_kitten_names):
     sc_role.warn_dangling = True
     sc_role.process_link = process_shortcut_link
 
-    def generate(all_options, name='kitty'):
+    def generate_default_config(all_options, name):
         from kitty.conf.definition import as_conf_file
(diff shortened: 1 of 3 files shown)


blueyed:
The main motivation here is to allow for easy diffing against your current config.
With this you can do so after `make docs`, using `docs/generated/conf/kitty.conf`.

It might be worth having a Makefile target to only generate the config files (since Sphinx might not be required for this after all), but this PR is a good step anyway IMHO.

kovidgoyal:
You can see what options you have that are different from the default config by running

kitty --debug-config

No need to diff against kitty.conf

You can also have kitty generate the default conf file for you by temporarily renaming your kitty.conf file and pressing the shortcut for editing the conf file in kitty (which IIRC is ctrl+shift+f2)

Might still be worth having a separate downloadable conf file, not sure.

PR #645, opened 2018-06-16
docs: as_conf_file: no colon, desc before

blueyed:
- do not use `#:`, but only `#` for comments (the former is required for
  Sphinx to pick it up, but this is a literal file).
- handle_option: render `long_text` above the option(s), not below.
  This is more common for configuration files, and looks like it might
  have been done to please Sphinx (where this is required for Python
  documentation strings).

### DIFF
--- a/kitty/conf/definition.py
+++ b/kitty/conf/definition.py
@@ -145,7 +145,7 @@ def wrapped_block(lines):
     if wrapper is None:
         import textwrap
         wrapper = wrapped_block.wrapper = textwrap.TextWrapper(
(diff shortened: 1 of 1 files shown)


kovidgoyal:
The reason for #: is to distinguish documentation/examples from commented out settings. It doesn't have anything to do with sphinx. 

The description/examples are below the settings, because that's the style I prefer. It puts the actual setting front and center. Otherwise, you start out by reading some long sentence that can take a while to actually tell you what the setting is. Which is fine if you want to understand the setting, but annoying if you are, for instance looking for the setting value.

blueyed:
Fair enough, thanks for your explanation.

PR #656, opened 2018-06-18
WIP/RFC: output feedback with font_family substitution

blueyed:
When using "font_family = UbuntuMono" (without space in the name) kitty
silently falls back to "Liberation Mono".

This changes it to output an error instead, and also outputs info for
when "font_family = auto" is used.

TODO:

- [ ] align/refactor with ./kitty/fonts/core_text.py


blueyed:
Ok. How should it be logged then?  There is nothing like log_info, is there?

PR #657, opened 2018-06-18
Pre-compile static patterns using `re.compile` once

blueyed:
Not sure where to put the variables though - should they be at the top of the module, above the function/class, or even within the class(es)?

### DIFF
--- a/kittens/diff/search.py
+++ b/kittens/diff/search.py
@@ -8,6 +8,8 @@ from kitty.fast_data_types import wcswidth
 
 from ..tui.operations import styled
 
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Compiling regexps before hand is a tradeoff. it means you are gaining some run-time performance at the cost of startup performance. And given that the regexp module caches the last 100 compiled regexps anyway, it is rarely worth the tradeoff. So unless you have some evidence (as in performance data) that a particular regexp is causing performance issues, I suggest leaving them alone.

PR #660, opened 2018-06-19
Support Mac OS X 10.10

justinnhli:
The current master uses two features not available for OSX <10.12:

* `os/log.h` (in `kitty/logging.c`)
* `NSEventModifierFlagNumericPad`, which replaced the deprecated `NSNumericPadKeyMask`

Kudos to @metayan (https://github.com/metayan/kitty/tree/osx11) for identifying the cleanest changes to continue supporting OSX 10.10.


kovidgoyal:
I'm afraid this is not something I am interested in supporting. The proposed patches simply allow kitty to be built from source on 10.10, they dont allow a built kitty to run on 10.10. Not to mention that officially trying to support 10.10 means I can never use any new APIs Apple might release in the future. And I then have to assume the burden of making sure kitty works on older macOS releases.

Sadly Apple does not believe in backwards compatibility, kitty is simply following the consequences of that policy. My goal with kitty compat on macOS is to stay compatible with approx 2-3 releases behind the current macOS release,

justinnhli:
The changes run fine on my Air (10.10.5). But not supporting older versions make sense - I will just maintain this set of patches on my fork.

Thanks!

PR #772, opened 2018-07-30
osx: focus for keyboard input follows mouse

### LINKED ISSUE #754: feature request: focus_follows_mouse doesn't work across OS windows
Thanks for the great work on Kitty!

It seems like `focus_follows_mouse` works across Kitty split windows within a single OS window, but not across multiple Kitty OS windows. It would be really great if that would work as well, as coming from iTerm 2 I'm used to that behavior.

Any chance that's possible to implement?

jordanlewis:
This patch adds (unconditional on preferences) focus follows mouse for
kitty windows on macOS. It doesn't bring moused-over windows to the
foreground - it only changes where keypresses are sent.

Closes #754.

Here's a GFYCat showing the behavior: https://gfycat.com/gifs/detail/RemarkableEachFlea

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -695,6 +695,8 @@ static GLFWapplicationshouldhandlereopenfun handle_reopen_callback = NULL;
     if (window->cursorMode == GLFW_CURSOR_HIDDEN)
         hideCursor(window);
 
+    [[self window] makeKeyWindow];
+
     _glfwInputCursorEnter(window, GLFW_TRUE);
 }
 
@@ -743,6 +745,7 @@ static GLFWapplicationshouldhandlereopenfun handle_reopen_callback = NULL;
                                           NSTrackingEnabledDuringMouseDrag |
                                           NSTrackingCursorUpdate |
                                           NSTrackingInVisibleRect |
+                                          NSTrackingActiveAlways |
                                           NSTrackingAssumeInside;
 
     trackingArea = [[NSTrackingArea alloc] initWithRect:[self bounds]


kovidgoyal:
I'd ideally like to minimize the changes made to glfw so as to deviate from upstream as little as possible. So I think it would be better to only add NSTrackingActiveAlways in glfw. Then in mouse.c in kitty in handle_move_event under OPT(focus_follows_mouse) something like this:

```c
if(!global_state.callback_os_window->is_focused) focus_os_window(global_state.callback_os_window);
```

that might also raise the window as it calls glfwFocusWindow(). If so you can add a function in kitty/cocoa_window.m to only change focus without raising the window and use that instead.

kovidgoyal:
Also if you are specify NSTrackingActiveAlways, you should remove NSTrackingActiveInKeyWindow. According to the docs only one can be specified. https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/TrackingAreaObjects/TrackingAreaObjects.html

jordanlewis:
Thanks @kovidgoyal!

benoit-pierre:
This patch result in the window being raised on focus on Linux.

kovidgoyal:
On X11 whether a window is raised or not is not really in kitty's control, it depends on the window manager. IIRC glfw sends the window manager the _NET_ACTIVE_WINDOW message, whether the window manager chooses to raise the window or not is then a matter of policy/config for the window manager. Sadly linux has no well defined API for managing window focus/raise. There is XRaiseWindow and XSetInputFocus which bypass the window manager, and NET_ACTIVE_WINDOW which does not neither of which are exactly what is needed.

benoit-pierre:
Yes! Focus management is the job of the window manager.

PR #829, opened 2018-08-26
WIP Completion for fish

scorphus:
**WIP – Please do not merge yet**

I'm submitting this PR in its early stage to get some preliminar feedbacks and reviews, which are more than welcome.

- [x] Completion function and command line parser reviewed
- [x] Documentation added
- [ ] Tested by another Fish user

### DIFF
--- a/kitty/complete.py
+++ b/kitty/complete.py
@@ -57,6 +57,14 @@ kitty_completions() {
 
 complete -o nospace -F kitty_completions kitty
 ''',
+    'fish': '''
+function __kitty_completions
+    # Send all words up to the one before the cursor
+    commandline -cop | kitty +complete fish
+end
+
+complete -f -c kitty -a "(__kitty_completions)"
+''',
 }
 
 
@@ -84,6 +92,11 @@ def bash_input_parser(data):
     new_word = data.endswith('\n\n')
     words = data.rstrip().splitlines()
     return words, new_word
+
+
+@input_parser
+def fish_input_parser(data):
+    return data.rstrip().splitlines(), True
 
 
 @output_serializer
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I'm not familiar with fish so I cannot comment on that part of it. In general the way completion is supposed to work is that the shell should send a list of words each separated from the next by a newline upto the word the cursor is currently on. If the cursor is not on a word (i.e. it is after a psace) there should be two newlines at the end. You can change this scheme as you like by implementing an appropriate input_parser in complete.py for your shell.

kitty then responds by printing out a shell script on stdout that is evaled by the shell to set the completions.

After you have implemented this scheme you should also add some instructions to docs/index.rst about how to enable the completions in fish, just as there are for bash and zsh.

scorphus:
Great! Looks like I'm on track.

Fish's completion works quite differently from Bash and Zsh. I've integrated many other tools to Fish in the past so I'll hopefully end up with a decent, working completion.

Thanks for your initial review, @kovidgoyal 👍 I'll finish this ASAP (on my free time, of course).

skorokithakis:
As a fish user, I would love it if this got merged.

scorphus:
Me too, @skorokithakis! I think I'll have some free time tonight to advance on this! Stay tuned 😃

scorphus:
@kovidgoyal, Please review again. Would you like a test and feedback from another Fish user?

BTW, @skorokithakis, did you have the chance to test this?

Thanks!

kitty/complete.py
```diff
@@ -86,6 +94,11 @@ def bash_input_parser(data):
     return words, new_word
 
 
+@input_parser
+def fish_input_parser(data):
+    return data.rstrip().splitlines(), True
```
kovidgoyal: This does not look right to me. new_word should be True only if the cursor is after a space.

For example:

kitty +kitt<cursor>

should have new_word False, while

kitty +kitten <cursor>

should have new_word True

scorphus: Thing is that completions work quite differently in Fish than in other command line shells. As Fish interprets and filters the completions suggestions, there's no need to work on completing the current partial token. Only if we're doing something really tricky with the partial token. In that sense, `kitty +complete fish` will always receive complete tokens.

kovidgoyal:
OK as I said I dont know anything about fish -- so that was just a comment based on the other shells. In any case I have merged, since even if it does not work perfectly it is better than nothing.

scorphus:
Great! I'll ask my dear friend @wpjunior to test it and post his feedback!

No problem :-) I just wanted to explain that. I should have done this in the commit message – will do so the next time ;-)

PR #839, opened 2018-08-28
Minor code refactoring

Luflosi:
Is there any reason for calculating the absolute value and then correcting the sign again?


kovidgoyal:
It's because I can never rememebr what C does with division when the numerator is negative and the divisor is not an exact multiple is -3.5 /3 -1 or 0? i.e. how does the division operator truncate the result in case of negative numbers?

PR #840, opened 2018-08-28
Silence tic warnings during compilation

Luflosi:
This pull request redirects stderr of tic to `/dev/null` to ignore the annoying `"terminfo/kitty.terminfo", line 1, col 22, terminal 'xterm-kitty': older tic versions may treat the description field as an alias` warning.
According to https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL, `subprocess.DEVNULL` is only available since python 3.3. Does kitty need to support older python versions? If so, `/dev/null` can be manually opened as a file.
Could there be other errors or warnings that we might miss if we ignore stderr or is the check in the next line for the existence of the file enough?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -598,7 +598,7 @@ def package(args, for_bundle=False, sh_launcher=False):
     for x in (libdir, os.path.join(ddir, 'share')):
         odir = os.path.join(x, 'terminfo')
         safe_makedirs(odir)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
kitty needs to support python >= 3.5

However I deliberately did not do this because I dont want to miss any actually useful warnings from tic incase I make a mistake in the terminfo file.

PR #841, opened 2018-08-29
dont -> don't

Luflosi:
I found every word dont and replaced it with don't. Is there any reason for leaving the apostrophe out?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -359,7 +359,7 @@ Changelog
 
 - When triple-clicking select all wrapped lines belonging to a single logical line.
 
-- hints kitten: Detect bracketed URLs and dont include the closing bracket in the URL.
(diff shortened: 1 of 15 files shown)


kovidgoyal:
The apostrophe is a bit of an anachronism that I personally find useless. The convention of using apostrophes to denote contractions may have made sense when the use of contractions was new -- it doesnt really anymore. That said I dont have any strong feelings about it.

kovidgoyal:
Yeah they are, they were made because of compiler warnings. And are a
good idea in any case. If you care about matching upstream glfw, you
should see about getting them into upstream.

Luflosi:
Thanks.
I'm linking to the place where this comment should have been for future reference: https://github.com/kovidgoyal/kitty/commit/841c907efc643a4f4f06b2686d0d1e6022710600#r34453289

kovidgoyal:
Catches more errors at compile time rather than runtime.

Issue #837, opened 2018-08-28
Scrolling speed is capped in programs like vim and tmux


Luflosi:
I noticed that I couldn't scroll as fast as I wanted in tmux and vim, less however does not seem to be affected. After digging a bit in the code, I noticed, that kitty only ever sends at most one escape code per call to `scroll_event()` to the child. When `s` (the number of lines to be scrolled) is bigger than 1 (or smaller than -1), kitty ignores the value and just sends one escape code to the child. To improve this I tried to simply put the call to `write_escape_code_to_child()` in a while loop like this:
```diff
@ mouse.c:568 @ scroll_event(double UNUSED xoffset, double yoffset, int flags) {
     } else {
         if (screen->modes.mouse_tracking_mode) {
             int sz = encode_mouse_event(w, upwards ? GLFW_MOUSE_BUTTON_4 : GLFW_MOUSE_BUTTON_5, PRESS, 0);
-            if (sz > 0) { mouse_event_buf[sz] = 0; write_escape_code_to_child(screen, CSI, mouse_event_buf); }
+            if (sz > 0) {
+                mouse_event_buf[sz] = 0;
+                while (s != 0) {
+                    write_escape_code_to_child(screen, CSI, mouse_event_buf);
+                    if (s > 0) {
+                        s--;
+                    } else {
+                        s++;
+                    }
+                }
+            }
         } else {
             fake_scroll(abs(s), upwards);
         }
```
This worked for the most part, except when scrolling very fast, broken escape codes seem to get sent to the child. In tmux, when scrolling down while already at the bottom of the scroll buffer, characters like this appear in the shell inside tmux:
`65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M65;62;19M`

[...]

PR #843, opened 2018-08-29
Allow for faster scrolling in alternate screen mode

Luflosi:
Thanks to 0e248b3faa1551555320a7522e8297229ad72788, I can now make #837 a pull request.


SolitudeSF:
after this change kakoune scrolls 3 times faster than `wheel_scroll_multiplier` option value. i dont know if thats was intended speed all along but it feels absolutely wrong.

PR #878, opened 2018-09-06
Correct spelling mistakes.

EdwardBetts:


kovidgoyal:
The files in the glfw sub-directory actually come from the glfw project so you should send the fixes to them and they will get merged into kitty automatically. I'm happy to merge the rest of the fixes if you create a separate PR for them.

PR #913, opened 2018-09-07
Replace "control" in the docs by "ctrl"

Luflosi:
Shouldn't `control` here be `ctrl` or did you deliberately write `control`?


kovidgoyal:
It's deliberate since onmac keyboards the ctrl key is labelled as control and here we are tlaking explicitly about macs

Luflosi:
At least on my Mac the key is labelled ctrl but apparently it's labelled control on newer Keyboards, which I didn't realise before, sorry

Issue #106, opened 2017-08-04
Every version of Terminus I try..  Will not work...


Anonrate:
So I have been going at this for hours now.  I'm sure you all have heard of the font **Terminus**.  It's the only font I will use in the Terminal.  I just discovered **kitty** about 12 hours ago, and in that 12 hours all I've been trying to do is get the font **Terminus** to work.

#### Versions that I have used are as follows:
* Font package from [Debian Stretch][1] **xfonts-terminus**
* I've manually built the [original][2] font (Both **PSF** and **PCF**)
* I've tried the [Powerline][3] version of the font in the format of **PCF**
* Lastly..  I tried the very *shitty* looking **TTF** format from [ax86][4] <sup>[\[1\]](#error)</sup>

I will also put a code version of this arrow so you can navigate to the locations with ease.
![image](https://user-images.githubusercontent.com/11949153/28974525-b3623daa-7904-11e7-8ccf-3f77922098ce.png)

[1]:https://packages.debian.org/stretch/xfonts-terminus (Package Details for xfonts-terminus)
[2]:http://terminus-font.sourceforge.net/ (Terminus Font Home Page)
[3]:https://github.com/powerline/fonts/tree/master/Terminus (Terminus Powerline)
[4]:https://files.ax86.net/terminus-ttf/ (Terminus \(TTF\))

Every single one failed to show any signs of being present.

#### Trying to resolve the issue myself, I have also tried:
* I have got all the required packages to build **Kitty** and have successfully built it without error.
* Other fonts will work, and if it wasn't for the **TTF** format of **Terminus** not working as well...  I would have guessed that it is only bitmap fonts that wont work.
* I've rebuilt that font cache via `fc-cache -fv`
* Tried with a fully configured `kitty.conf`
* Tried with and without haveing `bold_font`, `italic_font`, `bold_italic_font` with still no avail.
* Tried using the full qualified name of the font, rather than just the *alias*.
* Increasing the size of the font
* Specified full font family(*?*) name (with Regular, Medium, etc..) with and without the same casing
* I've enabled bitmap fonts by removing **70-no-bitmaps.conf** from the font configure directy.
* I've tried changing the `TERM` variable to `xterm-256color`
* I've tried with and without system wcwidth
* I've ran **Kitty** from a different directoy

<a name="error">\[1\]</a>: Got this error while running `python3 .src/kitty` with only `font_family   Terminus (TTF)` in the config file.
<pre>
Traceback (most recent call last):
[...]

PR #934, opened 2018-09-08
Add support for embedded bitmap glyphs

isaz:
Tested with [Terminus (TTF) ](https://files.ax86.net/terminus-ttf/) using the recommended font sizes, seems to fix issues mentioned in #106.
This _doesn't_ add support for bitmap fonts.

### DIFF
--- a/kitty/freetype.c
+++ b/kitty/freetype.c
@@ -236,7 +236,7 @@ load_glyph(Face *self, int glyph_index, int load_type) {
     // Embedded bitmap glyph?
     if(self->face->glyph->bitmap.pixel_mode == FT_PIXEL_MODE_MONO) {
         FT_Bitmap bitmap;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I have merged. However, I have once concern. When iterating over the pixels in the bitmap should you be using bitmap->pitch? See for example the downsample_function which uses a stride (calculated from the bitmap pitch)

isaz:
Yes, however, according to the [docs](https://www.freetype.org/freetype2/docs/reference/ft2-bitmap_handling.html#FT_Bitmap_Convert), the `pitch` of the bitmap is a multiple of `alignment` (in our case it's 1), making it equal to the bitmap `width`.
Thanks.

kovidgoyal:
OK, just to be explicit I have changed it to use pitch, since that is its stated purpose.

Issue #595, opened 2018-06-01
Testers needed for Linux binaries


kovidgoyal:
I have made new binary builds of kitty available for Linux (with bundled dependencies), along with a one-liner to install them. See https://sw.kovidgoyal.net/kitty/binary.html for details. This way people wanting to be up-to-date with the latest release version of kitty dont need to build from source.

It would be nice if we could get some testing on different Linux distributions.

P.S. The one-liner works on macOS also, where it simply downloads the dmg and installs it into /Applications auto-magically.

reed-jones:
Ubuntu 18.04, the binary install went great. I made a symlink from `~/.local/kitty.app/bin/kitty` to `~/.local/bin/kitty`, and copied `~/.local/kitty.app/share/applications/kitty.desktop` to `~/.local/share/application/kitty.desktop`, and had to change the Icon in the desktop file to its full absolute path, but then everything worked great, as expected!

kovidgoyal:
Good to know :)

kovidgoyal:
This is a bug in python versions before 3.6, should be an easy workaround. Basically versions of python between 3 and 3.5 only used a version of json.loads that could not decode bytes.

kovidgoyal:
https://github.com/kovidgoyal/kitty/commit/2aba3e249c6d45270f3278a0430082e68d9e01f8

brontosaurusrex:
That did the trick

    curl -L https://sw.kovidgoyal.net/kitty/installer.sh | sh /dev/stdin
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed
    100  8249  100  8249    0     0  28705      0 --:--:-- --:--:-- --:--:-- 28742
    Using python executable: /usr/bin/python3
    Checking for latest release on GitHub...
    Will download and install kitty-0.11.1-x86_64.txz
    Downloaded 8520320 bytes 
    Extracting tarball...
    kitty successfully installed to /home/ticho/.local/kitty.app
    Use /home/ticho/.local/kitty.app/bin/kitty to run kitty
    Traceback (most recent call last):
      File "python3.6/runpy.py", line 280, in run_path
      File "python3.6/runpy.py", line 85, in _run_code
      File "kitty/__main__.py", line 96, in <module>
      File "kitty/__main__.py", line 89, in main
      File "kitty/main.py", line 10, in <module>
      File "kitty/borders.py", line 8, in <module>
[...]

PR #938, opened 2018-09-09
Add Ubuntu with Gnome Shell in binary.rst

tanrax:


kovidgoyal:
The desktop files and icons are already there in ~/.local/kitty.app you just need to symlink them into appropriate places on your system, see for instance the first comment at #595 (and according to that you may need to edit the desktop file to use an absolute path for the icon -- which can probably be moved into the installer).

tanrax:
Done!

PR #949, opened 2018-09-10
Mouse selection based on edges

### LINKED ISSUE #945: Mouse selection should be based on horizontal center of characters
I'm using Kitty 0.12.1 on OS X 10.12.6.

When I use the mouse to select text moving from left to right, the right-most character appears to get selected as soon as the mouse X position is beyond the left edge of the character. It would be better if it gets selected only when the mouse is beyond the horizontal center of the character. The same goes for the other end of the selection

For what it's worth, that is how Terminal.app works.

Here is some gratuitous ASCII art:

```
 __        __
 \ \      / /
  \ \ /\ / / 
   \ V  V /  
    \_/\_/             
 ^    ^
 |    |
 |    +--------- It would be nice if W only got selected when mouse crosses this point
 |
 +------------ W currently gets selected as soon as mouse crosses
               this point (coming from left)

```

Let me know if this doesn't make sense and I'll make a screencast.

jscheid:
Base mouse selection on the edges between cells rather than the cells themselves.

For finding the start and end line in regular selection, the vertical cell index is used rather than the edge.  However, the vertical edge is used in rectangular selection mode.

Now allows selection of a single character.

This changes the semantics of `screen_start_selection` and `screen_update_selection` to accept an edge index instead of cell index for x, and an edge or cell index for y depending on whether rectangular selection is enabled.

Closes #945

Note: "extend word" mode is wonky when extending forward (to the right) but it was wonky before already so I've left it alone for now.

Also note that rectangular selection mode would feel better with a crosshair cursor. One for a future PR.

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -130,11 +130,11 @@ distance_to_window(Window *w, OSWindow *os_window) {
 static bool clamp_to_window = false;
 
 static inline bool
-cell_for_pos(Window *w, unsigned int *x, unsigned int *y, OSWindow *os_window) {
+cell_for_pos(Window *w, unsigned int *x, unsigned int *y, unsigned int *edge_x, unsigned int *edge_y, OSWindow *os_window) {
     WindowGeometry *g = &w->geometry;
     Screen *screen = w->render_data.screen;
     if (!screen) return false;
-    unsigned int qx = 0, qy = 0;
+    double qx = 0, qy = 0;
     double mouse_x = global_state.callback_os_window->mouse_x;
     double mouse_y = global_state.callback_os_window->mouse_y;
     double left = window_left(w, os_window), top = window_top(w, os_window), right = window_right(w, os_window), bottom = window_bottom(w, os_window);
@@ -143,12 +143,15 @@ cell_for_pos(Window *w, unsigned int *x, unsigned int *y, OSWindow *os_window) {
         mouse_y = MIN(MAX(mouse_y, top), bottom);
     }
     if (mouse_x < left || mouse_y < top || mouse_x > right || mouse_y > bottom) return false;
-    if (mouse_x >= g->right) qx = screen->columns - 1;
-    else if (mouse_x >= g->left) qx = (unsigned int)((double)(mouse_x - g->left) / os_window->fonts_data->cell_width);
-    if (mouse_y >= g->bottom) qy = screen->lines - 1;
-    else if (mouse_y >= g->top) qy = (unsigned int)((double)(mouse_y - g->top) / os_window->fonts_data->cell_height);
+    if (mouse_x >= g->right) qx = screen->columns - 0.5;
+    else if (mouse_x >= g->left) qx = ((double)(mouse_x - g->left) / os_window->fonts_data->cell_width);
+    if (mouse_y >= g->bottom) qy = screen->lines - 0.5;
+    else if (mouse_y >= g->top) qy = ((double)(mouse_y - g->top) / os_window->fonts_data->cell_height);
     if (qx < screen->columns && qy < screen->lines) {
-        *x = qx; *y = qy;
+        *x = (unsigned int)qx;
+        *y = (unsigned int)qy;
+        *edge_x = (unsigned int)(qx + 0.5);
+        *edge_y = (unsigned int)(qy + 0.5);
         return true;
     }
     return false;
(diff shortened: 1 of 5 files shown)


jscheid:
For testing this I recommend a large font size, something like `font_size 48.0`.

Luflosi:
I tested this and the selection behaves as you described it 🎉, which I prefer over the current behaviour of the selection.

Luflosi:
Actually, I have a problem, where the character in the top left or the bottom right corner gets selected, for no apparent reason.

jscheid:
I could reproduce the top-left corner bug and I've pushed a fix for it. Any idea how I can reproduce the bottom-right corner bug?

Luflosi:
I can reproduce the bottom-right corner bug by opening tmux with an empty .tmux.conf, selecting a bit of text and then clicking anywhere.

jscheid:
Thanks, it should work better now. Can you give it another try?

maximbaz:
Love this PR, it makes a lot of sense

Luflosi:
There are still some issues. When I start selecting text and move my mouse to the top to expand the selection beyond the top by scrolling it into view, the new text is not selected and the one end of the selection where the mouse cursor is supposed to be becomes offset. When I then scroll the selection out of view, the character in the bottom-right corner becomes selected. Here is a youtube video demonstrating this behaviour: https://youtu.be/ZFhTEz5MP1I
Any suggestions how I could share something like this in the future? The compression makes it look pretty bad and I can't use asciinema in this case since it doesn't record the selection.

Luflosi:
@jscheid It would be really nice if you could find the bug since I very much prefer the selection behaviour of this patch over the current behaviour.

martinetd:
I've had a look at this, the most visible part of the bug is fixable (add an `anchor_scrolled_by`, and set start/end `scrolled_by` from `self->scrolled_by` and `self->anchor_scrolled_by` appropriately -- with a trick point that you also need to compare the `y`s taking `scrolled_by`s into account... which isn't trivial, because we need to subtract them, and index_type is unsigned so just subtracting in the check will overflow -- I'll leave this as an exercise to the reader :P)... but that just uncovers more bugs with `EXTEND_WORD` due to recomputing `screen_selection_range_for_word` for both edges all the time and it's a bit harder to make it handle scrolled_by at all (not impossible but I didn't try)

My pending work actually ha(s|d) the same problem with scrolled_by in the leftwards check, it's kind of fun because I also had added a very close logic with anchors to deal with leftwards/rightwards... But I added two anchors: they're either the first point in extend cell, the edges of the first word in extend word, or the edges of the first line in extend_line.
That lets screen_update_selection only ever compute the moving edge of the selection, and makes the later part of this function quite a bit simpler.

There's also problems with the edges of the screen, whether it's -1 in screen.c or ±0.5 in mouse.c, this PR can have x overflow in either direction; I'd like to introduce some helper to get the previous/next char that deal with this kind of stuff a bit better (it's important for wide chars as well as they're rather tricky to get right; as well as restoring the proper visual_line_ as thees functions often needs to change line to check; the later part might be made easier by adding a second temporary linebuf)


@jscheid, how would you like to move forward? I'm not saying my code is better, I just have the luxury to be here later, and my code hasn't had scrutiny so it likely has its own share of new bugs that will be found once I show it.
If you'd like to finish this PR I'll sit tight and try not to be too annoying, or too influencing, as long as the known bugs are nipped here I'll be happy to wait and build on top of whatever gets merged.
If however you're a bit short on time, I could start over my changes on top of your three commits starting with the bug fix I described then rebuilding the other changes (adding helpers, two anchors etc); the end result will be fairly different from this but since these changes fix bugs that are present here I'd rather do that than try to fix things here then do that anyway (as I'll need them for smart selection :P)

Just say the word, either way, I won't do/say more until you reply :)

jscheid:
> I could start over my changes on top of your three commits 

That sound great actually, I'm afraid the whole selection handling is a lot more hairy than I thought at first; it sounds like you're aiming on improving that, so if you want to take it from here that would be fantastic. Let me know if there's anything else I can help with at this point. Thanks!

I've also just pushed some WIP improvements I've done a few days ago, but never finished... feel free to use them or ignore them. The commit is breaking the tests and also there is still a problem with word-based selection when it is extended beyond the window edge.

martinetd:
Yeah, it is a lot more complex than what I thought at first as well. @kovidgoyal made it looks simple but there's a lot going on! :)

I've had a look at your WIP patch, I'm not sure I understand all this implies at glance so I will probably stay with the previous commit up until which I think I understand what's going on... I'm actually not too sure how to "take over" the PR, I can make a PR for your branch in your repo and that will end up here but I wouldn't be able to remove the WIP patch with that. I don't know how to give someone access to just a branch either... Should I just open a new PR?

Either way I probably won't push anything immediately, I'll be busy tomorrow and disconnected most of Sunday, but this should give me plenty to keep myself busy Sunday and I'll push something early Monday one way or another )

kovidgoyal:
Selection handling is indeed fairly messy, might be ripe for a refactoring. Sadly I dont have time for that now, but feel free to ask and I will provide any guidance I can.

jscheid:
@martinetd good question... if you had push access to this repository (not my fork) you could just push onto my branch, but I guess you don't. Whatever is easy for you, if you go with a new PR just cc me and I'll close this one. Or if you want to create PRs against my branch I'll try to merge them in a timely manner. Thanks again for helping out with this.

kovidgoyal:
closing in favor of #1007

PR #965, opened 2018-09-13
 More detailed "kitty --version" when new commits are present since the latest version tag

Luflosi:
`kitty --version` will now print the output of `git describe --abbrev=4 --dirty --always --tags` when there are new commits since the latest version tag. The colour of the version number is still green like it is currently. `kitty --debug-config` will print the same version as `kitty --version`, and also the first 10 hex digits of the latest commit hash, just like before. The `v` of a git tag is not printed when there are no new commits since the latest version tag to preserve the current behaviour of `kitty --version` for releases, this can be changed if you prefer to always print a version number like `v0.12.1` instead of `0.12.1`. In the future, if you would decide to only rely on the version tags and not the version constant anymore, we could probably parse the version number but since you prefer to manage the version number manually I just compare the two versions and print a warning if they are different. Is `print()` like I use it here good or should that be done differently? This check and the parsing could also be done at compile time but I don't know how to do that, maybe you can help me a bit if you're interested in that.

### DIFF
--- a/kitty/cli.py
+++ b/kitty/cli.py
@@ -171,12 +171,12 @@ def prettify_rst(text):
     return re.sub(r':([a-z]+):`([^`]+)`(=[^\s.]+)', r':\1:`\2`:code:`\3`', text)
 
 
-def version(add_rev=False):
-    rev = ''
+def version(add_hash=False):
+    vcs_hash = ''
     from . import fast_data_types
-    if add_rev and hasattr(fast_data_types, 'KITTY_VCS_REV'):
-        rev = ' ({})'.format(fast_data_types.KITTY_VCS_REV[:10])
-    return '{} {}{} created by {}'.format(italic(appname), green(str_version), rev, title('Kovid Goyal'))
+    if add_hash and hasattr(fast_data_types, 'KITTY_VCS_HASH'):
+        vcs_hash = ' ({})'.format(fast_data_types.KITTY_VCS_HASH[:10])
+    return '{} {}{} created by {}'.format(italic(appname), green(str_version), vcs_hash, title('Kovid Goyal'))
 
 
 def wrap(text, limit=80):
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Thinking about it a little, I'm not sure what the utility of this patch is. All it does is tell you what the git commit and dirty state were the last time kitty's C parts were built. But it could easily be that there have been changes to only python files, so no rebuilding was done in the meantime and the information is out of date. 

kitty --version 

is supposed to be used by end users to know what version they are running. For anybody running kitty from source, it would be both easier and more reliable to use git log/git status to get that information.

Luflosi:
You're right, if `kitty/data-types.c` does not get recompiled when changing a python file, the version number will not be updated. However at least on my machine on macOS `kitty/data-types.c` does seem to get recompiled when changing a python file. This may not be the case on linux, I haven't tested that and I don't understand the build process enough to know that.
Steps I tried:
1. `git checkout` my `verbose_version` branch
2. Run `make app`
3. `kitty.app/Contents/MacOS/kitty --version` outputs `kitty v0.12.1-54-g2a9c created by Kovid Goyal`
4. Delete a comment in kitty/cmds.py
5. Run `make app` again
6. During compilation the line `Compiling kitty/data-types.c ...` is printed
7. `kitty.app/Contents/MacOS/kitty --version` now outputs `kitty v0.12.1-54-g2a9c-dirty created by Kovid Goyal`

If this patch reliably updates the version I would argue that it has some utility. `kitty --version` will then tell users the _exact_ version they are running and not just the latest stable release if they are building from source and the output from of latest stable release stays the same.

kovidgoyal:
If you are using make app then it will always be compiled. However, if you run kitty directly from the source checkout via python3 . as described in https://sw.kovidgoyal.net/kitty/build.htmlthen you can change python files and your changes will be live without needing to recompile and in this case te checkout information in --version will be stale.

kovidgoyal:
While I can appreciate this change is useful for people running via make app, it is useless, or worse, misleading for people running from source checkouts. As such I dont think it should be included.

Issue #970, opened 2018-09-14
Double-scrollback system


martinetd:
Hi, another testing the waters proposal; this one is more ambitious code-wise.

Basically, I like having a large buffer for search, but kitty just eats way too much memory for scrollback (for reference as of today, starting an empty kitty with 64k lines scrollback consumes 50MB, filling the 64k of the buffer adds another 300MB. There is no way I can sustain my 45 opened terms as of this minute with that kind of memory usage)

On the other hand, I don't ever go very far with mod+pgup, so I was thinking a hybrid system could be more efficient:
 - keep a few pages of scrollback as currently.
 - copy lines as they go out of that scrollback into a secondary scrollback buffer as plain text, or even better into a compressed buffer (with some fast algorithm like lz4 or zstd)
 - scrollback actions like mod+pgup cannot go past the normal buffer so there is no change here, but the pager scrollback buffer would get its input from both buffers concatenated

In my opinion, that secondary buffer should not be processed at all on resizes etc and just be kept as plain text; for correctness I might actually suggest to fold the scrollback lines back into one when they are line continuation.

code-wise this isn't as clear to me as the selection code, but I suspect `historybuf_push` could use `line_as_unicode`/`line_as_ansi` just before overwriting the string and pushing to the secondary buffer, controlling its size at this point ; not sure how to decide if unicode or ansi there?
On the display side, as_text would just concatenate both, the line count should work out on its own... And the secondary buffer would live happily only in the C side.

(And as bonus, it would become easy to have this secondary buffer backed by a file if someone wants an unlimited buffer like #884 asked ; not that I would do that part)

Anyway, just like the other issue, I'd like to get an opinion before starting anything.

kovidgoyal:
First, I am somewhat surprised that kitty memory usage on startup depends on scrollback buffer size. I thought I fixed the bug where it was accidentally pre-allocating the entire scrollback buffer. The expected behavior is that startup memory usage is independent of ``scrollback_lines`` are you saying that's not the case for you, even if you run from current master?

Now coming to the question of storing large amounts of scrollback:

1) I agree that if it is implemented, it should not be resized with the normal scrollback
2) I dont know if compressing it is worthwhile, might just be better to write it to disk straightaway. However, scrolling performance is a big priority for kitty, so the implementation of this with or without compression would need to not have a measurable effect on scrolling performance. That probably means using a separate thread for it.
[...]

PR #974, opened 2018-09-15
Secondary pager history buffer

martinetd:
PR as discussed in #970

This implements a straight circular buffer to dump lines as ansi text when they get ejected from the scrollback buffer

The behaviour can be disabled by setting a 0 size.

A couple of commits also look at `line_as_ansi()` as it is now performance-critical for scrollback, it might be possible to do better still.

There are no bugs/missing part of this that I am aware of but I'd like to test a bit more and probably clean up a bit

### DIFF
--- a/kitty/line.c
+++ b/kitty/line.c
@@ -235,7 +235,6 @@ line_as_ansi(Line *self, Py_UCS4 *buf, index_type buflen) {
     if (limit == 0) return 0;
     char_type previous_width = 0;
 
-    WRITE_SGR("0");
     Cursor c1 = {{0}}, c2 = {{0}};
     Cursor *cursor = &c1, *prev_cursor = &c2;
     char_type prev_attrs = 0;
--- a/kitty/line.c
+++ b/kitty/line.c
@@ -259,11 +259,8 @@ line_as_ansi(Line *self, Py_UCS4 *buf, index_type buflen) {
         previous_width = attrs & WIDTH_MASK;
     }
     return i;
-#undef CHECK_BOOL
-#undef CHECK_COLOR
 #undef WRITE_SGR
 #undef WRITE_CH
-#undef WRITE_COLOR
 }
 
 static PyObject*
--- a/kitty/line.c
+++ b/kitty/line.c
@@ -237,21 +237,25 @@ line_as_ansi(Line *self, Py_UCS4 *buf, index_type buflen) {
 
     WRITE_SGR("0");
     Cursor c1 = {{0}}, c2 = {{0}};
-    Cursor *cursor = &c1, *prev_cursor = &c2, *t;
+    Cursor *cursor = &c1, *prev_cursor = &c2;
+    char_type prev_attrs = 0;
 
     for (index_type pos=0; pos < limit; pos++) {
+        bool need_sgr = false;
         char_type attrs = self->gpu_cells[pos].attrs, ch = self->cpu_cells[pos].ch;
         if (ch == 0) {
             if (previous_width == 2) { previous_width = 0; continue; }
             ch = ' ';
         }
-        ATTRS_TO_CURSOR(attrs, cursor);
-        cursor->fg = self->gpu_cells[pos].fg; cursor->bg = self->gpu_cells[pos].bg;
-        cursor->decoration_fg = self->gpu_cells[pos].decoration_fg & COL_MASK;
+        if (attrs != prev_attrs) { ATTRS_TO_CURSOR(attrs, cursor); need_sgr = true; prev_attrs = attrs; }
+#define CMPSET(color) if (cursor->color != self->gpu_cells[pos].color) { cursor->color = self->gpu_cells[pos].color; need_sgr = true; }
+        CMPSET(fg); CMPSET(bg); CMPSET(decoration_fg);
 
-        const char *sgr = cursor_as_sgr(cursor, prev_cursor);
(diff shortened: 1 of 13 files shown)


kitty_tests/history.py
```diff
@@ -0,0 +1,29 @@
...
+        set_options(defaults)
+        s = self.create_screen()
+
+        for i in range(1, 1000*1000):
+            s.draw('aaa')
+            s.linefeed()
+            s.carriage_return()
+
+    def test_history_long(self):
+        defaults.scrollback_pager_history_size = 10
```
martinetd: Change that to 0 to compare with/without; 0 with PR isn't too far from without PR (need better measurements).
The test still shows a bigger difference than what I'm observing in a real kitty instance with a huge cat, though; I haven't gotten to that yet.

kitty/line.c
```diff
@@ -235,35 +235,35 @@ line_as_ansi(Line *self, Py_UCS4 *buf, index_type buflen) {
     if (limit == 0) return 0;
     char_type previous_width = 0;
 
-    WRITE_SGR("0");
```
martinetd: Forgot this one -- I cannot think of any case where resetting the mode is needed at start of line. In particular, I had a patch that carried the sgr mode from the previous line, until I noticed less completely ignores this and resets the mode to zero at the start of every line...

I guess that could have been done for compat with other pagers, or other versions of less? I'll remove that commit and the equivalent in the rewrap function if you think that is useful.

martinetd:
nah, you'd need to copy colors as well as attrs, and that was the slow part of line_as_ansi (always copying attr/colors into the cursor to compare) ; just looking at attributes and copying when needed sped it up quite a bit, so I believe a naive copy would be just slower than what I have right now.

I'll look a bit more at performance but probably not today, take your time and review when you can :)

martinetd:
FWIW, I've tried implementing lz4 compression just for fun (the code is waaay too ugly at this point to push here, and I haven't fixed rewrap for compression)

Performance-wise it's another 5-10% hit, so probably not worth it (or if I can refactor this nicely as an option perhaps -- see next)

Now what's interesting though is that there's a compression ratio; naively compressing line by line as a stream and it goes from 10 (long lines, journalctl) to 4 (short lines, ls) -- with some batching (4k at a time) the speed doesn't change but I get from 13! (journal) to 6.5 (ls)
That kind of makes me want that option, I could reduce the buffer I planned on having at around 10MB to something like 3MB (sorry, I'm a memory-first kind of person :D)

Anyway, it's out of scope for this PR; if I can clean it up nicely to make it optional (perhaps compile-time optional too, kitty aims to work on windows? well lz4 is probably not very difficult to package)


Now I've had my fun I'll look at some more cleanup here first, won't have too much time for the rest of the week but feel free to point out at anything here.

kovidgoyal:
Let's get this done first, and then we can discuss compression. Sadly I have some stuff going on right now that is eating up all my time, but I will try to review this ASAP

martinetd:
Reviewed the style a bit, no major change except renaming `cursor_as_sgr` to `cell_as_sgr` (since it no longer works on cursor - it might use moving from cursor.c to somewhere else as well but I wasn't sure where would be good?); it's probably not perfect but it won't get much better (removed WIP tag)


Fixed `screen_request_capabilities` that I broke when updating `cursor_as_sgr` to work on gpu cells, apparently the screen's cell at cursor position didn't contain the same attributes as the cursor so it's creating a dummy gpucell from the cursor now; that isn't called often so it's probably fine.

I've also removed the slow history.py tests, the scrollback itself is tested datatypes.py and with the options being set now the secondary scrollback gets basic tests at the same time there. I've also left `kitty_tests/bench_scrollback.py` in that can be run manually anytime, feel free to move it if required.

kovidgoyal:
Looks good in general. I'll review in detail asap.

kovidgoyal:
I have merged, since its looks fine. Some comments:

1) In pagerhist_push why are we limiting line length to 1024? Since the line length is known, we could just use that instead of hardocding 1024

2) Need some tests that exercise the pagerhist buffer code

3) Need to handle pagerhist when using the pipe action (see boss.py data_for_at())

martinetd:
1. Didn't give it much thought. We cannot use the line length as is though, the as ansi string can be much bigger than the line length itself (imagine a line changing all the background/foreground/bold/etc at every character) ;
I'd have a hard time coming up with a strict upper bound that isn't ridiculous (and end up wasting precious space at the end of the buffer, because it'd be much more than 1k); I'd rather just try to fit it with whatever we have left at the end of the buffer and if that wasn't enough recompute line_as_ansi from the beginning.

2. I was about to say it was tested in the normal tests, but if you changed the default that is no longer true. I'll duplicate some test.

3. hmm, then you probably also would want get_text to give the alternate scrollback if ansi & history is on? Might be simpler to just remove `add_pager_history` from `window.as_text`'s parameters and make it happen if `as_ansi` and `add_history` are set?

kovidgoyal:
1. I think a good design is something like:
  - use line size * 1.2 as the initial size
  - have line_as_ansi indicate if the output was tuncated and retry with a progressively larger buffer

2. Yeah no need to duplicate the tests, just turn iton for all the tests where it matters. You'd probably need to add an explicit test to test conditions around buffer filling

3. I'm ok with removing add_pager_history. it would be nice to be able to get pager history without ansi though. I can imagine some use cases for it.

kovidgoyal:
make that MAX(1024, line_size * 1.2)

PR #981, opened 2018-09-18
bash completion: make complete helper start with _

martinetd:
It is standard to start complete helpers with underscore to avoid
clobbering the function namespace ; like other shells do.

It is ironic for a completion helper to make completion less agreable...

### DIFF
--- a/kitty/complete.py
+++ b/kitty/complete.py
@@ -44,7 +44,7 @@ _kitty() {
 compdef _kitty kitty
 ''',
     'bash': '''
-kitty_completions() {
+_kitty_completions() {
     local src
     local limit
     # Send all words up to the word the cursor is currently on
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I doubt an underscore significantly decreases the probablity of a namespace collision with a function called kitty_completions, but sure, I dont really care either way.

martinetd:
It's not collision, it's that `kit<tab>` will complete `kitty ` with a space if there is no other kitty* function, but it'll stop at `kitty` without space with this one.

Thanks though!

martinetd:
(And I just had to write that to notice the `-o nospace` didn't help; I don't restart my shells often enough... And that has a good reason to be there. Oh, well, that won't change much - sorry for the noise)

Issue #359, opened 2018-03-06
iBeam cursor difficult to see on dark backgrounds


acheronfail:
When using kitty with a dark background on macOS, it's difficult to see where the cursor is since it is also dark - I find myself struggling to find the cursor and waving it around like mad until I find it.

This is what I mean:
<img width="200" alt="screen shot 2018-03-06 at 6 22 47 pm" src="https://user-images.githubusercontent.com/34289614/37019212-92494acc-216b-11e8-8615-329b1590fcb3.png">

The native Terminal app and iTerm2 both solve this by using a different image for the cursor ([you can find iTerm2's cursor here](https://github.com/gnachman/iTerm2/blob/master/images/IBarCursor%402x.png)).

I think that it would be nice if we had a hybrid dark/light cursor like iTerm2's, or at least a cursor that is easier to see with light/dark backgrounds. I'd also be interested in making the PR myself if you could point me in the right direction. 😄

kovidgoyal:
Doesn't Apple allow you to change your mouse theme? It should not be the job of every application to provide its own mouse cursor. Personally, I like to use white mouse cursors with a black outline and a slight black shadow. Works well on all background colors, like this: http://www.rw-designer.com/cursor-set/opensuse-oxygen

acheronfail:
Unfortunately I believe Apple restricts changing the mouse themes. It used to be possible in versions of macOS < 10.10, but since SIP they've cracked down a lot more on System restrictions.

I also prefer white cursors - and while I agree that it shouldn't be the job of every application to provide its own mouse cursor, that doesn't make me like the situation Apple places us in since we can't change the mouse theme.

kovidgoyal:
Wow, that's bad. Well OK, I suppose we have to workaround Apple's control freakyness. If you do want to implement a custom cursor in kitty, the relevant place would be in kitty/glfw.c see set_mouse_cursor()
 (googling should tell you what API to use for glfw custom cursors). 

Note that kitty uses three different cursor types (ibeam, hand, arrow). IMO, the way this should work is:

1) Allow user to define a mouse cursor theme in the config directory so they can customize the cursor images to whatever they want. I have the opposite philosophy to Apple :) 

2) On macOS only use special cursor images shipped with kitty by default.

acheronfail:
Okay, I'll give that a crack when I get the chance. 😄

kovidgoyal:
I didn't bother with making it configurable, I leave that as an exercise for whoever is interested.

[...]

Issue #794, opened 2018-08-06
mouse cursor invisible when inside the terminal windows in mac os x


egaburov:
Using master branch.
Built with `make app`

Once the mouse cursor is moved inside an active kitty terminal window, it disappears. I can still select text inside window. but no cursor is visible.

Image with selected text but no cursor:

<img width="1959" alt="screenshot 2018-08-05 21 50 55" src="https://user-images.githubusercontent.com/1485713/43697527-ac30f9ac-98f9-11e8-8f52-4b23b5f4d41e.png">

kovidgoyal:
I cannot replicate this using kitty built from current master.  Steps I tried:

1) Start kitty
2) Wait a few seconds until the mouse cursor disappers
3) Move mouse
4) Mouse cursor re-appears

Make sure your mouse_hide_wait setting i kitty.conf is correct. it should either be zero or some high number like 3 or 5.

egaburov:
When running on integrated GPU (macbook pro 2017,  15'') mouse disappears, but when forcing discrete gpu (by disabling  "automatic graphics switching" in System Preferences->Energy Saver) mouse appears. I have mouse_hide_wait set to 0.0, tried 3, did not try 5, same effect. no mouse cursor when using integrated gpu.

kovidgoyal:
I dont have a macbook pro so I cannot test with that, but on my air which has only an integrated GPU, I cannot reproduce. Presumably, there is some issue with using custom mouse cursor images and dual GPUs. Sadly, without the needed hardware it is hard for me to fix. You can test that by commenting out the call to set_custom_ibeam_cursor in main.py

egaburov:
yup, commenting call to `set_custom_ibeam_cursor` fixes the issue. Thanks

PR #985, opened 2018-09-18
[WIP] add support for custom cursor on mac

esell:
First off, I am not a C or Python person so please excuse any obvious or stupid mistakes :).

This is in reference to issue #359 and is an attempt to create a config option for folks to specify a custom ibeam cursor when using macos. Currently the code expects a `beam-cursor.png` file to live in the local kitty config directory. I'm not really sure the best way to handle this as I'm not sure if you want to include a custom set with the distributed app or not.

The feature can be turned on by setting the `custom_cursor` flag in kitty.conf to true/yes.

Thoughts? Suggestions?

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -291,7 +291,8 @@ def cursor_text_color(x):
         return
     return to_color(x)
 
-
+o('macos_custom_beam_cursor', False, long_text=_('''Enable/disable custom cursor for macOS. WARNING, 
+this might make your cursor invisible on dual GPU machines.'''))
 o('cursor', '#cccccc', _('Default cursor color'), option_type=to_color)
 o('cursor_text_color', '#111111', option_type=cursor_text_color, long_text=_('''
 Choose the color of text under the cursor. If you want it rendered with the
--- a/kitty/main.py
+++ b/kitty/main.py
@@ -120,9 +120,7 @@ def get_new_os_window_trigger(opts):
 
 def _run_app(opts, args):
     new_os_window_trigger = get_new_os_window_trigger(opts)
-    if False and is_macos:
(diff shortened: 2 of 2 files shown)


kovidgoyal:
Dont worry about your lack of python/C. We all start somewhere. kitty already distributes custom beam cursor images on macOS. The feature is turned off because of #794. All you need to do is add a config option for it that defaults to off. The option name for it should be something like "macos_custom_beam_cursor" and it should be a boolean option, with help text that explains the risk of turning it on (namely that it might make the cursor invisible on dual GPU machines).

esell:
Wow, that is much easier compared to what I had! 

I backed out all of the previous changes and added new code based on your suggestions. Limited testing on my end seems to work.

scorphus:
Thanks, @esell!

dylan-chong:
@esell Was the macos_custom_beam_cursor  option supposed to be added to the template configuration file?

esell:
@dylan-chong it was supposed to but i'm assuming it isn't based on your question? :)

dylan-chong:
It is not

dylan-chong:
Judging by the pull request changes it wasn't

esell:
Hmm, i'm not quite sure where it would go... It was added here:

https://github.com/kovidgoyal/kitty/blob/master/kitty/config_data.py#L817

@kovidgoyal, is this the right location?

dylan-chong:
@esell ah i just realised! The default kitty configuration must be generated only when installing kitty. When upgrading kitty, i assume that there is no way of copying the new changes across to the existing configuration . I didn't realise the file was generated from python code, so i assumed to kitty was copying across a file that looks the same as a configuration file

PR #1007, opened 2018-09-22
[WIP] Selection overhaul

martinetd:
Build on top of @jscheid's work in #949 and do a light refactoring of selection; I can't push there so opening a new PR instead.

For now I just pushed the scroll_anchor fix for the bug discussed in the other PR, @Luflosi please give it a shot.

I might do a bit more tonight but the bulk of it will come Monday, just didn't want to sit on that one fix.

TODO:

 - [x] Fix scrolled_by handling (partially fix for EXTEND_CELL)
 - [ ] Fix line continuation handling (x < 0 or >= screen->columns) ; will also fix wide chars at the same time for free.
 - [ ] Rework update_selection to only ever care about one edge (will fix EXTEND_WORD with scrolling)

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -130,11 +130,11 @@ distance_to_window(Window *w, OSWindow *os_window) {
 static bool clamp_to_window = false;
 
 static inline bool
-cell_for_pos(Window *w, unsigned int *x, unsigned int *y, OSWindow *os_window) {
+cell_for_pos(Window *w, unsigned int *x, unsigned int *y, unsigned int *edge_x, unsigned int *edge_y, OSWindow *os_window) {
     WindowGeometry *g = &w->geometry;
     Screen *screen = w->render_data.screen;
     if (!screen) return false;
-    unsigned int qx = 0, qy = 0;
+    double qx = 0, qy = 0;
     double mouse_x = global_state.callback_os_window->mouse_x;
     double mouse_y = global_state.callback_os_window->mouse_y;
     double left = window_left(w, os_window), top = window_top(w, os_window), right = window_right(w, os_window), bottom = window_bottom(w, os_window);
@@ -144,12 +144,15 @@ cell_for_pos(Window *w, unsigned int *x, unsigned int *y, OSWindow *os_window) {
     }
     w->mouse_pos.x = mouse_x - left; w->mouse_pos.y = mouse_y - top;
     if (mouse_x < left || mouse_y < top || mouse_x > right || mouse_y > bottom) return false;
-    if (mouse_x >= g->right) qx = screen->columns - 1;
-    else if (mouse_x >= g->left) qx = (unsigned int)((double)(mouse_x - g->left) / os_window->fonts_data->cell_width);
-    if (mouse_y >= g->bottom) qy = screen->lines - 1;
-    else if (mouse_y >= g->top) qy = (unsigned int)((double)(mouse_y - g->top) / os_window->fonts_data->cell_height);
+    if (mouse_x >= g->right) qx = screen->columns - 0.5;
+    else if (mouse_x >= g->left) qx = ((double)(mouse_x - g->left) / os_window->fonts_data->cell_width);
+    if (mouse_y >= g->bottom) qy = screen->lines - 0.5;
+    else if (mouse_y >= g->top) qy = ((double)(mouse_y - g->top) / os_window->fonts_data->cell_height);
     if (qx < screen->columns && qy < screen->lines) {
-        *x = qx; *y = qy;
+        *x = (unsigned int)qx;
+        *y = (unsigned int)qy;
+        *edge_x = (unsigned int)(qx + 0.5);
+        *edge_y = (unsigned int)(qy + 0.5);
         return true;
     }
     return false;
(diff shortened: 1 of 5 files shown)


martinetd:
Not done as much as I would have liked today - or rather, I didn't understand the edge change properly so I'm left with a few off-by-ones in some situations I need to look at, and I really cannot think straight in a plane so delayed this for a bit longer.. No time estimate to give right now, it's not missing too much for a first appreciation of how badly I've messed it up but I'd like to iron out the off-by-ones I'm aware of at least and I don't really know how much time I'll have in the next few days.

There's also one more bug with wide chars that I think I'm missing something obvious, basically if wide chars wants to be drawn one before the end of the line an early carriage return/line feed is entered early to not 'truncate' the wide char between both lines in `screen_draw`/`if (UNLIKELY(self->columns - self->cursor->x < (unsigned int)char_width))` -- the problem is that this leaves a space (32, ' ') at the end of the line where the cursor was, and that space is selectable so I can't skip it in `screen_selection_range_for_word` but it also shows in copy/paste (this can be reproduced on master)
I don't see where that space comes from; from what I can see the cursor is represented as an empty cell (contains 0/BLANK_CHAR) so it's not just the carriage return/linefeed leaving junk behind.. For example I also see a space if I pre-enter aligned wide-chars and just delete one single char to create the hole... So I assume it's a space on purpose? But I don't see where? And I think it'd work better with a blank char (at least that'd fix selection for me)?

kovidgoyal:
IIRC blanks in-between contiguous text (and a wide char moving oto the next line is contiguous) are treated as spaces. This is probably incorrect (they should be removed instead) however there are problems with doing that. It is a little less efficient code wise and if you let selections skip these blank chars it means your selections could have weird holes in them.

Luflosi:
The new commit definitely improved the selection. However I found two more issues with selection. Here is how to reproduce them:
1. 1. Generate some text in the scrollback buffer, e.g. press enter a bunch of times
   2. Scroll up a bit
   3. Select some text that will be out of screen when scrolling down again
   4. Scroll down until the text is no longer visible
   5. The character in the top right corner becomes selected
2. 1. Generate some text in the scrollback buffer, e.g. press enter a bunch of times
   2. Scroll up a bit
   3. Select some text over multiple lines that will be at least partially out of screen when scrolling down again. The mouse button needs to be released on the right side of the window, either outside of the window or very close to the edge
   4. The selection becomes invisible
   5. Scroll down a bit until the selection is partially out of the screen
   6. The selection becomes visible again
   7. Scroll down a bit further to see the first bug again

martinetd:
Thanks for the bug finding.
I've given it some more thought and I think I don't like how the edge-level selection is being done right now as it makes the selection function really quite complex, I've got a few ideas to make it much simpler:
 - keep both anchors but only ever set one end in update_selection from x,y; then decide which comes first like the old code used to do (it was always setting end, keeping start where it was) -- that'll remove the need to check for extending direction in most cases
 - make the selection really mean edge for the rest of the code, so update_selection won't have to deal with if same cell put the selection all the way to uint max and deal with +1 offsets etc as the meaning of x given from mouse will be correct

That should fix most of my off-by-ones automagically and that might very well fix the first of your two bugs; the later one is probably something else, but I might figure it out whe nchanging the rest of the code to be based on edges.

I'm still busy for another day, should have something nice for the end of the week though!

@kovidgoyal: not sure I understood your answer, I do have weird holes atm e.g. selecting `テス \r\nト` will paste `テス ト` with a space. More to the point, I have no way to decide if the user really entered the first two chars then a space then the third chard or if they entered all three in a row and the widechar shifting happened; so it's not optional to clear that space into a blank char.

kovidgoyal:
The hole you are talking about is on pasting, the hole I am talking about is on selecting. When he blank char is at the end of the line, there is no hole on selecting, but imagine if it was instead in the middle of the line, then the selection would skip over it, leaving a hole.

martinetd:
Ah, yeah but in the middle of the line there is no shifting due to wide char, a wide char can start at an odd position, and we can just hilight the blank-char-after-wide-char there as usual.
My problem is `wide char - blank char - space || (end of line, continued to next wide char)`

I don't really care if it's hilighted or not, but it should not be considered for the selection text at least.

kovidgoyal:
I am fine with blanking the left-over char when moving a wide char onto the next line. But yo uhave to be careful you dont break something else, such as line wrap/pasting/etc.

martinetd:
Sorry for slow progress, really cannot code as much as I intended to while traveling - finally home, should be able to finish this now.

Just an update on the wide char spilling over next line, at least for bash it turns out the space comes from the shell (bash apparently transforms `テスト` into `テス ト\rト` if the `ト` would be split!...)
zsh doesn't though but we still have a space there, so that might be "fixable" to not break the word. I haven't looked at why but putting it on hold for now, will try to finish the rest first and eventually deal with that later if I still care (besides, anything related to that might conflict with the other wide glyph PR, so I have a good excuse :p)

kovidgoyal:
No worries, take your time. Best way to test wide char things is to use cat, shells do funky stuff with unicode chars.

martinetd:
So the reason this is slow (mentionned on another issue), might as well describe what I'm trying to do now to simplify the whole thing:
 - going back to how the current code works of start always being the initial click position and end being the update position; that simplifies a lot of swapping variables around in the code, the only places that need to check are the final drawing/getting text from selection helpers as they currently already do.
 - @jscheid's approach working on edge is logical but implementation-wise I find it very hacky to decide "oh there's nothing so let's place all elements to UINT_MAX"; I want the end part of the interval to be exclusive. That should simplify quite a few things as well, but it needs some refurbishing..
 - meanwhile I found a new bug around scrolling text (selecting when new output is being produced), haven't looked where to hook for that, probably going to be in another PR anyway; similarily I've reviewd my ambition down and probably won't finish the wide char issue I talked about. I'll try to get back to just the edge stuff and the refactoring I needed for "smart" selection extension; I really picked a bad timing to do that one :P

kovidgoyal:
:) selection is a pain I remember it caused me endless headaches when I was developing it, so take your time.

Luflosi:
Would you mind resolving the conflict?

martinetd:
@Luflosi sorry missed your message; I've just rebased this; just renamed some variables in first commit. All commits compile, the last one looks like it's working about as much as the previous HEAD -- and the cobwebs are intact :D

Luflosi:
Could you please resolve the conflict again?

Issue #809, opened 2018-08-16
Cannot map shift+ctrl+tab to previous_tab


Luflosi:
I was trying to do basically the same thing as described in #801 but with `shift+ctrl+tab` instead. `kitty --debug-keyboard` also did not receive the tab key while pressing shift and control at the same time. Looking at a143faaf056fcf85098b80758199163c9450068b, it seems that you don't check for this combination.

chdsbd:
@Luflosi Could you check if f9b1f6b fixed this? I just tried the newest Kitty release and `shift+ctrl+tab` doesn't seem to work even with this change

I have this config

```
clear_all_shortcuts yes
map cmd+t new_tab
map ctrl+tab next_tab
map shift+ctrl+tab prev_tab
```

`ctrl+tab` works, `shift+ctrl+tab` doesn't.

Luflosi:
@kovidgoyal Do you think `prev_tab` should be an alias for `previous_tab `?

kovidgoyal:
I have no objections to that, but if you do it, do it for the pther previous commands as well, such as previous_window

Luflosi:
Is there already some command which is an alias for another command? If not how would I implement this?

kovidgoyal:
Just add prev_tab = previous_tab under the definition of the previous_tab method

PR #1008, opened 2018-09-22
Add prev_tab and prev_window as aliases for previous_tab and previous_window

Luflosi:
As discussed in #809.
I'm pretty sure the only two commands with `previous` in them are are `previous_tab` and `previous_window` and there are no others.
Is the change in `docs/index.rst` ok or do you want a different separator, like `,` instead of `or`?


kovidgoyal:
:sc:`XXX` reloves tot he keyboard shortcut for that action, so there is no point adding it.

Luflosi:
Ok, should I revert the changes to `docs/index.rst` then?

kovidgoyal:
No need I did it for you

PR #1026, opened 2018-09-26
docs/build: extend launch script to use development exe

blueyed:
This is important for `kitty_exe()` to use the development script also,
e.g. for any kittens being launched.

[ci skip]

### DIFF
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -52,7 +52,15 @@ If that works, you can create a script to launch |kitty|:
 
     #!/usr/bin/env python3
     import runpy
-    runpy.run_path('/path/to/kitty/dir', run_name='__main__')
+    import sys
+
+    kitty_path = '/path/to/kitty/dir'
+
+    sys.path.insert(0, kitty_path)
+    from kitty.constants import kitty_exe  # noqa: E402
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I dont understand the purpose of this. If you are puttting a script named kitty in ~/bin or elsewhere on your PATH kitty should pick it up automatically in kitty_exe()

kovidgoyal:
Are you trying to have multiple kitty's on your path and have kitty_exe pick one that is lower precedence?

blueyed:
Oh, sure - I've named my wrapper script `kitty-git`.
I see that it is more code to copy into your own script, but in general this avoids some `os.access` calls and also handles a wrapper script that is not in `$PATH` at all.
To be clear: it took me a while to figure out why changes to `.so` files were not picked up, because it used the system-wide installation for the kitten.

kovidgoyal:
Well the instructions do say to name the script kitty and put it in PATH. I prefer to keep basic instructions/scripts as simple as possible. Otherwise there is more scope for things to go wrong. For instance, kitty_exe() is not really a public API, it may change in the future and I'd need to remember to update the script in that case.

PR #1036, opened 2018-09-28
[WIP/RFC] Handle multiple spaces after PUA glyphs

blueyed:
This handles multiple spaces after a PUA glyph (instead of just one).

Before:
multi-pua-before

After:
multi-pua-after

It still does not work as in (patched) urxvt, because kitty appears to shrink it to fit its height into the cell.  In urxvt it looks like follows:
multi-pua-urxvt

Where does the shrinking / fitting to cells happen?  I would rather have it take the full width of the 3 cells, and be cut at the top/bottom.

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -1008,12 +1008,25 @@ render_line(FONTS_DATA_HANDLE fg_, Line *line) {
         CPUCell *cpu_cell = line->cpu_cells + i;
         GPUCell *gpu_cell = line->gpu_cells + i;
         ssize_t cell_font_idx = font_for_cell(fg, cpu_cell, gpu_cell);
+
         if (is_private_use(cpu_cell->ch)
                 && cell_font_idx != BOX_FONT
                 && cell_font_idx != MISSING_FONT) {
+            int cells;
+            if (cell_font_idx > 0) {
+                Font *font = (fg->fonts + cell_font_idx);
+                glyph_index glyph_id = glyph_id_for_codepoint(font->face, cpu_cell->ch);
+
+                int width = get_glyph_width(font->face, glyph_id);
+                cells = ceilf((float)width / fg->cell_width);
+            } else {
+                cells = 1;
+            }
+
             int j = 0;
             while ((line->cpu_cells[i+j+1].ch == ' ' || line->cpu_cells[i+j+1].ch == 0)
                     && j < MAX_NUM_EXTRA_GLYPHS_PUA
+                    && j < cells
                     && i + j + 1 < line->xnum) {
                 j++;
                 // We have a private use char followed by space(s), render it as a multi-cell ligature.
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -1010,12 +1010,12 @@ render_line(FONTS_DATA_HANDLE fg_, Line *line) {
                 && cell_font_idx != BOX_FONT
                 && cell_font_idx != MISSING_FONT) {
             int j = 0;
-            while ((line->cpu_cells[i+1+j].ch == ' ' || line->cpu_cells[i+1+j].ch == 0)
+            while ((line->cpu_cells[i+j+1].ch == ' ' || line->cpu_cells[i+j+1].ch == 0)
                     && j < MAX_NUM_EXTRA_GLYPHS
-                    && i + 1 + j < line->xnum) {
+                    && i + j + 1 < line->xnum) {
                 j++;
                 // We have a private use char followed by space(s), render it as a multi-cell ligature.
-                GPUCell *space_cell = line->gpu_cells + i+j;
+                GPUCell *space_cell = line->gpu_cells + i + j;
                 // Ensure the space cell uses the foreground/background colors from the PUA cell.
                 // This is needed because there are stupid applications like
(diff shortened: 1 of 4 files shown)


kitty/fonts.c
```diff
@@ -1006,21 +1006,33 @@ render_line(FONTS_DATA_HANDLE fg_, Line *line) {
...
+                space_cell->fg = gpu_cell->fg;
+                space_cell->bg = gpu_cell->bg;
+                space_cell->decoration_fg = gpu_cell->decoration_fg;
+            }
+            if (j) {
+                RENDER;
+                render_run(fg, line->cpu_cells + i, line->gpu_cells + i, j + 1, cell_font_idx, true);
+                run_font_idx = NO_FONT;
+                first_cell_in_run = i + 1 + j;
+                prev_width = gpu_cell->attrs & WIDTH_MASK;
```
blueyed: Re `prev_width`: I think this should be 1 always (regardless of this patch) to not `continue` above in any case.

blueyed:
> Where does the shrinking / fitting to cells happen? I would rather have it take the full width of the 3 cells, and be cut at the top/bottom.

This appears to happen in `place_bitmap_in_canvas` (for fontconfig).

kovidgoyal:
Fitting into the cell is not going to change -- it is needed for proper rendering of glyphs from fallback fonts, which can have different metrics than the main font.

kovidgoyal:
As for handling multiple spaces, IIRC kitty supports a max ligature length of 9 so you should ensure that the fake ligature you create is no longer than that, or bad things will happen down the road (see MAX_NUM_EXTRA_GLYPHS)

blueyed:
MAX_NUM_EXTRA_GLYPHS is check here: https://github.com/kovidgoyal/kitty/pull/1036/files#diff-bf6698a6663f1ce908538b90d1fbe1e7R1014

blueyed:
> Fitting into the cell is not going to change -- it is needed for proper rendering of glyphs from fallback fonts, which can have different metrics than the main font.

I've looked into a bit more and the bitmap created from fontconfig is only 25 pixels wide already (`render_bitmap: rows=23 width=25 max_width=30, cell_height=20`).  This is likely due to internally setting the desired font height I assume.  (Not sure why `rows > cell_height`?)
In this case this is already a fallback font btw (via `symbol_map`).
What I would like to achieve is that the bitmap would take the max_width, and gets cropped at top/bottom - but I can see that this is nothing you would like to have in general.

blueyed:
> No, I prefer scaling to cropping. But I am fine with having an option to control that behavior.

Great.  Do you have some pointers for me where this needs to happen?
Is it related to `set_font_size` being based on (preferred) height?

kovidgoyal:
Basically you'd need to change the code that fits the glyph(s) into the canvas to crop instead of scaling based on the value of the option

blueyed:
That would be `place_bitmap_in_canvas` I assume??
render_bitmap before loads the glyph with width=25 already though (for max_width=30 (3 cells)).  Therefore I am assuming that font size setting is causing this.
(There appear to be several places where width should be used instead of height then maybe)

kovidgoyal:
sorry dont remember off the top of my head would have to read the code, and I dont have the time for that right now

blueyed:
Ok, will look into this a bit more.

As for this PR: it should be independent from what we discussed above, but I've just noticed that it is broken (added space before):
multi-pua-broken

kovidgoyal:
yes glyphs are currently centered in the canvas. this works well for most glyphs such as emoji and flags and  east asian characters. If you want left alignment for pua glyphs, you will need to change the freetype rendering api to pss in a bool controlling alignment. rememberthat these functions are also used in coretext so make the changes there as well.

Finally, the more cells there are in a ligature, the worse lookup performance is. so I would prefer restricting the max number of spaces to 2 or 3 ince I doubt nerd fonts need more than that.

blueyed:
I've added some commit to use the actual number of cells (currently only for freetype, based on the bitmap).  Does this make sense?

blueyed:
The metrics for U+e0c2 are this btw:
```
glyph: 1581 bitmap.width: 25 bitmap.rows: 23 horiAdvance: 1600 horiBearingX: 0 horiBearingY: 1152 vertBearingX: -832 vertBearingY: -192 vertAdvance: 1088 width: 1600 height: 1472
```

(cell height is 20, cell width 10)

blueyed:
If `get_glyph_width` makes sense like this, it could be used with `is_glyph_empty`, of course.

kovidgoyal:
I'll review this when I have a little more time.

kovidgoyal:
I'll review this sometime in the next few days. Apologies for the long delay, my personal life has been crazy and I am trying to catch up now.

blueyed:
Wow, thanks for merging - I would have expected some review/comments on it first, and would have squashed at least the fixup then, but ok.. :)
Thanks!

kovidgoyal:
I made some minor changes, but otherwise it looked good, and I am not fussy about commit history :)

I find merging things quickly that require only minor cleanups is better for my long term productivity.

blueyed:
Cool, thanks again!

PR #1041, opened 2018-09-30
Don't guard against alternate_screen in as_text

maximbaz:
Complementary to 15f534a8f9d44d9873b07d8309db498aca7c6cd1

### DIFF
--- a/kitty/window.py
+++ b/kitty/window.py
@@ -438,6 +438,7 @@ class Window:
 
     def as_text(self, as_ansi=False, add_history=False, add_pager_history=False, add_wrap_markers=False, alternate_screen=False):
         lines = []
+        add_history = add_history and not (self.screen.is_using_alternate_linebuf() ^ alternate_screen)
         if alternate_screen:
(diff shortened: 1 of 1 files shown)


kitty/window.py
```diff
@@ -438,7 +438,6 @@ def destroy(self):
 
     def as_text(self, as_ansi=False, add_history=False, add_pager_history=False, add_wrap_markers=False, alternate_screen=False):
         lines = []
-        add_history = add_history and not self.screen.is_using_alternate_linebuf() and not alternate_screen
```
kovidgoyal: The secondary screen (the one used by fullscreen applications) has no history buf. So if we are getting the text from that screen add_history must be false. which means that either the screen is not using alternate_linebuf and alternate screen is false or both are true. basically the condition should be

not (self.screen.is_using_alternate_linbebuf() ^ alternate_linebuf)

maximbaz: Got it, thanks

PR #1050, opened 2018-10-05
I added a ignore *.swp files under the git management

matoruru:
I think this modify is useful for developer using vim editor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.so
 *.pyc
 *.pyo
+*.swp
 tags
 build
 linux-package


kovidgoyal:
Why not just use a centralized backup directory in vim. See :help backupdir  then you dont need to go around adding *.swp to the gitignore of every project. Or if you dont want to do that, add it to the global .gitignore file on your computer.

matoruru:
Oh I am so sorry, you are right. I actually didn't that vim's feature. Thank you for teaching me.

PR #1055, opened 2018-10-09
Fix Wayland key repeat frequency

tarmack:
The Wayland protocol specification uses a Hz value in the key repeat rate field. The code however uses this value as a millisecond interval. This pull request makes the code calculate the interval from the Hz value.

This should probably be upstreamed to glfw but I'm not entirely sure how to go about that since the code seems to have seen quite a few changes there (but looks like it still does the wrong thing here).


kovidgoyal:
kitty's copy of glfw has diverged from upstream, so if you want to send it upstream, it would probably be a different patch.

tarmack:
I noticed that the code has diverged, will you be syncing with upstream again at some point in the near future? Because If you do not, I don't really care to go and test with upstream glfw I suppose. And if you will sync I may just go about fixing it then I suppose.

kovidgoyal:
No the divergence is permanent for many reasons, principally https://github.com/glfw/glfw/issues/1140

Issue #1066, opened 2018-10-15
Configuration value to disable showing active process in Mac menu bar


sleach:
To start, thank you for you efforts on Kitty, great app!

Wanted to request adding a configuration variable that would disable showing the active process on the Mac menu bar (assume that would be a "window" property). Currently, all it basically does is "flash text" as I go about doing work (showing current process) and it's distracting.

Attaching a screenshot to show the component I am referring to. Thanks!

![2018-10-15 at 7 30 am](https://user-images.githubusercontent.com/33944/46953923-6aa39f00-d04c-11e8-9a3e-17266245fe4b.png)

kovidgoyal:
that title will change in several places, the tab bar, the window title and the global menubar. I dont see how removing it from one place is going to make a significant difference to distraction levels. You can instead use the --title option to specify some fixed title, which will override whatever titles are set by the running programs in all places.

Luflosi:
What is that menu bar title for anyways? We already have the window title to show us the text. I also don't know of any other app that uses the menu bar to display text. For a newcomer to kitty this might also be confusing since that person will be used to be able to click on everything in that menu bar; clicking on this item will either do nothing at all or make that menu bar item flash blue, depending on how short the click is. Also, as I explained in #1072, using the `--title` option will not work for new tabs and OS windows.

Luflosi:
Could you please explain, why this feature exists? Or why there is no option to turn this off? I have never seen the menu bar being used to display stuff in this way with any other app. And the information displayed there is also always displayed in the normal window title and sometimes a in a few other places, making this redundant IMO. Your suggestion to set the title with `--title`, presumably to an empty string, does not have the same effect as optionally not displaying this menu bar title, because setting the title in this way will also change it in tabs, the normal window title and in Mission Control, all places, where I personally _do_ want to see it.

kovidgoyal:
The suggestion to use --title was for the OPs complaint to avoid distraction. Let me ask you instead, why does making use of hitherto unused and useless space offend you so much? It is nice to have one single place that never changes to check for the title of the current window. Whether other applications do it or not is irrelevant, if I was interested in only doing things other's did, I would never have started kitty.

Luflosi:
> The suggestion to use --title was for the OPs complaint to avoid distraction.

Ok, makes sense.

> why does making use of hitherto unused and useless space offend you so much?

[...]

PR #1068, opened 2018-10-15
Add option to hide the menu bar item which contains the window title

Luflosi:
Implements #1066. I hope the option name makes sense. Also, what does the line I added in `kitty/state.c` do?


kovidgoyal:
see discussion in #1066

PR #1069, opened 2018-10-15
Filter tic output to remove "older tic versions may treat the description field as an alias"

Luflosi:
Since you mentioned in #840 that you don't want to miss actually useful warnings from tic, I implemented a simple filter that ignores all messages that match a very specific regex.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -599,9 +599,9 @@ def package(args, for_bundle=False, sh_launcher=False):
     for x in (libdir, os.path.join(ddir, 'share')):
         odir = os.path.join(x, 'terminfo')
         safe_makedirs(odir)
-        proc = subprocess.run(['tic', '-x', '-o' + odir, 'terminfo/kitty.terminfo'], check=True, text=True, stderr=subprocess.PIPE)
+        proc = subprocess.run(['tic', '-x', '-o' + odir, 'terminfo/kitty.terminfo'], check=True, stderr=subprocess.PIPE)
         regex = '^"terminfo/kitty.terminfo", line [0-9]+, col [0-9]+, terminal \'xterm-kitty\': older tic versions may treat the description field as an alias$'
-        err_msg = proc.stderr.rstrip()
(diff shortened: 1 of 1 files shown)


Luflosi:
I don't quite understand why the test failed. It does not seem to know the `text` argument of `subprocess.run` even though the docs say that function with the parameter exists since python 3.5 (which you told me is the minimum supported python version): https://docs.python.org/3/library/subprocess.html

Luflosi:
And what is build-terminfo used for/when is it used?

kovidgoyal:
according to the python docs, text was added in 3.7 not 3.5

```
Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines. Added the capture_output parameter.
```

Dont use text anyway. pythons encoding handling magic is a bad design. Just decode stderr using utf-8

Luflosi:
Ah, I didn't notice that changelog. Now I just decode stderr using utf-8 like you suggested. I decode it after the `subprocess.run()` since the `encoding` parameter was introduced in python 3.6. I now also use a loop in case `tic` output multiple lines.

Luflosi:
What is build-terminfo used for/when is it used?

kovidgoyal:
to build the kitty terminfo files from source. It is used during the build process.

kovidgoyal:
It is run manually to update the compiled terminfo files that are checked into version control, which is what allows kitty to be run directly from source. In setup.py it is run to generate the terminfo files that are added to the built kitty package.

Luflosi:
Why does `setup.py` run `tic`? Can't it just copy the stuff from the `terminfo` directory? Now, that would not compile `kitty.terminfo` every time `package()` of `setup.py` is called, but do you want that? And if you want that, then wouldn't it be better to put the terminfo-compilation functionality of `package()` and `build-terminfo` into its own function in `terminfo.py`?

kovidgoyal:
I do want it, and it is not complex enough to put into its own function.

PR #1072, opened 2018-10-16
Add option to hide the window title

Luflosi:
I like the way kitty looks without a title. I know that I can configure my shell and probably all my other programs like vim to not update the title but that would also hide the title from the tabs, where I very much prefer having a title since I find it hard to navigate the tabs without it. This is why I added an option to do just that.
I'm a bit unsure if I implemented this in the best way. The description of the option might also need an explanation which is a little more verbose than just `Hide the kitty window title.`, but I could not think of one. I also did not test this on any OS other than macOS yet. Does creating new OS windows with an empty title have any unforeseen side effects?

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -139,7 +139,7 @@ class Boss:
                 os_window_id = create_os_window(
                         initial_window_size_func(opts_for_size, self.cached_values),
                         pre_show_callback,
-                        appname, wname or self.args.name or cls, cls)
+                        '' if self.opts.hide_title else appname, wname or self.args.name or cls, cls)
         tm = TabManager(os_window_id, self.opts, self.args, startup_session)
         self.os_window_map[os_window_id] = tm
         return os_window_id
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -544,7 +544,7 @@ update_window_title(Window *w, OSWindow *os_window) {
     if (w->title && w->title != os_window->window_title) {
         os_window->window_title = w->title;
         Py_INCREF(os_window->window_title);
-        set_os_window_title(os_window, PyUnicode_AsUTF8(w->title));
+        if (!OPT(hide_title)) set_os_window_title(os_window, PyUnicode_AsUTF8(w->title));
 #ifdef __APPLE__
         if (os_window->is_focused) cocoa_update_title(w->title);
 #endif
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -524,6 +524,9 @@ Negative values will cause the value of :opt:`window_margin_width` to be used in
 o('window_padding_width', 0.0, option_type=positive_float, long_text=_('''
 The window padding (in pts) (blank area between the text and the window border)'''))
 
+o('hide_title', False, long_text=_('''
+Hide the kitty window title.'''))
+
 o('active_border_color', '#00ff00', option_type=to_color, long_text=_('''
 The color for the border of the active window'''))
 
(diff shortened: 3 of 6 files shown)


kovidgoyal:
there is no need for an option for this, use

kitty --title " "

that will effectively be a blank window title and will override whatever title is set by the program running inside kitty. Note that you might need to use a unicode space character or other invisble character instead as kitty might have code to strip spaces when processing titles somewhere.

or better yet, if you dont want to see th title, just hide the titlbar itself, since without the title, it is largely useless. See the macos_hide_titlebar (or some such) exiting option.

Luflosi:
How do I specify a command line option for a macOS App? The only way I could find was to create another app with Applescript and use it to launch kitty, but that seems very inconvenient.

kovidgoyal:
see the kitty FAQ

Luflosi:
Oh, I didn't notice that piece of info in the FAQ yet, sorry.
Unfortunately `kitty --title " "` does not do what I'd like to do since it sets the title everywhere, in the window title, the menu bar and in tabs. Also, when I open a new tab or OS window, that tab or window is unaffected by the option.
Here is an example on macOS:
1. Open kitty with `kitty --title "asdf"`
2. The OS window title is `asdf` and the menu bar title is `:: asdf`
3. Open a new tab
4. The original tab title is `asdf` but the new tab title, window title and menu bar title are what my shell sets the title to

This pull request only hides the window title and nothing else as far as I can tell.
I would still like to see the title in tabs and that's something kitty can't do atm without also showing it in the OS window title. Without the title in the tab I find it harder to find a tab I want to switch to.

Luflosi:
And I don't want to hide the entire titlebar since then I'm unable to drag the window around, I don't have the three buttons, no rounded window corners and no window shadow (yes, I like pretty).

kovidgoyal:
The only reason to not have the title in the menubar/titlebar is if you fond it distracting. But if you have it in the tabbar, then it is just as distracting there. SO the reason does not apply. Using --title " " will remove it from everywhere. 

And if you want new tab titles set to " " use the kitty remote control feature to do it by add a call to set-title in you shell rc files.

Asheq:
FWIW, I would also like to disable text in the titlebar because I like to keep my UI as uncluttered as possible and without redundancy.

Putting `--title " "` inside `~/.config/kitty/macos-launch-services-cmdline` works for me, but still has the same limitations that @Luflosi described above. I would certainly appreciate a `macos_show_window_title` option that applies to all kitty windows.

Luflosi:
You can try my fork, it has the `hide_title` option. Or you can cherry-pick the commit where I implemented this feature.

kovidgoyal:
Sigh :) Well @Luflosi send a PR and I will merge it. But make the option control displaying titles in both the window and the menubar. That is rename the macos_show_window_title_in_menubar option to
macos_show_window_title_in with default value:

window_titlebar,global_menubar

and can be set to none for neither.

Luflosi:
In my fork, I implemented the option to work on other platforms too. Do you just want a macOS specific option?

kovidgoyal:
But does it actually work on other platforms. On my linux box if the window sets not title then the iwndow manager uses the program name as
the title.

Luflosi:
Setting the title to an empty string works for me.

Luflosi:
So should I make the pull request with an option that works on all platforms? I would need to rename the option to `show_window_title_in` since `macos_show_window_title_in` would no longer be fitting.

kovidgoyal:
> Setting the title to an empty string works for me.

That will depend on the individual window manager in question I guess

kovidgoyal:
> So should I make the pull request with an option that works on all platforms? I would need to rename the option to `show_window_title_in` since `macos_show_window_title_in` would no longer be fitting.

well given there is no menubar functionality on linux and the titlebar function may or may not work, I am inclined to leave it to macos only.

Luflosi:
@Asheq there is now a new config option `macos_show_window_title_in` in the latest master that allows you do hide the window title in addition to the menu bar title.

PR #1082, opened 2018-10-21
Do not send duplicated wakeups to main loop

blahgeek:
## Problem

On macOs, when a process produces lots of outputs (e.g. `seq 1 1000000000`), Kitty's refresh rate would go down significantly (down to ~1fps).

## Cause

The `wait_for_events()` function in main loop would take much more time when there's lots of outputs, to handle multiple wakeup events from IO thread.

## Solution

Only send single wakeup to main thread on each loop.

### DIFF
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -92,6 +92,10 @@ static int signal_fds[2], wakeup_fds[2];
 // before ticking over the main loop. Negative values mean wait forever.
 static double maximum_wait = -1.0;
 
+// Loop count of the main loop, increment in each iteration.
+// It's used to de-duplicate wakeups
+static long main_loop_id = 0;
+
 static inline void
 set_maximum_wait(double val) {
     if (val >= 0 && (val < maximum_wait || maximum_wait < 0)) maximum_wait = val;
@@ -773,11 +777,15 @@ main_loop(ChildMonitor *self, PyObject *a UNUSED) {
 #define main_loop_doc "The main thread loop"
     bool has_open_windows = true;
 
+    main_loop_id = 0;
     while (has_open_windows) {
         double now = monotonic();
         if (global_state.has_pending_resizes) process_pending_resizes(now);
         render(now);
         wait_for_events();
+        // increment loop id right after wait_for_events()
+        // so that new wakeups would arrive
+        main_loop_id += 1;
 #ifdef __APPLE__
         if (cocoa_pending_actions) {
             if (cocoa_pending_actions & PREFERENCES_WINDOW) { call_boss(edit_config_file, NULL); }
@@ -985,6 +993,7 @@ io_loop(void *data) {
     size_t i;
     int ret;
     bool has_more, data_received;
+    long last_wakeup_mainloop_id = -1;
     Screen *screen;
     ChildMonitor *self = (ChildMonitor*)data;
     set_thread_name("KittyChildMon");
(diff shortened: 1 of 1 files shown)


kovidgoyal:
incrementing and comparing integers is not atomic, so you cant really use it for cross-thread synchronization. And while you could possibly use atomic operations for integer increment and comparison, wait_for_events() and wakeup_main_loop() themselves are not atomic. which means you could end up missing wakeups if the timing is not right

kovidgoyal:
also as far as I know, wait_for_events() should wakeup just once even if there are multiple calls to wakeup_main_loop(). Is that not the case on macOS?

blahgeek:
Each call to wakeup_main_loop() dispatches an event to main loop and wait_for_events() would need to handle all of them, which takes a long time (~0.5 seconds in macOS while running `seq`, not sure about linux).

Atomic integer is not required here since only one thread would alter the value, the other thread just reads.

There's indeed small chances of missing wakeups (when a wakeup is to be sent between `wait_for_events()` and `main_loop_id += 1`), but there's always timeout for waiting.

I would try to work out another solution without causing missing wakeups if I got some free time.

kovidgoyal:
yeah but the event is an empty event, it does not actually require any handling, beyond reading it off the queue. I'm surprised that is expensive in cocoa. It might be possible to further optimize it in _glfwPlatformPollEvents in cocoa_window.m Check the event type and if it is NSEventTypeApplicationDefined dont call sendEvent with it.

blahgeek:
I tried to not send out NSEventTypeApplicationDefined events in _glfwPlatformPollEvents, it doesn't help much.

kovidgoyal:
Ok well I suppose we can use something along the lines of what you propose, the chances of delayed input processing are pretty low even without using mutexes, which might be expensive.

Luflosi:
Even if only one thread is writing and one thread is reading, without using mutexes there will still be data races and thus undefined behaviour. See https://en.wikipedia.org/wiki/Race_condition#Software

kovidgoyal:
An alternate approach is to make use of the fact that the main thread only parses input if input_delay time has passed. So there is no need for the IO loop to wakeup the main loop more often than that. I dont however know if the input delay is large enough to materially affect performance or not.

blahgeek:
Does this commit increases maxium input delay to be `input_delay * 2`?

kovidgoyal:
poll now takes a timeout. So the second poll call can take at most 1ms in your example. and why would maximum input delay double? just as poll now takes a timeout in the I/O loop, the main loop check for input delay also takes a timeout (sets maximum wait time).

blahgeek:
Ah, my bad, I though both IO loop and main loop (parse function) would wait for `input_delay`, I was wrong, they use new_input_at timestamp.

👍

blahgeek:
For the poll timeout, I would suggest using `ceil(time_delta)` to prevent unnecessary multiple timeouts.

kovidgoyal:
the poll timeout is wrong anyway right now, the value needs to be converted from seconds to milliseconds

kovidgoyal:
fixed and also used ceil

blahgeek:
> Does this commit increases maximum input delay to be `input_delay * 2`?

> Ah, my bad, I though both IO loop and main loop (parse function) would wait for `input_delay`, I was wrong, they use new_input_at timestamp.

Consider the following event sequences with input_delay = 100:

t=1: received some output, new_input_at = 1
t=100: send wakeup to main loop. Parsing function checks `now - new_input_at` which is still less than `input_delay`, do nothing
t=200: send another wakeup to mainloop. Parsing function now can do the work.

The input delay in this scenario is 199ms.

kovidgoyal:
Oh and not to mention that in your scenario above the main_loop would set max_wait_time to 1 ms and wakeup itself 1ms later

Issue #1088, opened 2018-10-23
Kitty does not respect the default system double-click interval


Luflosi:
I wrote some code that allows kitty to accept the special value `system` for `click_interval` instead of a number on macOS. My goal is to set the click_interval to the system default if the option is missing or set to `system`. Can I open a [WIP] pull request to implement that also for other platforms? I would need some help with that.

kovidgoyal:
sure feel free.

PR #1090, opened 2018-10-24
[WIP] Respect the default system double-click interval

Luflosi:
As discussed in #1088. The goal is to set the click_interval to the system default if the option is missing or set to system.
The current code will currently most likely break horribly on any platform that is not macOS.
Where should I put the platform specific code that gets the value from the system? glfw? If so, in which file does it fit best?

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1585,6 +1585,11 @@ void _glfwPlatformGetWindowContentScale(_GLFWwindow* window,
         *yscale = (float) (pixels.size.height / points.size.height);
 }
 
+double _glfwPlatformGetDoubleClickInterval(_GLFWwindow* window)
+{
+    return [NSEvent doubleClickInterval];
+}
+
 void _glfwPlatformIconifyWindow(_GLFWwindow* window)
 {
     [window->ns.object miniaturize:nil];
--- a/glfw/glfw3.h
+++ b/glfw/glfw3.h
@@ -2930,6 +2930,32 @@ GLFWAPI void glfwGetWindowFrameSize(GLFWwindow* window, int* left, int* top, int
  */
 GLFWAPI void glfwGetWindowContentScale(GLFWwindow* window, float* xscale, float* yscale);
 
+/*! @brief Returns the double click time interval.
+ *
+ *  This function returns the maximum time between clicks to count as a
+ *  double click.
+ *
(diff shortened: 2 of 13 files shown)


kovidgoyal:
the best place for this kind of thing is in glfw. Look at glfw/*_window.c

you will need to add new API to glfw to return the platform double click interval, which should go into glfw3.h. Once you add the declaration to that file run ./glfw/glfw.py to auto-update glfw-wrapper.h

To start with just have the functions for wayland and x11 return default values.

Luflosi:
Ok, I got that. Now what's the best way to use that new function? Do I somehow call it directly from the parser in python or do I just let the parser put a negative number into the `click_interval` variable in `state.h` and then overwrite it immediately using c code?

kovidgoyal:
use a negative number

kovidgoyal:
probably do it in create_os_window() for the first time the window is created, yu can be sure glfw is properly initialized then.

Luflosi:
Could you review the code please?

kovidgoyal:
Looks fine in general. I suggest changing the default in config_data.py for click_interval and adding some help text about negative values meaning to use system default. It would also be nice to figure out how to get this value from the system for X11 and wayland, but not critical.

kovidgoyal:
Oh and I meant change the default to -1

Luflosi:
Is the following description good or could it be improved?
The interval between successive clicks to detect double/triple clicks (in seconds). Negative numbers will use the system default instead (this only works on macOS currently, X11 and Wayland will use the default value of 0.5 in this case)

kovidgoyal:
Dont bother with mentioning where it does not work, just say, will use the system default, if available or fallback to 0.5. That way we dont have to remember ot change the description if someone later adds support for wayland/x11

Luflosi:
Like this?

Luflosi:
Any idea on how to do it in X11 and/or Wayland? I didn't find anything useful but I also know pretty much nothing about writing code for X11 and Wayland. I did find a function that should do the right thing on windows though.

kovidgoyal:
IIRC X11 uses the X resources system, you'd need to read the mouseClickTimes resource onthe window and the root window, but it is not widely used, so I wouldn't worry about it. No idea about wayland.

Luflosi:
Would be nice to have feature parity between the different platforms, maybe you could look into it when you have some time again and nothing more important to do

Issue #1106, opened 2018-10-29
The "macos_hide_titlebar" option should be combined with "x11_hide_window_decorations"


Luflosi:
The `macos_hide_titlebar` config option doesn't just hide the titlebar, it hides all the window decorations including the window title, the red, yellow and green buttons, the border around the window, the rounded corners and the window shadow. Therefore a name like `macos_hide_window_decorations` seems more appropriate. I assume `x11_hide_window_decorations` does a very similar thing. But then what's the point in having separate options for macOS and X11? Why not just call the option `hide_window_decorations` and have it do nothing for Wayland?

kovidgoyal:
I'm afraid rationalizing config option names is not worth the disruption it would cause, with people's config not working after updating. I am willing to break compat in kitty.conf if there is some actual new functionality/critical bug, but notsimply for the sake of rationalization.

Luflosi:
Would creating the new option and making the old options be aliases of the new option work? We could then deprecate the old options. One issue I can see with this approach is if the user has an option that does not apply to the current operating system as with this change the option will become active for their platform. Another option would be to leave the old options like they are (and deprecate them) and just add the new option. What do you think?

PR #1127, opened 2018-11-09
[WIP] Combine the "x11_hide_window_decorations" and "macos_hide_titlebar" options

Luflosi:
As discussed in #1106.
I'd like to only let the three options change the `hide_window_decorations` variable if they are different than the default. How could I do that? Currently the value will just be overwritten by the next option.
Are there any other things you'd like to be different?


kovidgoyal:
I'll review this when i have a bit more time.

Luflosi:
Your commit looks much nicer than mine, thank you!

PR #1128, opened 2018-11-09
Micro-optimisation

Luflosi:
The code in the `GA` define already has a check `if (ret == NULL) return NULL;`, therefore another check isn't needed, right?
I have two more little questions: Why do you sometimes use `Py_DECREF()` and sometimes `Py_CLEAR()`? And what are the curly braces in `#define S(name, convert) { GA(name); global_state.opts.name = convert(ret); Py_DECREF(ret); if (PyErr_Occurred()) return NULL; }` for?

### DIFF
--- a/kitty/state.c
+++ b/kitty/state.c
@@ -395,7 +395,7 @@ PYWRAP1(set_options) {
     S(macos_hide_from_tasks, PyObject_IsTrue);
     S(macos_thicken_font, PyFloat_AsDouble);
 
-    GA(tab_bar_style); if (!ret) return NULL;
+    GA(tab_bar_style);
     global_state.tab_bar_hidden = PyUnicode_CompareWithASCIIString(ret, "hidden") == 0 ? true: false;
     Py_CLEAR(ret);
     if (PyErr_Occurred()) return NULL;


Luflosi:
> Py_Clear sets the pointer being cleared to NULL and also handles a NULL input pointer, py_decref does neither, but is faster, for obvious reasons.

I'm just wondering, why you don't/can't use `Py_DECREF` in
```C
GA(kitty_mod);
kitty_mod = PyLong_AsLong(ret); Py_CLEAR(ret); if (PyErr_Occurred()) return NULL;
```
and
```C
GA(tab_bar_style);
global_state.tab_bar_hidden = PyUnicode_CompareWithASCIIString(ret, "hidden") == 0 ? true: false;
Py_CLEAR(ret);
if (PyErr_Occurred()) return NULL;
```
like you do in
```C
#define S(name, convert) { GA(name); global_state.opts.name = convert(ret); Py_DECREF(ret); if (PyErr_Occurred()) return NULL; }
```
> the curly braces simply create a new block, so any variable declarations inside the expanded macro do not leak into the outside scope.

I know what curly braces can do but as far as I can see, you don't declare any new variables in
```C
#define GA(name) ret = PyObject_GetAttrString(opts, #name); if (ret == NULL) return NULL;
#define S(name, convert) { GA(name); global_state.opts.name = convert(ret); Py_DECREF(ret); if (PyErr_Occurred()) return NULL; }
```
right?

kovidgoyal:
It's simply defensive codng. The curly braces are there because the macro refers to another macro, which may in the future add a variable declaration. The use of Py_Clear is also safer than py_decref and since this is not performance sensitive code (run only once) it is better to be safe.

Issue #1112, opened 2018-11-01
Scrolling sensitivity in macOS with retina display


michaelkix:
(I'm not sure this has anything to do with retina display.)

Scrolling speed/sensitivity seems low compared to other terminals - about half the expected one.

I use a locally modified version of kitty with this change in scroll_event() in mouse.c even though I'm not sure this is the correct way to do it:
```
-        double pixels = global_state.callback_os_window->pending_scroll_pixels + yoffset;
+        double pixels = global_state.callback_os_window->pending_scroll_pixels + yoffset * global_state.callback_os_window->viewport_y_ratio;
```

On my machine viewport_y_ratio is 2.0.

Unrelated: thanks for kitty - I like it very much and use it on a daily basis. Other terminals are way too slow for my liking.

Luflosi:
I think it has everything to do with the retina display. I don't have a retina display and for me `viewport_y_ratio` is 1.0. You can make a pull request if you want but if you do, put `global_state.callback_os_window->pending_scroll_pixels + yoffset` in parenthesis
 so that `global_state.callback_os_window->pending_scroll_pixels` is also multiplied by the factor.

michaelkix:
Wouldn't this lead to `pending_scroll_pixels` being multiplied by 2 twice?
The result of the computation is accumulated into `pending_scroll_pixels` if I understand correctly, everything these is already pre-multiplied.

Luflosi:
Yeah, you're right, my bad. I thought about it a bit more and I think the best thing to do is to multiply `yoffset` by `viewport_y_ratio` before it's used for the first time. @kovidgoyal should `yoffset` be multiplied by `viewport_y_ratio` even when `is_high_resolution` is false? @michaelkix depending on the answer the best place to put this multiplication is probably either before or after `if (is_high_resolution) {`. @kovidgoyal or should the multiplication be done outside of `scroll_event()` so that `scroll_event()` is called with the already multiplied value?

kovidgoyal:
I'll look at it when I have a moment

Luflosi:
What about Linux? X11 and Wayland also seem to be able to return a value from `_glfwPlatformGetWindowContentScale()` other than 1.

Luflosi:
Does this help reproduce it on linux for anyone: https://wiki.archlinux.org/index.php/HiDPI ?

maximbaz:
[...]

PR #1129, opened 2018-11-09
Make high resolution scrolling more responsive and configurable

maximbaz:
I was testing high resolution scrolling under Wayland and found that scrolling becomes acceptable if I multiply `yoffset` by `10` and awesome if I multiply it by `20`.

Since `wheel_scroll_multiplier` is not used under high resolution scrolling at all, I thought we could put it to good use here, so that responsiveness becomes configurable.

Under Wayland, my `viewport_y_ratio` is `2`. I added it to this patch because it was suggested by @Luflosi in #1112, if you want to keep it, https://github.com/kovidgoyal/kitty/commit/82f9aecacbc0f76ca9e7aec809fc8cf3fd73ff5a probably needs to be reverted.

With this formula, scrolling is acceptable with default settings and I can further improve it if I change `wheel_scroll_multiplier` to `10`.

If we remove `viewport_y_ratio` from the formula, the default value of `wheel_scroll_multiplier` (`5`) is too slow, and I would need to change it to at least `10`, but better to `20`.

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -356,6 +356,12 @@ precision scrolling devices, not for high precision scrolling on platforms such
 as macOS and Wayland. Use negative numbers to change scroll direction.'''))
 # }}}
 
+o('touch_scroll_multiplier', 1.0, long_text=_('''
+Modify the amount scrolled by a touchpad. Note this is only used for high
+precision scrolling devices on platforms such as macOS and Wayland.
+Use negative numbers to change scroll direction.'''))
+# }}}
+
 g('mouse')  # {{{
 
 o('url_color', '#0087BD', option_type=to_color, long_text=_('''
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -549,7 +549,7 @@ scroll_event(double UNUSED xoffset, double yoffset, int flags) {
     int s;
     bool is_high_resolution = flags & 1;
     if (is_high_resolution) {
-        yoffset *= global_state.callback_os_window->viewport_y_ratio * OPT(wheel_scroll_multiplier);
+        yoffset *= OPT(touch_scroll_multiplier);
         if (yoffset * global_state.callback_os_window->pending_scroll_pixels < 0) {
             global_state.callback_os_window->pending_scroll_pixels = 0;  // change of direction
         }
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -549,6 +549,7 @@ scroll_event(double UNUSED xoffset, double yoffset, int flags) {
     int s;
     bool is_high_resolution = flags & 1;
     if (is_high_resolution) {
+        yoffset *= global_state.callback_os_window->viewport_y_ratio * OPT(wheel_scroll_multiplier);
         if (yoffset * global_state.callback_os_window->pending_scroll_pixels < 0) {
             global_state.callback_os_window->pending_scroll_pixels = 0;  // change of direction
         }
--- a/kitty/state.c
+++ b/kitty/state.c
@@ -373,6 +373,7 @@ PYWRAP1(set_options) {
     S(tab_bar_edge, PyLong_AsLong);
     S(mouse_hide_wait, PyFloat_AsDouble);
     S(wheel_scroll_multiplier, PyFloat_AsDouble);
+    S(touch_scroll_multiplier, PyFloat_AsDouble);
     S(open_url_modifiers, convert_mods);
     S(rectangle_select_modifiers, convert_mods);
     S(click_interval, PyFloat_AsDouble);
--- a/kitty/state.h
+++ b/kitty/state.h
@@ -13,7 +13,7 @@
 typedef enum { LEFT_EDGE, TOP_EDGE, RIGHT_EDGE, BOTTOM_EDGE } Edge;
 
 typedef struct {
-    double visual_bell_duration, cursor_blink_interval, cursor_stop_blinking_after, mouse_hide_wait, click_interval, wheel_scroll_multiplier;
+    double visual_bell_duration, cursor_blink_interval, cursor_stop_blinking_after, mouse_hide_wait, click_interval, wheel_scroll_multiplier, touch_scroll_multiplier;
     bool enable_audio_bell;
     CursorShape cursor_shape;
     unsigned int open_url_modifiers;


Luflosi:
At least on macOS a value of 1 makes kitty scroll as fast as all other apps.

maximbaz:
Just to confirm, are you testing with https://github.com/kovidgoyal/kitty/commit/82f9aecacbc0f76ca9e7aec809fc8cf3fd73ff5a reverted? Because both that commit and this PR increase the speed on macOS.

In any case, I don't really care about the default value, I'm totally fine changing it in my local config as long as it's configurable 😉

Luflosi:
That commit doesn't make a difference for me since `yscale` is 1 for me since I don't have a retina display. And I didn't suggest changing the default value for everyone, only where high precision scrolling is supported. I think that 5 lines per scroll event are pretty reasonable when scrolling with a mouse that doesn't support high precision scrolling.

maximbaz:
Maybe it wasn't a good idea then to re-use `wheel_scroll_multiplier` option and we better have a different one for high precision scrolling...

Luflosi:
I think the semantics of that option make perfect sense for low and high precision scrolling and we should not create a new option.

maximbaz:
If the option will have different values based on whether the device is "high" or "low" precision, users would need to have a way to override _both_ these values, and this will be impossible I think...

In other words, I would like to override `wheel_scroll_multiplier` to `10` to get the speed as in sakura with touchpad, but I don't want to change the default `5` for when I use a mouse.

kovidgoyal:
It would be interesting to see what VTE does on wayland. Why is it scrolling faster than kitty? Does it multiply the pixels by something and if so where does the value of that something come from? @egmontkob you know the answer?

egmontkob:
GTK+ gives us a smooth scrolling delta which hides from us the X11 vs. Wayland differences, and is probably already scaled by some global GTK+ or GNOME setting (I'm not sure about it). I'm not aware of these details.

Then in VTE it's scaled according to ceil(height / 10.0) [here](https://gitlab.gnome.org/GNOME/vte/blob/0.54.2/src/vte.cc#L9872), which may not be a good approach as per [748012](https://bugzilla.gnome.org/show_bug.cgi?id=748012). See also [769696](https://bugzilla.gnome.org/show_bug.cgi?id=769696).

kovidgoyal:
@egmontkob thanks for the pointers

ducis:
Hello, I am wondering, in the current master branch, when the last line is updated (or a new line is added to the bottom), does kitty redraw the whole window? like if there is an "A" in the second line from the bottom, then the first line from the bottom changed, does kitty go through the process of rasterizing the letter "A" again?  I assume if it does not need to (by holding a buffer of some previously rasterized parts of the terminal window), both high-prec scrolling would be easy and less computation would be needed?

ducis:
So far as I know, a terminal program either is only able to change some lines from the bottom (normal terminal program, and tab completion) or takes control of the whole (virtual) terminal (like vim, tmux, or clear), it does not change lines from the top or in the middle. In the latter case you don't have scrolling anyway. So we can always assume a terminal program changes a certain number of lines from the bottom. Is it correct?

kovidgoyal:
kitty only ever rasterizes a character once, that's part of the magic
behind its performance. And no nay terminal program
can make changes to any part of the screen at any time, for example by
simply changing the background color, or directly by moving the cursor
around the screen.

ducis:
Thanks. Then where shall I start looking into the scrolling/rendering stuff? In particular, which variables in the source refer to the previously rasterized content?

kovidgoyal:
there is no previously rasterized content. rendering happens on the GPU.
If you want to get previously rendered content you have to render into a
framebuffer. There is code that does that for other reasons in shaders.c
but you are going to need familiarity with OpenGL to make it work

ducis:
So by "rasterizes a character once" you meant there are caches like many images with the size of individual characters in the graphics memory?

I guess by "does that for other reasons in shaders.c" you mean draw_cells_interleaved_premult().
Is this function triggered in, say, a fresh install of kitty without customized configuration file or passed commandline options? If not, how should I enable it so that I can test if it affects the latency first?

ducis:
Does the framebuffer to render into live in the graphics memory or the main memory? 
Usually for this kind of things it should live in the graphics memory and the application has pointer-like stuff with which the app can tell the GPU to do things like memcpy() purely in the graphics memory.
But I am not sure if you meant we have to fetch the rendered results into the main memory by "you have to render into a framebuffer", which would be much worse performance-wise.

kovidgoyal:
On Tue, Mar 10, 2020 at 08:07:16PM -0700, ducis. wrote:
> So by "rasterizes a character once" you meant there are caches like many images with the size of individual characters in the graphics memory?

There is a sprite map stored on the GPU.

> 
> I guess by "does that for other reasons in shaders.c" you mean draw_cells_interleaved_premult().
> Is this function triggered in, say, a fresh install of kitty without customized configuration file or passed commandline options? If not, how should I enable it so that I can test if it affects the latency first?

It will be used when you have a transparent window and images displayed
in the window.

kovidgoyal:
On Tue, Mar 10, 2020 at 08:20:13PM -0700, ducis. wrote:
> Does the framebuffer to render into live in the graphics memory or the main memory? 
> Usually for this kind of things it should live in the graphics memory and the application has pointer-like stuff with which the app can tell the GPU to do things like memcpy() purely in the graphics memory.
> But I am not sure if you meant we have to fetch the rendered results into the main memory by "you have to render into a framebuffer", which would be much worse performance-wise.

framebuffers live in GPU memory

ducis:
Hi, I just tested with typometer.
The environment is compiz (0.8.*, not the Ubuntu one) 60Hz frame rate globally, 
both delays turned to 0, VSync enabled in compiz but disabled in kitty (so no tearing).
Transparency within kitty roughly doubles the latency from ~17ms to ~31ms,
**with or without icatted images.**
But transparency with compiz plugins (which can make any window translucent) results in latency ~26ms.
Would you think that the 17ms->31ms is caused by going through draw_cells_interleaved_premult()?
Am I missing anything?

kovidgoyal:
Put a printf() in draw_cells_interleaved_premult() to check if it is going through that. IIRC without images it will not be used, see line 636 onwards of shaders.c for details. Either you need transparency + any image or non-transparent and negative z-index images.

ducis:
I put in the printf() and tested again. Yes, draw..premult() is only triggered by both background_opacity and icat.
How transparency affects latency really depended on what windows were below it.
It seems that the more complex stuff are below a translucent window the more likely the computation will take more than one frame.
But in general transparency,
as well as draw_cells_interleaved_premult() when triggered by icat, do not force an additional frame.

kovidgoyal:
You should be able to trigger it with just

kitty -o background_image=whatever.png

and no transparency (assuming you are running from master)

ducis:
No, I am running from 0.16.0 which does not seem to have background_image,
but I just forced draw_cells_interleaved_premult() at the end of draw_cells().
Somehow it made the font bolder and much uglier when background_opacity=1.
The speed was just as fast as unmodded whether background_opacity = 1 or not,
so long as there was nothing below the window to be actually blended.

Maybe there should be an option to explicitly force rendering to a framebuffer
regardless of transparency or whatever? 
So everyone can benchmark. After some time we may find out that 
we can use an intermediate framebuffer with negligible penalty in performance,
which would then enable more fancy stuff.

I am even using Intel graphics, an additional copy would be even more forgiving on a real GPU.

kovidgoyal:
Additional copies may have negligible performance penalty, in terms of
render time, but that's not my philosophy. My goal is to reduce energy
consumption, not just render time. That is why kitty has various
settings that actually insert latency into the render loop. And
performing unnecessary operations on all renders just for smooth
scrolling is not a worthwhile tradeoff.

I dont really see why that would be needed anyway. You could just switch
rendering to using a framebuffer while scrolling and turn it off again
after. It would mean writing a bit more code which you would have to do
anyway, since to actually implement scrolling you need to render stuff
above and below the window limits.

ducis:
It is hard to say without actual implementation and benchmarking, but keeping a large buffer of results painted in the last frame might actually reduce computation at the expense of more VRAM.

But whatever, what is the current situation of smooth scrolling in the master branch? 
I saw that you merged something from this issue and the Luflosi branch in
https://github.com/kovidgoyal/kitty/pull/1454,
but I couldn't figure out whether we can enable some smooth scrolling in the current master.

kovidgoyal:
there is no smooth scrolling https://github.com/kovidgoyal/kitty/pull/1454

Luflosi:
@ducis We can work on this together, if you like.

ducis:
@Luflosi 
Once I switch completely from tmux.

PR #1147, opened 2018-11-17
Fix set_background_opacity with increment/decrement

kdarkhan:
There is a logic error in `set_background_opacity` which breaks increment/decrement handling.
This commit fixes it.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -459,10 +459,10 @@ class Boss:
                     _('You must set the dynamic_background_opacity option in kitty.conf to be able to change background opacity'))
         os_window_id = window.os_window_id
         if opacity[0] in '+-':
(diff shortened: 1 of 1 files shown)


maximbaz:
Can you describe this logic error? It's not obvious to me even now 🙂

kdarkhan:
Sure. Putting this into my config
```map kitty_mod+9    set_background_opacity +0.1
map kitty_mod+8    set_background_opacity -0.1```
I would expect the first shortcut to increment, and second to decrement the current opacity.
However because
`opacity = background_opacity_of(os_window_id)`
overrides the passed in `opacity` parameter, and later
`opacity += float(opacity)` just doubles the effective opacity instead of incrementing/decrementing it.

PR #1179, opened 2018-11-29
Fix memory leak

Luflosi:
When the call to `realloc` in kitty/gl-wrapper.c fails, the pointer `exts_i` is overwritten, even though the original memory needs to be freed. I'm not sure wether you prefer the fix in the commit or this one:
```C
if (num_exts_i > 0) {
    char **tmp = NULL;
    tmp = (char **)realloc((void *)exts_i, (size_t)num_exts_i * (sizeof *exts_i));
    if (tmp == NULL) free(exts_i);
    exts_i = tmp;
}
if (exts_i == NULL) {
    return 0;
}
```
The fix in the commit doesn't check for null twice but this one might look nicer.
I found this using static analysis with cppcheck.

### DIFF
--- a/kitty/gl-wrapper.c
+++ b/kitty/gl-wrapper.c
@@ -173,9 +173,14 @@ static int get_exts(void) {
         num_exts_i = 0;
         glGetIntegerv(GL_NUM_EXTENSIONS, &num_exts_i);
         if (num_exts_i > 0) {
-            exts_i = (char **)realloc((void *)exts_i, (size_t)num_exts_i * (sizeof *exts_i));
-        }
-        if (exts_i == NULL) {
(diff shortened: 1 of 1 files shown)


kovidgoyal:
It's a leak only if that function is called more than once, which it isn't. It is called only from init_glad which is called only once. Also gl-wrapper.c is generated code, so changing it directly wont work, you'd have to patch it in glad/generate.py

Luflosi:
Seems like this is a bug in glad, which was "fixed" in https://github.com/Dav1dde/glad/pull/178 but they didn't really fix it.

kovidgoyal:
well ping me whenever they fix it properly and I will regenerate gl-wrapper.c

PR #1182, opened 2018-11-30
parser: accept `reset` sequence to exit accumulate functions

martinetd:
Input looks stuck until the accumulate buffer fills up; after displaying
random bytes (e.g. cat a binary) nothing would print anymore but input
would still work.

In that situation, one would normally expect the `reset` command to "fix
things" - so getting out of the accumulate loop with this sequence would
make sense.

I'd actually be tempted to bail out even more easily (e.g. when getting
color codes as well, so a colorful PS1 prompt would already get out of
the loop) but it might not really go well with the logic behind these
accumulating sequences.

You can try with `echo -e '\xc2\x9e'` for PM

(re-reading this all now, I guess it is subjective; but I don't think it'd hurt anyone)

### DIFF
--- a/kitty/parser.c
+++ b/kitty/parser.c
@@ -905,6 +905,7 @@ accumulate_osc(Screen *screen, uint32_t ch, PyObject DUMP_UNUSED *dump_callback)
         case DEL:
             break;
         case ESC_ST:
+        case ESC_RIS:
             if (screen->parser_buf_pos > 0 && screen->parser_buf[screen->parser_buf_pos - 1] == ESC) {
                 screen->parser_buf_pos--;
                 return true;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
ESC_RIS is 'c' c is an allowed character inside OSC and OTH sequences, you cannot abort on receiving a c. That would break handling of escape codes that want to use c

kovidgoyal:
never mind, I see you are aborting on ESC + c not C alone.

kovidgoyal:
still esc + c is also an allowed sequence inside OSC/OTH escape codes. They are basically defined as any sequence of bytes until an ST, BEL, NULL or ESC+\, so the same objection still applies. I dont know of any actual escape codes that use esc+c but, that might well change in the future. I dont think we should be unilaterally changing the semantics of these escape codes.

The proper fix for this is to change the rs1 sequence in the terminfo file, to emit an empty OSC code before \Ec that will automatically close any ongoing accumulate. reset is supposed to use rs1 from terminfo, so that should work.

martinetd:
Ok, will look at updating the terminfo file - I did think of 'c' but figured esc+c was unlikely enough, but didn't go as far as checking.

PR #1184, opened 2018-11-30
Ignore mouse modes option

martinetd:
This adds a `mouse_modes` option defaulting to yes, which when set two no:
 - disables entering the various mouse modes when receiving the appropriate sequences
 - scrolls the main scrollback buffer even in alt screen mode (mouse wheel no longer does anything within any program, so might as well map it to something slightly more useful)

In theory the terminfo should be changed to just say these aren't supported, but in practice this just works: applications think the mode is supported and send the switch sequence but it's just ignored, and since mode isn't actually changed in kitty we never send the mouse event sequences back on mouse activity.
I guess some app might behave differently if they know it's not supported but I don't think I've seen any like that.

Happy to discuss the option name or whatever else, I just can't get used to in-app mouse events especially when it messes with copying into selection buffer - I'm sure there's a modifier you can press to work around that but I don't want to (and I know I could spend time figuring out how to configure every program that support these, but there are way too many on way too many servers and some don't have obvious disabling options).
Speaking of that modifier I guess I'd also be happy inverting the meaning of the modifier (no modifier = dumb mouse, modifier = mouse event), but there is no value for me here.

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -419,6 +419,9 @@ o('focus_follows_mouse', False, long_text=_('''
 Set the active window to the window under the mouse when
 moving the mouse around'''))
 
+o('mouse_modes', True, long_text=_('''
+Ignore terminal mouse modes if set to no'''))
+
 # }}}
 
 g('performance')  # {{{
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -606,8 +606,10 @@ scroll_event(double UNUSED xoffset, double yoffset, int flags) {
                     write_escape_code_to_child(screen, CSI, mouse_event_buf);
                 }
             }
-        } else {
+        } else if (OPT(mouse_modes)) {
             fake_scroll(abs(s), upwards);
+        } else {
+            screen_history_scroll(screen, abs(s), upwards);
         }
     }
 }
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -604,7 +604,7 @@ set_mode_from_const(Screen *self, unsigned int mode, bool val) {
 
 #define MOUSE_MODE(name, attr, value) \
     case name: \
-        self->modes.attr = val ? value : 0; break;
+        if (OPT(mouse_modes)) self->modes.attr = val ? value : 0; break;
 
     bool private;
     switch(mode) {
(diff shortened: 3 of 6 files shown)


kovidgoyal:
I'm not in favor of this. You can simply press Shift (this is standard across many terminal emulators) to select with the mouse ignoring application mouse handling. Adding this option means applications now cannot rely on mouse handling in the terminal.

kovidgoyal:
I'd really rather not do this, sorry. Historically, mouse support in terminals has been a royal pain. There are four different incompatible protocols for reporting mouse events. There is no easy way to detect if a terminal supports one of the other of these protocols. If you look at, for example, :help ttymouse in vim you will get an idea of the complexity around this. I dont want to add to that complexity by creating a terminal where mouse events may or may not work based on runtime config.

It is true that most (all?) current terminal programs are designed to work equally well with ot without mouse support, but that is an artifcat of the historically terrible mouse support terminals have had. I dont want to freeze the entire ecosystem in that paradigm.

martinetd:
Ok, np - I'll keep that as a local patch, won't be the first or last program where I do that :)

While we're on royal pains that I like to have differently, do you feel the same about alt screen?
I'd like to add an option that basically ignores alt screen and does everything in main screen - there are various use cases (like `vim -t some_common_symbol` with cscope, will blurt out a huge list that easily spills off screen, so having it in scrollback buffer lets me search with pager instead of.. . not being able to do anything with it), but basically I just like being able to run some command, open a file and be able to scroll back a bit and look at the output of the previous command. (not helping: that's an option in urxvt, `urxvt*secondaryScreen: false` in .Xdefaults)

Anyway, long story short - I also have a local patch that does this and works well enough for me (I scroll the current screen up to not overwrite text when entering alt mode and do not do anything when exiting; there might be quirks I haven't noticed so far), but haven't made this into an option yet.
I guess it's not worth spending time on that either?

kovidgoyal:
I just added the feature to pipe a little while ago that allows you to see the hidden screen with scrollback ina  new window.

martinetd:
Yeah, I noticed that but it's not as good - to go back to my `vim -t` example the text I want to see is stuff that scrolled up out of the alt buffer, so it's not going in any scrollback history at all right now.
There are cases where I want the main buffer, and this option would likely be good enough (maybe not as comfortable as just scrolling one page up, but it'd work); but I don't see any way of accessing obsolete alt buffer data: either stuff from within vim/whatever before I exited it, or scrolled out stuff while it is being accessed.

kovidgoyal:
yeah I'm afraid this is one for your local patch queue as well :)

PR #1185, opened 2018-11-30
ibus: deactivate overlay line on draw (unless drawing it)

martinetd:
When entering text with ibus, a commit that is handled asynchronously can be processed after the next preedit overlay is displayed.
That needs the peredit overlay hidden then shown again

This commit isn't optimal: it doesn't really do the "then shown again" part, so typing something with that overlay in a constantly moving buffer (e.g. irssi with a clock ticking for example) will just hide the buffer whenever the clock ticks and only print it again when one continues typing.
I'd need to remember the overlay line content at the start of `screen_draw` and redisplay it at the end... But the overlay line content isn't really kept around; currently that'd mean scanning the overlay positions to reconstruct the utf8 text and it's more than meh.
If anything, I'd prefer to store the utf8 text in a variable in the screen and use that for redrawing, but I'm not actually sure if that'd work well in practice (what about cursor movements? I'm afraid it'd just draw the overlay buffer over something else just after the clock update, and not where it should really be)

It's still a lot better than messing the screen up, so I've been using this for a while now and in practice it's not so bad, but I don't know how you feel about partial fixes.

### DIFF
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -377,6 +377,7 @@ draw_combining_char(Screen *self, char_type ch) {
 
 void
 screen_draw(Screen *self, uint32_t och) {
+    if (self->overlay_line.is_active) deactivate_overlay_line(self);
     if (is_ignored_char(och)) return;
     uint32_t ch = och < 256 ? self->g_charset[och] : och;
     bool is_cc = is_combining_char(ch);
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I think you can save the utf8_text (and the cursor position at the time it was drawn) and use it to redraw the overlay. Restore the cursor position to whatever it was before redrawing. Note that you willhave to strdup the utf8_text.

kovidgoyal:
screen_draw is not called for cursor movement, only actual text.

page-down:
Due to recent IME-related changes, is this PR no longer needed?

PR #1186, opened 2018-11-30
glfw xkb: debounce compile_keymap

martinetd:
Only allow reloading keymap once every 2 seconds at most.

Some programs like ibus generate a lot of XkbMapNotify events, which can be very slow with multiple kitty windows hammering the X server

(unrelated: that's my last PR for today, I have a few more commits that need cleanup at some point, but will try to force myself to finish selection first)

### DIFF
--- a/glfw/xkb_glfw.c
+++ b/glfw/xkb_glfw.c
@@ -261,6 +261,11 @@ load_compose_tables(_GLFWXKBData *xkb) {
 GLFWbool
 glfw_xkb_compile_keymap(_GLFWXKBData *xkb, const char *map_str) {
     const char *err;
+    double time = glfwGetTime();
+    /* skip back-to-back keymap reloads */
+    if (xkb->reload_time != 0.0 && time - xkb->reload_time < 2.0)
+        return GLFW_TRUE;
+    xkb->reload_time = time;
     release_keyboard_data(xkb);
     err = load_keymaps(xkb, map_str);
     if (err) {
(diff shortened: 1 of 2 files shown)


kovidgoyal:
but what if there are two genuine keyboard layout changes within 2 secs? kitty will ignore the second one and therefore break. If you do want to implement debouncing, it has to be genuine debouncing, which means that there has to be a call to compile_keymap *after* the last received mapnotify event. You can probably use the timer infrastructure I added to the glfw linux backends to do this. Look for addTimer/removeTimer

martinetd:
Can't wait that long or the experience will also be sluggish, and would need a corner case for the initial loading.

If you want to be correct (which is something I can agree with, there's a reason I didn't submit this bunch of patches earlier :D), I can run it once, then if it comes a second time within 2s fire a timer, and run it a second time at the end of that timer?

kovidgoyal:
Hmm does normal layout changing create only a single MapNotify event, or multiple events? Is it only ibus that creates so many events, and how do other programs handle this? Since MapNotify is delivered to all (visible?) windows, there should be a lot of traffic with any programs, not just kitty.

kovidgoyal:
weird, maybe try reporting this behavior to ibus, it looks like a bug, a 147! changes, surely they can be batched up into one or at most a few changes.

martinetd:
It's really weird, firefox doesn't seem to get this many events. I'm having trouble getting the intercepted window to actually appear on screen so I'm not sure, but events show up anyway and an intercepted firefox got one or two map events... where an intercepted kitty got 21.

Meanwhile, I found an ibus setting ("use system keyboard layout" in advanced) that makes it not send any xkbMapNotify event at all, so I guess that can work -- but even something like `setxkbmap us` triggers 7 events on a single kitty window so I think we're doing something wrong.
I'll try to figure out why we receive 7 events with setxkbmap, that doesn't seem healthy

kovidgoyal:
maybe there is some response the X server expects that we are not making.

PR #1188, opened 2018-11-30
reset terminfo: add an empty OSC sequence to rs1 string

martinetd:
In OSC/OTH mode, the parser accumulate bytes until it fills up or
encounters some end-of-osc sequence.
After entering in either of these mode by mistake, for example by
displaying random bytes (e.g. cat a binary) nothing would print anymore
but input would still work.

In that situation, being able to hit `reset` and recover a usable
terminal helps.

You can try with `echo -e '\xc2\x9e'` for PM

----

That works. As you said I need to output a full OSC sequence, e.g. enter one in case I wasn't in one earlier, to not get a parser warning.


kovidgoyal:
can you add a comment to terminfo.py explaining the need for the empty OSC code, since it is unusual.

martinetd:
force pushed: `# Reset string1 (empty OSC sequence to exit OSC/OTH modes, and regular reset)`

Issue #1190, opened 2018-11-30
fontconfig: does not find bold/italic for font with multiple family names


blueyed:
`fc-query Ligaiosevka-term-regular.ttf` displays `fc-family: "Iosevka Term"(s) "Liga Iosevka Term"(s)`, which I interpret as mutliple families.

In kitty.conf I am using `font_family Liga Iosevka Term`, i.e. the second name.

Kitty loads the font properly, but then fails to find the bold etc variants for it:
```
[334 23:07:53.043393] normal face: /home/user/.fonts/Ligaiosevka-term-regular.ttf:0
[334 23:07:53.043475] bold face: /home/user/.fonts/Ligaiosevka-term-regular.ttf:0
[334 23:07:53.043554] italic face: /home/user/.fonts/Ligaiosevka-term-regular.ttf:0
[334 23:07:53.043625] bi face: /home/user/.fonts/Ligaiosevka-term-regular.ttf:0
[334 23:07:53.043703] Symbol map faces:
[334 23:07:53.043772] /home/user/.fonts/Hack Regular Nerd Font Complete.ttf:0
[334 23:07:53.043842] /usr/share/fonts/TTF/DejaVuSansMono.ttf:0
```

It uses `Ligaiosevka-term-regular.ttf` for all of them.

From a quick look this appears to be kitty not finding/handling the "Liga
Iosevka Term" family.

```
> …/Vcs/kitty/kitty/fonts/fontconfig.py(36)

  34     @lru_cache()                                                                                                                                                                         
  35     def all_fonts_map(monospaced=True):                                                                                                                                                  
  36  ->     return create_font_map(fc_list(monospaced))                                                                                                                                      
(Pdb++) [x for x in l if x['path'] == '/home/user/.fonts/Ligaiosevka-term-bold.ttf']
[{'path': '/home/user/.fonts/Ligaiosevka-term-bold.ttf', 'family': 'Iosevka Term', 'style': 'Bold', 'full_name': 'Liga Iosevka Term Bold', 'postscript_name': 'LigaIosevka-Term-Bold', 'weight': 200, 'slant': 0, 'index': 0, 'scalable': True, 'outline': True, 'color': False, 'spacing': 'MONO'}]
```

Relevant C code: https://github.com/kovidgoyal/kitty/blob/819ee94a70163be5317be09660907dd01d709a97/kitty/fontconfig.c#L39-L44

[...]

PR #1191, opened 2018-11-30
fontconfig: fix find_best_match for multiple font families

blueyed:
Fixes https://github.com/kovidgoyal/kitty/issues/1190.

Just a request for comments - maybe it should be added to some cache?!


kovidgoyal:
this approach isn't right, discuss further on the issue.

PR #1245, opened 2018-12-17
Add `resize_window` configuration action

eepp:
This patch adds the `resize_window` configuration action, which has the following arguments:

1. Quality of resize amongst `wider`, `narrower`, `taller`, `shorter` (mandatory).
2. Increment in number of cells (optional, default: 1).

This makes it possible to configure keys as such:

    map ctrl+shift+left resize_window narrower
    map ctrl+shift+right resize_window wider
    map ctrl+shift+up resize_window taller
    map ctrl+shift+down resize_window shorter

and have a behaviour which is somewhat close to Terminator's.

Should you accept this idea, I can add documentation in a subsequent patch.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -223,6 +223,22 @@ example, in the Tall layout you can make the first window wider/narrower, but
 not taller/shorter. Note that what you are resizing is actually not a window,
 but a row/column in the layout, all windows in that row/column will be resized.
 
+You can define shortcuts in :file:`kitty.conf` to make the active window
+wider, narrower, taller, or shorter by mapping to the ``resize_window``
+action, for example::
+
+   map ctrl+left resize_window narrower
+   map ctrl+right resize_window wider
+   map ctrl+up resize_window taller
+   map shift+down move_window shorter
+   ...
+
+The ``resize_window`` action has a second, optional argument to control
+the resizing increment (a positive integer)::
+
+   map ctrl+up resize_window taller 3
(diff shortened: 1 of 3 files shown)


kovidgoyal:
is there some reason you dont like the existing facility for resizing?

eepp:
> is there some reason you dont like the existing facility for resizing?

Yes: I need to hit Ctrl+Shift+R, wait the 250+ ms this Python app takes to load, then use W/N/T/S, keys that are unrelated and scattered, then hit Escape.

Meanwhile I can just hit Ctrl+Shift+(Left/Right/Up/Down) in Terminator to quickly resize the current terminal instantaneously. And so can I in kitty with this patch now.

kovidgoyal:
Strange, is python particularly slow on you rmachine? On my four year old laptop, running this kitten takes less than 100ms which is perceptually almost instantaneous.

But OK, I'm fine with adding this action, please add a patch for the docs and ping me to review.

eepp:
Where should I document this? Because on https://sw.kovidgoyal.net/kitty/conf.html, it looks like only commands with a default keyboard shortcut are documented. For example, `neighboring_window` and `nth_window` are not there, but they are mentioned on https://sw.kovidgoyal.net/kitty/.

kovidgoyal:
The best place would be the layouts section in index.rst where it talks about resizing windows.

eepp:
Done.

Issue #1242, opened 2018-12-16
new_os_window outside kitty shortcuts


ggsalas:
Hi, I would like to know if there is a way to create a new kitty window while I'm not using kitty. 

The usual approach on macos is click the app icon on the dock and select "new window" but I see not option for that: 
![image](https://user-images.githubusercontent.com/8403993/50059654-7ef93a80-0168-11e9-9808-6666888331fe.png)

Alternative: Can I add a global shortcode to open a new kitty window? 

Thanks

Luflosi:
Yeah, that's not implemented currently. But you should be able to click on the kitty icon and use cmd+n to open a new window.

ggsalas:
Oh, thanks for the clarification. I was trying to figure out a way to open a new window without need to go to the kitty app. I can wait then to the new feature. Thanks

kovidgoyal:
Not sure what you are asking for, an item in the dock menu? You can use kitty --single-instance then running kitty --single-instance from any terminal will open a new os window in the existing kitty instance, automatically.

ggsalas:
Hi @kovidgoyal let me explain better:
At this moment I'm using macos and I use many desktops. If I'm at desktop 3 and I need a new kitty window I need to:
1- go to kitty app (on desktop 1, for example)
2- create a new kitty window
3- move the kitty window to the desktop 3

My doubt is if currently there is a quicker way to do that.
Thanks

Luflosi:
Or, while in desktop 3, click the kitty icon, press and hold the command key and press the n key. That should be easier than `kitty --single-instance`.

Luflosi:
Does it switch to desktop 1 right when you click the icon? We might have some settings set differently. I have the second checkbox in mission control not set. That option translates to something like "switch to space containing program when switching to it". @kovidgoyal With that option set I think it's currently impossible to do what @ggsalas would like to do. Do you know how to add an option to open a new window in that context menu?
This is what it looks like for my text editor:
![screenshot](https://user-images.githubusercontent.com/15217907/50112689-2f863d80-0240-11e9-88cb-e58a8f52caf3.png)

jacwah:
I have disabled the mission control setting. The switch occurs when the window is opened. After noticing that it only happened with the app version, I realized that Options->Assign To This Desktop was set, causing all windows to be opened in desktop 1. Stupid me. Maybe the same for @ggsalas ?

[...]

PR #1248, opened 2018-12-17
macOS: Add "New OS Window" to dock menu

jacwah:
screen shot

This is my first ever with Objective-C, any feedback welcome. I had to do some silly runtime stuff to to inject the `applicationDockMenu` method, but it works.

Discussed in #1242.


kovidgoyal:
Looks ok to me, except that dockMenu leaks on application exit. I have added a cleanup function for that.

PR #1252, opened 2018-12-19
`change_font_size current ...` changes the current OS window's font size

eepp:
When I first read this in the docs:

> To setup shortcuts to change only the current window’s font size:
> `map kitty_mod+f6 change_font_size current 10.0`

I thought _current window_ meant the current layout's window (the window that you create with Ctrl+Shift+Enter), in that it would resize only this window's font size. Terminator has this feature, hence the confusion for me:

image

I thought this was a bug in kitty, but after reading the code I can see that it's meant to resize the whole OS window's font size:

```
        if all_windows:
            current_global_size = global_font_size()
            new_size = calc_new_size(current_global_size)
            if new_size != current_global_size:
                global_font_size(new_size)
            os_windows = tuple(self.os_window_map.keys())
        else:
            os_windows = []
            w = self.active_window
            if w is not None:
                os_windows.append(w.os_window_id)
```

Therefore I am suggesting to make it more explicit in the documentation.

Would changing a single "subwindow"'s font size be possible in kitty considering the way it's designed?

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -169,14 +169,14 @@ Similarly, to switch back to the previous layout::
 ''')],
     'shortcuts.fonts': [
         _('Font sizes'), _('''\
-You can change the font size for all top-level kitty windows at a time
+You can change the font size for all top-level kitty OS windows at a time
 or only the current one.
 '''), _('''\
 To setup shortcuts for specific font sizes::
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
No font sizes can only be changed for top level OS windows, not kitty
windows.

PR #1270, opened 2018-12-26
Document switch cases of momentum_data

Luflosi:
I thought the code would be a bit more understandable with these comments documenting, what each case means.


Luflosi:
We could also just use an enum.

Issue #1005, opened 2018-09-22
Setting background_opacity to 0.0 does not make the area around characters transparent


Luflosi:
I was trying to implement a transparent background blur on macOS using `NSVisualEffectView`. For this to work, kitty needs to have a transparent background. When setting `background_opacity 0.0` in `kitty.conf` the text looks very bad:
![text transparent bright bg](https://user-images.githubusercontent.com/15217907/45916073-2d3e5e00-be60-11e8-9608-74c2267b6d11.png)
![text transparent dark bg](https://user-images.githubusercontent.com/15217907/45916071-257eb980-be60-11e8-8eaa-99c5047ece98.png)
This is roughly how I would expect it to look:
![expected](https://user-images.githubusercontent.com/15217907/45916084-67a7fb00-be60-11e8-8b35-6368885739d6.png)
Am I missing something or is this a limitation of the current text rendering implementation?

kovidgoyal:
I dont have time to look at the code right now, but I suspect that is because the text anti-aliasing does not work with background opacity of 0. I suggest instead using a low but non-zero value.

Luflosi:
The text anti-aliasing seems to get worse as the `background_opacity` approaches 0. I stacked five kitty instances with one window each on top of each other. Every kitty instance only has `background_opacity` set in `kitty.conf`. The value of `background_opacity` of the lowest (leftmost) window is `1.0` and the background is completely black. Every other window has at least some transparency, through which the black color of the lowest window is visible.
![blocky text](https://user-images.githubusercontent.com/15217907/45917458-e8bebc80-be77-11e8-88ba-fbdf80259af4.png)
As you may be able to see every window except the one with a value of `0.0` has at least some anti-aliasing going on although the text seems to expand in thickness as the opacity approaches 0 and an opacity value of 0.01 looks almost as bad as a value of 0:
0.01:
![0 01](https://user-images.githubusercontent.com/15217907/45917535-a1d1c680-be79-11e8-9369-bbf8f83369ab.png)
0.0:
![0 0](https://user-images.githubusercontent.com/15217907/45917536-a9916b00-be79-11e8-92c7-a3c93421769b.png)

kovidgoyal:
I suspect the problem is that anti-aliasing works by mixing the foreground and background colors using a mixing function that is different than the mixing function used by the OS to blend pixels in the window buffer. I can' t think of a good way around that.

Luflosi:
[...]

Issue #1221, opened 2018-12-10
Margin different color with transparency


RosalesJ:
## Issue
The margin displays as a lighter color than the background when background_opacity < 1.

![shot](https://user-images.githubusercontent.com/17154302/49759204-02112100-fc8f-11e8-819a-f3147ab79e54.png)

## System
macOS Mojave 10.14.2
kitty 0.13.1

Versions affected: 0.13.1, 0.13.0

## Replication
Minimal configuration in ```kitty.conf```:
```sh
background_opacity 0.85
background #002b36
window_margin_width 10
```

Luflosi:
I noticed this too, also without the `window_margin_width` or `background` config options. There are three screenshots of this happening in https://github.com/kovidgoyal/kitty/issues/1005#issuecomment-438880434

PR #1291, opened 2019-01-07
Fix white background color not allowing transparency

### LINKED ISSUE #1285: Semi-transparency does not work with colors other than black.
I am trying to use a white background with 50% opacity, but it only becomes transparent when I configure it to 0 opacity.

Luflosi:
Fixes #1285.
I played around with the cell_fragment shader code some more and while I don't really understand what's going on, this patch fixes the issue (while introducing a slight problem, which you should look at, I'll explain it later). This patch essentially undoes 074614bc8e940725939aa7a316f0a7086e01e05c and 96229becd1bef8bd69b412e61fac96146300646e except that it multiplies except of dividing. `final_color`, the result of the computation seems to need pre-multiplication and `final_color` before my new multiplication seems to not be pre-multiplied. Otherwise I can't explain why this patch works.
Here are some screenshots with various `background_opacity` settings and `background #ffffff` and `background #000000` colors in `kitty.conf` with and without the fix in front of a checkerboard pattern:

- Without the fix
  - `background #000000`
    - `background_opacity 1.0`
black 1 0
    - `background_opacity 0.5`
black 0 5
    - `background_opacity 0.0`
black 0 0
  - `background #ffffff`
    - `background_opacity 1.0`
white 1 0
    - `background_opacity 0.5`
white 0 5
    - `background_opacity 0.0`
white 0 0
- With the fix
  - `background #000000`
    - `background_opacity 1.0`
black 1 0
    - `background_opacity 0.5`
black 0 5
    - `background_opacity 0.0`
black 0 0
  - `background #ffffff`
    - `background_opacity 1.0`
white 1 0
    - `background_opacity 0.5`
white 0 5
    - `background_opacity 0.0`
white 0 0

Here is the problem: If you look closely at the bottom-right side of the left triangle in the bottom screenshot (a little less noticeable in the screenshot above that), you can see, that anti-aliasing is creating a few while-ish pixels where there should only be different mixes of blue and transparent ones, letting the black background shine through. That also happens with the other triangle but that isn't visible because of the white color behind the kitty window in that spot. I also think this is happening with the text but I'm not sure. That would mean, that #1005 is only partially fixed.

If my theory at the top with the pre-multiplication is correct, that means, that `alpha_blend_premul()` is taking the pre-multiplied (according to one comment in the code) color from `calculate_foreground()` and converting it to non-pre-multiplied color. I furthermore assume, that the `vec4` datatype uses values 0-255 for the color information. That would explain the weird white pixels because the back and forth calculation looses some color information.

There are also still some other issues with transparency like the weird borders with white background color but that should be looked at another time IMHO since I find #1005 and #1285 being fixed more important for now.

Wow, this pull request took several hours to create with all the testing and screenshots and the text and stuff 😅.

### DIFF
--- a/kitty/border_fragment.glsl
+++ b/kitty/border_fragment.glsl
@@ -4,5 +4,5 @@ in vec3 color;
 out vec4 final_color;
 
 void main() {
-    final_color = vec4(color, background_opacity);
+    final_color = vec4(color * background_opacity, background_opacity);
 }
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -101,7 +101,7 @@ void main() {
 #else
     // SIMPLE
 #ifdef TRANSPARENT
-    final_color = alpha_blend_premul(fg.rgb, fg.a, background, bg_alpha);
+    final_color = alpha_blend_premul(fg.rgb, fg.a, background * bg_alpha, bg_alpha);
     final_color = vec4(final_color.rgb, final_color.a);
 #else
     // since background alpha is 1.0, it is effectively pre-multiplied


Luflosi:
Wait, don't merge this yet, maybe I found a different solution:
```diff
diff --git a/kitty/cell_fragment.glsl b/kitty/cell_fragment.glsl
index 971b2881..504ea9ec 100644
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -101,7 +101,7 @@ void main() {
 #else
     // SIMPLE
 #ifdef TRANSPARENT
-    final_color = alpha_blend_premul(fg.rgb, fg.a, background, bg_alpha);
+    final_color = alpha_blend_premul(fg.rgb, fg.a, background * bg_alpha, bg_alpha);
     final_color = vec4(final_color.rgb, final_color.a);
 #else
     // since background alpha is 1.0, it is effectively pre-multiplied
```
I don't think `background` is pre-multiplied, so I just did that and everything I tested so far seems to work fine. I'll make a new commit tomorrow if you don't mind and create another bunch of screenshots if you like.

kovidgoyal:
That would break #1221

kovidgoyal:
vec4 is a floating point type, so stores alpha as number between 0 and 1.

Luflosi:
Can #1221 be fixed in a different way? I think the way you fixed it was wrong.

Luflosi:
I did that, can you please review and merge if everything is good?

PR #1341, opened 2019-01-28
Rename OS X -> macOS

Luflosi:
I renamed some instances of `OS X` and `osx` to `macOS`, since that is the correct name since macOS 10.12 Sierra, which was released in September of 2016. `setup.py` can be called with `osx-bundle` as an argument. I added a new argument `macos-bundle` but kept the old one for backwards compatibility since I don't want to break anything. Should I keep that option in the list of `choices`?
Some instances of `osx` remain, mainly in `publish.py`, `.circleci/config.yml` and both in the filename and content of `update-on-ox` but I didn't change those because I'm too afraid, that that could break something.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,7 @@ jobs:
           - run: python3 logo/make.py
           - run: python3 setup.py build --debug --verbose
           - run: python3 test.py
-          - run: python3 setup.py osx-bundle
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Leave the remaining osx mentions alone, since they integrate with my build system, which I am not going to change anytime soon.

PR #1362, opened 2019-02-05
Respect $SHELL environment variable

bennofs:
According to `login(1)`:

The value for $HOME, $USER, $SHELL, $PATH, $LOGNAME, and $MAIL are set according to the appropriate  fields  in
the password entry.

Prefering `$SHELL` to the entry read from the password database makes it easy
for users to override the interactive shell without changing their login shell.

If you use a non-bash-compatible shell like "fish" as your shell, setting it as the login
shell might break some programs while just overriding SHELL can be done in a controlled way
and nicely scoped to only programs launched from the window manager for example (whereas setting the login shell is a somewhat global option that is not so easily changed for just a few programs).

### DIFF
--- a/kitty/constants.py
+++ b/kitty/constants.py
@@ -119,7 +119,7 @@ logo_data_file = os.path.join(base_dir, 'logo', 'kitty.rgba')
 logo_png_file = os.path.join(base_dir, 'logo', 'kitty.png')
 beam_cursor_data_file = os.path.join(base_dir, 'logo', 'beam-cursor.png')
 try:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
kitty already has a shell config option that makes it easy to override the shell it uses. As far as I know the SHELL env var is not in any standard, so I dont think kitty should support it. Not to mention that since kitty reads its env vars from the system environment, the variable would have to be set in system or session rc files, which I dont think is significantly more convenient that setting it in kitty.conf

Issue #1379, opened 2019-02-13
On macOS closing a window should not switch to another workspace


aequitas:
The behaviour of Kitty when closing windows in workspaces is different from other applications on macOS.

When using workspaces on macOS. If a Kitty terminal window is open in one workspace. Closing the last Kitty window in another workspace should not shift focus to the window in the other workspace (and thereby switching to that workspace). But rather stay on the same workspace.

Also Kitty seems to focus on the previous window when closing a window, disregarding which workspace the window is in. So closing a window in a workspace which still has other Kitty windows open also causes a switch of workspace.

If you like I can make a video to clarify what I mean.

Luflosi:
I can reproduce this issue and I agree it should be fixed.
If I remove the two lines from `_glfwPlatformFocusWindow()` from `glfw/cocoa_window.m`, this issue disappears, so we should figure out where this function is called when it shouldn't be called.

kovidgoyal:
This is by design, since otherwise closing kitty windows does not focus other windows at all on macOS. See destroy_os_window() in glfw.c. If you want to fix this you would need to make that code workspace aware and restrict focusing switching to other windows on the same worksace.

Luflosi:
Hmm, you're right. I'll try to look into it when I have some time.

kovidgoyal:
You'd basically need to use private APIs, like here: https://chromium.googlesource.com/chromium/src/+/c2818d44191bc4342c2ae322f3e0cb218dd2ec66/chrome/browser/cocoa/nswindow_additions.mm

aequitas:
@kovidgoyal I figured the behavior is because of that. I think making the code workspace aware would be the best solution. @Luflosi is this something you want to pick up or should I have a go at it?

Luflosi:
@aequitas feel free to have a go at it, I'm quite busy atm.

aequitas:
After some experimenting I ended up with a working PoC to read workspaces: https://github.com/kovidgoyal/kitty/pull/1383

There is still some work to be done. The linkedlist stack for focus history might not be ideal for this use case so I'd have to see how to solve that. As now you might have to look deeper into the stack to find the next window (in the current workspace) to focus to.

I'll try to get it sorted out over the weekend.

aequitas:
@kovidgoyal awesome, looks a lot better than what I was going for. 

I've been testing it out and in concept your changes work as expected, only this function seems to return 0 for all workspaces and 1 for the first workspace:  
[...]

PR #1396, opened 2019-02-19
#1379 highest_focus_counter and loop index should be initialized to 0

koekeishiya:
Fixes a minor variable initialization mistake that causes the recent workspace-aware focus history on macOS to fail. See my comment in the referenced issue: #1379


aequitas:
Tested this change and for me its also working as expected.

PR #1400, opened 2019-02-22
Fix function returning the wrong type

Luflosi:
Im pretty sure this function is not supposed to return a boolean. I think this function is only used for `memcpy(ans, addr, MIN(strlen(addr), sizeof(ans)));`. Using a boolean here means that `memcpy()` copies at most one byte (I think). What sort of bugs would this cause?


kovidgoyal:
Wont cause any errors, since GLFWbool is a bool which is typically an int which is more than wide enough for the use it is used for here.

Luflosi:
Well, a quick regex search for `typedef .* GLFWbool` found only `typedef bool GLFWbool;` in `glfw/internal.h`.

PR #1403, opened 2019-02-22
Check if a pointer is null earlier

Luflosi:
If `lines` is `NULL` but `sizes` is not `NULL`, we have a memory leak here. To prevent that, I check if `lines` is `NULL` earlier. I'm not sure if this is the correct place to put this line or if it should be placed a few lines below, just before `size_t num_lines = PyList_GET_SIZE(lines);`. I don't know what the few lines between the two places do and if they should be executed if `lines` is `NULL`.


kovidgoyal:
lines can never be NULL, guaranteed by PyAr_ParseTuple

PR #1410, opened 2019-02-23
Fix memory leak

Luflosi:
If either `chars.data` or `candidates.data` is `NULL`, then the other one isn't freed, this pull request fixes that. I didn't put a semicolon behind `FREE_VEC(candidates)` because the definition of `FREE_VEC(vec)` already has a semicolon at the end.

### DIFF
--- a/kittens/choose/main.c
+++ b/kittens/choose/main.c
@@ -128,8 +128,9 @@ run_search(Options *opts, GlobalData *global, const char * const *lines, const s
     Chars chars = {0};
 
     ALLOC_VEC(text_t, chars, 8192 * 20);
+    if (chars.data == NULL) return 1;
     ALLOC_VEC(Candidate, candidates, 8192);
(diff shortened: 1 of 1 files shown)


kittens/choose/main.c
```diff
@@ -128,8 +128,9 @@ run_search(Options *opts, GlobalData *global, const char * const *lines, const s
     Chars chars = {0};
 
     ALLOC_VEC(text_t, chars, 8192 * 20);
+    if (chars.data == NULL) return 1;
     ALLOC_VEC(Candidate, candidates, 8192);
-    if (chars.data == NULL || candidates.data == NULL) return 1;
+    if (candidates.data == NULL) { FREE_VEC(candidates) return 1; }
```
Luflosi: Sorry, `FREE_VEC(candidates)` should be `FREE_VEC(chars)`. This was a leftover from some old code of mine where I freed both `candidates` and `chars`, which isn't needed here. I evidently deleted the wrong statement.

PR #1421, opened 2019-02-25
Add cell_width sanity check

Luflosi:
`cell_height` has a sanity check but `cell_width` doesn't so I added one.
We should probably add the four numbers as constants somewhere.

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -403,6 +403,8 @@ calc_cell_metrics(FontGroup *fg) {
     if (ch >= 4 && ch <= 1000) cell_height = ch;
     else log_error("Cell height invalid after adjustment, ignoring adjust_line_height");
     int line_height_adjustment = cell_height - before_cell_height;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Not needed because of line 403

Luflosi:
I don't quite understand. What does `if (ch >= 4 && ch <= 1000) cell_height = ch;` have to do with checking `cell_width`?

kovidgoyal:
sorry line 401

Luflosi:
Well, just like with `cell_height`, `cell_width` could already be invalid, so that the check in line 401 fails and `cw` isn't copied to `cell_width`, even if the config options aren't set. But if the config options aren't set, `cw` is already equal to `cell_width` and it makes no difference if the value is copied or not. Both checks print an error message if the check fails but the checks in lines 401 and 403 only print the error message with `log_error()` while the other checks fail with `fatal()`. That's why the other checks are needed.

Issue #995, opened 2018-09-19
OSC 52 handler appends to clipboard instead of setting it


PhilSwift7:
not sure if this is intended but if you run this code twice:

`printf "\033]52;c;$(printf "%s" "test" | base64)\a"`

your clipboard content will be `testtest` instead of `test`.

it can be fixed replacing += for = on the line 403 https://github.com/kovidgoyal/kitty/blob/master/kitty/window.py#L403

as a workaround running `printf "\033]52;c;$(printf "%s" "" | base64)\a"` clears the buffer but I don't know if that should be needed.

kovidgoyal:
It is by design. See https://sw.kovidgoyal.net/kitty/protocol-extensions.html#pasting-to-clipboard

agriffis:
@kovidgoyal I appreciate the extra functionality this provides. Would you consider making it configurable though? It violates the principle of least surprise, and becomes problematic when the user doesn't have the ability to send invalid codes to the terminal. For example I'm using mosh which supports OSC 52 pass-through, but since mosh is itself a terminal emulator in a sense, it doesn't permit a non-conforming payload.

kovidgoyal:
Why not just get mosh to output the invalid sequences as well, they do no harm on any other terminal emulator.

Luflosi:
Mosh does not seem to have active development right now, the latest release is from 22 Jul 2017. And people asking what's going on in https://github.com/mobile-shell/mosh/issues/974 and https://github.com/mobile-shell/mosh/issues/1021 don't make that look any better. So I'm afraid changing mosh is not a valid option at the moment.
I'm not saying that you should make this behaviour configurable, I'm just saying that your solution isn't practical for the foreseeable future.

kovidgoyal:
Considering that mosh is used to log securely into remote systems (as a
replacement for ssh), if it is actually un-maintained, I would not feel
comfortable using it at all, let alone for OSC 52.

agriffis:
I'm less interested in the mosh angle than the principle of least surprise. The current implementation of OSC 52 in kitty does something that's different from what users expect. I wonder if there would be a better way to approach it, to get the best of both worlds. For example, rather than requiring an invalid payload to clear the clipboard, instead support a specific invalid payload to explicitly concatenate. Then the default would be what people expect, but you could still get concatenation (which is quite a nice idea)

vphantom:
[...]

PR #1432, opened 2019-03-01
Allow no-append in clipboard_control

agriffis:
If no-append is present in clipboard_control, this will disable kitty's OSC 52 extension of concatenating to the clipboard. See #995

This is a compromise that I hope will be acceptable, since it doesn't change kitty's default behavior, and won't catch existing users off-guard.

I also considered using a separate option, but it seems unnecessary to add yet another option for something that only tweaks the behavior of OSC 52. Let me know if you'd prefer a separate option and I'm happy to update the PR.


agriffis:
I just force-pushed a doc fix. (I was using md syntax for quoting instead of rst.)

PR #1434, opened 2019-03-02
Replace the "78" terminfo dir with a symlink

Luflosi:
Since `78` and `x` always contain the same `xterm-kitty` file, I replaced `78` with a symlink.


Luflosi:
Sorry, this won't work at the moment because `build-terminfo` executes `shutil.copy2('terminfo/x/xterm-kitty', 'terminfo/78')`, which will fail because it's the same file. I guess I'll just remove that line.

Luflosi:
I fixed it, sorry for that.

PR #1435, opened 2019-03-02
Improve terminfo compilation code

Luflosi:
`setup.py` and `build-terminfo` do essentially the same thing when compiling the terminfo file by calling `tic` so I moved the common functionality into a new function in `kitty/terminfo.py` and hopefully improved things along the way.
The code in `setup.py` had a check `if not glob.glob(os.path.join(odir, '*/xterm-kitty'))` that checked if `tic` actually wrote the file to the filesystem. `build-terminfo` did not have this functionality. Here the same check would not work because the file would already exist before executing `tic`. I solved this by simply creating an temporary directory and letting `tic` write the file there. I can then check if the file exists and move it to the correct location.
I was also very careful to not assume if the output directory is called `x` or `78`; I simply create the destination directory (e.g. `terminfo/x` or `terminfo/78`) if it doesn't exist already and copy the compiled terminfo file into it. `build-terminfo` was setting the output directory with the `TERMINFO` environment variable while `setup.py` set it via the `-o` command line option. I chose to use the second option.
I also kept my stderr filtering code so that the output from `build-terminfo` is now also filtered. Let me know what you think about that change.
You should also look at the new imports and if they are in a good place or if they should be somewhere else (if the imports in the middle of the code should be at the beginning of the file and if the imports at the beginning of a file should be in a function).

### DIFF
--- a/build-terminfo
+++ b/build-terminfo
@@ -2,17 +2,15 @@
 
 import os
 import sys
-import shutil
-import subprocess
 
 base = os.path.dirname(os.path.abspath(__file__))
 os.chdir(base)
 sys.path.insert(0, base)
 
 from kitty.terminfo import generate_terminfo  # noqa
+from kitty.terminfo import compile_terminfo
 
 with open('terminfo/kitty.terminfo', 'w') as f:
     f.write(generate_terminfo())
 
-os.environ['TERMINFO'] = os.path.join(base, 'terminfo')
-subprocess.check_call(['tic', '-x', 'terminfo/kitty.terminfo'])
+compile_terminfo(base)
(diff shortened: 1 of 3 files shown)


Luflosi:
The CI found that I missed to import `sys` and had one blank line where there should have been two according to some python style guide. I fixed those issues.

wimstefan:
Since that fix xterm-kitty is installed into `/usr/share/terminfo/terminfo/x` instead of `/usr/share/terminfo/x` ...
My distribution is Gentoo and I'm using the live ebuild. This is simply said just a wrapper around `setup.py linux-package` where the resulting linux-package is copied to `/usr` ...

PS: Is it OK to report this in the closed issue or do you want me to create a new issue?

Luflosi:
#1440 should fix your issue. I somehow missed that error, sorry.
You could have created a new issue and linked to this pull request but it's not important.

wimstefan:
Thank you so much for the lightning fast respone!

Issue #1123, opened 2018-11-05
Scroll by pixels, not by lines, and support acceleration


maximbaz:
Extracting https://github.com/kovidgoyal/kitty/issues/1112#issuecomment-435876514 in a separate thread.

I noticed that sakura terminal scrolls by pixels:

![sakura-scroll](https://user-images.githubusercontent.com/1177900/48001263-58c39200-e108-11e8-86a0-07a099be5a1d.gif)

While kitty is only able to scroll by lines:

![kitty-scroll](https://user-images.githubusercontent.com/1177900/48001306-7d1f6e80-e108-11e8-9b0f-0e02655bead3.gif)

Scrolling by pixels feels a lot smoother, plus `sakura` has acceleration, which helps a lot. 

-------

Kitty doesn't seem to have acceleration, so if I set `wheel_scroll_multiplier 1` in config, scrolling will be very slow even if I violently move my fingers on touchpad — the absence of acceleration prevents me from permanently setting `wheel_scroll_multiplier` to `1`.

kovidgoyal:
Given the way kitty renders supporting pixel scrolling is way too much work. As for acceleration, not a fan personally, but pull requests are welcome. Note that on high resolution scrolling devices such as touchpads on macOS or with proper drivers in linux, there should be no need to implement acceleration, since it is automatic. You'd only needeeeed it for touchpad device that dont have propeer drivers or wheel mice, and at least personally, I dont wat acceleration for wheel mice.

maximbaz:
Pixel scrolling - OK, I anyway expected this answer 😄 

Acceleration - definitely not wanted for wheel mice, but it doesn't work work for me on laptop. If you run linux on laptop, does acceleration work for you?

Acceleration works everywhere else for me, so I'm really surprised that it might be a driver issue, did all other apps truly re-implemented acceleration?

maximbaz:
Is there a place I can maybe add some debug prints to prove that my touchpad doesn't produce accelerated values?

Luflosi:
I would _**really**_ love to have smooth and precise scrolling in kitty as I know of no terminal emulator on macOS that supports it and only two on linux. I might make a wip pull request to attempt to start implementing this in the near future. Here is a great article that goes into great detail about many problems, solutions and more when implementing something like this for anyone interested: https://pavelfatin.com/scrolling-with-pleasure/.
[...]

Issue #1406, opened 2019-02-23
ModuleNotFoundError: No module named 'kitty.update_check'


blueyed:
When using the method described at https://github.com/kovidgoyal/kitty/blob/e44b331cc3048c0d921e83ab836dbc6167d82e50/docs/build.rst#install-and-run-from-source to run kitty from source (`python .`) it displays this error/warning:
```
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/bin/../lib/kitty/__main__.py", line 117, in <module>
    main()
  File "/usr/bin/../lib/kitty/__main__.py", line 108, in main
    namespaced(['+', first_arg[1:]] + sys.argv[2:])
  File "/usr/bin/../lib/kitty/__main__.py", line 68, in namespaced
    func(args[1:])
  File "/usr/bin/../lib/kitty/__main__.py", line 27, in runpy
    exec(args[1])
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'kitty.update_check'
[054 07:54:28.731077] Failed to process update check data '', with error: invalid literal for int() with base 10: ''
```

Luflosi:
This is probably the same error as described in #1352.

kovidgoyal:
Indeed, you likely have a kitty on your path already that points to a different one than the built from source one. Remove that or change it to point to the correct kitty and you will be fine.

blueyed:
I have this wrapper - shouldn't it work?

```
#!/usr/bin/env python3
import runpy
import sys

kitty_path = '/home/daniel/Vcs/kitty'

# try:
#     sys.path.insert(0, kitty_path)
#     from kitty.constants import kitty_exe  # noqa: E402
#     kitty_exe.ans = __file__
# except ImportError:
#     pass

runpy.run_path(kitty_path, run_name='__main__')
```

`kitty '+runpy' 'from kitty.update_check import *'` appears to work with it.

kovidgoyal:
Remember that the PATH you see inside a shell is not going to
necessarily be the PATH that kitty sees, because any changes to the PATH
you make inside shell startup scripts will not be seen by kitty.

blueyed:
Hmm.. I am running `python .` (which also triggers it) from the same shell.
[...]

PR #1454, opened 2019-03-10
[WIP] Implement smooth and precise scrolling

Luflosi:
I started to implement smooth and precise scrolling, see #1123. My plan is to render to some kind of buffer (I'm currently using a texture) and then copy the right part of that buffer to the normal framebuffer to allow pixel-precise scrolling. This is still very much a WIP, my code currently doesn't even do anything with scrolling. The code is also very hacky with some of it copied straight from learnopengl.com, that I don't even know exactly what it does. A lot of the new code for example in `init_cell_program()` should probably be somewhere else. I also hardcoded the size of the texture. This also literally the first time I've done anything with OpenGL, so this will be difficult for me and probably take a while to finish implementing.
My current problem is getting the information that tells the shader by how much to shift the content up and down from `after_render()` into the shader. I'm currently just using `0.1` as a constant. I think you're doing something like that in `create_cell_vao()` with the `A1` define but I don't know how that works. Could you help me with that please?
Any other feedback is also welcome.


kovidgoyal:
Dont worry about the awkwardness of your OpenGL code -- OpenGL is a pain to use. I will help you clean up your code after you have something working.

As for your question: If you want to convert a pixel scroll offset into an OpenGL offset, you basically divide the pixel scroll offset by the viewport height (os_window->viewport_height). Then divide that by 2 because the opengl co-ord system goes from -1 to 1 which means the full screen has length 2.

Some general comments. This will as far as I can see render blank at the bottom line when scrolling. If that is all you want to do, you could simply offset the rendering of the main shader and then draw a blank rectangle at the bottom. No need to use framebuffers at all.

Luflosi:
I didn't ask about converting the offset from pixels to OpenGL but thank you anyways because I'll need that info later. But first I need to actually copy that offset from the C code to the shader and I don't know how to do that.

As for the blank sections, my first proof of concept will have those, yes. But it won't be the same as drawing a blank rectangle at the bottom or the top because the content will be shifted by a specific number of pixels. I also plan to render more than what is currently visible on screen and then keep what was rendered as a sort of cache. This will allow me to pre-render parts that are not yet visible. It should also save a little bit of GPU time and power because I don't have to render the whole screen on every frame. This will become more important than it currently is because kitty currently only needs to render one frame per line scrolled and the user doesn't scroll 60 lines per second (or whatever FPS their monitor has) most of the time while 60 pixels per second is going to be a lot more likely, at least with high resolution input devices.

Luflosi:
> you could simply offset the rendering of the main shader and then...

Oh, I missed that. But my other reason for doing it like this, still applies.

kovidgoyal:
Use a uniform to pass values from C to the shader. See glUniform

Luflosi:
Your tip was very helpful.
My first prototype is sort of working, here is a short video: https://youtu.be/6n31SO1HJAI.

Luflosi:
@maximbaz 
> if you ever get to implementing improvements for scrolling, be sure to ping me for testing

Doing it now 😁🎉

maximbaz:
Nice contribution, I've been following this PR 😝 The demo on the video is awesome!

I don't have high hopes as I assume it won't work on X (since it doesn't even have precise scrolling), but I just tried anyway — when kitty opens from your branch, all I see is black screen.

I'm launching via `$ PATH= /usr/bin/python . /usr/bin/bash`, to avoid #1406 and to make sure my zsh config is not causing any trouble, and here's what I see in the console:

```
❯ PATH= /usr/bin/python . /usr/bin/bash
[071 00:59:12.726902] Failed to run kitty for update check, with error: [Errno 13] Permission denied: 'kitty'
1899 2044
0.000000
0.000000
0.000000
0.000000
0.000000
```

Does it give you any clue?

In the meantime, I'll quickly install sway and try it out there 😉

Luflosi:
Yeah, it won't work on X but no idea why there is a black screen. But I'd consider this code pre-alpha and to be honest I'm almost surprised it worked for me 😅.
#1406 should be fixed by #1408.

Luflosi:
@maximbaz the black background was probably caused by the hardcoded size. It's still not very good, resizing the window or opening a tab so that the tab bar is shown doesn't adjust the size of the texture but that will have to wait a bit...

maximbaz:
At least the black background issue is fixed now, kitty launches fine on X11 😉 👍 

I also wanted to try on sway, but for unknown to me reasons kitty doesn't work there, even on master branch, kitty's window opens but remains empty. If I provide `KITTY_DISABLE_WAYLAND=1`, then it launches fine (and is blurry, indication of it being running via XWayland). Might be something wrong with sway on my side, I don't usually use it.

Anyway, good job once again, I'll go back now to crying about the absence of smooth scrolling for X11 in glfw 😢

Luflosi:
@kovidgoyal what do `prepare_to_render_os_window()` and `render_os_window()` do?

kovidgoyal:
prepare_to_render() updates all the window global state and checks if the
window needs to be rendered (i.e. if anything has changed since the last
time it was rendered). render_os_window() actually renders the window by
calling various functions from shaders.c

Luflosi:
Where is `screen_resize()` from `kitty/screen.c` called? If I search for it, that's the only occurrence in the entire source tree.

maximbaz:
Just to give you some input, this is really nice work @Luflosi! I'm testing on sway, and although there are some grey artifacts, in general the code works, and smooth scrolling is very pleasant! 🙂 

Do let me know when you want a more thorough testing / feedback!

record_1586431224

Luflosi:
Thanks for the nice words @maximbaz. The grey artefacts are there because I didn't manage to make the rendered region bigger than the window yet. It needs to be at least one line taller to fill the grey regions. Basically everything else is still broken, like resizing, OS windows, tabs, kitty windows, background images, selections, ... This will need a lot more work and while I haven't done much for the past year, I now want to work on it again.

kovidgoyal:
On Thu, Apr 09, 2020 at 03:41:14AM -0700, Luflosi wrote:
> Where is `screen_resize()` from `kitty/screen.c` called? If I search for it, that's the only occurrence in the entire source tree.

It becomes the resize() methos of the screen object

jclsn:
Pity that this features seems to have been abandoned.

serhez:
@jclsn I think this wouldn't work for vim as it takes over the terminal and performs its own scrolling simulation AFAIK, but I'm not 100% sure.

jclsn:
@serhez Yeah, probably not

Akari202:
@serhez there is vim-smoothie that adds smooth scrolling to navigation via ctrl-f and ctrl-d

serhez:
> @serhez there is vim-smoothie that adds smooth scrolling to navigation via ctrl-f and ctrl-d

There are many such plugins but they do not implement pixel-wise scrolling, the do smooth scrolling in the sense that they use a non-constant (mostly non-linear too) function to calculate the delay between each line being scrolled, I think. But it's still line-wise scrolling.

alba4k:
Is this 100% abbandoned? it would be one of the biggest features for quality of life, especially with trackpads

NeilGirdhar:
@jclsn Per your comment: https://github.com/neovim/neovim/issues/27583

jclsn:
@NeilGirdhar I am one of these guys who refuse to use Neovim 😉

simonlearnscoding:
looks like this won't be pursued any further? too bad, now that there is a smooth cursor animation, smooth scrolling is the only thing that separates neovide from kitty

hernancerm:
+1. Buttery smooth scrolling like in Neovide would be a very significant quality of life improvement, as it makes navigation in Vim/Neovim much less disorienting. The cursor animation is cool, but I'd say less practical than smooth scrolling.

I don't even know if having a scroll as smooth as Neovide is possible in kitty.

dmtrKovalenko:
Oh my god you already have a PR for that! That is going to be huge. Is there anything community can help with to move this features forward? I am ready to spend a lot of time to move this forward.

kovidgoyal:
Before you get too excited this is smooth scrolling for kitty's own scrollback. It wont work in TUI apps like neovim. Implementing smooth scrolling for TUI apps is a whole other kettle of fish.

Doroszewski:
Still better than nothing and good enough, especially for people who use CLI more often than TUI :). (Also, isn't that what VT320 was doing?) Hopefully this will finally make it to Kitty.

sophronesis:
any updates?

alicealysia:
> Oh my god you already have a PR for that! That is going to be huge. Is there anything community can help with to move this features forward? I am ready to spend a lot of time to move this forward.

Step 1 is to get this updated. As you can see, this has a single commit from 2020 that will need to be forward ported, this will require a rewrite as a codebase changes a lot over the course of 5 years.

To more or lesss explain the code as I understand it, per pixel scrolling is done via a shader which generates a texture of the current viewport, then positions it based on the current per pixel scroll position.

The biggest outstanding challenge that wasn't solved in the original pr is rendering new lines above the current line, I'm also unsure if a lerping animation was introduced for digital scrolling systems, so if not, that will need to be introduced too.

We can look to neovide for inspiration on both, because contrary to popular belief, reviewing the issues and pr for smooth scrolling shows that it actually did have the same limitations that make smooth scrolling so difficult in a terminal emulator?

Nvim introduced the ability to set the number of lines in the viewport. Neovide simply increases the viewport by the number of lines scrolled, then uses a shader to render a rectangle the size of the window offset by a value that lerps between the current line pre-scroll and post-scroll.

idelice:
Here's my attempt to enable smooth scroll: https://github.com/kovidgoyal/kitty/pull/9330

kovidgoyal:
This is now implemented in master.

PR #1459, opened 2019-03-12
Fix typo

Luflosi:


Luflosi:
Well, it probably is a copy-paste error, not a typo but whatever.

PR #1478, opened 2019-03-16
Guard against negative screen scale.

tarmack:
This prevents a crash when an invalid scaling value is reported by the compositor.

### DIFF
--- a/kitty/glfw.c
+++ b/kitty/glfw.c
@@ -379,8 +379,8 @@ get_window_content_scale(GLFWwindow *w, float *xscale, float *yscale, double *xd
         if (monitor) glfwGetMonitorContentScale(monitor, xscale, yscale);
     }
     // check for zero or NaN values of xscale/yscale
-    if (!*xscale || *xscale != *xscale) *xscale = 1.0;
-    if (!*yscale || *yscale != *yscale) *yscale = 1.0;
+    if (*xscale <= 0 || *xscale != *xscale) *xscale = 1.0;
+    if (*yscale <= 0 || *yscale != *yscale) *yscale = 1.0;
 #ifdef __APPLE__
     const double factor = 72.0;
 #else


kitty/glfw.c
```diff
@@ -379,8 +379,8 @@ get_window_content_scale(GLFWwindow *w, float *xscale, float *yscale, double *xd
         if (monitor) glfwGetMonitorContentScale(monitor, xscale, yscale);
     }
     // check for zero or NaN values of xscale/yscale
-    if (!*xscale || *xscale != *xscale) *xscale = 1.0;
-    if (!*yscale || *yscale != *yscale) *yscale = 1.0;
+    if (*xscale < 0 || *xscale != *xscale) *xscale = 1.0;
+    if (*yscale < 0 || *yscale != *yscale) *yscale = 1.0;
```
Luflosi: I think this should be `<=` instead of `<`.

tarmack: You are absolutely correct

kitty/glfw.c
```diff
@@ -379,8 +379,8 @@ get_window_content_scale(GLFWwindow *w, float *xscale, float *yscale, double *xd
         if (monitor) glfwGetMonitorContentScale(monitor, xscale, yscale);
     }
     // check for zero or NaN values of xscale/yscale
```
Luflosi: You may also want to adjust this comment.

tarmack:
It seems that a little more testing uncovered that this is not a full fix. I'll be working on this some more.

kovidgoyal:
@tarmack it may make sense to do value sanitization inthe glfw backend as well (see wl_window.c or x11_window_c or glfw/cocoa_window.m)

kovidgoyal:
In the meantime I have committed code to make the scale sanitization more comprehensive, since it is a good idea.

tarmack:
Hi Kovid, good that you added a max as well, that is the actual problem I found out. Sometimes when an monitor is removed there is a race and the monitor reports a scale of 49152 (2^15 + 2^14). Given the specific value I suspect this is actually the result of a bug somewhere but can't really pinpoint the origin.
I work on a low ram machine and noticed that when I disabled the external output for testing, it was quite common for Kitty to end up crashing with an error allocating memory for a history segment. I am not sure if this is directly related to the scaling though

In addition to this scale issue I noticed that Kitty on Wayland (Sway) doesn't actually scale its drawing, only the output is upscalled by the compositor. This is probably the result of the "fake" scaling to allow fractional scales (the output protocol only does integer scales) used by my compositor. I would like to improve this in a separate PL.

kovidgoyal:
Sure Wayland related improvements are most welcome. I dont use it, so I
am not the best person to work on it.

Issue #461, opened 2018-04-15
Do not render ligatures under cursor (optionally?)


maximbaz:
I'm playing with font ligatures using Fira Code, it is fun to see them rendered and they convey a meaning quite clearly, however it is sometimes confusing to actually _edit_ the rendered ligature.

For example, looking at this, the meaning is clear:

![image](https://user-images.githubusercontent.com/1177900/38778473-5655727a-40ba-11e8-8f0b-78063f007375.png)

However, what's not immediately obvious is what will happen if I press <kbd>Del</kbd> or <kbd>Backspace</kbd> in this case:

![image](https://user-images.githubusercontent.com/1177900/38778479-7a2a8afa-40ba-11e8-8fcc-d1fe96726250.png)

In fact, in this particular case the code is:

```javascript
if (a !== b) {
```

What I suggest to do is to **not render a ligature if a cursor is currently over it**, just render `!==`. 
Then it is very obvious what <kbd>Del</kbd> or <kbd>Backspace</kbd> will do 🙂 

I guess it should be relatively easy to achieve, because a ligature always takes the same number of cells as all of the the underlying "raw" characters.

I have seen people express this opinion for font creators (e.g. https://github.com/be5invis/Iosevka/issues/257), where they suggest to edit the font and choose a better design. However, I think rendering "raw" characters instead of a ligature when the cursor is above the glyph is a more generic and font-agnostic solution.

I'm not sure if this requires a configuration option or should become a new behavior for everyone.

----------------

UPDATE: [looks like](https://github.com/jwilm/alacritty/issues/50#issuecomment-277468592) I'm not the first one who thought of this (hehe), and Qterminal and Konsole already have such behavior:

![decomposition](https://cloud.githubusercontent.com/assets/20362502/22620945/a7e15214-eb0f-11e6-8f01-de7c4256f9b4.png)

kovidgoyal:
Patches are welcome

maximbaz:
Thanks, this is interesting. 

The first Konsole's issue [361547](https://bugs.kde.org/show_bug.cgi?id=361547) will not be a problem in kitty due to the way how ligatures are implemented, the characters will never move because by design the "ligaturized" text takes exactly the same number of cells as the "raw" text.

![ligature_size](https://user-images.githubusercontent.com/1177900/38783847-6ea95a32-4108-11e8-8d84-93d843528f6b.gif)

Very interesting observation with colors 🙂

[...]

PR #1479, opened 2019-03-17
Do not render ligatures under cursor

Luflosi:
In #461 you gave a few helpful tips, and after a couple hours of trying different things, I came up with the following code that doesn't even touch any CPU or GPU cells and seems to work as a proof of concept. There is a line `render_ligature = (i != 4)` with a constant (4 in this case). All ligatures beginning in the third cell from the left of the screen will be rendered as individual characters.
Does this look promising to you? Is this the right approach?

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -256,6 +256,11 @@ Syntax is::
 
 '''))
 
+o('disable_ligatures_under_cursor', False, long_text=_('''
+Render the characters of a ligature under the cursor individually
+to make editing more intuitive.
+'''))
+
 
 def box_drawing_scale(x):
     ans = tuple(float(x.strip()) for x in x.split(','))
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -58,6 +58,7 @@ typedef struct {
 
 
 static hb_buffer_t *harfbuzz_buffer = NULL;
+static hb_feature_t no_liga_feature;
 static char_type shape_buffer[4096] = {0};
 static size_t max_texture_size = 1024, max_array_len = 1024;
 
@@ -717,7 +718,7 @@ num_codepoints_in_cell(CPUCell *cell) {
 }
 
 static inline void
-shape(CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, index_type num_cells, hb_font_t *font) {
+shape(CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, index_type num_cells, hb_font_t *font, bool disable_ligature) {
     if (group_state.groups_capacity <= 2 * num_cells) {
         group_state.groups_capacity = MAX(128, 2 * num_cells);  // avoid unnecessary reallocs
         group_state.groups = realloc(group_state.groups, sizeof(Group) * group_state.groups_capacity);
@@ -741,7 +742,13 @@ shape(CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, index_type num_cells, hb
     group_state.last_cpu_cell = first_cpu_cell + (num_cells ? num_cells - 1 : 0);
     group_state.last_gpu_cell = first_gpu_cell + (num_cells ? num_cells - 1 : 0);
     load_hb_buffer(first_cpu_cell, first_gpu_cell, num_cells);
-    hb_shape(font, harfbuzz_buffer, NULL, 0);
+
+    if (!disable_ligature) {
+        hb_shape(font, harfbuzz_buffer, NULL, 0);
+    } else {
+        hb_shape(font, harfbuzz_buffer, &no_liga_feature, 1);
+    }
+
     unsigned int info_length, positions_length;
     group_state.info = hb_buffer_get_glyph_infos(harfbuzz_buffer, &info_length);
     group_state.positions = hb_buffer_get_glyph_positions(harfbuzz_buffer, &positions_length);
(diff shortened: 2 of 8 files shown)


kovidgoyal:
Yes, that is basically the correct approach. There are various complications though, what if the cursor is on a width two character, or if a font has ligatures for == and === and the cursor is on the first =, does the ligature change to = followed by the ligature for ==? Or three ===?

And of course you still need to figure out how to get the information about the current cursor position into the render function.

And another complication is that you need to check that moving the cursor over a ligature only causes that instance of the ligature to be changed back, not all of them on the screen. I think that is alreadythe case, but it needs to be verified.

Luflosi:
What happens in the code when the cursor position changes? How is that handled?

Luflosi:
I got something sort of working.
The ligature doesn't render, if the cursor is one cell to the right of the last cell of the ligature. If there is a ligature for `===` and the cursor position is one cell to the right of that, the `=` will all render individually, even if there is a ligature for `==`.
I found a way to get the current cursor position.
Multiple instances of the same ligature are handled correctly.
Now I just need to optimise the three places where I added ` || true` to just render if the cursor position actually changed. Is there already some functionality in place to detect a changed cursor position? Otherwise I'll just add a new boolean `screen->cursor_pos_changed` or something like that.
I also need to not render the ligature if the cursor is anywhere in the ligature, not just if it's next to the ligature.

kovidgoyal:
There is no notification for cursor position changing, lots and lots of
things can change the cursor position.

Luflosi:
I can now detect the changing cursor position.
Here is a short video of how it looks so far: https://youtu.be/HUcsYLaQUH8.

Luflosi:
It pretty much works now:
recording

Luflosi:
The next steps are to get feedback from you, add a config option to configure this and clean up the code.

kovidgoyal:
Cool, I am going to be away for a few days, will look at it when I
return.

kitty/fonts.c
```diff
@@ -741,7 +741,19 @@ shape(CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, index_type num_cells, hb
...
     load_hb_buffer(first_cpu_cell, first_gpu_cell, num_cells);
-    hb_shape(font, harfbuzz_buffer, NULL, 0);
+
+    if (!render_ligature) {
+        hb_feature_t feature = {
+            .tag = hb_tag_from_string("calt", strlen("calt")),
+            .start = 0,
+            .end = UINT_MAX,
+            .value = 0
+        };
```
khaledhosny: You can simplify this and use `hb_feature_from_string("-calt")`.

Luflosi: Cool, thanks.

kovidgoyal:
@Luflosi: Can you squash this into one or two commits, I will review it then.

Luflosi:
I squashed the commits.

kitty/config_data.py
```diff
@@ -256,6 +256,11 @@ def adjust_line_height(x):
 
 '''))
 
+o('show_ligatures_under_cursor', False, long_text=_('''
```
kovidgoyal: This should be named disable_ligatures_under_cursor and default to False, to not change existing behavior.

kitty/data-types.h
```diff
@@ -195,11 +195,10 @@ typedef struct {
     PyObject_HEAD
 
     bool bold, italic, reverse, strikethrough, blink, dim;
-    unsigned int x, y;
+    index_type x, y, last_x, last_y;
```
kovidgoyal: These should be named last_rendered_X/y to make it clear what their purpose is.

kitty/fonts.c
```diff
@@ -741,7 +741,16 @@ shape(CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, index_type num_cells, hb
     group_state.last_cpu_cell = first_cpu_cell + (num_cells ? num_cells - 1 : 0);
     group_state.last_gpu_cell = first_gpu_cell + (num_cells ? num_cells - 1 : 0);
     load_hb_buffer(first_cpu_cell, first_gpu_cell, num_cells);
-    hb_shape(font, harfbuzz_buffer, NULL, 0);
+
+    if (!render_ligature) {
+        hb_feature_t feature;
+        const char* feature_str = "-calt";
+        if (!hb_feature_from_string(feature_str, strlen(feature_str), &feature)) fatal("hb_feature_from_string() failed");
```
kovidgoyal: These seems unnecessarily ineeficient. Does it really need to be called on every shape() run. Cant it instead use  a static variable? And there should be no need to use strlen() here, sizeof() should suffice.

kitty/data-types.h
kovidgoyal: And these should really be variables in the screen struct rather than the Cursor struct. Cursor objects can come and go (see for example screen_save_cursor). In fact, the information of where the cursor was last rendered on the screen logically belongs to the screen, notth ecursor.

kitty/fonts.c
```diff
@@ -1009,14 +1018,20 @@ render_run(FontGroup *fg, CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, inde
...
+render_line(FONTS_DATA_HANDLE fg_, Line *line, index_type lnum, Cursor *cursor) {
+#define RENDER if (run_font_idx != NO_FONT && i > first_cell_in_run) { render_run(fg, line->cpu_cells + first_cell_in_run, line->gpu_cells + first_cell_in_run, i - first_cell_in_run, run_font_idx, false, center_glyph, render_ligature); }
     FontGroup *fg = (FontGroup*)fg_;
     ssize_t run_font_idx = NO_FONT;
     bool center_glyph = false;
+    bool render_ligature = true;
     index_type first_cell_in_run, i;
     attrs_type prev_width = 0;
     for (i=0, first_cell_in_run=0; i < line->xnum; i++) {
+        if (cursor != NULL && !OPT(show_ligatures_under_cursor)) {
```
kovidgoyal: this can be made more efficient by checking cursor != NULL && show_ligatures && lnum == cursor->y outside the loop and storing the result in bool.

kitty/fonts.c
Luflosi: Where is the best place to initialise that variable? `sizeof()` won't work here, as it will return the size of the pointer (8 bytes on my machine) instead of the length of the actual string.

kitty/fonts.c
Luflosi: Yeah, that's better.

kitty/config_data.py
Luflosi: Originally I wanted to also have this option be inverted compared to `show_ligatures_under_cursor` but I couldn't think of a good short name. I thought of `hide_ligatures_under_cursor` but I found that to be confusing since it would suggest that nothing is visible at all. I like your suggestion `disable_ligatures_under_cursor` better. 👍

kitty/data-types.h
Luflosi: 👍

kitty/fonts.c
Luflosi: Maybe I found a good place.

Luflosi:
I think I made all the changes you suggested. Is `init_fonts()` the right place for the `hb_feature_t` variable? And is it ok to use `fatal()` in that function? A couple lines above, an error is reported via `PyErr_NoMemory()`.

kitty/fonts.c
kovidgoyal: you know the relation between sizeof and strlen() it is fixed, so there is not need to use strlen.

PR #1481, opened 2019-03-17
Fix scale change detection at runtime on wayland.

tarmack:
Without this any changes to the output scale after kitty is started is never applied.

### DIFF
--- a/glfw/wl_window.c
+++ b/glfw/wl_window.c
@@ -41,6 +41,34 @@
 
 #define URI_LIST_MIME "text/uri-list"
 
+static void checkScaleChange(_GLFWwindow* window)
+{
+    int scale = 1;
+    int i;
+    int monitorScale;
+
+    // Check if we will be able to set the buffer scale or not.
+    if (_glfw.wl.compositorVersion < 3)
+        return;
+
+    // Get the scale factor from the highest scale monitor.
+    for (i = 0; i < window->wl.monitorsCount; ++i)
+    {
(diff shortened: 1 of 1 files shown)


glfw/wl_window.c
```diff
@@ -1041,6 +1048,7 @@ void _glfwPlatformSetWindowSize(_GLFWwindow* window, int width, int height)
 {
     window->wl.width = width;
     window->wl.height = height;
+    checkScaleChange(window);
```
kovidgoyal: this should go into the handleConfigure functions ratherthan inside the set window size function. I believe other parts of glfw can call this function as well.

tarmack: You are absolutely right, it was not working at all like this before (application would hang). Substituting `_glfwInputWindowContentScale` for the direct resize fixed that, I think this solution is much cleaner in general.

kovidgoyal: But doesn't this mean that if there is a scale change but not a size change resizeWindow will not be called? I think this entire thing needs to be re-factored. The size and scale should all be checked independently in configure callbacks. And then the appropriate input window callback should be called. And if either size or scale has changed, the resizeWindow function should be called.

tarmack: You are right, this thing needs a refactor really badly. There is actually a ton of duplication that needs removing as well. I'll get back when I have something.

kovidgoyal:
This should be fixed in current master.

PR #1491, opened 2019-03-23
Option to disable drawing of active window border

akhilles:
`draw_active_window_borders` option added, defaults to True. Unaffected by draw_minimal_borders and window_margin_width.

### DIFF
--- a/kitty/layout.py
+++ b/kitty/layout.py
@@ -17,6 +17,7 @@ cell_width = cell_height = 20
 all_borders = True, True, True, True
 no_borders = False, False, False, False
 draw_minimal_borders = False
+draw_active_window_borders = False
 
 
 def idx_for_id(win_id, windows):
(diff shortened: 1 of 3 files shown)


kitty/layout.py
```diff
@@ -26,8 +27,9 @@ def idx_for_id(win_id, windows):
 
 
 def set_draw_minimal_borders(opts):
```
akhilles: Maybe this function name can be generalized to set_draw_border_flags()?

kovidgoyal:
I dont understand how this is supposed to work. As far as I cant tell, it will simply draw the borders around the active window in the inactive border color instead of the active border color. If that is what you want to do, why not just set the active border color to be the same as the inactive border color in kitty.conf?

akhilles:
This is to completely disable rendering of active window border. If I change the inactive border color to active border color, it still renders the extra edges as you switch active window.

You said I could submit a PR for this, but then just rejected it .... :(

akhilles:
Adding a screenshot:
image

kovidgoyal:
Oh sorry my mistake, I missed a part of the patch.

PR #1529, opened 2019-04-07
Respect macOS selection color

Luflosi:
In macOS under General, there is a setting for "Highlight color", which is used for all sorts of selection highlighting, e.g. selecting text in Safari or selecting a file in Finder.
This pull request aims to change the `selection_background` config option to accept a new value like "system" that makes kitty use the system default.
Most apps allow dynamically changing the highlight color while running by just changing the setting. This pull request does not aim to do that. Changing the color will require a restart for now.
In `kitty/glfw.c` I'm currently only printing the color. How do I actually set it? It is not available with the `OPT()` macro. I think I need to set it in the Screen struct with `screen->color_profile->overridden.highlight_bg`. Could you give me a few tips on how to do that exactly?
I'm also not sure if a I should change the GLFW API like I did it now or if I should just create one new function in the cocoa code. The cursor blink interval for example can be read with `cocoa_cursor_blink_interval()`, while the double click interval can be read with `glfwGetDoubleClickInterval()`.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1521,10 +1521,10 @@ double _glfwPlatformGetDoubleClickInterval(_GLFWwindow* window)
 
 
 int _glfwPlatformGetSelectionColor(_GLFWwindow* window,
-                                    uint8_t* red,
-                                    uint8_t* green,
-                                    uint8_t* blue,
-                                    uint8_t* alpha)
+                                   uint8_t* red,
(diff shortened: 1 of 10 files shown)


kovidgoyal:
I dont think this is a good idea. Terminal colors have no relation to
system colors, therefore randomly choosing one of the system colors is
likely to create poor contrast/mismatched colors.

Luflosi:
Ok

PR #1538, opened 2019-04-14
Option for hide mouse on keypress event

### LINKED ISSUE #1534: Feature request: option for hide mouse on keypress event
It would be really nice if the user could enable an option to instantly hide the mouse cursor on any keypress event in the active window. That way you could click inside a window, and as soon as you start typing in your terminal, the mouse cursor disappears. This feature is enabled by default in many other terminal applications like iTerm2. Right now the user has to hack this feature in by setting the mouse_hide_wait param to be very small, which has unintended effects if you're actively using the mouse.

Luflosi:
Fixes #1534.
Is this generally the right approach?
Does `global_state.callback_os_window` always have the correct value or should I get the OS window some other way?
Maybe you can think of a better name besides `mouse_hide_key`.
Or maybe we should change the behaviour of the `mouse_hide_wait` option to accept a new value.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -108,6 +108,9 @@ To update |kitty|, :doc:`follow the instructions <binary>`.
 - Allow setting :opt:`active_border_color` to ``none`` to not draw a border
   around the active window (:iss:`805`)
 
+- Use negative values for :opt:`mouse_hide_wait` to hide the mouse cursor
+  immediately when pressing a key (:iss:`1534`)
+
 
 0.13.3 [2019-01-19]
 ------------------------------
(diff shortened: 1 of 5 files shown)


Luflosi:
I'll change the default to no at some point.

kovidgoyal:
yes callback_os_window is always correct in callbacks such as key events. I would prefer just modifying mouse_hide_wait with negative values meaning, hide on keypress.

Luflosi:
Does everything look good?
I have some concerns about this feature behaving slightly differently than the native macOS implementation in most other macOS apps. In macOS, arrow keys and other keys/key combinations that move the (text) cursor also hide the mouse cursor.

Luflosi:
I would have liked to rebase the commits but oh well...
What's your opinion on keys that change the cursor position? Should they hide the mouse cursor?

kovidgoyal:
they do hide the mouse, in kitty as well

Luflosi:
I didn't realise that you moved that line up a bit. While kitty does now react to all keys, including arrow keys, it now also reacts to modifier keys like <kbd>ctrl</kbd> and <kbd>cmd</kbd>, which is not the case in most other macOS apps. What is the most common behaviour in Linux?

PR #1559, opened 2019-04-23
Convert HTTP URLs to HTTPS URLs where possible

Luflosi:
I converted those URLs because HTTPS should provide a bit more security than HTTP.
I checked every single URL by hand and ran `make linkcheck` to make sure that the site supports HTTPS.
Is it ok to change the URLs in the LICENSE? Not sure.
I didn't change the URL for `http://man7.org/linux/man-pages/man7/shm_overview.7.html` as `man7.org` doesn't even allow connections on port 443. I didn't change `http://pygments.org/` because the certificate is only valid for a different domain. The same is true for `http://cocoadev.com/index.pl?GameKeyboardHandlingAlmost` but that site seems to be totally dead atm, showing a 404 for that specific page and the nginx welcome page in the root directory. There are also some HTTP URLs in the tests, which I did not touch because I'm pretty sure that no connection is actually made. I also left the URLs in `logo/kitty.svg` alone as it's an automatically generated file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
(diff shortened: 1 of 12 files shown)


kovidgoyal:
Changing license URLS is fine as far as I know, since the actual license is contained inthe LICENSE file itself.

Luflosi:
At first I was a bit scared because the LICENSE states
> Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.

but then I noticed, that on https://www.gnu.org/licenses/gpl-3.0.en.html, all the links are changed as well.

PR #1562, opened 2019-04-24
Update changelog.rst to reflect change to disable_ligatures config option

Luflosi:
In 934336e64278dae88ab87aa75c9bf553a5931fa6, the config option
`disable_ligatures_under_cursor` was renamed to `disable_ligatures` and the
behaviour of that option was altered so that it is now possible to
disable ligatures completely. This commit reflects that change in changelog.rst


Luflosi:
Why was there a warning in CircleCI?

PR #1571, opened 2019-04-25
Rewrite compilation code

Luflosi:
I found the compilation process to be too slow, so I rewrote it to be more parallel and achieved a decent reduction in compilation time on my multicore system.
This is still a WIP and I mainly want to test my code using the CI at this point in time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,14 +16,22 @@ import sys
 import sysconfig
 import time
 from contextlib import suppress
+import queue
+import pty
+import struct
+import fcntl
+import termios
+import asyncio
+import signal
+from kitty.enums import BuildType
 
 base = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.join(base, 'glfw'))
 glfw = importlib.import_module('glfw')
 verbose = False
 del sys.path[0]
-build_dir = os.path.join(base, 'build')
-constants = os.path.join(base, 'kitty', 'constants.py')
+build_dir = 'build'
+constants = os.path.join('kitty', 'constants.py')
 with open(constants, 'rb') as f:
     constants = f.read().decode('utf-8')
 appname = re.search(r"^appname = '([^']+)'", constants, re.MULTILINE).group(1)
@@ -177,7 +185,9 @@ def first_successful_compile(cc, *cflags, src=None):
 
 class Env:
 
-    def __init__(self, cc, cppflags, cflags, ldflags, ldpaths=[]):
+    def __init__(self, cc, cppflags, cflags, ldflags, ldpaths=None):
+        if ldpaths is None:
+            ldpaths = []
         self.cc, self.cppflags, self.cflags, self.ldflags, self.ldpaths = cc, cppflags, cflags, ldflags, ldpaths
 
     def copy(self):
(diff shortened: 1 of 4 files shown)


Luflosi:
Here are two screenshots of my CPU utilisation:
before:
before
after:
after
The arrows mark where the compilation process starts and ends.
As you can see, without this patch, most of the time, the compilation process is only using one of the four cores. With this patch more time is spent utilising more cores. The overall time reduction is also visible, the distance between the arrows is reduced.

Luflosi:
Compiling `kittens/unicode_input/unicode_names.c` takes an extremely long time. In fact, with my patch, I start compiling it first and it finishes last. Maybe there is some easy way to speed this up. Does `names.h` need to be compiled into the binary? Maybe there is another way that does not require too much effort.

kovidgoyal:
Why do you care so much about compile times? kitty's build system
supports incremental compilation, so while developing you just need to
build the changed files.

Luflosi:
You're right, maybe I shouldn't care too much about that one file. And yeah, most of the time, I can use the incremental compilation (which this patch broke a little but I'll fix that before marking it ready for merging).

Luflosi:
In the current code, you output some files with filenames ending in `.temp.so` and then rename them to `.so` later. What is the reason for doing this compared to directly using `.so`?

kovidgoyal:
atomic updates.

Luflosi:
I assume this is not needed for the `.o` files?

kovidgoyal:
No it is not. .so files can be read by existing or newly launched kitty processes, so it is important they are always in a consistent state.

Luflosi:
In `glfw.build_wayland_protocols()` one of the command line arguments to `wayland-scanner` starts with a `//` (at least in my Manjaro VM). Is that intended?
```
['/usr/bin/wayland-scanner', 'client-header', '//usr/share/wayland-protocols/stable/xdg-shell/xdg-shell.xml', '/path/to/kitty/glfw/wayland-xdg-shell-client-protocol.h']
```

kovidgoyal:
I doubt it's intended, // is exactly the same as / as far as path
interpretation goes, so I see no reason why it would be intended.

Luflosi:
How can I write tests for this? Do you even want tests for this?

kovidgoyal:
No, I dont see a need for tests for the build system.

setup.py
```diff
@@ -493,38 +573,40 @@ def files(kitten, output, extra_headers=(), extra_sources=(), filter_sources=Non
...
     }
     env = init_env(args.debug, args.sanitize, native_optimizations, args.profile, args.extra_logging)
     try:
-        compile_c_extension(
-            kitty_env(), 'kitty/fast_data_types', args.incremental, compilation_database, all_keys, *find_c_files()
-        )
-        compile_glfw(args.incremental, compilation_database, all_keys)
-        compile_kittens(args.incremental, compilation_database, all_keys)
-        for key in set(compilation_database) - all_keys:
-            del compilation_database[key]
```
Luflosi: @kovidgoyal could you please explain what the purpose of these two lines is?

kovidgoyal: they remove stale entries from the compilation database. For example if a file was deleted/reamed.

Luflosi:
When an error occurs while compiling something, your code runs the command again. I wanted to avoid that, so I use a pipe with every call to the compiler so that I can print the error message later. Unfortunately this results in the text not being formatted anymore (no colors/bold). This is probably caused by the compiler detecting that it is outputting to a pipe and not a tty. If this is the case, there must be a command line option to force the formatted output. I haven't looked for such an option yet. I'd like to know if you see a better option.

kovidgoyal:
Well if you dont want to run the particular compilation again (I dont quite understand why), you have two options:

1) Find a compiler specific option to force output formatting
2) Allocate a new tty and set the stdout/stderr of the compiler process to that and set its size using TIOCWINSZ to be the same as the current terminal and then read output from that.

Luflosi:
I don't want to run the command again because when testing, I want to get error messages as soon as possible and running a command twice takes more time.
I went with the first option, gcc has the option `-fdiagnostics-color=always` and clang has `-fcolor-diagnostics`. The output still looks slightly different though (at least for clang) because the compiler doesn't know the width of the screen and formats the output a little differently. Is that an issue?

kovidgoyal:
To me getting proper output on errors is far more important than the
extra time taken to compile a single file twice. So I would prefer
either running it again or using a pseudo-tty. You might also be able to
set the LINES and COLS environment variables. If clang reads them then
that would solve the issue.

Luflosi:
I tried implementing the second option, however I ran into a problem. I created a new pty with `master, slave = pty.openpty()` and used `w = subprocess.Popen(cmd, stdout=subprocess.DEVNULL, stderr=slave)` instead of `subprocess.Popen(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE)`. After that change `os.wait()` no longer works as before, probably because there is some buffer that becomes full or even no buffer at all. Replacing it with `select.select(workers, [], [])` almost works except that when the child process exits, I don't know how to detect that and not get stuck in the `select()` call forever. Could you help me with that please?

kovidgoyal:
You need to create a thread to drain the master fd. You also need to use
the TIOCWINSZ ioctl to set the geometry of the pseudo-terminal to match
that of the real terminal.

Luflosi:
Setting the geometry works. Is there no other way to drain the master fd than to use one thread per fd?

kovidgoyal:
No there are many ways to do it. Setup a single drain thread and pass
the fds to it. If you dont want to use threads at all, then instead of
using os.wait() use a SIGCHLD signal handler with asyncio which monitors
all the master fds as well. This is the "best" approach IMO

See https://docs.python.org/3/library/asyncio-eventloop.html#unix-signals

Luflosi:
When I add the SIGCHLD signal handler, how can I see which child exited and what the exit status was?

kovidgoyal:
https://stackoverflow.com/questions/2595503/determine-pid-of-terminated-process or just poll all children

Luflosi:
Is it ok to move `compile_commands.json` into the `build` directory?

kovidgoyal:
Should be fine

Luflosi:
Can I use relative paths for compilation? It would simplify the code a little bit and shorten the compilation command and a few lines in `compile_commands.json`. But maybe there are some edge cases or side effects I didn't see yet.

kovidgoyal:
I prefer absolute paths, it means things are safe against anything doing
a chdir() anywhere. However, if you can ensure there are no chdirs()
anywhere in the build code, it should be safe.

Luflosi:
I looked at the four `os.chdir()`s in `setup.py`. One is there to correctly set the directory before everything else, one is only used for `make clean`, and two are used after everything else is done, when assembling the macOS app bundle. I also looked at all other instances of `os.chdir(` anywhere else in the code but didn't see anything that might get called when building.

kovidgoyal:
OK, I guess we can use relative paths.

Luflosi:
The CI fails with the error message
```
Error: Your Homebrew is outdated. Please run `brew update`
```
How do I fix this?

kovidgoyal:
Add brew update to the CI file

kovidgoyal:
Never mind I took care of it.

Luflosi:
I think this is pretty much how I'd like it to be. More testing would be good though.
Since this is a massive change to review and quite a few changes are independent from each other, I will make separate pull requests for those changes first. That way these small changes will be easy to review and discuss.
I'd also like to discuss a few things with you, maybe you have a few ideas on how to make my code better. There is also a 125 lines long function, which could be improved.
Please don't merge this yet.

kovidgoyal:
@Luflosi sure, go ahead. Close this PR and make separate smaller ones.

PR #1574, opened 2019-04-26
Add command to disable/enable ligatures per window

blahgeek:
Font ligatures are good for code editing, but not good for some other terminal applications (e.g. htop).

This patch allows user to disable/enable ligatures dynamically, per window.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -1051,3 +1051,9 @@ class Boss:
             dbus_notification_activated(*args)
         else:
             dbus_notification_created(*args)
+
+    def set_disable_ligatures(self, value, window=None):
+        if window is None:
+            window = self.active_window
+        window.screen.disable_ligatures = value
+        window.screen.dirty_sprite_positions()
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -1052,7 +1052,7 @@ class Boss:
         else:
             dbus_notification_created(*args)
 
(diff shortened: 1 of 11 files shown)


kovidgoyal:
Looks fine, some feedback:

1) This should store a per window "ligature strategy" variable that should be initialised form the global option

2) There should be a way to define a keyboard shortcut to change the ligature strategy for the current window/tab/all windows

3) The remote control command should allow defining windows to change via  --match option, just like many other commands that affect windows take.

4) Not clear to me why you have exposed dirty_sprite_positions to python code, probably for debugging? That should be removed.

blahgeek:
Thanks for your feedback.

> * This should store a per window "ligature strategy" variable that should be initialised form the global option
> * There should be a way to define a keyboard shortcut to change the ligature strategy for the current window/tab/all windows
> * The remote control command should allow defining windows to change via  --match option, just like many other commands that affect windows take.

OK, I would implement these when I got time.

> 4. Not clear to me why you have exposed dirty_sprite_positions to python code, probably for debugging? That should be removed.

It's for redrawing the screen. Is there a better way already?

kovidgoyal:
Ah I did not notice you were using it in boss.py -- yes, that is the correct way to get kitty to redraw the screen when the option value changes.

blahgeek:
@kovidgoyal updated

Issue #1098, opened 2018-10-26
Some font families can't be loaded


matj1:
When I set `font_family` in config to some values, it doesn't load the fonts and loads DejaVu Sans Mono instead. I've checked the names with `fc-list` and they should be correct.

Fonts families that were loaded:
- `DejaVu Sans Mono`
- `Source Code Pro`
- `Liberation Mono`
- `Fira Code`
- `Fira Mono`
- `Ubuntu Mono`

Font families that I couldn't get to load:
- `Linux Libertine Mono`
- `Latin Modern Mono`
- `Noto Sans Mono`

kovidgoyal:
kitty will only use fonts that are detecxted as monospace by fontconfig, i.e. have spacing=100. Use kitty list-fints to see which fonts are correct or fc-list : family spacing | grep spacing=100

matj1:
Is there a way to load fonts not listed as monospaced?

Luflosi:
I'm pretty sure they need to be monospaced and there is no way around that.

PR #1576, opened 2019-04-27
Mentioned `kitty list-fonts` command

g4s8:
It was really hard to find why some font didn't work, I tried to use `Hack Nerd Font` as in `fc-list` output, but kitty accepts another name: `Hack Regular`.
I found this command `kitty list-fonts` in https://github.com/kovidgoyal/kitty/issues/1098#issuecomment-433576219 but not in docs and not in `man kitty`. So I think it should be mentioned in font related docs.

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -199,6 +199,7 @@ o(
     'monospace',
     long_text=_('''
 You can specify different fonts for the bold/italic/bold-italic variants.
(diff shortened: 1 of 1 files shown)


g4s8:
@Luflosi thanks, please check https://github.com/kovidgoyal/kitty/pull/1576/commits/83482873dc456856274de7d3faa552413de37657

PR #1584, opened 2019-05-01
Micro-optimization

Luflosi:
I'm pretty sure `(void)exe_dir_;` doesn't actually do anything besides suppress a warning about an unused variable. Since `exe_dir_` is used in the `#ifdef __APPLE__` block, this statement can be moved into an `#else` block.


kovidgoyal:
(void)exe_dir_; will generate no assembly instructions. It is a no-op, so there is no need to put it in an else block.

PR #1627, opened 2019-05-19
Rename setup.py argument kitty.app -> app to match argument of make (make app)

Luflosi:
I kept the `kitty.app` argument for backwards compatibility.


kovidgoyal:
I prefer kitty.app it is more clear. app is generic and could mean
anything.

PR #1638, opened 2019-05-24
Fix warning with gcc 4.8

Luflosi:
I think this is the correct fix because #1637 fails the CI tests.
@jreybert please test this.


kovidgoyal:
I dont see why this would work, the warning as far as I understand is because on old systes the bitmap rows/width members are using signed integer types. So we need to cast them for compatibility.

Luflosi:
The bitmap rows/width members are of type `size_t`, so also using `size_t` in the loops should fix the warning, no matter if `size_t` is signed or unsigned.

kovidgoyal:
size_t is always unsigned and FT_Bitmap uses unsigned int in current
freetype versions https://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_bitmap

Luflosi:
Sorry, I thought that `bitmap` was of type `ProcessedBitmap` instead of `FT_Bitmap`, having `size_t` as member type for the two fields.

PR #1643, opened 2019-05-24
Add `@autoreleasepool` blocks to Cocoa code from glfw upstream

Luflosi:
I basically searched for all occurrences of `autoreleasepool` in the glfw code and looked if kitty also had them. I added them where they were missing and also in a few glfw methods that are kitty specific. I hope I caught all those methods and didn't miss any. Unfortunately I feel like there is a fairly high probability that I missed one. Is there like a leak checker or something for Objective-C?


Luflosi:
lol, I pinged the user autoreleasepool.

kovidgoyal:
kitty does not need autorelease pools. This is because it uses the native cocoa run loop. kitty/glfw code is executed only in callbacks from that run loop. The run loop sets up the autorelease pool for us.

PR #1657, opened 2019-05-28
fix ansi pager history santizer

### LINKED ISSUE #1642: The formatting for `map f1 pipe @ansi overlay less +G -R` is off on macos
Using `kitty_mod+h` to `show_scrollback`, my scrollback looks like:

Screen Shot 2019-05-24 at 09 43 13

Using `f1` to `pipe @ansi overlay less +G -R -n`, my scrollback looks like:
Screen Shot 2019-05-24 at 09 43 02

nerdrew:
fixes #1642

### DIFF
--- a/kitty/window.py
+++ b/kitty/window.py
@@ -123,8 +123,8 @@ def text_sanitizer(as_ansi, add_wrap_markers):
     def remove_both(line):
         return pat.sub('', line.replace('\r', ''))
 
-    if as_ansi:
-        return remove_both if add_wrap_markers else remove_sgr
+    if not as_ansi:
+        return remove_both if not add_wrap_markers else remove_sgr
     return remove_wrap_markers
 
 
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -874,7 +874,7 @@ class Boss:
                         'KITTY_PIPE_DATA':
                         '{scrolled_by}:{cursor_x},{cursor_y}:{lines},{columns}'.format(**pipe_data)
                     }
-                stdin = stdin.encode('utf-8')
+                    stdin = pipe_data['text']
         return env, stdin
 
     def special_window_for_cmd(self, cmd, window=None, stdin=None, cwd_from=None, as_overlay=False):


kitty/window.py
```diff
@@ -123,7 +123,7 @@ def remove_sgr(line):
     def remove_both(line):
         return pat.sub('', line.replace('\r', ''))
 
-    if as_ansi:
+    if not as_ansi:
         return remove_both if add_wrap_markers else remove_sgr
```
nerdrew: I can't tell if the logic for `add_wrap_markers` also needs to change.

nerdrew:
The pager wrap still has `^M` characters in it for me. The `stdin = pipe_data['text']` change fixed it for me, though I'm not sure what deleting `stdin = stdin.encode('utf-8')` will break.

Here are screen shots of the various states for me. They all use the following invocation:

```
python3 . -o "map=f1 pipe @text_wrap overlay less -R -n" -o "scrollback_lines 50" -o "map=f2 pipe @ansi_wrap overlay less -R -n" -o "map=f3 pipe @text overlay less -R -n"
```

And this command to create the scrollback:

```
echo aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa; for i in $(seq 0 100); echo $i
```

Regular `show_scrollback` (looks correct):
Screen Shot 2019-05-28 at 21 15 44

`pipe @text_wrap overlay less -R -n` (has spurious `^M` characters and text is not wrapped, less does the wrapping):
Screen Shot 2019-05-28 at 21 16 15

`pipe @ansi_wrap overlay less -R -n` (color looks right, but again has spurious `^M` characters and text is not wrapped, less does the wrapping):
Screen Shot 2019-05-28 at 21 17 13

`pipe @text overlay less -R -n` (looks right):
Screen Shot 2019-05-28 at 21 17 24

kovidgoyal:
As noted here: https://sw.kovidgoyal.net/kitty/pipe.html

adding _wrap means that the output will have carriage returns in it. This is meant for programs that want to do wrapping themselves, typically when the screen they are writing to is of different size than the kitty window. If you want to pie to less in an overlay, dont use _wrap just use @ansi and you should be fine. There wil be no wrap markers causing less to do the wrapping itself, which should exactly match the kitty window, since the overlay window has the same size.

nerdrew:
Ah! I missed that part of the documentation. That totally makes sense. Thanks for the awesome software!

PR #1668, opened 2019-05-31
Correctly draw spaces around tab separators.

ddddavidmartin:
Good day,

Before, the tab bar would end right before the separator to the next tab.
We have to take the actual separator with spaces on both sides into account though. Basically if the tab separator is '  |  ' it was before:

 `shortened_tab_title…|  next_tab_title_with_leading_spaces`

And it is now

 `shortened_tab_tit…  |  next_tab_title_with_leading_spaces`

This now respects what the tab separator is set to.

Thanks!


kovidgoyal:
Thanks, I fixed it slightly differently.

PR #1670, opened 2019-06-02
Stack layout: Add align_top_left option

### LINKED ISSUE #505: Vertically align content to top instead of center
As reported in #416, when the window height is not a multiple of cell height the content is centered veritcally, and thus produces a row of empty pixels at the top and bottom.

I think it would be better if it were aligned to the top, and the empty pixels were all at the bottom. This seems to be how other terminals handle this.

copy:
Fixes #505.

Needs some documentation, but I'll wait for a review first.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -28,7 +28,7 @@ from .fast_data_types import (
     toggle_maximized
 )
 from .keys import get_shortcut, shortcut_matches
-from .layout import set_draw_borders_options
+from .layout import set_layout_options
 from .remote_control import handle_cmd
 from .rgb import Color, color_from_int
 from .session import create_sessions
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I dont think this makes sense as a layout option. Why should it be available only for the stack layout? If you do want to make it per layout, then the there should be a global setting, overridable, per layout.

copy:
You're right, I hadn't considered that layouts may be switched. I've moved it to a global option.

blunckr:
I'm confused, do I just add `align_top left yes` to my kitty.conf. I've done that and it seems to have no effect. I'm running 0.14.4, which has this code in it.

copy:
@blunckr The option was later renamed to [`placement_strategy`](https://sw.kovidgoyal.net/kitty/conf.html#opt-kitty.placement_strategy).

blunckr:
That works, thank you!

PR #1674, opened 2019-06-03
Use "with suppress()" to suppress python exceptions

Luflosi:
Using
```Python
with suppress(OSError):
    os.remove('somefile.tmp')
```
instead of
```Python
try:
    os.remove('somefile.tmp')
except OSError:
    pass
```
makes the code more compact and more readable IMO.

This pattern was recommended by Raymond Hettinger, a Python Core
Developer in his talk "Transforming Code into Beautiful, Idiomatic Python" at https://www.youtube.com/watch?v=OSGv2VnC0go. The transcript is available at https://github.com/JeffPaine/beautiful_idiomatic_python

### DIFF
--- a/kittens/ask/main.py
+++ b/kittens/ask/main.py
@@ -3,6 +3,7 @@
 # License: GPL v3 Copyright: 2018, Kovid Goyal <kovid at kovidgoyal.net>
 
 import os
+from contextlib import suppress
 
 from kitty.cli import parse_args
 from kitty.constants import cache_dir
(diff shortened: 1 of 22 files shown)


Luflosi:
Did you know about this feature? I had never seen it before I watched the talk.

kovidgoyal:
Yes, though I am not particularly impressed by it :)

If you want to discover lots of little goodies like this, read the docs
for the contextlib, functools and iterators modules.

PR #1675, opened 2019-06-03
Add more detailed instructions for packagers to change `update_check_interval`

Luflosi:
I think this would be very helpful for packagers that don't know the code base.
I'm a total rst noob and don't know how to put the `code-block` inside the `note`, maybe you can do that.


kovidgoyal:
I dont think a patch is a good idea, too much might change in that file. Instead just use something like, find update_check_interval in the config_data.py and change the default value from 24 to 0.

nightuser:
Maybe it should be set from environment variable during compilation?

Luflosi:
Yeah, that's a better solution.

PR #1689, opened 2019-06-05
Remove double assignment

Luflosi:
The `payload` variable is assigned twice.
Static program analysis FTW!


kovidgoyal:
Which static analyzer are you using?

Luflosi:
The one built into PyCharm CE.

PR #1731, opened 2019-06-21
Move definition of GL_SILENCE_DEPRECATION

Luflosi:
Move definition of GL_SILENCE_DEPRECATION from `setup.py` and `glfw/glfw.py` to `cocoa_platform.m`. This is where GLFW defines it.

### DIFF
--- a/glfw/cocoa_platform.h
+++ b/glfw/cocoa_platform.h
@@ -28,6 +28,11 @@
 #include <dlfcn.h>
 
 #include <Carbon/Carbon.h>
+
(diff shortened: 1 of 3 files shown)


kovidgoyal:
There is no guarantee that that is the only place opengl functions are used. Indeed, kitty itself uses plenty of OpenGL calls which will cause deprecation warnings. I prefer to turn it on/off globally.

Luflosi:
Weird, there are no warnings on my machine or in the CI. Do you know why this happens?

PR #1734, opened 2019-06-22
Use SPECIAL_SOURCES to set the git hash

Luflosi:
Using the already existing functionality of `SPECIAL_SOURCES` makes the code cleaner IMO.
The `rev` variable pollutes the global namespace. Is there an easy way to prevent that or should we just ignore it?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -306,6 +306,9 @@ def run_tool(cmd, desc=None):
 SPECIAL_SOURCES = {
     'kitty/parser_dump.c': ('kitty/parser.c', ['DUMP_COMMANDS']),
 }
+if os.path.exists('.git'):
+    rev = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()
+    SPECIAL_SOURCES.update({'kitty/data-types.c': ('kitty/data-types.c', ['KITTY_VCS_REV="{}"'.format(rev)])})
 
 
 def newer(dest, *sources):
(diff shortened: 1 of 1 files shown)


Luflosi:
Yes, this is better than my patch.
I do however see one small issue with it. Are you sure, it's a good idea to manually read from `.git/refs/heads/master` if git is not installed? git has probably no guarantees, that the git repo will always have exactly this structure. I think we should just return an empty list, like when there is no git repository. The probability of someone downloading a git repository without having git installed is probably pretty low anyways.

kovidgoyal:
If a git repo does not have that structure, which I find highly unlikely, it will fail, and we can fix it. I dont want to depend on the git binary being installed if we dont have to.

trygveaa:
`.git` can be a file pointing to the git directory, for instance if it's a secondary worktree. Not likely if git isn't installed though, but for completeness:

```
trygve@george:~/dev/kitty$ git worktree add ../tmp-kitty
trygve@george:~/dev/kitty$ cd ../tmp-kitty
trygve@george:~/dev/kitty$ cat .git
gitdir: /home/trygve/dev/kitty/.git/worktrees/tmp-kitty
trygve@george:~/dev/kitty$ ./setup.py
  File "./setup.py", line 309, in get_vcs_rev_defines
    rev = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()
  File "/usr/lib/python3.7/subprocess.py", line 395, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.7/subprocess.py", line 472, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'git': 'git'

During handling of the above exception, another exception occurred:

  File "./setup.py", line 926, in <module>
    main()
  File "./setup.py", line 896, in main
    build(args)
  File "./setup.py", line 530, in build
    kitty_env(), 'kitty/fast_data_types', args.incremental, compilation_database, all_keys, *find_c_files()
  File "./setup.py", line 408, in compile_c_extension
    defines = defines()
  File "./setup.py", line 311, in get_vcs_rev_defines
    with open('.git/refs/heads/master') as f:
NotADirectoryError: [Errno 20] Not a directory: '.git/refs/heads/master'
```

PR #1739, opened 2019-06-23
Always clean files manually instead of using git

Luflosi:
I see a couple problems when using git to determine the files to delete:
- git needs to be installed
- The manual cleaning code is supposed to function correctly, so why not always use it? That way bugs in the manual cleaning code are also more likely to get discovered
- If some files ignored by git should not be deleted, they need to be added to a sort of blacklist. If some files should be deleted by the manual cleanup code, they need to be added to a sort of whitelist. If someone forgets to add files to one of these lists, then not deleting files that should be deleted is better than deleting files that should not be deleted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -780,17 +780,6 @@ Categories=System;TerminalEmulator;
 
 def clean():
     os.chdir(os.path.dirname(os.path.abspath(__file__)))
-    if os.path.exists('.git'):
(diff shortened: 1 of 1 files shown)


Luflosi:
I'm opening a new pull request as the branch name was not what I intended it to be.

PR #1740, opened 2019-06-23
Always clean files manually instead of using git

Luflosi:
I see a couple problems with using git to determine the files to delete:
- git needs to be installed
- The manual cleaning code is supposed to function correctly, so why not always use it? That way bugs in the manual cleaning code are also more likely to get discovered
- If some files ignored by git should not be deleted, they need to be added to a sort of blacklist. If some files should be deleted by the manual cleanup code, they need to be added to a sort of whitelist. If someone forgets to add files to one of these lists, then not deleting files that should be deleted is better than deleting files that should not be deleted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -780,17 +780,6 @@ Categories=System;TerminalEmulator;
 
 def clean():
     os.chdir(os.path.dirname(os.path.abspath(__file__)))
-    if os.path.exists('.git'):
-        for f in subprocess.check_output(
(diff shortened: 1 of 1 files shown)


Luflosi:
I had another good reason to not use git for cleaning, that I forgot to mention:
git doesn't really track directories. The git command used for listing the files won't list any empty directories. It could happen that after cleaning, an empty `build` directory and an empty `kitty.app` (with an empty directory structure inside) were left.

PR #1751, opened 2019-06-26
Micro-optimization

Luflosi:


kovidgoyal:
How is this an optimization?

Luflosi:
When `args.action` is `test`, it isn't checked if it is also `clean` or any of the later options. The compilation database is also not read from disk and written back when `args.action` is `test`. I was also able to get rid of the return statement.
Instead of what I did in this commit, I could also add a `return` to the end of the `if args.action == 'test':` block, if you prefer that.

kovidgoyal:
Nothing beyond os.execlp is executed, the current process gets entirely
replaced by a new one at that point, so there is no need for this.

Luflosi:
Oh, ok, I didn't actually really read that line 😅.
Sorry about the noise then.

Issue #893, opened 2018-09-06
Scrollback search


prawnsalad:
Coming from iterm2 the biggest thing I'm missing is iterms buffer search.
![image](https://user-images.githubusercontent.com/725880/45170954-13aedc80-b1f9-11e8-9105-d3eb0efe1a11.png)

While any process is running in the shell, it highlights any occurrence of the search term so that we can easily find parts of text we are interested in in real-time without effecting the commands being run in the shell.

spl:
I was also looking for this answer. It might help to reorganize this bit of documentation. Having it under the heading “The scrollback buffer” was not obvious to me. Perhaps it putting it as a FAQ with a title that included the (key-)words “search,” “find,” and “text” would help.

kovidgoyal:
Feel free to send a PR with improvements, relevant file is docs/faq.rst

spl:
@prawnsalad Oh right, that is a useful feature in iTerm. It's also helpful to have [triggers](https://www.iterm2.com/documentation-triggers.html). In particular, I use them to highlight words like “error” and ”warning” in the running output.

kovidgoyal:
That's never going to happen. Searching for text live in program output is a huge performance killer. It means you have to re-run the search expression on every received byte. kitty prioritizes performance.

And if you really want to do it for an already running process, simply use a tool like https://github.com/jerome-pouiller/reredirect/

Though really, if you want to monitor a process you should pipe its output into a monitor program when you invoke it. The monitor program can do arbitrarily powerful things.

kovidgoyal:
Thinking about it, one way to do this without impacting kitty performance at all is to create a live output kitten, which would work like the current scrollback viewing facility works but would have kitty tee the output it receives from the process into its stdin, so it would be live and you could configure it to run any program/script of your choice to search for things/highlight things whatever.

felixfbecker:
So <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>H</kbd> opens the scrollback in the default pager, e.g. `less`. But what if I have a command running (like a watching compiler, or a server) that is streaming output? iTerm lets me easily search the output of those too, but it doesn't seem to work with Kitty.

kovidgoyal:
You can search the current contents of the scrollback buffer and the
screen at anytime by pressing kitty_mod+h kitty does not care if the
output is live or not. You will simply get to search whatever the output
is at the moment you pressed the keys.

s1341:
[...]

PR #1780, opened 2019-07-05
Implement basic support for 'live' kittens

s1341:
This type of kitten receives a 'tee fd' of the output of a given window as stdin. Example uses are 'live highlighting' similar to what iterm2 does.

See discussion on issue #893

### DIFF
--- a/kittens/live_demo/main.py
+++ b/kittens/live_demo/main.py
@@ -2,27 +2,42 @@
 # vim:fileencoding=utf-8
 # License: GPL v3 Copyright: 2018, Kovid Goyal <kovid at kovidgoyal.net>
 
+from contextlib import suppress
 import fcntl
 import os
 import select
 import sys
-from ..tui.operations import styled
+from ..tui.operations import alternate_screen, styled
 
 
 def main(args):
-    text = ''
     if sys.stdin.isatty():
         if '--help' not in args and '-h' not in args:
             print('You must pass the text to be hinted on STDIN', file=sys.stderr)
             input('Press Enter to quit')
             return
 
+    global readline
+    import readline as rl
+    readline = rl
+    from kitty.shell import init_readline
+
+    init_readline(readline)
+
+    sys.stdin = open(os.ctermid())
+    term = ""
+    with alternate_screen():
+        with suppress(KeyboardInterrupt, EOFError):
+            term = input("Enter search term: ")
+    if term == "":
+        return
+
     fcntl.fcntl(sys.__stdin__.fileno(), fcntl.F_SETFL, os.O_NONBLOCK)
     try:
         while True:
             if select.select([sys.__stdin__], [], [], 0) == ([sys.__stdin__], [], []):
                 text = sys.__stdin__.read()
-                text = text.replace("hello", styled("hello", bold=True, bg="green", fg="black"))
+                text = text.replace(term, styled(term, bold=True, bg="green", fg="black"))
                 sys.stdout.write(text)
     except KeyboardInterrupt:
         pass
@@ -34,7 +49,7 @@ def handle_result(args, data, target_window_id, boss):
     pass
 
 
-handle_result.type_of_input = 'live'
+handle_result.type_of_input = 'live-history'
 
 
 if __name__ == '__main__':
--- a/kittens/live_demo/main.py
+++ b/kittens/live_demo/main.py
@@ -5,6 +5,7 @@
 from contextlib import suppress
 import fcntl
 import os
+import re
 import select
 import sys
 from ..tui.operations import alternate_screen, styled
@@ -28,18 +29,23 @@ def main(args):
     term = ""
     with alternate_screen():
         with suppress(KeyboardInterrupt, EOFError):
-            term = input("Enter search term: ")
+            term = input("Enter search regex: ")
     if term == "":
         return
 
     fcntl.fcntl(sys.__stdin__.fileno(), fcntl.F_SETFL, os.O_NONBLOCK)
+
+    def replace_func(matchobj):
+        return styled(matchobj.group(0), bold=True, fg="black", bg="green")
+
     try:
         while True:
             if select.select([sys.__stdin__], [], [], 0) == ([sys.__stdin__], [], []):
                 text = sys.__stdin__.read()
-                text = text.replace(term, styled(term, bold=True, bg="green", fg="black"))
+                text =  re.sub(term, replace_func, text, count=0)
                 sys.stdout.write(text)
     except KeyboardInterrupt:
+        sys.stderr.write("exiting\n")
         pass
     return
 
--- a/kittens/live_demo/main.py
+++ b/kittens/live_demo/main.py
@@ -2,6 +2,8 @@
 # vim:fileencoding=utf-8
 # License: GPL v3 Copyright: 2018, Kovid Goyal <kovid at kovidgoyal.net>
 
+import fcntl
+import os
 import select
 import sys
 from ..tui.operations import styled
@@ -15,6 +17,7 @@ def main(args):
             input('Press Enter to quit')
             return
 
+    fcntl.fcntl(sys.__stdin__.fileno(), fcntl.F_SETFL, os.O_NONBLOCK)
     try:
         while True:
             if select.select([sys.__stdin__], [], [], 0) == ([sys.__stdin__], [], []):
--- a/kittens/live_demo/main.py
+++ b/kittens/live_demo/main.py
@@ -6,6 +6,7 @@ import select
 import sys
 from ..tui.operations import styled
 
+
 def main(args):
     text = ''
     if sys.stdin.isatty():
--- a/kittens/live_demo/main.py
+++ b/kittens/live_demo/main.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+# vim:fileencoding=utf-8
+# License: GPL v3 Copyright: 2018, Kovid Goyal <kovid at kovidgoyal.net>
+
+import sys
+from ..tui.operations import styled
+
+
+def main(args):
+    text = ''
+    if sys.stdin.isatty():
+        if '--help' not in args and '-h' not in args:
+            print('You must pass the text to be hinted on STDIN', file=sys.stderr)
+            input('Press Enter to quit')
+            return
+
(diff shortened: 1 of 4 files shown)


s1341:
History is not working for some reason... Not sure why...

kovidgoyal:
I'll look over the code when I have a moment, but if what you really want to do is highlight text in the output of whatever is running in the window, then the right way to do it, is instead of having a live kitten, have a facility for running filter programs. So launch a filter program, replace the childs output fd by the filter programs output fd in the Child struct and set the filter programs input fd to the childs output fd.

The filter program can be any arbitrary executable and can do whatever you like, highlight text, transform it, etc. There will need to be some facility for killing the filter program and resetting the childs output fd in Child struct as well. So that filtering mode can be turned off. The existing screen and scrollback contents can be written to a temp file and passed to the filter program.

s1341:
Filter programs sound like a neat idea. I can give it a try next Friday... Do you have anything specific in mind?

kovidgoyal:
You'd have to write your own filter programs, I dont know of any
existing off-the-shelf solutions.

s1341:
I guess that the simplest filter program would be `grep --passthrough <pattern>`.

I think we should add a 'filter' command to kitty, so that you can install/remove a filter quickly. Something like this:
```
kitty + filter install 'grep --passthrough hello'
```
Then to kill the filter, you could do:
```
kitty + filter remove
```
What do you think?

kovidgoyal:
You would want to install the filter on a specific window, not kitty
generally, so there would need to be a way to define mappings for
installing filters that are defined in kitty.conf and maybe a kitten to
ask the user to specify a filter executable. Although the last is not
really needed IMO.

s1341:
I figured the 'install' command would work on the current window.

A mapping is a good idea, but how would you specify things like the search term in the mapping?

kovidgoyal:
map f1 install_filter some-filter-program search-term


if you want to dynamically specify the search term then

map f1 install_filter some-filter-program 

which will run the ask kitten to ask for the command line
to pass to the filter program

s1341:
Ok. I have some time now. I'll look into adding filter support.

I think that the live kitten has merit in any case, so I'd appreciate it if you'd review the code and give me feedback so we can merge. 

I'm still not 100% clear as to how to go about using the response from the ask kitten in the install_filter mapping, but that's something that can be dealt with later, I suppose.

kovidgoyal:
Sadly I am totally swamped, so it is going to be a while before I
can review your code.

s1341:
Ok. No worries. I'll look into filters meanwhile.

s1341:
Ok. Here is my plan for the filters:
1. Add a 'set_child_fd' python-accessible function, with id and fd parameters, to child-monitor.c. This function will:
  * set the 'fd' in the Child struct to the passed fd.
  * replace the appropriate fd in the fds variable with the passed fd.
2. Modify the Child.fork function so that if self.stdin is an int, it will install that int as the stdin_read_fd for the child. In addition, change the "if stdin is not None" after the spawn to also ignore numeric stdin values.
3. Add a new 'kill' function to the Child class in child.py. This function will:
  * send a SIGKILL signal to the child's fd, and wait for it to die.
4. Add a new 'filter' member to the Window class in tab.py.
5. Add a new function 'launch_filter(cmd, cwd=None, env=None)' to the Window class in tab.py. This function will:
   * create a new Child for the filter, passing self.child.child_fd as the stdin, and the cmd, cwd and env arguments.
   * save this new child as the filter member on the Window.
   * call fork() on the filter Child.
6. Add a new 'install_filter' function to the Boss class in boss.py. This function will:
  * call launch_filter on the specified window.
  * call child_monitor.set_child_fd passing the id of the specified window and the window.filter.child_fd
7. Add a new 'remove_filter' function to the Boss class in boss.py. This function will receive a window id, and do the following:
  * call child_monitor.set_child_fd, passing the id of the specified window and the window.child.child_fd.
  * call window.filter.kill()
  * set window.filter to None
8. Add new 'cmd_install_filter' and 'install_filter' functions to cmd.py, allowing one to set a filter program from a kitty repl/cli. These functions will call the install_filter function on the Boss object. They will allow one to specify a 'match' for a window, and a filter command/args.
9. Add new 'cmd_remove_filter' and 'remove_filter' functions to cmd.py, allowing one to remove the current filter program from a kitty repl/cli. These functions will call remove_filter on the Boss object. They will allow one to specify a 'match' for a window.

How does this look?

A few things:
1. I don't know how I'm going to pass the history/screen to the filter... need to think about that. 
2. I haven't yet understood exactly how to allow configuration of key-mappings to run install_filter/remove_filter.
3. I'm not 100% certain about the kill mechanism... do you have a better idea than just SIGKILLing the filter program?

s1341:
I didn't manage to make much progress, mostly because I got thoroughly confused between input and output, and because my filter programs seem to be dying immediately as they are spawned (using the Child.fork function to spawn them).

This raises two issues:
1. Why are my filter programs dying?
2. How are we planning to deal with premature death of a filter program (which would leave stale fd's in kitty's state)?

I'll work on this more next week.

kovidgoyal:
When a program dies the kernel closes fds automatically, so the end of
the pipe kity has will fail with EPIPE or similar, at which point you
just assume the filter has dies and restore the old fds.

s1341:
ok. I'll look into it next Friday.

s1341:
Hi,

So I'm a little stuck, when I try to use the spawn function in child.c to spawn my filter program, it dies immediately, and I can't figure out what I'm doing wrong. Do you have any suggestions to help debug this? I'm on OSX...

kovidgoyal:
dies with what error code? and what program are you runnig?

s1341:
I don't get an error code... I've tried running:
```
ag --passthrough hello
```
And 
```
python highlight.py
```

Where highlight.py is:
```
#!/usr/bin/env python3
import select
import sys


STANDARD_COLORS = {name: i for i, name in enumerate(
    'black red green yellow blue magenta cyan gray'.split())}
STANDARD_COLORS['white'] = STANDARD_COLORS['gray']
UNDERLINE_STYLES = {name: i + 1 for i, name in enumerate(
    'straight double curly'.split())}


def color_code(color, intense=False, base=30):
    if isinstance(color, str):
        e = str((base + 60 if intense else base) + STANDARD_COLORS[color])
    elif isinstance(color, int):
        e = '{}:5:{}'.format(base + 8, max(0, min(color, 255)))
    else:
        e = '{}:2:{}:{}:{}'.format(base + 8, *color)
    return e


def colored(text, color, intense=False, reset_to=None, reset_to_intense=False):
    e = color_code(color, intense)
    return '\033[{}m{}\033[{}m'.format(e, text, 39 if reset_to is None else color_code(reset_to, reset_to_intense))


def faint(text):
    return colored(text, 'black', True)


def styled(text, fg=None, bg=None, fg_intense=False, bg_intense=False, italic=None, bold=None, underline=None, underline_color=None, reverse=None):
    start, end = [], []
    if fg is not None:
        start.append(color_code(fg, fg_intense))
        end.append('39')
    if bg is not None:
        start.append(color_code(bg, bg_intense, 40))
        end.append('49')
    if underline_color is not None:
        if isinstance(underline_color, str):
            underline_color = STANDARD_COLORS[underline_color]
        start.append(color_code(underline_color, base=50))
        end.append('59')
    if underline is not None:
        start.append('4:{}'.format(UNDERLINE_STYLES[underline]))
        end.append('4:0')
    if italic is not None:
        s, e = (start, end) if italic else (end, start)
        s.append('3'), e.append('23')
    if bold is not None:
        s, e = (start, end) if bold else (end, start)
        s.append('1'), e.append('22')
    if reverse is not None:
        s, e = (start, end) if reverse else (end, start)
        s.append('7'), e.append('27')
    if not start:
        return text
    return '\033[{}m{}\033[{}m'.format(';'.join(start), text, ';'.join(end))

out = open("/tmp/kitty.out", "wt+")
out.write("starting...\n")
while True:
    term = "hell"
    if select.select([sys.__stdin__], [], [], 0) == ([sys.__stdin__], [], []):
        text = sys.__stdin__.read()
        text = text.replace(term, styled(term, bold=True, bg="green", fg="black"))
        # sys.stdout.write(text)
        out.write(text)
        out.flush()
```

s1341:
Doh! I had a typo! the highlight.py is now not dying...

s1341:
Ok. Now I'm confused about which is the input and which is the output fd for Child processes. Is the following correct in child-monitor.c:

 - fds[EXTRA_FDS + i] is the output of the process
- child.fd is the input of the process

s1341:
Ok. For some reason, changing the fds[EXTRA_FDS + i].fd in my set_child_fd function doesn't change it in the io_loop thread. Do I need to make fds volatile?

s1341:
Ok. I finally got it to use my fd... had to add an 'update_chlidren' function and associated array to the io_loop. Now I get this error:

```
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "./__main__.py", line 117, in <module>
    main()
  File "./__main__.py", line 113, in main
    func(sys.argv[1:])
  File "./__main__.py", line 22, in remote_control
    main(args)
  File "./kitty/remote_control.py", line 175, in main
    response = do_io(global_opts.to, send, no_response)
  File "./kitty/remote_control.py", line 123, in do_io
    received = io.recv(timeout=10)
  File "./kitty/remote_control.py", line 103, in recv
    read_command_response(self.tty_fd, timeout, ans)
TimeoutError: Timed out while waiting to read command response. Make sure you are running this command from within the kitty terminal. If you want to run commands from outside, then you have to setup a socket with the --listen-on command line flag.
```
I don't really understand what is going wrong...

s1341:
Ok. I've got the filter pipeline working now. I'll open a pull request for it.

kovidgoyal:
Cool :)

s1341:
Created PR #1809

kovidgoyal:
Closing in favor of new marks feature.

PR #1785, opened 2019-07-07
Replace glad with glad2

Dav1dde:
Replaces the in Kitty embedded glad code with a glad2 build.

The changes in glad2 make the post processing of glad in the `glad/generate.py` unnecessary, since it supports installing/uninstall glad debug functions out of the box now. Also due to a header only generation the includes don't have to be modified anymore.

I did test the changes with `make test`/`test.py` and by manually starting Kitty (`__main__.py`).

Why you maybe would not want to merge this:

* I still consider glad2 "beta", it all around works pretty well but I might end up changing API once more (renaming functions)
* It's hard to test all functionality (extensions etc.) and to make sure nothing unexpected broke


Luflosi:
FYI the CI is complaining about the extra line at the end of `glad/generate.py`.

kovidgoyal:
I am always happy to get rid of code from kitty, where possible. And congrats on glad2 :)

PR #1809, opened 2019-07-19
Add the output filter capability

s1341:
This PR allows one to add an arbitrary filter program to a kitty window. All terminal output will go through the filter function, which can modify its input in any way. For example, this could be used to colorize or highlight text as it is output.

The PR adds two new kitty commands:
- launch-filter starts the passed program as a filter for a window. An arbitrary window can be specified with the --match argument, the default is the current window. The filter program should process data it receives on its stdin, and should output processed data to stdout.
- remove-filter removes the currently running filter from the specified window, defaulting to the current window.

This patch is experimental, but working. There are some known issues:
- Trying to close the window with "exit" after a filter has been specified causes the window to hang.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -126,6 +126,7 @@ class Boss:
         self.child_monitor = ChildMonitor(
             self.on_child_death,
             DumpCommands(args) if args.dump_commands or args.dump_bytes else None,
+            self.on_filter_death,
             talk_fd, listen_fd
         )
         set_boss(self)
@@ -323,15 +324,11 @@ class Boss:
             if window is not None:
                 window.send_cmd_response(response)
 
-    def _cleanup_tab_after_window_removal(self, src_tab):
-        if len(src_tab) < 1:
-            tm = src_tab.tab_manager_ref()
-            if tm is not None:
-                tm.remove(src_tab)
-                src_tab.destroy()
-                if len(tm) == 0:
-                    if not self.shutting_down:
-                        mark_os_window_for_close(src_tab.os_window_id)
+    def on_filter_death(self, window_id):
+        window = self.window_id_map.pop(window_id, None)
+        if window is None:
+            return
+        window.filter = None
 
     def on_child_death(self, window_id):
         window = self.window_id_map.pop(window_id, None)
@@ -684,7 +681,20 @@ class Boss:
             output += str(s.linebuf.line(i))
         return output
 
-    def _run_kitten(self, kitten, args=(), input_data=None, window=None, custom_callback=None, action_on_removal=None):
+    def _launch_filter(self, command, window=None):
+        if window is None:
+            window = self.active_window
+        window.launch_filter(command)
+        self.child_monitor.set_child_fd(window.id, window.filter.child_fd)
+
+    def _remove_filter(self, window=None):
+        if window is None:
+            window = self.active_window
+        self.child_monitor.set_child_fd(window.id, window.child.child_fd)
+        window.filter.kill()
+        window.filter = None
+
+    def _run_kitten(self, kitten, args=(), input_data=None, window=None):
         orig_args, args = list(args), list(args)
         from kittens.runner import create_kitten_handler
         end_kitten = create_kitten_handler(kitten, orig_args)
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -235,7 +235,6 @@ set_child_fd(ChildMonitor *self, PyObject *args) {
     }
     for (unsigned int i = 0; i < self->count; i++) {
         if (children[i].id == id) {
-            
             update_queue[update_queue_count] = children[i];
             update_queue[update_queue_count].input_fd = fd;
             update_queue_count++;
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -46,7 +46,7 @@ typedef struct {
 typedef struct {
     PyObject_HEAD
 
-    PyObject *dump_callback, *update_screen, *death_notify;
+    PyObject *dump_callback, *update_screen, *death_notify, *filter_death_notify;
     unsigned int count;
     bool shutting_down;
     pthread_t io_thread, talk_thread;
@@ -61,7 +61,9 @@ typedef struct {
 typedef struct {
     Screen *screen;
     bool needs_removal;
-    int fd;
+    bool needs_temp_removal;
+    int input_fd;
+    int output_fd;
     unsigned long id;
     pid_t pid;
 } Child;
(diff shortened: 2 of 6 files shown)


s1341:
An example filter script which highlights the term passed as an argument is shown below:
```
#!/usr/bin/env python3
import fcntl
import os
import select
import sys


STANDARD_COLORS = {name: i for i, name in enumerate(
    'black red green yellow blue magenta cyan gray'.split())}
STANDARD_COLORS['white'] = STANDARD_COLORS['gray']
UNDERLINE_STYLES = {name: i + 1 for i, name in enumerate(
    'straight double curly'.split())}


def color_code(color, intense=False, base=30):
    if isinstance(color, str):
        e = str((base + 60 if intense else base) + STANDARD_COLORS[color])
    elif isinstance(color, int):
        e = '{}:5:{}'.format(base + 8, max(0, min(color, 255)))
    else:
        e = '{}:2:{}:{}:{}'.format(base + 8, *color)
    return e


def colored(text, color, intense=False, reset_to=None, reset_to_intense=False):
    e = color_code(color, intense)
    return '\033[{}m{}\033[{}m'.format(e, text, 39 if reset_to is None else color_code(reset_to, reset_to_intense))


def faint(text):
    return colored(text, 'black', True)


def styled(text, fg=None, bg=None, fg_intense=False, bg_intense=False, italic=None, bold=None, underline=None, underline_color=None, reverse=None):
    start, end = [], []
    if fg is not None:
        start.append(color_code(fg, fg_intense))
        end.append('39')
    if bg is not None:
        start.append(color_code(bg, bg_intense, 40))
        end.append('49')
    if underline_color is not None:
        if isinstance(underline_color, str):
            underline_color = STANDARD_COLORS[underline_color]
        start.append(color_code(underline_color, base=50))
        end.append('59')
    if underline is not None:
        start.append('4:{}'.format(UNDERLINE_STYLES[underline]))
        end.append('4:0')
    if italic is not None:
        s, e = (start, end) if italic else (end, start)
        s.append('3'), e.append('23')
    if bold is not None:
        s, e = (start, end) if bold else (end, start)
        s.append('1'), e.append('22')
    if reverse is not None:
        s, e = (start, end) if reverse else (end, start)
        s.append('7'), e.append('27')
    if not start:
        return text
    return '\033[{}m{}\033[{}m'.format(';'.join(start), text, ';'.join(end))


fcntl.fcntl(sys.__stdin__.fileno(), fcntl.F_SETFL, os.O_NONBLOCK)
term = sys.argv[1]
while True:
    if select.select([sys.__stdin__], [], [], 0) == ([sys.__stdin__], [], []):
        text = sys.__stdin__.read()
        text = text.replace(term, styled(term, bold=True, bg="green", fg="black"))
        sys.__stdout__.write(text)
        sys.__stdout__.flush()
```

s1341:
@kovidgoyal I'd appreciate feedback on this PR.

kovidgoyal:
As I said, I am swamped at the moment, but rest assured that your PRs
are high up in my queue :)

s1341:
@kovidgoyal awespme. If you have any ideas about the "exit" issue, I'd be glad to hear them...

kovidgoyal:
I need more detail. What exit() function and where is it hanging?

s1341:
What I mean is trying to exit a window by typing "exit" or ctrl-d. It seems to reap_children fine, but I couldn't figure out where it actually dies...

s1341:
I mean, where it get's stuck...

kovidgoyal:
Unless you have turned on close_on_child_death in kitty.conf the filter program has to
quit for kitty to auto-close the window.

s1341:
Uh. OK. Unfortunately, this happens also when the filter program has died or been killed by remove-filter. So it's not like the filter is still blocking...

kovidgoyal:
Then I will have to read the code to tell you more.

s1341:
Ok. Thanks.

s1341:
Another thing we should consider is whether to use the master pty fd, as I've done now, or whether we should instead create a standard pipe. The standard pipe approach would allow the use of grep like utilities as filter programs. What do you think?

s1341:
In fact, we can continue to use the pty, and tools like grep/ripgrep will work, so long as we don't set the fd to NON_BLOCK. However, as these programs are line-wise, you don't see the text you are typing into the terminal until you hit return. Can you think of a workaround for this?

kovidgoyal:
No not really, you cant pattern match while live typing because then your filter can only process one character at a time. Either it buffers the characters to match patterns in which case you will have the same issue as with grep or it passes every sigle character through which means it wont work for live typing.

kovidgoyal:
If you want to support live typing then filters are the wrong architecture for this. Instead you would have to do much more in kitty. Basically have an infrastructure to scan lines as they are created/modified for an expression and mark matches in them.  You would need to add hooks to the screen_draw() funtion and the various editing functions in screen.c such as screen_erase_* to run the matcher over the modified lines.

s1341:
I'm quite happy with the filter mechanism and hightlight.py (which echos the input chars, one by one, simulating live typing but not highlighting typed input obviously). 

My only issue is that control-chars (ctrl-z, ctrl-c etc.) don't seem to work. What do I need to do to get those to work?

kovidgoyal:
ctrl-c and ctrl-z are special. When the tty device is not in raw mode,
the kernel interprets them and sends a signal SIGINT and SIGSUSP
respectively to the process connected to the terminal. since the filter
is not connected to a terminal they are passed through with no sognals.
highlight.py should pass them on and things shuld just work when the
bytes are written to the actual child processes tty fd.

s1341:
i pass on all bytes as read from the sys.__stdin__, so it should work, no?

kovidgoyal:
Are you reading from the text file or the binary, i.e.
sys.__stdin__.buffer

s1341:
Oh. Didn't know that .buffer is the binary. That seems to work. Thanks!

Should I commit highlight.py to the repo? If so, where? It doesn't belong in kittens, and I don't think it belongs in the root folder either...

kovidgoyal:
hold off for now i will decide when i review

s1341:
Ok. No problem.

s1341:
I've been using the filter support with highlight.py (which I modified to accept regex's) all week, and it's really useful. There is currently a bug where removing the filter kills the window, maybe I'll look into that tomorrow.

kovidgoyal:
Cool, will review it, I promise :) Just swamped right now working on
calibre 4 so dont have time to do any real work on kitty.

s1341:
No worries.

s1341:
Any chance you'll merge this at some point? Or at least review it?

kovidgoyal:
It's on my TODO list, apologies for the delay, I simply dont have the
time required to review a large change like this one at the moment. But
I will get to it.

s1341:
Ok. Apologies for nagging.

kovidgoyal:
Closing this in favor of the newly added marks functionality since it is
a strictly superior solution for highlighting text in a terminal window.

map f1 toggle_marker text 1 ERROR

will highlight the word ERROR in the current window when you press F1.
And the highlight will be removed when you press F1 again.

s1341:
Cool new feature. It would be great if it could ask for the regex somehow. So you’d have a generic key trigger for mark, which would then ask you what to mark. Is this doable?

kovidgoyal:
Sure, use create_marker see marks.rst

PR #1821, opened 2019-07-21
WGL: Cleanup

Luflosi:
From https://github.com/glfw/glfw/commit/6bd264244a9efbcf89dcb19541ad47d6c7f11041.


Luflosi:
On second thought, I'm wondering if this file is even used in kitty.

Luflosi:
I made #1822 in case this belongs to the windows backend, which isn't used.

PR #1828, opened 2019-07-22
Only show the hand cursor shape over a URL if the `open_url_modifiers` are pressed

Luflosi:
Don't merge this yet please.
The hand cursor suggests that whatever the mouse cursor is over at the moment can be clicked on. This is not the case for URLs, where pressing one or more modifier keys is necessary to be able to click on a URL. This pull request makes changes that only show the hand cursor if the right modifier keys are pressed.
This is very hacked together atm but I'd like to hear your opinion on this functionality first and then get some help to improve the code quality.

### DIFF
--- a/kitty/data-types.h
+++ b/kitty/data-types.h
@@ -297,6 +297,7 @@ void copy_color_table_to_buffer(ColorProfile *self, color_type *address, int off
 void colorprofile_push_dynamic_colors(ColorProfile*);
 void colorprofile_pop_dynamic_colors(ColorProfile*);
 
+void update_url_cursor_shape(MouseTrackingMode mouse_tracking_mode, int modifiers);
 void set_mouse_cursor(MouseShape);
 void enter_event(void);
 void mouse_event(int, int, int);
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I'm not really a fan of this concept. I deliberately did not implement it this way because I think of the hand cursor as an indication that something is clickable underneath, the fact that one has to press modifiers for the click to register is immaterial. This can be particularly useful in situations where the link underlining is not very visible.

derekschrock:
I currently work around this by commenting `detect_url(...)` in `kitty/mouse.c` `HANDLER(handle_move_event)` and rebuild locally.   I work with a lot of files that contain a lot of URLs and the constant flipping of the cursor is rather annoying.  I'm never going to click these URLs nor do I need to be told I'm over one.   But I'd also rather not have to rebuild on updates too.  

I was planning on creating a PR that would create a new option that would wrap that `detect_url` call in a check.

Would this be an alt. way to get to the same solution?  An opt in option that would let kitty run as expected but give the user the option to disable the hand.  Might even limit CPU too since you're not always doing string checks for url prefixes.

kovidgoyal:
In master you can now open urls without needing to press modifiers at all, so this is obsolete.

PR #1837, opened 2019-07-24
Add option to hide the window title on macOS

Luflosi:
See #1072.
Don't merge this yet please.
I have a couple questions:
Do I need the autoreleasepool in `cocoa_hide_window_title()`?
Why are there seemingly two places for the deprecated config options? One below
```Python
@deprecated_handler('macos_show_window_title_in_menubar')
```
and one here:
```Python
if not getattr(opts, 'macos_show_window_title_in_menubar', True):
    opts.macos_show_window_title_in = 0b01
```
Is there some way to have one config option but two independent variables where that option is stored? `macos_show_window_title_in_menubar` and `macos_show_window_title_in_window` are really two different, independent options. I agree that having one config option for the two things is better but in the code I'd like to treat them separately, so I don't need to use bit masks and bit shifts when I'm using the options in the code. The way I partially solved it now is a bit ugly, still requires not-so-readable bit operations and doesn't allow me to access `macos_show_window_title_in_menubar` and `macos_show_window_title_in_window` from the python code. It also doesn't allow me to properly handle the case where both `macos_show_window_title_in` and `macos_show_window_title_in_menubar` are present in the config.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -212,7 +212,7 @@ To update |kitty|, :doc:`follow the instructions <binary>`.
 
 - macOS: Fix :kbd:`cmd+period` key not working (:iss:`1318`)
 
-- macOS: Add an option :opt:`macos_show_window_title_in_menubar` to not
+- macOS: Add an option `macos_show_window_title_in_menubar` to not
   show the current window title in the menu-bar (:iss:`1066`)
 
 - macOS: Workaround for cocoa bug that could cause the mouse cursor to become
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -518,7 +518,7 @@ update_window_title(Window *w, OSWindow *os_window) {
         Py_INCREF(os_window->window_title);
         set_os_window_title(os_window, PyUnicode_AsUTF8(w->title));
 #ifdef __APPLE__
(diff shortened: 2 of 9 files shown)


Luflosi:
The CI complains about the option `macos_show_window_title_in_menubar` being referenced in the changelog. How do I fix that?

kovidgoyal:
Remove the :opt: prefix before it in the changelog

kovidgoyal:
yes, it is best to have the autoreleasepool because this function is called during window construction which happens before the event loop is started.

No you dont need to do it in two places. 

As for option parsing use option_type=choices('all', 'window', 'menubar') and in state.c create a parser function that sets an actual bit field.

Luflosi:
> As for option parsing use option_type=choices('all', 'window', 'menubar') and in state.c create a parser function that sets an actual bit field.

Implementing it like that will still require me to have only one variable for both options and access them by accessing the correct bit in bit field. I'd prefer to avoid that. Is there another way that would allow me to use two variables directly in the config parser code?

kovidgoyal:
No, there isn't.

Luflosi:
Sorry that I marked this as ready for review, this isn't actually working properly yet.

Luflosi:
Please review.

PR #1840, opened 2019-07-24
Format the @autoreleasepool blocks in kitty/cocoa_window.m like in GLFW

Luflosi:
I think this is slightly more readable.


Luflosi:
I pinged autoreleasepool again 🙄.

Issue #1869, opened 2019-07-30
0.14.3 crashes when launched via AwesomeWM's default launcher


ndarwincorn:
kernel: linux-zen 5.2.3
display manager: lightdm 1.30.0
window manager: awesome 4.3

When run via <kbd>mod4</kbd>+<kbd>r</kbd>`kitty`<kbd>enter</kbd>, kitty fails to start. Trace from my awesome log:

```
Traceback (most recent call last):
  File "/usr/bin/../lib/kitty/kitty/main.py", line 314, in main
    _main()
  File "/usr/bin/../lib/kitty/kitty/main.py", line 303, in _main
    setup_environment(opts, args)
  File "/usr/bin/../lib/kitty/kitty/main.py", line 227, in setup_environment
    shell_env = read_shell_environment(opts)
  File "/usr/bin/../lib/kitty/kitty/main.py", line 220, in read_shell_environment
    return read_shell_environment.ans
AttributeError: 'function' object has no attribute 'ans'
```

Launching Kitty from another terminal emulator works fine. Downgrading to 0.14.2 also works.

Let me know if you need further debugging/can't repro.

ndarwincorn:
[Downstream bug](https://bugs.archlinux.org/task/63325), in case it's a packaging issue (wasn't immediately obvious from a skim of the 0.14.2/0.14.3 diff).

Luflosi:
This looks like it's an issue in kitty.

Luflosi:
Which shell are you using? And can you try running `<your-shell> -l -c env` and see what the exit code is and if there are any error messages?

PR #1871, opened 2019-07-30
Fix a crash when trying to read the environment from a shell

Luflosi:
Fixes https://github.com/kovidgoyal/kitty/issues/1869.
If the shell from which the environment variables are being read exits with a non-zero exit code, kitty crashes. This commit fixes the crash by returning an empty environment in that case.

Note that I have not tested the fix.
I'm also not 100% sure if this is the right way to fix the crash.


kovidgoyal:
I fixed it slightly differently.

PR #1881, opened 2019-08-01
Avoid catching all exceptions

Luflosi:
Catching all exceptions just to know when `read()` didn't read anything is not necessary.

### DIFF
--- a/kitty/main.py
+++ b/kitty/main.py
@@ -219,8 +219,9 @@ def read_shell_environment(opts=None):
                     ret = p.wait(0.01)
                 except TimeoutExpired:
                     ret = None
-                with suppress(Exception):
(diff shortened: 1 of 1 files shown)


Luflosi:
Fair enough. Can I combine your code and my code? I find that catching the exception when trying to append `None` to the string is a bit ugly.

kovidgoyal:
To me, that would be unnecessary code duplication. The reason the
suppress() is there is to avoid non-blocking read related errors. It also
takes care of read() returning something unexpected, like None. So why
have extra code just to check for that?

Luflosi:
Ok, I understand your point of view and respect your decision. I have no strong feelings about that piece of code being either way.

Issue #1885, opened 2019-08-01
Let monotonic() return uint64_t instead of double?


Luflosi:
Can I make `monotonic()` return the number of nanoseconds since the start in an `uint64_t` instead of the number of seconds in a `double`? That would ensure a constant nanosecond accuracy for the next 580 years or so. The `double` will slowly get less accurate over time although that may not be a problem in practice. But I'd like to make sure that it never becomes a problem.

kovidgoyal:
monotonic subtracts the start time for the returned time anyway. Simply have it do that subtraction in the numerator before dividing, then there are no dimishing accuracy issues with using a a double.

kovidgoyal:
Actually since monotonic() is not too widely used, it might make sense to change it to use nanosecs rather than secs, but I would need to see how invasive the patch is

Luflosi:
I changed `monotonic()` in `kitty/data-types.c` and got it to compile after changing less than 30 lines. I need to change a bit more to make it actually work again though. But I don't think this patch will be too invasive.
Should I introduce a new datatype or hardcode the datatype to `uint64_t`?

kovidgoyal:
Probably best to have a monotonic_type since in the future we may get
better than nansecond resolution and having a dedicated type will make
it easier to track down and change.

PR #1897, opened 2019-08-08
Use datatype monotonic_t instead of double to keep track of time

Luflosi:
See #1885.
This patch is a *little* bigger than 30 lines now but I think I have changed pretty much all the time variables except for those in the GLFW code.
The code should compile without the second commit, I'll have to figure out how to make it compile with the second commit. I will also have to figure out what I broke because kitty renders very infrequently now. Maybe I just forgot to change a `double` somewhere.
I chose to use `int64_t` instead of `uint64_t` because a few pieces of the code rely on the value being negative and I couldn't be bothered to use two types, carefully track which values can be negative and casting the types all the time. This means that the variable will "only" be good for about 290 years.

### DIFF
--- a/glfw/backend_utils.c
+++ b/glfw/backend_utils.c
@@ -22,19 +22,6 @@
 #define ppoll pollts
 #endif
 
-static inline double
-monotonic(void) {
-    struct timespec ts = {0};
-#ifdef CLOCK_HIGHRES
-    clock_gettime(CLOCK_HIGHRES, &ts);
-#elif CLOCK_MONOTONIC_RAW
-    clock_gettime(CLOCK_MONOTONIC_RAW, &ts);
-#else
-    clock_gettime(CLOCK_MONOTONIC, &ts);
-#endif
-    return (((double)ts.tv_nsec) / 1e9) + (double)ts.tv_sec;
-}
-
 void
 update_fds(EventLoopData *eld) {
     for (nfds_t i = 0; i < eld->watches_count; i++) {
@@ -109,7 +96,7 @@ update_timers(EventLoopData *eld) {
 }
 
 id_type
-addTimer(EventLoopData *eld, const char *name, double interval, int enabled, bool repeats, timer_callback_func cb, void *cb_data, GLFWuserdatafreefun free) {
+addTimer(EventLoopData *eld, const char *name, monotonic_t interval, int enabled, bool repeats, timer_callback_func cb, void *cb_data, GLFWuserdatafreefun free) {
     if (eld->timers_count >= sizeof(eld->timers)/sizeof(eld->timers[0])) {
         _glfwInputError(GLFW_PLATFORM_ERROR, "Too many timers added");
         return 0;
@@ -117,7 +104,7 @@ addTimer(EventLoopData *eld, const char *name, double interval, int enabled, boo
     Timer *t = eld->timers + eld->timers_count++;
     t->interval = interval;
     t->name = name;
-    t->trigger_at = enabled ? monotonic() + interval : DBL_MAX;
+    t->trigger_at = enabled ? monotonic() + interval : MONOTONIC_T_MAX;
     t->repeats = repeats;
     t->callback = cb;
     t->callback_data = cb_data;
@@ -144,7 +131,7 @@ void
 toggleTimer(EventLoopData *eld, id_type timer_id, int enabled) {
     for (nfds_t i = 0; i < eld->timers_count; i++) {
         if (eld->timers[i].id == timer_id) {
-            double trigger_at = enabled ? (monotonic() + eld->timers[i].interval) : DBL_MAX;
+            monotonic_t trigger_at = enabled ? (monotonic() + eld->timers[i].interval) : MONOTONIC_T_MAX;
             if (trigger_at != eld->timers[i].trigger_at) {
                 eld->timers[i].trigger_at = trigger_at;
                 update_timers(eld);
@@ -155,7 +142,7 @@ toggleTimer(EventLoopData *eld, id_type timer_id, int enabled) {
 }
 
 void
-changeTimerInterval(EventLoopData *eld, id_type timer_id, double interval) {
+changeTimerInterval(EventLoopData *eld, id_type timer_id, monotonic_t interval) {
     for (nfds_t i = 0; i < eld->timers_count; i++) {
         if (eld->timers[i].id == timer_id) {
             eld->timers[i].interval = interval;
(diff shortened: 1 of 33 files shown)


kovidgoyal:
I dont think the conversion between time units should be functions. Either they can be macros, or if you want to use functions make them static inline functions in a header file. 

And this patch is pretty large/far ranging, not surprised it broke something.

Luflosi:
I was planning to make the functions static inline later, when everything works correctly but I can also do it now. Which header file would be the best place for them?

kovidgoyal:
I suggest using either data_types.h or a separate monotonic.h header.

Luflosi:
I decided to create a new header file. I put it in the `kitty` directory. How can I import it in both the kitty and the GLFW code?

kovidgoyal:
Just use relative paths,

../kitty/whatever.h

Luflosi:
Should I define `monotonic_t` in `monotonic.h` or `data-types.h`?

kovidgoyal:
monotonic.h

Luflosi:
At least it complies again. The content of kitty windows is now not rendered at all but I can open and close new tabs, kitty windows and OS windows. This will take some more fixing.

How can I make the functions `static inline`? I haven't really used that before. If I just add `static inline` in front of all the functions in `monotonic.c` and `monotonic.h`, kitty doesn't compile anymore.

kovidgoyal:
Move the funtion definition into the .h file and remove the .c file.

Luflosi:
Yaayyy, it compiles again. And running kitty for a couple minutes, I couldn't find anything broken. Now I just need to figure out why this one test is failing. Do you have a tip why that test might be failing?

Luflosi:
Never mind, I found the bug.
I'll read over all the changes again and then I'll mark this as ready for review.

Luflosi:
`kitty/child-monitor.c` has a line
```c
if (time_delta >= 0) ret = poll(fds, self->count + EXTRA_FDS, (int)ceil(1000 * time_delta));
```
in it. Is the `ceil()` important? If I replace the line with
```c
if (time_delta >= 0) ret = poll(fds, self->count + EXTRA_FDS, monotonic_t_to_ms(time_delta));
```
is that ok?

kovidgoyal:
Should be fine.

Luflosi:
I'm using `clock_gettime(CLOCK_UPTIME_RAW, &ts);` on macOS instead of the `((double)(mach_absolute_time() * timebase.numer) / timebase.denom)/SEC_TO_NS;` we had before. This allows me to not initialise the `timebase` variable and makes the code more compact and readable. From the man page of `clock_gettime`:
```
CLOCK_UPTIME_RAW   clock that increments monotonically, in the same manner as
                   CLOCK_MONOTONIC_RAW, but that does not increment while the system
                   is asleep.  The returned value is identical to the result of
                   mach_absolute_time() after the appropriate mach_timebase conver-
                   sion is applied.
```
As far as I know `CLOCK_UPTIME_RAW` is only available on macOS so this code should behave exactly as before.
Are you fine with this change?

Luflosi:
Please review.

kovidgoyal:
Are you sure there is nowhere that will break with a clock that does not
increment while the system sleeps? For example, the update notification
checks every 24 hours.


On Mon, Sep 23, 2019 at 04:53:44AM -0700, Luflosi wrote:
> I'm using `clock_gettime(CLOCK_UPTIME_RAW, &ts);` on macOS instead of the `((double)(mach_absolute_time() * timebase.numer) / timebase.denom)/SEC_TO_NS;` we had before. This allows me to not initialise the `timebase` variable and makes the code more compact and readable. From the man page of `clock_gettime`:
> ```
> CLOCK_UPTIME_RAW   clock that increments monotonically, in the same manner as
>                    CLOCK_MONOTONIC_RAW, but that does not increment while the system
>                    is asleep.  The returned value is identical to the result of
>                    mach_absolute_time() after the appropriate mach_timebase conver-
>                    sion is applied.
> ```
> As far as I know `CLOCK_UPTIME_RAW` is only available on macOS so this code should behave exactly as before.
> Are you fine with this change?
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/1897#issuecomment-534066325

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

Luflosi:
I am not sure. Are you sure that this worked before? Didn't kitty use clocks that do not increment while the system sleeps before? Using `clock_gettime(CLOCK_UPTIME_RAW)` instead of `mach_absolute_time()` will give the exact same result as stated in the man page.

kovidgoyal:
The man page says: "clock that increments monotonically, in the same manner as
                   CLOCK_MONOTONIC_RAW, but that does not increment while the system
                   is asleep."

which I think means that UPTIME is identical to RAW except that it does not increment while the system sleeps.

Luflosi:
Yes, that is correct. However `CLOCK_UPTIME_RAW` should only be available on macOS. And on macOS `clock_gettime(CLOCK_UPTIME_RAW)` and `mach_absolute_time()` return the same time after conversion. If you fear that `CLOCK_UPTIME_RAW` might be available on other platforms I can add an `#ifdef __APPLE__` similar to how it was before. Alternatively I could remove the use of `CLOCK_UPTIME_RAW` completely or move it to the end of the `#ifdef` chain since `CLOCK_MONOTONIC_RAW` and `CLOCK_MONOTONIC` should be available on macOS.

kovidgoyal:
On Mon, Sep 23, 2019 at 07:51:58AM -0700, Luflosi wrote:
> Yes, that is correct. However `CLOCK_UPTIME_RAW` should only be available on macOS. And on macOS `clock_gettime(CLOCK_UPTIME_RAW)` and `mach_absolute_time()` return the same time after conversion.  


I dont see how that's possible if one does not increment while sleeping
and the other does? I assume they return the same value only if the
computer does not sleep.

kovidgoyal:
Should be easy to test with a 5 line C program.

Luflosi:
I'll test it.

jacwah:
Found in `usr/include/mach/mach_time.h`:

```
uint64_t                        mach_absolute_time(void);

__OSX_AVAILABLE_STARTING(__MAC_10_9, __IPHONE_8_0)
uint64_t                        mach_approximate_time(void);

/*
 * like mach_absolute_time, but advances during sleep
 */
__OSX_AVAILABLE(10.12) __IOS_AVAILABLE(10.0) __TVOS_AVAILABLE(10.0) __WATCHOS_AVAILABLE(3.0)
uint64_t                        mach_continuous_time(void);
```

Seems clear that `mach_absolute_time` does not advance during sleep.

kovidgoyal:
Well ok then should not be a regression.

Luflosi:
While this may not be a regression, I feel like I should still change it. `monotonic()` should either be affected or not affected by a sleeping system, no matter the platform. Should I move the `#ifdef CLOCK_UPTIME_RAW` to the bottom of the `#ifdef` chain or remove it completely?

kovidgoyal:
Do you know what its semntics are w.r.t. sleep on other OSes? Linux?
BSDs?

Luflosi:
As far as I know `CLOCK_UPTIME_RAW` is only available on macOS/Darwin. https://bugs.python.org/issue35702 confirms this.

kovidgoyal:
Ah ok, sorry, I see what you are saying. Yes it should probably be
removed. Since monotonic() should have the same semantics on all
platforms.

Luflosi:
I removed it.

Luflosi:
It was a good idea to use `-Wfloat-conversion`, I didn't think of that.

kovidgoyal:
Well reviewing it was sufficiently painful that I decided to get some help from the compiler :)

Luflosi:
Wow, so many macOS conversion fixes.
> Hopefully the last macOS float conversion fix

Yup, that was the final one 😄.

kovidgoyal:
Yeah was using CI for them as I dont have access to a mac right now

On Fri, Sep 27, 2019 at 07:49:18AM -0700, Luflosi wrote:
> Wow, so many macOS conversion fixes.
> > Hopefully the last macOS float conversion fix
> 
> Yup, that was the final one 😄.
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/1897#issuecomment-535972011

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

PR #1921, opened 2019-08-23
macOS: implement drag and drop of text into kitty

### LINKED ISSUE #1368: Support drag and drop for text
I select and drag text to and from Terminal windows all the time. kitty doesn’t seem to allow dragging text either direction.

Thanks.

Luflosi:
Closes #1368.
Since two people asked for this feature and I agree that this would be nice to have, I implemented it.

Is calling `_glfwInputDrop()` for dropped text the right thing to do?
I also thought about adding a `flags` parameter to `_glfwInputDrop()` to signal if a string was a file URL or normal text. But since it's possible that the array of strings contains both types, I didn't do that. I would have to add this information to every string individually.
Is calling `_glfwInputError()` in the `else` case the right thing to do or can you think of a better solution? This code should never be called but I just want to make sure nothing bad can happen.
I'm not sure which convention to follow regarding the curly braces. GLFW seems to have them in new lines but your code seems to not do that. Just tell me if you would like me to change something regarding the curly braces.

This took me waaay too many hours to implement 😅.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -630,7 +630,7 @@ static GLFWapplicationshouldhandlereopenfun handle_reopen_callback = NULL;
         markedRect = NSMakeRect(0.0, 0.0, 0.0, 0.0);
 
         [self updateTrackingAreas];
-        [self registerForDraggedTypes:@[NSPasteboardTypeFileURL]];
+        [self registerForDraggedTypes:@[NSPasteboardTypeFileURL, NSPasteboardTypeString]];
     }
 
     return self;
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Using glfwinputerror is fine. As for calling glfwinputdrop ideally we
should change the signature of the function to add information about the
kind of strings url/text, but at the moment, since kitty makes no use
of that information, it is OK to leave it as is.

kovidgoyal:
And note this does not really close #1368 since that is requesting the ability to drag text from within kitty as well, which is the painful part to implement.

Luflosi:
You're right, I only implemented half of the feature request. I won't try implementing the other half because it's not easy as you mentioned and I don't think it's not worth the effort.

PR #1926, opened 2019-08-25
Correct indentation for two _glfwInputError() messages

Luflosi:
This commit changes the indentation of two error messages to match the indentation in the rest of the code.

Mostly unrelated: Why are the functions `_glfwPlatformSetClipboardString()` and `_glfwPlatformGetClipboardString()` in `wl_window.c` so different between the GLFW version in kitty and the upstream version? I thought we could maybe combine the two to get the best from both and then try to get the new version upstream as well.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -692,8 +692,8 @@ static GLFWapplicationshouldhandlereopenfun handle_reopen_callback = NULL;
             [window->context.nsgl.object update];
         } @catch (NSException *e) {
             _glfwInputError(GLFW_PLATFORM_ERROR,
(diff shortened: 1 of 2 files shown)


Luflosi:
Would you mind answering my question 😀?
> Why are the functions `_glfwPlatformSetClipboardString()` and `_glfwPlatformGetClipboardString()` in `wl_window.c` so different between the GLFW version in kitty and the upstream version? I thought we could maybe combine the two to get the best from both and then try to get the new version upstream as well.

kovidgoyal:
As far as I recall the kitty version is designed not to hang in the face of misbehaving software, the glfw version does not bother.

Issue #1919, opened 2019-08-23
Mapping Special Keys (Ctrl+ö)


bschnitz:
Some mappings seem not to be detected by kitty.

How to reproduce:
1. put `map ctrl+odiaeresis paste_from_clipboard` (or `map ctrl+0xf6 paste_from_clipboard`) into the kitty.conf .
2. open kitty, copy something into clipboard and try to paste using ctrl+ö

Expected result:
The contents from clipboard appear.

Actual result:
; is written to the terminal

Luflosi:
I tried it with `map ctrl+ö paste_from_clipboard` and had the same problem. When starting kitty, it prints the following error message:
```
Shortcut: ctrl+ö has unknown key, ignoring
```
`0xf6` is not the same key as `ö` and kitty does not print any error message.

PS: Also using a german keyboard?

bschnitz:
0xf6 is the output of  `kitty --debug-keyboard` for the key-press event of ö. I think You can't use `map ctrl+ö paste_from_clipboard` for mapping inside kitty.conf. The manual states:

> For a list of key names, see: GLFW keys. The name to use is the part after the GLFW_KEY_ prefix. For a list of modifier names, see: GLFW mods. On Linux you can also use XKB key names to bind keys that are not supported by GLFW. See XKB keys for a list of key names. The name to use is the part after the XKB_KEY_ prefix. Note that you should only use an XKB key name for keys that are not present in the list of GLFW keys. Finally, you can use raw system key codes to map keys. To see the system key code for a key, start kitty with the kitty --debug-keyboard option. Then kitty will output some debug text for every key event. In that text look for native_code the value of that becomes the key name in the shortcut.

There are links in the manual. I couldn't copy them. I suggest to look into it directly:
https://sw.kovidgoyal.net/kitty/conf.html#keyboard-shortcuts

kovidgoyal:
You cannot map arbitrary key names, since these tend to be platform dependent. Use the native code/name to map it instead which you can get from --debug-keyboard. Sadly I dont have a german keyboard, so I cant test this, but the relevant code is in keys.py see the get_shortcut function

bschnitz:
[...]

PR #1928, opened 2019-08-26
Allow mapping special keys

Luflosi:
Closes https://github.com/kovidgoyal/kitty/issues/1919.
The function is from https://stackoverflow.com/questions/8263618/convert-virtual-key-code-to-unicode-string.

`keyCodeToString()` translates a key code to an NSString. It would be nice to completely replace `translateKey()` so that we don't need to use a hardcoded mapping anymore but I don't think that's possible because for example pressing any function key results in a String with the UTF-8 representation having a value of 16 in the first and only byte. This does not allow distinguishing between those keys so we need to come up with another solution, possibly combining `keyCodeToString()` and `translateKey()`.


kovidgoyal:
You cant use strings from UCKeyTranslate to convert keys to key codes. You have no idea how to convert those strings back into key codes. Also since keyDown already calls UCKeyTranslate do we really need to call it again? And is we do why cant we use _glfw.ns.unicodeData?

Luflosi:
> You cant use strings from UCKeyTranslate to convert keys to key codes. You have no idea how to convert those strings back into key codes.

Can't we just use something like
```objective-c
NSNumber* charToKeyCode(NSString *keyChar)
{
    static NSMutableDictionary* dict = nil;

    if (dict == nil) {
        dict = [NSMutableDictionary dictionary];

        // For every keyCode
        for (CGKeyCode i = 0; i < 128; i++) {
            NSString* str = keyCodeToString(i);
            if (str != nil && ![str isEqualToString:@""]) {
                [dict setObject:[NSNumber numberWithInt:i] forKey:str];
            }
        }
    }

    return [dict objectForKey:keyChar];
}
```
to build a reverse mapping?
<br>

> Also since keyDown already calls UCKeyTranslate do we really need to call it again?

We most likely don't need to call it again, I just wanted to be somewhat confident that this idea is even possible before doing any kind of optimisation.

kovidgoyal:
I dont think I understand, wont that just end up mapping the special
char bak to semi-colon?

Luflosi:
It shouldn't since the reverse mapping function uses `keyCodeToString()` and not `translateKey()`. And `keyCodeToString()` properly maps the key code `0x29` to the character `ö` with my keyboard.
Maybe I didn't really understand. What do we actually need the reverse mapping for? I assumed that we needed it when parsing keyboard mappings in the config file. Is that true? And do we need it for anything else?
I'm also not actually sure what you mean by "special char".

kovidgoyal:
keyCodeToString() will map 0x29 to odiaresis and so the reverse map will
map odiaresis back to 0x29 and 0x29 is mapped to GLFW_KEY_SEMICOLON

The way key mapping works is that when the key event reaches kitty, if
the glfw key code is not GLFW_KEY_UNKNOWN then the glfw key code is
used. Only if the key in unknown is the native code used. 

So what you would need to do is detect when keyCodeToString() is mapping
keys to something other than their canonical value, then either set the
glfw key to unknown or change the keyboard api to pass the key name as a
string and have kity match against that preferentially. And when using
keyCodeToString() you have to be very careful about IME events when
normal keys get used to generate arbitrary text, which will totally mess
up key mapping if it is based on strings.

Basically, this is a bug in Apple's API. They should not have reused key
codes for different keys. Any workaround we can come up with will not be
robust as a result.

Luflosi:
I implemented something that I thought would work but it doesn't yet. I tried adding new `ä`, `ö` and `ü` GLFW keys. Do you know what I could have missed?
At least the `Shortcut: ctrl+ö has unknown key, ignoring` error message isn't printed anymore, so I must have gotten something right.
Is there a way to declare a unicode character constant like `'ä'` with UTF-8 encoding and `uint32_t` type or something in C? Right now I just use the hexadecimal values but that's not very readable.

kovidgoyal:
There is a bunch of stuff to do when adding a new key, check the previous commits for adding the plus key for example.

kovidgoyal:
No there is no way to use anythin other than a numeric constant in a switch statement. You can make it symbolic with a #define if you like.

Luflosi:
What do the constants in `ENCODING` in `kitty/key_encoding.py` mean?

kovidgoyal:
They are used to handle the kitty extended keyboard protocol

Luflosi:
How can I figure out what to map ä, ö, ü and ß to?

kovidgoyal:
You dont need to figure it out, just run update_encoding() it will
update the file automatically.

Luflosi:
Why does `translateKey()` in `glfw/cocoa_window.m` contain upper case characters? At least on my keyboard only lower case characters are passed to the switch statement when pressing shift and a letter.

kovidgoyal:
It contains both, since there is no reason not too and who knows what
apple will do in the future.

Luflosi:
It's quite painful to add new keys tbh.
So far I added `ÄäÖöÜüß<#`.
Don't merge this yet please, I'm not done yet.

Luflosi:
18000 lines changed in `kitty/keys.h` 😳. So many changes for just a few new keys.

Luflosi:
If I add dead keys, they won't work as keyboard shortcuts, right?

kovidgoyal:
Whether a key is a dead key or not depends on the currently active
keyboard layout, and yes dead keys are not passed on to kitty fom glfw.

Luflosi:
Do you know what the test error message means?

kovidgoyal:
It means interpret_key_evet() is not giving an expected results for the
F1 key

Luflosi:
What do the `SHIFTED_PRINTABLE` and `UN_SHIFTED_PRINTABLE` variables in `kitty/keys.py` do? This is keyboard layout dependent and there are a couple mistakes in the definition of those variables.

kovidgoyal:
they are used to interpret what alt+shift+key should become in the
traditional keyboard terminal modes which dont really support alt+shift

Luflosi:
When you added `UNDERSCORE` to `UN_SHIFTED_PRINTABLE` in c2fd7005cb72fca996bdb668098fd0d1d4ec6d29, you didn't add it to `SHIFTED_PRINTABLE`. Was that intentional?

kovidgoyal:
SHIFTED_PRINTABLE is a superset of UNSHIFTED_PRINTABLE, it start as a
copy of it and gets stuff added to it.

Luflosi:
Yeah but that means, that `UNDERSCORE` is mapped to `_` in both variables. All the other ones except `SPACE` have a different mapping for the shifted variant.

kovidgoyal:
Yeah because there is no unshifted underscore on a US keyboard.

Luflosi:
Does the test fail because I added too many keys? I had to increase the limit in `kitty/keys.py` because 128 keys were not enough anymore.

Luflosi:
I see that `kitty/keys.py` only takes the least significant seven bits. Is the eighth bit used for anything? Can I just add one more bit?

kovidgoyal:
I dont think it is used, but am not 100% sure, try it and see.

On Sun, Sep 01, 2019 at 11:07:13AM -0700, Luflosi wrote:
> I see that `kitty/keys.py` only takes the least significant seven bits. Is the eighth bit used for anything? Can I just add one more bit?
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/1928#issuecomment-526940919

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

Luflosi:
Please review.

Luflosi:
I added the keys in literally 11 places and had to regenerate two files resulting in tens of thousands of changes. This is not a very good situation. Maybe in the future we can reduce the number of places where keys are defined.

Luflosi:
kitty should now have all keys from the following keyboard layouts:
- US English
- British
- Danish
- French
- German
- Italian
- Portuguese
- Russian
- Spanish
- Swiss

kovidgoyal:
Keyboard handling is optimized for performance, not ease of
modification. And automated file changes can be trivially ignored.

On Mon, Sep 02, 2019 at 04:38:10PM -0700, Luflosi wrote:
> I added the keys in literally 11 places and had to regenerate two files resulting in tens of thousands of changes. This is not a very good situation. Maybe in the future we can reduce the number of places where keys are defined.
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/1928#issuecomment-527257969

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

Luflosi:
Will you review this when you have some time?

kovidgoyal:
yes

Luflosi:
After rebasing onto the latest master, I got the following traceback when regenerating the automatically generated files:
```
  File "/nix/store/4i0nln542psiic6hcbj99x3r3rsrg1zg-python3-3.7.4/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/nix/store/4i0nln542psiic6hcbj99x3r3rsrg1zg-python3-3.7.4/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "./__main__.py", line 117, in <module>
    main()
  File "./__main__.py", line 108, in main
    namespaced(['+', first_arg[1:]] + sys.argv[2:])
  File "./__main__.py", line 68, in namespaced
    func(args[1:])
  File "./__main__.py", line 27, in runpy
    exec(args[1])
  File "<string>", line 1, in <module>
  File "./kitty/keys.py", line 8, in <module>
    from .key_encoding import KEY_MAP
  File "./kitty/key_encoding.py", line 356, in <module>
    config_key_map.update({k: g[v] for k, v in key_name_aliases.items()})
  File "./kitty/key_encoding.py", line 356, in <dictcomp>
    config_key_map.update({k: g[v] for k, v in key_name_aliases.items()})
KeyError: 'EXCLAM'
```
Do you know why this happens? What is the purpose of the `g` variable? It looks like it just maps a bunch of key names to the same name, e.g. `{'MINUS': 'MINUS'}`. What is the difference to the `config_key_map` variable?

kovidgoyal:
g is the module globals, it contains all the key names defined at global
scope. And it is there mainly as a performance optimization so you can
check for a key with the is operator instead of the ==

Luflosi:
Do you know why the exception happens? What can I do to fix it? Do I somehow need to add the keys to `g` as well?

kovidgoyal:
If you have added them to ENCODING they will be automatically added to g

Luflosi:
Running the two functions now obviously works because the keys are already in `ENCODING` but undoing the changes to the two generated files and trying to recreate them won't work. Adding new keys will lead to the same problem. Is there anything that can be done about that? Does my patch make sense or does it break anything?

kovidgoyal:
Of course it works. You simply have to run update_encoding() before adding
the new keys to key_name_aliases. And this can easily be fixed, by only
updating config_key_map with those entries in key_name_aliases that
exist in ENCODING.

Luflosi:
Thanks for 41b0f8852f31037e432836b8cd922a319abac9d5. It works as expected.

PR #1931, opened 2019-08-27
Fix undefined behavior

Luflosi:
I'm fairly certain that this bug can't actually be triggered in practice with the current way `format_text()` is used but I'm fixing it just to be sure.

From the manpage of `snprintf()`:

> The snprintf() and vsnprintf() functions will write at most size-1 of the characters printed into the output string (the size'th character then gets the terminating '\0'); if the return value is greater than or equal to the size argument, the string was too short and some of the printed characters were discarded.

and

> These functions return the number of characters printed (not including the trailing '\0' used to end output to strings), except for snprintf() and vsnprintf(), which return the number of characters that would have been printed if the size were unlimited (again, not including the final '\0'). These functions return a negative value if an error occurs.

When `buf` is too short to hold the string, `snprintf()` may return a value so big, that `p` points outside of `buf`. `if (p != buf) *(--p) = 0;` will then write outside of `buf`, which is undefined behavior.
I fixed this by checking if `buf` was too short to hold the string and then clamping `p`. If the `while` loop continues after that, `max_new_chars` will be `0` and `snprintf()` should not do anything anymore.
I also handeled the case where the return value of `snprintf()` is negative.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -263,7 +263,16 @@ format_text(const char *src) {
     char *p = buf;
     if (!src[0]) return "<none>";
     while (*src) {
-        p += snprintf(p, sizeof(buf) - (p - buf), "0x%x ", (unsigned char)*(src++));
+        int max_new_chars = sizeof(buf) - (p - buf);
+        int new_chars = snprintf(p, max_new_chars, "0x%x ", (unsigned char)*(src++));
+        if (new_chars < 0) {
+            _glfwInputError(GLFW_PLATFORM_ERROR,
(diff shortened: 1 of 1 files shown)


kovidgoyal:
If you want to make that function really robust, you also need to
terminate the loop when p goes beyond the end of buf.

Luflosi:
If p is at the end of buf, `max_new_chars` should be `0` and `snprintf()` with a size argument of zero doesn't do anything, right?

kovidgoyal:
max_new_chars could be negative and you really shouldn't rely on the
behavior of snprintf() in such a case. See my latest commit.

PR #1939, opened 2019-08-30
Fix NSLog() printing of unicode format strings

Luflosi:
See https://stackoverflow.com/questions/720052/nslog-incorrect-encoding.
The `%s` format placeholder for `NSLog()` expects an encoding other than UTF-8, which leads to garbled Unicode characters when trying to print a UTF-8 encoded string:
```Objective-C
NSLog(@"Ä %s %@", "Ä", [NSString stringWithUTF8String:"Ä"]);
```
prints `Ä √Ñ Ä`.
As can be seen in the example above, the workaround is to convert the UTF-8 encoded C-string to an `NSString` and print that instead.
`debug_key()` calls `NSLog()`.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -952,13 +952,13 @@ is_ascii_control_char(char x) {
                     &char_count,
                     text
                     ) != noErr) {
-            debug_key(@"UCKeyTranslate failed for scancode: 0x%x (%s) %s\n",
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Rather than using stringWithUtf8String just use @()

Luflosi:
Thank you, that looks much nicer.

PR #1942, opened 2019-08-31
Add missing named_keys

Luflosi:


kovidgoyal:
+ and ' ' wont work because they are used as separators in the config
file. Backslash is OK I think.

Luflosi:
I removed `+` and ` `.

PR #1943, opened 2019-09-01
Suppress more specific exception

Luflosi:


kovidgoyal:
Why? Are you sure that int() only ever raises ValueError and not say TypeError? What advantage is there to suppressing ValueError over Exception?

Luflosi:
I prefer to catch the least generic python exception type possible so that bugs resulting from catching errors that were not anticipated are not silently ignored.

This is also the exception that the example code in https://docs.python.org/3/tutorial/errors.html#handling-exceptions uses.

From https://docs.python.org/3/library/exceptions.html#ValueError:
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

From https://docs.python.org/3/library/exceptions.html#TypeError:

> Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.
> 
> This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be. If an object is meant to support a given operation but has not yet provided an implementation, NotImplementedError is the proper exception to raise.
> 
> Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

I think that this makes it pretty clear that something like `int(1, 16)` throws a `TypeError` while something like `int('z', 16)` throws a `ValueError`. Experimentation confirms that this is true at least for my python version. I think the first argument passed to `int()` is guaranteed to be a string so a `TypeError` should never be thrown.

If you think that `int()` may conceivably throw other exceptions, you may ignore this patch. I have no strong feelings about it.

kovidgoyal:
I see no point in making the code less robust for an abstract principle. I dont have a lot of faith in int() actually throwing only a single exception type for ever. It would be different if int()'s documentation claimed it only threw a single exception type. But without that contract it does not make sense to rely on it. Not to mention that if int() does indeed only ever throw ValueError, then you lose nothing by catching Exception, since you do not in practice catch any exceptions other than ValueError.

PR #1946, opened 2019-09-01
Make adding more printable keys slightly easier

Luflosi:
When adding keys after `GLFW_KEY_UNDERSCORE`, one now needs to change a `#define` right below the last printable key instead of changing it elsewhere in the code.
This commit now also marks `GLFW_KEY_PLUS` and `GLFW_KEY_UNDERSCORE` as printable characters.

Why does `GLFW_KEY_SPACE` not count as a printable character in `glfwGetKeyName()`?

### DIFF
--- a/glfw/glfw3.h
+++ b/glfw/glfw3.h
@@ -394,6 +394,8 @@ extern "C" {
 #define GLFW_KEY_PLUS               163
 #define GLFW_KEY_UNDERSCORE         164
 
+#define GLFW_KEY_LAST_PRINTABLE     GLFW_KEY_UNDERSCORE
+
 /* Function keys */
 #define GLFW_KEY_ESCAPE             256
 #define GLFW_KEY_ENTER              257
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I dont know what you mean in glfw/input.c GLFW_KEY_SPACE is categorised as a printable key in _glfwGetKeyName()

Luflosi:
I meant in `glfwGetKeyName()` not in `_glfwGetKeyName()`.
From `glfwGetKeyName()`:
```C
if (key != GLFW_KEY_KP_EQUAL &&
    (key < GLFW_KEY_KP_0 || key > GLFW_KEY_KP_ADD) &&
    (key < GLFW_KEY_APOSTROPHE || key > GLFW_KEY_LAST_PRINTABLE))
{
    return NULL;
}
```
`GLFW_KEY_SPACE` is defined as `32` and `GLFW_KEY_APOSTROPHE` is defined as `39`. Since 32 < 39, `glfwGetKeyName()` will return `NULL` for `GLFW_KEY_SPACE`.

kovidgoyal:
No idea, not a function I have written. I'd guess its because space when
printed out is invisibel and is typically classified as a control
character historically.

Issue #1980, opened 2019-09-18
key input issue - clean_sym vs composed_sym ?


bew:
Hello,

I'm having an issue with the keybind `Alt+/` not doing anything in **terminal apps** (cat, zsh, nvim, ..)

I'm using the **french keyboard layout**, which has the slash and colon symbols on the same key (to get the slash I do `Shift+colon`).

The actual keys I have to press are `Alt Shift colon` (useful for the debug output below).

Here is the relevant debug output:
```
Press scancode: 0x40 clean_sym: Alt_L composed_sym: Alt_L mods: none glfw_key: 342 (LEFT ALT) xkb_key: 65513 (Alt_L)
on_key_input: glfw key: 342 native_code: 0xffe9 action: PRESS mods: 0x0 text: '' state: 0 sent key to child

Press scancode: 0x32 clean_sym: Shift_L composed_sym: Shift_L mods: alt glfw_key: 340 (LEFT SHIFT) xkb_key: 65505 (Shift_L)
on_key_input: glfw key: 340 native_code: 0xffe1 action: PRESS mods: 0x4 text: '' state: 0 sent key to child

Press scancode: 0x3c clean_sym: colon composed_sym: slash mods: alt+shift glfw_fallback_key: 46 (PERIOD) xkb_key: 46 (period)
on_key_input: glfw key: 46 native_code: 0x2e action: PRESS mods: 0x5 text: '' state: 0 sent key to child

Release scancode: 0x3c clean_sym: colon mods: alt+shift glfw_fallback_key: 46 (PERIOD) xkb_key: 46 (period)
on_key_input: glfw key: 46 native_code: 0x2e action: RELEASE mods: 0x5 text: '' state: 0 ignoring as keyboard mode does not allow release events

Release scancode: 0x32 clean_sym: Shift_L mods: alt+shift glfw_key: 340 (LEFT SHIFT) xkb_key: 65505 (Shift_L)
on_key_input: glfw key: 340 native_code: 0xffe1 action: RELEASE mods: 0x5 text: '' state: 0 ignoring as keyboard mode does not allow release events

Release scancode: 0x40 clean_sym: Alt_L mods: alt glfw_key: 342 (LEFT ALT) xkb_key: 65513 (Alt_L)
on_key_input: glfw key: 342 native_code: 0xffe9 action: RELEASE mods: 0x4 text: '' state: 0 ignoring as keyboard mode does not allow release events
```

The most interesting lines seems to be:
```
[...]

PR #1998, opened 2019-09-24
Extract glfw key event data to a struct

bew:
Initial work for https://github.com/kovidgoyal/kitty/issues/1980#issuecomment-533861847

Notes:

- I changed the struct `KeyEvent` used by IBUS handling to use embed a GLFWkeyevent struct, to avoid duplication, but I'm not sure it's a good idea after all, we can discuss this.
- the words `scancode` and `keycode` seems to be used interchangeably, I'd suggest to choose only one and stick with it. I think that `scancode` is more platform-agnostic. There is also `device_code` which could be used in the GLFWkeyevent struct, while we keep `scancode` in the plateform-specific KB handling code..
  What do you think?
- :warning: I did **not** test IBUS handling (not sure how..)

Is there a reason why there are almost no newlines nor comments in functions? It feels hard to read the code sometime.. (and splitting them in multiple functions might not always help)


bew:
@Luflosi I'll need help for macos if you have time :pray:

Luflosi:
You managed to make it work for macOS yourself, GJ 👍.

bew:
Thanks @kovidgoyal, that was fast! Do you have any comments on the questions I raised in the PR description?

kovidgoyal:
> * I changed the struct `KeyEvent` used by IBUS handling to use embed a GLFWkeyevent struct, to avoid duplication, but I'm not sure it's a good idea after all, we can discuss this.

It's fine by me, I made the code more robust, to handle null text and ensure it is zero terminated.

> * the words `scancode` and `keycode` seems to be used interchangeably, I'd suggest to choose only one and stick with it. I think that `scancode` is more platform-agnostic. There is also `device_code` which could be used in the GLFWkeyevent struct, while we keep `scancode` in the plateform-specific KB handling code..

I would actually prefer native_keycode but it's not that big of a deal.

>   What do you think?
> * warning I did **not** test IBUS handling (not sure how..)

You can just start the ibus daemon if it is not already running and turn on some keyboard layout such as the one for abbreviations or unicode  input or similar. It's been a while since I wrote that code so dont remember the details.

> 
> Is there a reason why there are almost no newlines nor comments in functions? It feels hard to read the code sometime.. (and splitting them in multiple functions might not always help)

newlines are matter of taste, I happen to find the number of newlines well suited to the way I read code. As for comments, I prefer to make code as clear as possible. I only add comments when something is counter-intuitive or complex or could benefit from a high level description. If something is unclear, feel free to ask and add a comment.

glfw/xkb_glfw.c
```diff
@@ -547,91 +548,108 @@ glfw_xkb_key_from_ime(KeyEvent *ev, bool handled_by_ime, bool failed) {
...
     );
-    key_event.action = action; key_event.glfw_modifiers = sg->modifiers;
-    key_event.keycode = scancode; key_event.keysym = glfw_sym;
-    key_event.window_id = window->id; key_event.glfw_keycode = glfw_keycode;
-    key_event.ibus_sym = syms[0];
-    if (ibus_process_key(&key_event, &xkb->ibus)) {
-        debug("↳ to IBUS: keycode: 0x%x keysym: 0x%x (%s) %s\n", key_event.ibus_keycode, key_event.ibus_sym, glfw_xkb_keysym_name(key_event.ibus_sym), format_mods(key_event.glfw_modifiers));
+
+    glfw_ev.action = action;
+    glfw_ev.scancode = xkb_sym;
```
bew: Actually, this line is wrong!!! Because this is not a scancode but the native keysym instead... when it should be `xkb_keycode` instead of `xkb_sym` here (for coherence with the name of the receiving field).

This also means that `on_key_event` then kitty's keymap handling used the value of the native keysym all along.. this doesn't sound right.. do it?

Do you actually want the native keycode (scancode) or the native keysym in kitty's keyboard handling?
I would say the native keycode would make more sense, but to avoid breaking user configs, we _must_ (?) stick with keysym?

bew: Also note that keeping the keysym value as the "scancode" in GLFWkeyevent for linux would be inconsistent with macos, which seems to correctly put the native keycode in event struct..

Now sure what to do... break user's linux configs? (might be ok, since this feature is probably not that used.. ?)

bew: :warning: overkill?

We could also have a "migration" path (over N releases), where we add the keycode (good) and the keysym (bad) to the key event struct.

Then if a kitty shortcut like `ctrl+XX` matches the keysym, display a sort of toast message (in addition to doing the action) telling the user to run Kitty from a terminal, and print logs telling him to update the mapping with `ctrl+YY` instead (using the true keycode).

WDYT?

Luflosi: > but which code is it

What do you mean exactly?

bew: Reading more, it looks like having the keyevent's "scancode" be the keysym value on linux seems to be the expected behavior:
https://github.com/kovidgoyal/kitty/blob/4ec1a8d9c3180828ae5a03cc24967b126fbec6b2/glfw/x11_window.c#L2594-L2597
So maybe I can just change the keyevent field to `native_keycode`, and document somewhere that on linux the `native_keycode` is not the `xkb_keycode` but the `xkb_keysym`..

bew: @Luflosi forget what I said, I think I got confused by `native_keycode`, because X11 has something called a keycode and a keysym, and it's the keysym value that is to be assigned to `native_keycode`

kovidgoyal:
No, we want scancode to be the XKB keysym value. Actual scancodes are
kernel/driver/hardware dependent numbers that are useless for anything
portable. Currently, if you wish to create shortcuts for keys that dont
exist in glfw you can do so using XKB symbolic names on linux and actual
scancode numbers on macOS. This functionality needs to remain, as I
doubt we will ever end up adding all possible keys to glfw.

On macOS they might make sense since the hardware is Apple
made as well, but not on Linux. You can, if you wish add the scancode as
an extra field to the struct. 

So maybe to avoid confusion name it native_key instead of
native_keycode, that way there is no chance of mistaking it for an X11
keycode or a kernel scancode or whatever.

PR #2021, opened 2019-10-03
Adding powerline style tabs

brujoand:
Screenshot 2019-10-03 at 21 39 43
This PR adds support for powerline styled tabs. I had to add the inactive fg/bg colors to the draw_data to be able to produce the colour transitions properly. To avoid having the background colour continue after the last tab i added the `is_last` parameter to the `draw_tab_with_x` method. This can probably be reused to have the `separator` style stop at the last item too. 

 It's been a while since I've written any python so please let me know if here are things I should have done differently.

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -679,10 +679,10 @@ Which edge to show the tab bar on, top or bottom'''))
 o('tab_bar_margin_width', 0.0, option_type=positive_float, long_text=_('''
 The margin to the left and right of the tab bar (in pts)'''))
 
-o('tab_bar_style', 'fade', option_type=choices('fade', 'separator', 'hidden'), long_text=_('''
-The tab bar style, can be one of: :code:`fade`, :code:`separator` or :code:`hidden`. In the fade style,
-each tab's edges fade into the background color, in the separator style, tabs are
-separated by a configurable separator.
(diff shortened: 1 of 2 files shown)


Luflosi:
The style warnings should be easy to fix. Let me know if you need help with that.

kovidgoyal:
The logic seems broken when there is not enough space per tab (i.e. lots of tabs). See attached screenshots. Notice how first tab is different from all the others.
Screenshot_20191004_132004
Screenshot_20191004_131927

brujoand:
Yeah I've added guards against writing a negative value to cursor.x in the two places I write to it, but I see now that I have to tweak it a bit more to get consistent behaviour.

brujoand:
This was a lot more fun than I had expected, so I gave a go at creating the triangle too. And I just wanted to check in to get some feedback on that. And the a problem I'm facing when there are too many tabs to precent. The `…` ends up overwriting part of the last segment, and I can't wrap my head quite around how to avoid it. Any input on that would be great. I'll head over to fixing the tests now.

brujoand:
If you still want me to split this into two PRs I can do that, just wanted to check in to see if this approach was okay.

kovidgoyal:
I have implemented drawing of the separator symbols using proper antialiasing with the Wu algorithm. Remove that part of your PR and I will merge the rest, fixing the many tabs issue as well.

brujoand:
Cool, I've removed the changes to fonts and boxdrawing and squashed my commits.

PR #2029, opened 2019-10-06
Cocoa: Move app delegate and menu creation to init

Luflosi:
From https://github.com/glfw/glfw/commit/ea7eb2ddab823e21c846c68d112d90f91fcba318.

I'm trying to apply this commit from upstream but I ran into a little problem:
`glfwSetApplicationShouldHandleReopen()` and `applicationShouldHandleReopen()` in `glfw/cocoa_window.m` access the variable `handle_reopen_callback`. I need to move `applicationShouldHandleReopen()` to `glfw/cocoa_init.m`. One of the two functions will no longer be able to access the variable.
What is the best way to solve this problem? Can I somehow move `glfwSetApplicationShouldHandleReopen()` or make the variable visible in the two files?

### DIFF
--- a/glfw/cocoa_init.m
+++ b/glfw/cocoa_init.m
@@ -31,6 +31,9 @@
 #include <sys/param.h> // For MAXPATHLEN
 #include <pthread.h>
 
+// Needed for _NSGetProgname
+#include <crt_externs.h>
+
 // Change to our application bundle's resources directory, if present
 //
 static void changeToResourcesDirectory(void)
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Either mark that variable as extern or create a wrapper function for accessing it or make glfwSetApplicationShouldHandleReopen nont static and add it into the cocoa_platform.h

Luflosi:
Do you prefer any one of those options?

kovidgoyal:
Making glfwSetApplicationShouldHandleReopen non-static would be my
choice, but I dont really care.

Luflosi:
Ok, I simply moved `glfwSetApplicationShouldHandleReopen()` to `glfw/cocoa_init.m` and it appears to work without any problems.
I should have tried this obvious solution before 🙄😅.
Please review.

PR #2037, opened 2019-10-09
Cocoa: Move slightly towards modern Objective-C

Luflosi:
Please double-check that the two lines are actually equivalent. I tested it and it seems to work fine but I'm not entirely sure if this is correct. Is `[title release];` still needed?


Luflosi:
I guess I need to remove the `[title release];`, right?

Luflosi:
Nope, the error has nothing to do with my code, I think.

kovidgoyal:
yes you should remove release. In cocoa land you only need to release things you have created using alloc/init or retained.

Luflosi:
I fixed it.

Issue #1310, opened 2019-01-16
Feature request: Move window to another Tab


binor:
Sometimes I'd like to move a window to another Tab when organizing Tabs. Would this be a feature you could implement?

kovidgoyal:
should also have the possibility to make a new tab from the window.

torfsen:
I have the opposite need: A tab that I'd like to make into a window.

BTW, thanks for Kitty, it's awesome!

kovidgoyal:
Since a tab contains multiple windows how does it become a window?

torfsen:
@kovidgoyal Perhaps a better formulation would be: I'd like to move all windows from one tab into another tab and automatically close the first tab.

kovidgoyal:
Sorry I dont understand your use case? What does moving al lwindows from a tab into a new different tab and closing the old tab achieve that simply renaming the old tab/moving it around does not?

yurikhan:
I think @torfsen is asking for tab merging. At start: tab A has windows A1, A2, …, An, tab B has B1, B2, …, Bm; on completion: tab A contains windows A1, A2, …, An, B1, B2, Bm; tab B no longer exists.

kovidgoyal:
ah well, that is basically the same as move one window, just move all instead.

0x17de:
i am working on a patch to move windows into separate tabs (which is working for me so far). i just need to finish the documentation parts and check if some cleanups are required before i submit a PR.
once this is done it should be easy to implement the opposite direction.

do you have some idea for hotkeys for both directions?

kovidgoyal:
The hotkey space is kind of crowded. I think you will need to use
kitty_mod+function key or something similar. Or maybe dont give it a
shortcut by default, there are several such actions already.

PR #2042, opened 2019-10-10
move windows into separate tabs

0x17de:
This PR allows moving windows away into new tabs and is related to #1310. Merging windows into other tabs could build upon this PR.
Please check my assigned hotkey and maybe propose a different one. So far i thought about e(x)tract (Ctrl+Shift+x). Maybe we can come up with something better, that also supports the other direction of merging windows into existing tabs.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -39,6 +39,7 @@ from .utils import (
     log_error, open_url, parse_address_spec, remove_socket_file, safe_print,
     set_primary_selection, single_instance, startup_notification_handler
 )
+from .window import Window
 
 
 def listen_on(spec):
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I am travelling for a few days, I will look at this when I return.

kovidgoyal:
This looks generally fine, good job. Some issues:

1)  You have to deal with overlay windows. If the window being moved has an overlay, both the window and its overlay will have to be moved

2) Moving a window into a new tab can in general cause it to be resized, this might be handled automatically, not sure, but should be checked. IN particular if th etab is in a different os window which has a different font size -- this of course does not happen with your current code, but can in general.

3) I would like to make the action more general so rather than move_window_to_tab one should have instead a detach_window command that takes arguments and allows the window to be moved to a specific tab/OS window. This can be done after the basic PR is merged, however.

kovidgoyal:
I have merged the state.c part of your PR. I want to think about a better api for detaching and placing windows in boss.py/tabs.py when I get that done we can add controls for it as desired.

kovidgoyal:
I have added a pair of detach_window()/attach_window() functions that should serve as a general API for moving windows between tabs.

kovidgoyal:
And I have added a _move_window_to() function to boss.py that serves as an API for moving windows to existing or new tabs.

vdimir:
Could you tell please, if it was included to any release? Cannot find this feature in the current one.

kovidgoyal:
Yes it has, years ago. look for detach_window

PR #2046, opened 2019-10-11
Cleanup

Luflosi:
From upstream: https://github.com/glfw/glfw/commit/62b7fe831179428c3fa1d7939d48562772ee5647.


kovidgoyal:
How is this a cleanup? Looks like a dirty up to me.

Luflosi:
Well, I was trying to reduce the difference to upstream and simply chose the commit message of the upstream commit. I can certainly change it.
To be fair, the rest of the code in this area uses the same style of preprocessor macros so I can see why this is called "cleanup".

PR #2065, opened 2019-10-16
Rename all uses of 'scancode' to 'native_key' where relevant

bew:
Follow up of #1998, to have some consistency with the variable names related to the plateform-specific key identifier, previously named `scancode`, now renamed to `native_key` to show that it represent the final key event (after mods application) in a native-to-the-platform format.

I removed all occurences of `scancode` except in the cocoa impl, where it looks like that `scancode` is the correct terminology to represent a key press. @Luflosi can you confirm?


Luflosi:
Well, the function to get that information is called `keyCode`, so maybe `scancode` isn't the correct terminology. I suggest just changing the cocoa implementation as well for consistency.
@kovidgoyal, what do you think?

kovidgoyal:
On Wed, Oct 16, 2019 at 05:05:24AM -0700, Luflosi wrote:
> Well, the function to get that information is called `keyCode`, so maybe `scancode` isn't the correct terminology. I suggest just changing the cocoa implementation as well for consistency.
> @kovidgoyal, what do you think? 

Sounds ok to me. On linux the codes are XKB codes not actual scancodes.
On macOS they probably are actual scancodes, but from the perspective of
kitty that does not matter.

Luflosi:
I meant to suggest that you rename `scancode` to `native_key` so that it is called `native_key` everywhere. Feel free to change it yourself, I can test of course. But when it compiles fine in the CI it will probably work fine since you just renamed stuff.

bew:
In my opinion, the name `native_key` is supposed to be an abstraction over the platform specific name or terminology. 
So it doesn't make much sense to name a platform specific implementation detail as the abstraction..

Does it make any sense?

bew:
I updated the cocoa implementation to use `keycode` as the key terminology for macos, @Luflosi test :pray: :smiley:

bew:
Is there something else I can do on my side?

bew:
Thanks @kovidgoyal :smiley:

Issue #879, opened 2018-09-06
Please submit xterm-kitty terminfo to ncurses database


z0rc:
As kitty introduced its own terminfo, running ssh from it is getting quite difficult because this terminfo isn't widely available. I'm aware about `kitty +kitten ssh` solution, but it isn't suitable when you have big amount of remote servers that constantly changing or running automation like Ansible or Kitchen.

Ad-hock solution with compiling terminfo on remote server works, but long-term solution would be submitting xterm-kitty to ncurses. After inclusion this would be automatically picked up by all Linux distributions in some time. Wide adoption would take years, but it's right approach. tmux, st and multiple other terminal emulators went this way. Of course this would require to make xterm-kitty a stable interface, and don't change it frequently.

ncurses accepts contributions in quite old way and doesn't use vcs of any sorts. In order to submit terminfo, please reach out developer as instructed at https://invisible-island.net/ncurses/ncurses.faq.html#report_bugs. More info on terminfo database at https://invisible-island.net/ncurses/#download_database.

z0rc:
Missing xterm-kitty in ncurses is a show stopper for me. If xterm-kitty is included in ncurses it's safe to assume that it's available everywhere, just like `bash` is available everywhere. Otherwise, imagine a team of 10 developers and hundred of various server roles, no way that change that satisfies some desktop quirks of single developer would be accepted across the servers infrastructure.

That said I'll reach out ncurses upstream and we'll see, how it goes.

maximbaz:
Just curious, why `kitty +kitten ssh` is not suitable for when you have big amount of remote servers? It seems to work fine for me, I simply aliased that command to `ssh` and forgot about this issue altogether.

kovidgoyal:
Even if ncurses included the terminfo file today, it would be years before you could be confident of having it available everywhere. But as I said, I am happy to provide any info/support needed to get the terminfo into ncurses.

kovidgoyal:
So to summarize the ncurses maintainer's response:

1) He does not like kitty's license (I'm happy to change the license of just the terminfo file to  CC license, if needed)
2) He does not like kitty TERM variable. This is not going to change as it would break lots of programs that turn on various features when they see a TERM variable of the type xterm-whatever. A situation that was created by the ncurses maintainer refusing to add new capabilities for various things.
3) He does not like the fact that kitty has introduced new capabilities into the terminal ecosystem

[...]

PR #2066, opened 2019-10-16
Document setting terminfo on remote hosts

barraponto:
See #879


Luflosi:
> although you might miss some kitty features

Changing the `TERM` variable also breaks a few things, you're not just missing out on features.

barraponto:
@Luflosi so maybe let's not recommend it at all, right?

kovidgoyal:
This issue is not really installation related, so I dont think the documentation belongs on the installation page. Forinstance it will occureven if you build kitty from source instead ofinstalling the binaries. Or if you use a distro or brew package of kitty. It is already documented in the kitty FAQ, where I think it belongs.

PR #2074, opened 2019-10-18
Fix triangle glyphs scaling by drawing them manually

### LINKED ISSUE #2006: Dynamic font scaling (with font size change keys) isn't perfect
ezgif-1-e542106769cf
The text is:
```
████████
██████
████████
```
(rendering in the browser is pretty broken ^^)
The offending char symbol is (hex unicode): `e0be` with the font: `DejaVuSansMono Nerd Font Mono`.

Do you have any idea why the font scaling is doing that? Can it be avoided?

bew:
Fixes #2006

It works so well!
:tada: :tada: :tada: 
Thanks for the help @kovidgoyal 

---
I updated the testing functions to use the new format of the class `setup_for_testing` (was probably a function before, now it uses a `with ...:` syntax).

I also changed something in icat, because the testing functions for box drawing make use of the method `show` from icat, which previously gave an error saying that `screen_size` is None at this line:
https://github.com/kovidgoyal/kitty/blob/26b7b1ec63198979a3d7795eaa36ed28646e7f3d/kittens/icat/main.py#L125
(which is called from `show`)

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -604,21 +604,21 @@ box_glyph_id(char_type ch) {
 START_ALLOW_CASE_RANGE
     switch(ch) {
         case 0x2500 ... 0x259f:
-            return ch - 0x2500;
-        case 0xe0b0 ... 0xe0d4:
-            return 0xa0 + ch - 0xe0b0;
+            return ch - 0x2500; // IDs from 0x00 to 0x9f
+        case 0xe0b0 ... 0xe0b4:
+            return 0xa0 + ch - 0xe0b0; // IDs from 0xa0 to 0xa4
         case 0xe0b6: // 
-            return 0x0fd;
+            return 0xa5;
         case 0xe0b8: // 
-            return 0x0fe;
+            return 0xa6;
         case 0xe0ba: // 
-            return 0x0ff;
+            return 0xa7;
         case 0xe0bc: // 
-            return 0x100;
+            return 0xa7;
         case 0xe0be: // 
-            return 0x101;
+            return 0xa8;
         default:
-            return 0xfff;
+            return 0xff;
     }
 END_ALLOW_CASE_RANGE
 }
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -556,10 +556,14 @@ START_ALLOW_CASE_RANGE
             return BLANK_FONT;
         case 0x2500 ... 0x2570:
         case 0x2574 ... 0x259f:
-        case 0xe0b0:
-        case 0xe0b2:
-        case 0xe0b4:
-        case 0xe0b6:
+        case 0xe0b0: // 
+        case 0xe0b2: // 
+        case 0xe0b4: // 
+        case 0xe0b6: // 
+        case 0xe0b8: // 
+        case 0xe0ba: //   
+        case 0xe0bc: // 
+        case 0xe0be: //   
             return BOX_FONT;
         default:
             ans = in_symbol_maps(fg, cpu_cell->ch);
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -543,6 +543,14 @@ in_symbol_maps(FontGroup *fg, char_type ch) {
 }
 
 
+// Decides which 'font' to use for a given cell.
+//
+// Possible results:
+// - NO_FONT
+// - MISSING_FONT
+// - BLANK_FONT
+// - BOX_FONT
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Please resolve the conflicts then I will merge.

bew:
The builds fail because there is a trailing whitespace somewhere... I'm not on my computer right now, so It's pretty hard for me to find it (github UI is not helpful), could you take a quick look?

As a sidenote, it would be nice to see where the trailing whitespace is in the CI

kitty/fonts.c
```diff
@@ -586,14 +598,25 @@ set_sprite(GPUCell *cell, sprite_index x, sprite_index y, sprite_index z) {
...
 static inline glyph_index
 box_glyph_id(char_type ch) {
 START_ALLOW_CASE_RANGE
     switch(ch) {
         case 0x2500 ... 0x259f:
-            return ch - 0x2500;
-        case 0xe0b0 ... 0xe0d4:
-            return 0xa0 + ch - 0xe0b0;
+            return ch - 0x2500; // IDs from 0x00 to 0x9f
+        case 0xe0b0 ... 0xe0b4:
```
bew: Note: there was an error here: `0xe0d4` instead of `0xe0b4`

bew:
Note: I didn't test this last change on glyph ids as I'm not on my computer right now, will do later today if you don't

bew:
I get a test failure:
```
FAIL: test_box_drawing (kitty_tests.fonts.Rendering)
  File "/home/circleci/project/kitty/launcher/../../kitty_tests/fonts.py", line 55, in test_box_drawing
    self.assertEqual(len(self.sprites), prerendered + len(box_chars))
    line = '─━│┃╌╍┄┅┈┉╎╏┆┇┊┋╴╵╶╷╸╹╺╻╼╽╾╿\ue0b0\ue0b2\ue0b4\ue0b6\ue0b1\ue0b3\ue0b8\ue0ba\ue0bc\ue0be═║╞╡╥╨╪╫╬╠╣╦╩╱╲╳▀▁▂▃▄▅▆▇█▉▊▋▌▍▎▏▐░▒▓▔▕▖▗▘▙▚▛▜▝▞▟┌┍┎┏┐┑┒┓└┕┖┗┘┙┚┛╭╮╯╰┼┽┾┿╀╁╂╃╄╅╆╇╈╉╊╋├┝┞┟┠┡┢┣┤┥┦┧┨┩┪┫┬┭┮┯┰┱┲┳┴┵┶┷┸┹┺┻╒╕╘╛╓╖╙╜╔╗╚╝╟╢╤╧'
    prerendered = 9
    s = <fast_data_types.Screen object at 0x564484aad320>
    self = <kitty_tests.fonts.Rendering testMethod=test_box_drawing>
AssertionError: 178 != 179
```
I'm not sure what to do here..

kitty/fonts.c
kovidgoyal: It is deliberate, its the whole block of powerline glyph ids, incase anymore are added in the future.

kitty/fonts/box_drawing.py
```diff
@@ -554,57 +554,57 @@ def test_char(ch, sz=48):
...
-            rows.append(join_cells(row))
-            rows.append(space_row)
-        rgb_data = b''.join(rows)
-        width *= 32
-        height *= len(rows)
-        assert len(rgb_data) == width * height * 4, '{} != {}'.format(len(rgb_data), width * height * 4)
-        display_bitmap(rgb_data, width, height)
-    finally:
-        set_send_sprite_to_gpu(None)
+    with setup_for_testing(family, sz) as (_, width, height):
```
Luflosi: Is there an advantage in using
```Python
with setup_for_testing(family, sz) as (_, width, height):
```
instead of just
```Python
(_, width, height) = setup_for_testing(family, sz)
```
?

bew: Well, `setup_for_testing` is actually a class now, not a function (and the previous code hasn't been updated when the change happened, and simply didn't work):
https://github.com/kovidgoyal/kitty/blob/7d8edccb04422a16e40594a71fef36dbdba669bc/kitty/fonts/render.py#L222

This syntax with the `with ... as ... : and an indented block` will do some operation before entering the block (the `__enter__` method of the returned obj), then do some more thing at the end of the block (the `__exit__` method).

Luflosi: Thanks for the explanation.

Issue #2081, opened 2019-10-20
kitty icat stopped workin


rien333:
https://github.com/kovidgoyal/kitty/commit/10e5fcc37535fdf6eed7f03f1c5380a2b4f0db9c broke `kitty +kitten icat` for me: 

``` shell
> kitty +kitten icat foo.jpg
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/bin/../lib/kitty/__main__.py", line 117, in <module>
    main()
  File "/usr/bin/../lib/kitty/__main__.py", line 113, in main
    func(sys.argv[1:])
  File "/usr/bin/../lib/kitty/__main__.py", line 12, in icat
    run_kitten('icat')
  File "/usr/bin/../lib/kitty/kittens/runner.py", line 103, in run_kitten
    runpy.run_module('kittens.{}.main'.format(kitten), run_name=run_name)
  File "/usr/lib/python3.7/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/bin/../lib/kitty/kittens/icat/main.py", line 400, in <module>
    main()
  File "/usr/bin/../lib/kitty/kittens/icat/main.py", line 348, in main
    if screen_size().width == 0:
TypeError: 'Size' object is not callable
```

Apparently, aforementioned commit made changes to the value of `screen_size()`, but I'm not sure why the error occurs. (tagging @bew, as they probably know what's up if they have time to look into it) 

## System info

Arch Linux, with kitty build from `HEAD`. Same error occurs in other terminals.

Luflosi:
Please test my PR linked above.

rien333:
Works great, thank you!

PR #2082, opened 2019-10-20
Fix icat kitten

Luflosi:
Fixes https://github.com/kovidgoyal/kitty/issues/2081.
This bug was introduced in 10e5fcc37535fdf6eed7f03f1c5380a2b4f0db9c.

One place in the code requires the actual `screen_size_function`, not just the result of it.


bew:
Nice catch, thanks for the fix!

PR #2088, opened 2019-10-23
Add AppData manifest file

tim77:
Please add initial AppData manifest file for modern package managers and easy installation on searching via GNOME Software/Discover KDE.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

### DIFF
--- a/kitty/data/kitty.appdata.xml
+++ b/kitty/data/kitty.appdata.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Artem Polishchuk <ego.cordatus@gmail.com> -->
+<component type="desktop">
+  <id>kitty</id>
+  <metadata_license>CC0</metadata_license>
+  <project_license>GPL-3.0-or-later</project_license>
+  <name>kitty</name>
+  <summary>Cross-platform, fast, feature full, GPU based terminal emulator</summary>
+  <description>
+    <p>Qt5 image viewer. Fast, configurable, easy to use. Optional video support.
+
+Key features:
+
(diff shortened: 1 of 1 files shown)


kovidgoyal:
kitty does not come with any desktop files/appstream files or any other desktop related metadata. Those can be provided by distro packages if they want it. kitty is a terminal emulator for power users of the command line, desktop files/app stores and the like are not really relevant to its target user base. Hell it does not even have a scrollbar :)

tim77:
Also:

> or any other desktop related metadata

Not really. `.desktop` file = desktop related metadata

kovidgoyal:
On Wed, Oct 23, 2019 at 07:37:53AM -0700, Artem Polishchuk wrote:
> Also:
> 
> > or any other desktop related metadata
> 
> Not really. `.desktop` file = desktop related metadata

There is no .desktop file.

tim77:
> There is no .desktop file.

Jesus Christ... 🤦🏻‍♂️
https://github.com/kovidgoyal/kitty/blob/master/setup.py#L721

kovidgoyal:
You are welcome to contribute the addition of an appdata file to
the code that generates the linux bundle. There is not going to be a
desktop file or an appdata file in kitty's source tree as there are
just useless noise there. Though I dont know what use a generated
appdata file is to anyone anyway, since it will only be dynamically
generated for linux bundles and presumably not help with software
centers or their ilk, though i confess I haven't a clue how they scrape
their metadata.

kovidgoyal:
Sure distros are welcome to add appdata files if they wish. The baseline kitty package is pretty barebones in terms of desktop integration anyway.

PR #2092, opened 2019-10-24
Simplify access to os.environ

Luflosi:
The `if` can be removed by using `get()` with a default parameter to access `os.environ`. This also reduces the number of accesses to `os.environ`.

### DIFF
--- a/kitty/constants.py
+++ b/kitty/constants.py
@@ -49,9 +49,8 @@ def _get_config_dir():
     locations.append(os.path.expanduser('~/.config'))
     if is_macos:
         locations.append(os.path.expanduser('~/Library/Preferences'))
-    if 'XDG_CONFIG_DIRS' in os.environ:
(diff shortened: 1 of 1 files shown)


Luflosi:
Wait, this doesn't work properly, sorry.
Don't merge this yet.

Luflosi:
`''.split(os.pathsep)` returns `['']`, which isn't what I need. I can't think of a way to make this work, so I'm going to close this PR.

Luflosi:
Making this work somehow would be really handy for a different piece of code as well. Can you think of a solution?

kovidgoyal:
Add a filter()

Luflosi:
Like this?
The code is not exactly a simpler than it was before, you need to decide if you still want to merge this PR.

Luflosi:
Thanks for the tip. I read that piece of documentation too but I didn't know, that the empty string evaluated to False.

PR #2105, opened 2019-10-28
Fix box_drawing.py so that it now draws full corner traingles

### LINKED ISSUE #2104: More Weird font glitch with nerd font icons
Much like #550, I'm seeing some weird glitches between Alacritty and kitty:

Kitty:

image

Alacritty:

image

The unicode point in question here is:
\uE0BC

which seems to be magically missing a line of pixels in some magical way on its top row.

Is there something that could be done about this?

From the little I can tell, this is already a special handled case in [box_drawing.py](https://github.com/kovidgoyal/kitty/blob/f93b605b758f210b368c7b3585976aea52fab261/kitty/fonts/box_drawing.py#L180):
https://github.com/kovidgoyal/kitty/blob/f93b605b758f210b368c7b3585976aea52fab261/kitty/fonts/box_drawing.py#L180-L193

I've managed to "fix" this by replacing the `height - 1` with `height`...
But this looks too easy?
Am I missing something?
Would you like to get a simple PR for this?

BTW: Awesome project!

I've just started using Kitty and I absolutely love it!
I can't wait to start to seriously play with the graphical protocol!

damageboy:
- Previous corner_triangles where missing one line of pixels which
  was evident when they were drawn immediately after a charchter with
  a background color that was different than the color of the triangle
- Closes #2104

### DIFF
--- a/kitty/fonts/box_drawing.py
+++ b/kitty/fonts/box_drawing.py
@@ -179,17 +179,17 @@ def triangle(buf, width, height, left=True):
 
 def corner_triangle(buf, width, height, corner):
     if corner == 'top-right' or corner == 'bottom-left':
-        diagonal_y = line_equation(0, 0, width - 1, height - 1)
+        diagonal_y = line_equation(0, 0, width - 1, height)
         if corner == 'top-right':
             xlimits = [(0, diagonal_y(x)) for x in range(width)]
         elif corner == 'bottom-left':
-            xlimits = [(diagonal_y(x), height - 1) for x in range(width)]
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Does height-1 have to be changed in line_equation() as well? It seems to me that changing it in xlimits is all that is required.

bew:
Hello @damageboy! please see [my comment](https://github.com/kovidgoyal/kitty/issues/2104#issuecomment-546933870) in your original issue

kovidgoyal:
no followup

PR #2125, opened 2019-11-11
make canberra sound theme configurable

terinjokes:
On Linux, kitty uses libcanberra to play an audible bell following the
XDG Sound Theming Specification. The library expects the application to
include the sound theme in the context, otherwise it plays from the
"freedesktop" theme; it does not query the environment or configuration.

Instead of linking against Gtk and querying GtkSettings, I've chosen to
allow the sound theme to be configured with a new kitty setting
"bell_sound_theme".

Change-Id: Ib6168fdd4a9ed5a8fb635ba02340613fb35ac426

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -523,6 +523,11 @@ o('bell_on_tab', True, long_text=_('''
 Show a bell symbol on the tab if a bell occurs in one of the windows in the
 tab and the window is not the currently focused window'''))
 
+o('bell_theme_name', 'freedesktop', long_text=_('''
+If audio bell is enabled on Linux, sets the XDG Sound Theme kitty will
+use to play the bell sound.
+'''))
+
 o('command_on_bell', 'none', option_type=to_cmdline, long_text=_('''
 Program to run when a bell occurs.
 '''))
--- a/kitty/data-types.h
+++ b/kitty/data-types.h
@@ -312,7 +312,7 @@ void set_special_key_combo(int glfw_key, int mods, bool is_native);
 void on_key_input(GLFWkeyevent *ev);
 void request_window_attention(id_type, bool);
 #ifndef __APPLE__
-void play_canberra_sound(const char *which_sound, const char *event_id);
+void play_canberra_sound(const char *which_sound, const char *event_id, const char *theme_name);
 #endif
 SPRITE_MAP_HANDLE alloc_sprite_map(unsigned int, unsigned int);
 SPRITE_MAP_HANDLE free_sprite_map(SPRITE_MAP_HANDLE);
--- a/kitty/desktop.c
+++ b/kitty/desktop.c
@@ -132,13 +132,14 @@ load_libcanberra(void) {
 }
 
 void
-play_canberra_sound(const char *which_sound, const char *event_id) {
+play_canberra_sound(const char *which_sound, const char *event_id, const char *theme_name) {
     load_libcanberra();
     if (libcanberra_handle == NULL || canberra_ctx == NULL) return;
     ca_context_play(
         canberra_ctx, 0,
         "event.id", which_sound,
         "event.description", event_id,
+        "canberra.xdg-theme.name", theme_name,
         NULL
     );
 }
--- a/kitty/glfw.c
+++ b/kitty/glfw.c
@@ -864,7 +864,7 @@ ring_audio_bell(OSWindow *w UNUSED) {
         glfwWindowBell(w->handle);
     }
 #else
-    play_canberra_sound("bell", "kitty bell");
+    play_canberra_sound("bell", "kitty bell", OPT(bell_theme_name));
 #endif
 }
 
--- a/kitty/state.c
+++ b/kitty/state.c
@@ -401,6 +401,11 @@ convert_mods(PyObject *obj) {
     return resolve_mods(PyLong_AsLong(obj));
 }
 
+static const char *
+pyunicode_as_char(PyObject *str) {
+   return PyUnicode_AsUTF8(str);
+}
+
 static WindowTitleIn
 window_title_in(PyObject *title_in) {
     const char *in = PyUnicode_AsUTF8(title_in);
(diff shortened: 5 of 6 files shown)


kovidgoyal:
I dont think users should have to configure sound themes in every
application, that's insane. IIRC the freedesktop sound theme spec is
based on the icon naming spec, which means you should be able to
configure it for all applications by creating ~/.local/share/sounds
with a freedesktop theme that inherits from the actual theme in use.

http://0pointer.de/public/sound-theme-spec.html

Or alternatively if only configuration via GTK settings is supported
which wont surprise me given the general brokennes of the Linux desktop,
it should be trivial to read from gtk-sound-theme-name from $XDG_CONFIG_HOME/gtk-3.0/settings.ini
using the python builtin INI parser no need to link against GTK (which
is never going to happen in kitty anyway).

terinjokes:
I agree with the premise that users shouldn't need to configure each application. Unfortunately, libcanberra has no logic to determine this configuration, it relies entirely on the application to provide it.

I didn't think you'd accept a patch linking to Gtk, nor am I suggesting doing so. If I were to attempt to duplicate the behavior by reading Gtk configurations, reading from Gtk 3.0's documentation may not be enough, Gnome users will have configuration in gsettings instead.

> a freedesktop theme that inherits from the actual theme in use

Using inheritance on the "freedesktop" theme may not be a widely supported use case, as it is the fallback if another theme doesn't include a sound. Pointing it at your usual theme using inheritance can result in some sounds not playing at all.

kovidgoyal:
On Sun, Nov 10, 2019 at 06:39:22PM -0800, Terin Stock wrote:
> I agree with the premise that users shouldn't need to configure each application. Unfortunately, libcanberra has no logic to determine this configuration, it relies entirely on the application to provide it.

The problem is if we add a config for this to kitty now, it means we
have to support it in perpetuity, even if linux desktop ppl ever get
their act together long enough to create a way to configure the sound
theme in a desktop neutral way. At which point libcanberra would
presumably learn how to read it automatically.
I hate adding this kind of carry-on-forever cruft to kitty.

> 
> I didn't think you'd accept a patch linking to Gtk, nor am I suggesting doing so. If I were to attempt to duplicate the behavior by reading Gtk configurations, reading from Gtk 3.0's documentation may not be enough, Gnome users will have configuration in gsettings instead.

I dont use gtk/gnome, if gnome has its own settings for this, you would
presumably need to link to GNOME not gtk to read it anyway? And whatever
system gnome uses for configuration presumably it is in a file
somewhere? How does KDE do it? Or MATE or LXDE or X/Y/Z?

kovidgoyal:
And just to note one can get the theme value from gsettings by running:

```
gsettings get org.gnome.desktop.sound theme-name
```

kovidgoyal:
And according to this, XFCE has its own setting: https://wiki.archlinux.org/index.php/Xfce#Sound_themes

terinjokes:
> And whatever system gnome uses for configuration presumably it is in a file somewhere?

Gnome uses gsettings, which is backed by a provider. Almost everyone would be using the dconf provider, which is stored in a special database format. Think of this as the Windows Registry for Linux.

> How does KDE do it?

KDE doesn't implement the XDG Sound Theme Specification. You have a notification panel and are expected to map each possible notification to whatever action you want to take. Often this is per-application.

I'm not familar enough with MATE and LXDE to comment.

> At which point libcanberra would presumably learn how to read it automatically.

Perhaps, but libcanberra is a bit unmantained nowadays, it hasn't seen any updates since 2013. Maybe they consider it "done"?

> And just to note one can get the theme value from gsettings by running:

This is true if the users uses Gnome or another desktop environment that uses gsettings.

> I dont use gtk/gnome, if gnome has its own settings for this, you would presumably need to link to GNOME not gtk to read it anyway?

I'm admitted that I'm not an expert in how Gtk settings are populated. I know [the example application](https://github.com/Distrotech/libcanberra/blob/edbf1733563f37c650d99fae1e826bf4a46ac3f7/src/canberra-gtk-play.c) links against `<gtk/gtk.h>` but somehow getts the configuration from Gnome when running under Gnome.

terinjokes:
It looks like Gnome and friends may populate XSETTINGS, and parse the gtkrc files if no XSETTINGS were set. I'll have to investigate more. If so, maybe I can query that instead.

kovidgoyal:
Also on a technical note, you cant use PyUnicode_AsUtf8() to store the theme name, because the memory it points to is only valid as long as the unicode object it comes from is alive. Better to have
a char[] in the options structure and copy into that.

kovidgoyal:
According to the registry there is no sound theme xsetting https://www.freedesktop.org/wiki/Specifications/XSettingsRegistry/

and xsettings will not work with wayland, so...

terinjokes:
> on X11 GTK talks to a XSETTINGS daemon that usually does this for you, or when missing, falls back to the user's setting.ini file.

[`[0]`](https://github.com/swaywm/sway/wiki/GTK-3-settings-on-Wayland)

> At the same time, modern GTK has way more settings exposed through XSettings than are documented in the registry. To find out what all of them are, you basically need to fire up gnome-settings-daemon temporarily and run dump_xsettings to extract them all.

[`[1]`](https://utcc.utoronto.ca/~cks/space/blog/linux/XSettingsNotes)

> The XSettings support in the Wayland backend is rudimentary. The supported settings are just hardwired to default values. XSettings was not a great success story in terms of adoption by other environments. One possible way forward would be to just drop the concept and instead port GtkSettings to use !GSettings directly. Alternatively, a simple interface could be provided by the compositor for these settings.
> 
> For now, the Wayland backend has been changed to use GSettings directly, when available.  

[`[2]`](https://wiki.gnome.org/Initiatives/Wayland/GTK%2B#XSettings)

---

Don't worry, _next year_ is the year of Linux on the desktop.

I can work on a patchset to query gsettings if available, otherwise query Gtk 3.0 settings. Hopefully, this doesn't put me on the hook for maintaining this for whatever KDE, MATE, or enlightenment decide to do.

kovidgoyal:
ugh, I am so glad I ditched linux desktop environments over a decade ago. Well OK then dont bother trying to read the setting, its not worth the aggravation. Rename the option to use a linux_ prefix and add a note to its docs that it might disappear in the future if linux ever gains a way to specify a system sound theme in a desktop agnostic manner. And fix the name storage issue and I will merge.

terinjokes:
Sure. I'll make those requested changes.

PR #2166, opened 2019-11-26
Build termcap file

Luflosi:
Since infocmp apparently can't generate a termcap file with all the capabilities, I let kitty do that instead.
Do you think that's a good idea?

Note that I have not yet tested the resulting termcap file.


kovidgoyal:
There's no point writing it out to a termcap file, since as far as I know, termcap dbs cant be read from random files. Instead just modify the FAQ entry to recommend doing something like

```
kitty +runpy "code to generate termcap" >> /etc/termcap
```

or whatever.

Luflosi:
I just tested this, it doesn't work yet. I'll have to investigate a little more.

PR #2204, opened 2019-12-18
Cocoa: Fix pre-window-creation event processing

Luflosi:
From upstream: https://github.com/glfw/glfw/commit/6e6805000ac7ddf39c8c5f6be3e877770cba5083.

Please check if the code added to `_glfwPlatformPostEmptyEvent()` is appropriate for kitty's version of GLFW or alternatively don't merge this.


kovidgoyal:
No this is not appropriate. In kitty, everything happens inside the
normal cocoa event loop, unlike in glfw. So this patch does not apply.

PR #2207, opened 2019-12-19
Fix POSIX conformance issues for clock_gettime

Luflosi:
From upstream: https://github.com/glfw/glfw/commit/081484ed340da6ccab5d0e3442c9d10a1b385ec3.

Does this commit introduce any problems similar to what my `monotonic_t` patch introduced?
Should I create a similar patch to change the macros in kitty's code?


kovidgoyal:
This does not apply since we dont use glfw's timers anymore.

Issue #2079, opened 2019-10-20
artifacts between terminal content and padding/margin space


mighty9245:
hello, i've noticed some artifacts to appear between terminal content and padding/margin space if you display an image (`kitty +kitten icat ...`, originally found while playing with `neofetch --kitty ...`).

this issue seems to be related to background opacity, because if background is fully opaque, i can't reproduce the issue. this issue also seems to be related to window geometry, because i can reproduce it only with specific window sizes.

displaying an image with `kitty +kitten icat ...` then opening and closing nano, `window_padding_width 12.0`, `background_opacity 0.8`:
![1571604674](https://user-images.githubusercontent.com/52252283/67166062-cd80f200-f394-11e9-8917-fe188f02f0a6.png)
displaying an image with `kitty +kitten icat ...` then opening and closing nano, `window_padding_width 12.0`, `background_opacity 1.0`:
![1571604972](https://user-images.githubusercontent.com/52252283/67166104-38cac400-f395-11e9-9f1a-3253160edb50.png)

kitty config: https://hastebin.com/raw/tenageyoqe

kitty 0.14.6
compton v.7.4
both from arch linux repositories (not aur)

mighty9245:
i also noticed that in some transitions (when resizing a window, for example) margin space has different color/opacity, but i guess it may be related to compton

mighty9245:
initially i thought there is only vertical lines, but as you can see there is a kind of frame around the image. according to the `xwininfo` utility, the window size is 1014x911 (1010x907 w/o borders, i guess they are included) and the image size is 1400x1273, window padding is set to 12 and background opacity to 0.8. i hope you will be able to reproduce the issue with this information
![123](https://user-images.githubusercontent.com/52252283/67417459-7ddb3a00-f5d1-11e9-83fd-b4ae78b9c1ef.png)

Luflosi:
@mighty9245 can you try compiling kitty from source with the patch from #2214 applied and see if that fixes your issue?

mighty9245:
@Luflosi, nah, the issue is still here. i would recommend you to play with terminal window geometry (resizing it and running the test) and you'll probably see that the issue is still here even with your patch.

btw i was playing with these lines adding -1 and +1 a while ago with no luck.

PR #2214, opened 2019-12-19
Fix the scissor limits (again)

Luflosi:
When I execute the test case
```bash
kitty -o background=cyan -o background_opacity=0.7 -o window_margin_width=40 sh -c "kitty +kitten icat logo/kitty.png; read"
```
from the comment in the code, I get the following:
original
With this commit, it looks like this:
fixed
Please check that this patch actually makes sense and is the correct way to fix this as I don't know how the scissor limits work.

### DIFF
--- a/kitty/shaders.c
+++ b/kitty/shaders.c
@@ -512,10 +512,10 @@ draw_cells(ssize_t vao_idx, ssize_t gvao_idx, GLfloat xstart, GLfloat ystart, GL
     // kitty -o background=cyan -o background_opacity=0.7 -o window_margin_width=40 sh -c "kitty +kitten icat logo/kitty.png; read"
 #define SCALE(w, x) ((GLfloat)(os_window->viewport_##w) * (GLfloat)(x))
     glScissor(
-            (GLint)(ceilf(SCALE(width, (xstart + 1.0f) / 2.0f))), // x
-            (GLint)(ceilf(SCALE(height, ((ystart - h) + 1.0f) / 2.0f))), // y
-            (GLsizei)(floorf(SCALE(width, w / 2.0f))), // width
-            (GLsizei)(floorf(SCALE(height, h / 2.0f))) // height
+            (GLint)(round(SCALE(width, (xstart + 1.0f) / 2.0f))), // x
+            (GLint)(round(SCALE(height, ((ystart - h) + 1.0f) / 2.0f))), // y
+            (GLsizei)(round(SCALE(width, w / 2.0f))), // width
+            (GLsizei)(round(SCALE(height, h / 2.0f))) // height
     );
 #undef SCALE
     if (os_window->is_semi_transparent) {


mighty9245:
#2079 related

kovidgoyal:
I doubt there is a well defined way to do this, basically OpenGL worksin float co-ords but scissor works in pixels, so there is no way to precisely map one to the other. Your patch does not work on my machine, but the following does:

```
diff --git a/kitty/shaders.c b/kitty/shaders.c
index e62fff3d..f478dae5 100644
--- a/kitty/shaders.c
+++ b/kitty/shaders.c
@@ -512,10 +512,10 @@ draw_cells(ssize_t vao_idx, ssize_t gvao_idx, GLfloat xstart, GLfloat ystart, GL
     // kitty -o background=cyan -o background_opacity=0.7 -o window_margin_width=40 sh -c "kitty +kitten icat logo/kitty.png; read"
 #define SCALE(w, x) ((GLfloat)(os_window->viewport_##w) * (GLfloat)(x))
     glScissor(
-            (GLint)(ceilf(SCALE(width, (xstart + 1.0f) / 2.0f))), // x
-            (GLint)(ceilf(SCALE(height, ((ystart - h) + 1.0f) / 2.0f))), // y
-            (GLsizei)(floorf(SCALE(width, w / 2.0f))), // width
-            (GLsizei)(floorf(SCALE(height, h / 2.0f))) // height
+        (GLint)ceilf(SCALE(width, (xstart + 1.f)/2.f)),
+        (GLint)floorf(SCALE(height, (ystart - h + 1.f)/2.f)),
+        (GLsizei)ceilf(SCALE(width, w / 2.f)),
+        (GLsizei)(floorf(SCALE(height, h / 2.f))) // height
     );
 #undef SCALE
     if (os_window->is_semi_transparent) {
```

Luflosi:
Your patch doesn't work for me 😕
screenshot

mighty9245:
for me too :(

kovidgoyal:
I didn't expect it too, fixing this is kinda difficult/impossible, which is why I
have not tackled it so far.

kovidgoyal:
I think the best approach might be to use round() rather than ceil() and floor()

mighty9245:
i replaced ceilf and floorf with roundf and this seems to solve the issue...

Luflosi:
When I replace `ceil()` and `floor()`in your patch with `round()`, it works for me 🎉. I can even resize the window without artefacts.

Luflosi:
You added a printf comment and made the formatting a bit worse. Other than that, it works now 🎉.

kovidgoyal:
The printf() was instrumental in me figuring out that round() is what was needed :)

Luflosi:
Well, if you added it on purpose, you can leave it in. I just thought you may have added it accidentally.

PR #2218, opened 2019-12-20
regenerates glad to fix missing APIENTRY

Dav1dde:
There was a bug in glad2 that omitted the APIENTRY macro for callback functions. This is not a big deal for Linux nor Windows 64bit but on Windows 32bit this will cause issues. I recently fixed that in glad. This is the PR to fix this problem for kitty.

Sorry this happened, luckily doesn't seem like anyone was affected by that issue.


Luflosi:
Well, kitty doesn't run on Windows, so don't worry, nobody could possibly have been affected.

kovidgoyal:
No worries, and thanks for the PR :)

PR #2222, opened 2019-12-22
Add circumflex (^) key

Luflosi:
Please double-check that I didn't miss to change anything, especially in `kitty/keys.py`.


kovidgoyal:
I dont think it should be in unshifted printable, since it is not a unshifted key in most keyboard layouts.

PR #2226, opened 2019-12-23
Premultiply resizing text with the alpha value

Luflosi:
The resizing text needs to be premultiplied or the anti-aliasing will look wrong. The issue seems to be more pronounced on grey to white backgrounds.
To reproduce the issue start kitty with `kitty --config NONE -o background_opacity=0.3 -o resize_draw_strategy=size -o foreground=#899` and resize the window.
`graphics_fragment.glsl` has `#define ALPHA_TYPE` at the top, which will be replaced by one of `SIMPLE`, `PREMULT` or `ALPHA_MASK` compile time, so it is impossible to define both `PREMULT` and `ALPHA_MASK` at the same time. This is why I premultiply in the `#ifdef ALPHA_MASK`.

Without this fix:
before
With:
after
The text is hard to read because of the poor contrast between the text color and the wallpaper but this is expected. I chose this combination of colours because it was fairly easy to see the anti-aliasing issue.
Please double-check that you can reproduce the issue on your machine.

### DIFF
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -633,7 +633,7 @@ draw_resizing_text(OSWindow *w) {
     snprintf(text, sizeof(text), "%u x %u cells", width / w->fonts_data->cell_width, height / w->fonts_data->cell_height);
     StringCanvas rendered = render_simple_text(w->fonts_data, text);
     if (rendered.canvas) {
-        draw_centered_alpha_mask(w->gvao_idx, width, height, rendered.width, rendered.height, rendered.canvas);
+        draw_centered_alpha_mask(w->is_semi_transparent, w->gvao_idx, width, height, rendered.width, rendered.height, rendered.canvas);
         free(rendered.canvas);
     }
 }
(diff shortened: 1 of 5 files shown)


kovidgoyal:
I cannot reproduce on my machine, but then I am not very sensitive to anti-aliasing issues in the first place. It looks about equally bad with and without the patch. Also looking in shaders.c you probably need to change BLEND_ONTO_OPAQUE in draw_centered_alpha_mask as well.

Luflosi:
Can you elaborate please? I don't understand what may need to be changed as well. I also don't understand what exactly the parameters of the blend function mean. The [documentation](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendFunc.xhtml) wasn't very helpful for me.

kovidgoyal:
In a transparent window you would not use blend onto opaque instead use
pre-multiplied color and blend_premult. In an opaque window you would use
blend_opaque and notuse premultiplied colors.

Luflosi:
Did I fix it properly this time? Was this what you meant?

PR #2236, opened 2019-12-29
Add help menu to macOS menu bar

Luflosi:
Most macOS Apps have a Help menu as the last menu item in the menu bar. They usually have a menu item with the keyboard shortcut <kbd>⌘</kbd>+<kbd>?</kbd>, which shows some kind of help or documentation. I named the menu item "Visit kitty website" and let it open the kitty website as documentation.
macOS also magically adds a search feature to the help menu.

Is hardcoding the URL like this a good idea? Or can I get it from somewhere else?


kovidgoyal:
Hardcoding is fine.

PR #2248, opened 2020-01-04
Add `font_feature_settings`

### LINKED ISSUE #2247: Disabling ligatures disables other OpenType features (which are enabled by default) too
I discovered this while thinking about ctrlcctrlv/TT2020#3

The problem is this:

https://github.com/kovidgoyal/kitty/blob/70071fe1f679f8fbd514fb061df02b55628e7c9f/kitty/fonts.c#L779

It turns off all OpenType shaping if `disable_ligatures` is true, not only ligatures. `disable_shaping` is a more accurate name for this option.

It would be better if I could have more fine tuning, e.g.&mdash; 

```yaml
disable_ligatures always # Internally appends to below option "liga" 0, "clig" 0, "dlig" 0 
font_shaping_tags "calt" 1, "ss02" 1
```

> [`hb_feature_from_string()`](http://www.manpagez.com/html/harfbuzz/harfbuzz-1.0.1/harfbuzz-Shaping.php#hb-feature-from-string) now accepts CSS font-feature-settings format.

I am working on this right now so would appreciate being given time to submit a PR. If I fail to figure it out I'll let you know, but the code seems well done, and the problem simple (for me, since I know my way around OpenType fonts and also HarfBuzz).

ctrlcctrlv:
Close #2247

This doesn't look like very much code; but as they say, **reading code is much harder than writing code**, and I had to really understand how Kitty works to write this, so it took a long time. :smile: 

2020-01-04-141156_2960x294_scrot
2020-01-04-140802_2960x294_scrot
2020-01-04-141025_2960x294_scrot
2020-01-04-140923_2960x294_scrot

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -360,24 +360,24 @@ init_font(Font *f, PyObject *face, bool bold, bool italic, bool emoji_presentati
     f->bold = bold; f->italic = italic; f->emoji_presentation = emoji_presentation;
     f->num_hb_features = 0;
     const char *psname = postscript_name_for_face(face);
-    if (strstr(psname, "NimbusMonoPS-") == psname) {
-        copy_hb_feature(f, LIGA_FEATURE); copy_hb_feature(f, DLIG_FEATURE);
-    }
+    copy_hb_feature(f, LIGA_FEATURE);
+    copy_hb_feature(f, DLIG_FEATURE);
     copy_hb_feature(f, CALT_FEATURE);
     if (font_feature_settings != NULL){
-        const char* face = postscript_name_for_face(f->face);
-
-        PyObject* o = PyDict_GetItemString(font_feature_settings, face);
+        PyObject* o = PyDict_GetItemString(font_feature_settings, psname);
         if (o != NULL) {
             long len = PyList_Size(o);
-            hb_feature_t* hb_feat = calloc(len, sizeof(hb_feature_t));
+            if (len==0) return true;
+            f->num_ffs_hb_features = len + f->num_hb_features;
+            hb_feature_t* hb_feat = calloc(f->num_ffs_hb_features, sizeof(hb_feature_t));
             for (long i = len-1; i >= 0; i--) {
                 PyObject* item = PyList_GetItem(o, i);
                 const char* feat = PyUnicode_AsUTF8(item);
                 hb_feature_from_string(feat, -1, &hb_feat[i]);
             }
+            for (size_t i = 0; i < f->num_hb_features; i++)
+            hb_feat[len+i] = hb_features[i];
             f->ffs_hb_features = hb_feat;
-            f->num_ffs_hb_features = len;
         }
     }
     return true;
@@ -386,6 +386,8 @@ init_font(Font *f, PyObject *face, bool bold, bool italic, bool emoji_presentati
 static inline void
 del_font(Font *f) {
     Py_CLEAR(f->face);
+    if (f->num_ffs_hb_features > 0)
+    free(f->ffs_hb_features);
     free_maps(f);
     f->bold = false; f->italic = false;
 }
@@ -794,9 +796,9 @@ shape(CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, index_type num_cells, hb
     load_hb_buffer(first_cpu_cell, first_gpu_cell, num_cells);
 
     if (fobj->num_ffs_hb_features > 0)
-        hb_shape(font, harfbuzz_buffer, fobj->ffs_hb_features, fobj->num_ffs_hb_features);
+        hb_shape(font, harfbuzz_buffer, fobj->ffs_hb_features, fobj->num_ffs_hb_features - (disable_ligature ? 0 : fobj->num_hb_features));
     else
-        hb_shape(font, harfbuzz_buffer, fobj->hb_features, fobj->num_hb_features - (disable_ligature ? 0 : 1));
+        hb_shape(font, harfbuzz_buffer, fobj->hb_features, fobj->num_hb_features - (disable_ligature ? 0 : fobj->num_hb_features));
 
     unsigned int info_length, positions_length;
     group_state.info = hb_buffer_get_glyph_infos(harfbuzz_buffer, &info_length);
@@ -1079,7 +1081,7 @@ render_run(FontGroup *fg, CPUCell *first_cpu_cell, GPUCell *first_gpu_cell, inde
         default:
             shape_run(first_cpu_cell, first_gpu_cell, num_cells, &fg->fonts[font_idx], disable_ligature_strategy == DISABLE_LIGATURES_ALWAYS);
             if (pua_space_ligature) merge_groups_for_pua_space_ligature();
-            else if (cursor_offset > -1) {
+            else if (cursor_offset > -1) { // false if DISABLE_LIGATURES_NEVER
                 index_type left, right;
                 split_run_at_offset(cursor_offset, &left, &right);
                 if (right > left) {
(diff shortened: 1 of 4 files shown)


kitty/fonts.c
```diff
@@ -1184,11 +1204,11 @@ static PyObject*
...
     PyObject *sm;
     Py_CLEAR(box_drawing_function); Py_CLEAR(prerender_function); Py_CLEAR(descriptor_for_idx);
-    if (!PyArg_ParseTuple(args, "OOOIIIIO!d",
+    if (!PyArg_ParseTuple(args, "OOOIIIIO!dO",
                 &box_drawing_function, &prerender_function, &descriptor_for_idx,
                 &descriptor_indices.bold, &descriptor_indices.italic, &descriptor_indices.bi, &descriptor_indices.num_symbol_fonts,
-                &PyTuple_Type, &sm, &global_state.font_sz_in_pts)) return NULL;
-    Py_INCREF(box_drawing_function); Py_INCREF(prerender_function); Py_INCREF(descriptor_for_idx);
+                &PyTuple_Type, &sm, &global_state.font_sz_in_pts, &font_feature_settings)) return NULL;
+    Py_INCREF(box_drawing_function); Py_INCREF(prerender_function); Py_INCREF(descriptor_for_idx); Py_INCREF(font_feature_settings);
```
martinetd: If you're doing Py_INCREF it also needs a Py_CLEAR a few lines above to not leak when function is called again, as well as in finalize() ~100 lines below (see for example how descriptor_for_idx or any of the others are handled)

ctrlcctrlv: Thank you, fixed :)

kovidgoyal:
So before I review the actual code, a couple of comments about the design of the option:

1) disable_ligatures should not be disabled by this setting. This is because it can be used to turn off ligatures only under the cursor as well, a feature I would be loth to loose. To maintain backward compatibility, I think disable ligatures should remain and simply add -calt to the features as per its value + cursor position.

2) kitty runs on macOS as well where fontconfig is not used, so it would probably be best to modify kitty + list-fonts to have a --postscript flag that will cause it to output postscript names instead.

ctrlcctrlv:
Backwards compatibility is not threatened by this patch because if it's unset, `disable_ligatures` still works. It only works differently if set.

However, perhaps it is better to make it so that it is also possible to specify a set of feature tags on hover as well?

The reason I don't want to "simply add -calt" is because every font works differently. Fira Code uses `calt`, but some other fonts use `liga`

ctrlcctrlv:
For complaint &numero;2, how about I just add the information to `kitty --debug-font-fallback`?

ctrlcctrlv:
<sup>(Sorry for the triple post, more thoughts keep coming to mind.)</sup>

@kovidgoyal What do you think about this syntax?

```
font_feature_settings FiraCode-Retina: +onum
font_feature_settings FiraCode-Retina:hover +onum -liga -calt
```

kovidgoyal:
On Sat, Jan 04, 2020 at 12:44:46AM -0800, Fredrick Brennan wrote:
> Backwards compatibility is not threatened by this patch because if it's unset, `disable_ligatures` still works. It only works differently if set.

Yes, thats the backward compat I am referring to.

> 
> However, perhaps it is better to make it so that it is also possible to specify a set of feature tags on hover as well?

The problem with this is it is much harder to use compared to
disable_ligatures. You have to know the postscript names of all fonts
you use and set it for each individually. disable_ligatures is just
single simple setting.

> 
> The reason I don't want to "simply add -calt" is because every font works differently. Fira Code uses `calt`, but some other fonts use `liga`

No monospace font I know of uses anything other than CALT for
programming ligatures, which is what disable ligatures is about
disabling, not ligatures in general, which anyway dont render well
monospaced. 

If you like you can have disable ligatures optionally take feature
names, and when no feature names are specified, defaults to CALT.

kovidgoyal:
On Sat, Jan 04, 2020 at 12:47:22AM -0800, Fredrick Brennan wrote:
> For complaint &numero;2, how about I just add the information to `kitty --debug-font-fallback`?

Leave it as is, I will take care of adding psnames to list-fonts.

ctrlcctrlv:
> If you like you can have disable ligatures optionally take feature names, and when no feature names are specified, defaults to CALT.

I do like, but are you sure you don't want it to be font-keyed?

kovidgoyal:
On Sat, Jan 04, 2020 at 12:57:01AM -0800, Fredrick Brennan wrote:
> > If you like you can have disable ligatures optionally take feature names, and when no feature names are specified, defaults to CALT.
> 
> I do like, but are you sure you don't want it to be font-keyed?

Yes, I am fine with adding syntax to font_features to specify for cursor
or not in addition, but disable_ligatures should be global.

ctrlcctrlv:
Understood, I will do it. No need to review before then. Thanks Kovid

(Also, wow, you wrote Calibre too? That program uses Python extensions better than any software I've ever had the pleasure of working on. I tried to convince another FontForge developer of its superiority, but I haven't quite managed to do so yet. If I had it my way we'd copy how you did it.)

kovidgoyal:
On Sat, Jan 04, 2020 at 01:15:12AM -0800, Fredrick Brennan wrote:
> Understood, I will do it. No need to review before then. Thanks Kovid

Thanks :)

> 
> (Also, wow, you wrote Calibre too? That program uses Python extensions better than any software I've ever had the pleasure of working on. I tried to convince another FontForge developer of its superiority, but I haven't quite managed to do so yet. If I had it my way we'd copy how you did it.)

Been using Python+C for a long time now, and I was never happy with
pythons' default methods for building/integrating with C extensions, so
rolled my own. Fortunately my techniques have aged well :)

ctrlcctrlv:
@kovidgoyal I've pushed a change for your review; it makes `disable_ligatures` into a global flag which affects all fonts.

Just so you know, I have decided to shelve the hover syntax for now. I couldn't think of a case where it's actually useful, and to implement it would mean more work than I have time for right now for something I won't use myself. If someone has a real use idea for it, and I like it, I might open a new PR for it in future though. 

I hope that's okay with you.

Diaoul:
Would that be possible to default reading font_features from fontconfig on linux when available? It seems to be what it is done to read the font_family no?
Let me know if this is something you would consider.

kovidgoyal:
Sure, dont think I am interestedin writing the code for it, but a PR is
welcome.

ctrlcctrlv:
Hi @Diaoul. I don't understand what you mean. I might be interested if you explain. Are you saying that there is a way to instruct Fontconfig as to what features should be enabled whenever a font is used?

kovidgoyal:
yes, you can: https://protesilaos.com/codelog/2019-07-25-opentype-features-fontconfig/

ctrlcctrlv:
Glad to see fontconfig support has been done, now using it myself :-) Thanks @Diaoul! Always nice when I implement a PR then my PR is improved. :-D

note `0` glyph: 
image

```xml
<!-- $ cat ~/.config/fontconfig/conf.d/99-fira-code-fontfeatures.conf -->
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
  <description>Enable select opentype features for FiraGO.</description>

  <!--
	 Typographic features are provided on a per-typeface basis.
	 Tables with all available features are available here:
	 https://en.wikipedia.org/wiki/OpenType_feature_tag_list 

      Also read the fontconfig user spec:
      https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
  -->

  <match target="font">
    <test name="family" compare="eq" ignore-blanks="true">
      <string>Fira Code</string>
    </test>
    <edit name="fontfeatures" mode="append">
      <string>tnum on</string> <!-- tabular numbers -->
      <string>zero on</string> <!-- slashed zero -->
    </edit>
  </match>
</fontconfig>
```

PR #2250, opened 2020-01-04
Use a python f'' string instead of string.format()

Luflosi:
I think this is a bit more readable.


kovidgoyal:
f strings are not available in python 3.5

Luflosi:
Ok, sorry.

bew:
Could we update the minimal python version?

kovidgoyal:
Someday, sure.

Luflosi:
Python 3.5 is no longer supported according to https://devguide.python.org/#status-of-python-branches. Support was ended on 2020-09-13. Is now the time to update the minimal version to 3.6?

kovidgoyal:
The minimum python version is already 3.6

Luflosi:
I missed that, sorry.

Issue #163, opened 2017-11-05
background image


rtorresware:
now that we can show images in the kitty terminal (awesome), can we get support for background images?

kovidgoyal:
Not something I care about, personally. But it should be trivial to implement. Patches welcome.

coxley:
Would love if this had a toggle for current wallpaper + dim. Only feature keeping me from `kitty` full time :)

teratorn:
+1 need Fallout PipBoy-terminal background image support... going fully transparent against the desktop wallpaper results in horribly broken anti-aliasing and looks like garbage on MacOS.

(On a side note - defaulting to pure-black background hurts peoples eyes. You can't focus in to a black-hole, and you *should* have something for the eyes to focus on if you want to avoid eye-strain... a nice grey would be a better default, imho)

If I find time, I'll try to work on this. Are there PNG and JPG image loaders already in the codebase?

Luflosi:
@teratorn #1005 describes your anti-aliasing issue.
I agree that the black background color is not very good but I'm not sure it should be changed since this is personal preference and most people will probably change it anyways and might not like a different default color any more than the current one.

kovidgoyal:
@teratorn kitty already links against libpng and uses it for its graphics protocol. just search for the png headers to see where and how it is used. As for staring into black, I strongly disagree. I find it very relaxing for my eyes. Not to mention that monitors displaying a black color are not black holes, they still emit light.

giftig:
I've just swapped to using kitty today and I'm loving it; much better than all the other options I've tried in linux. I came here to ask for exactly this feature since I like having pokemon images in my terminal like a nerd.

If it's something you think is trivial to implement @kovidgoyal  I'll gladly have a crack at it sometime soon if you can give me some vague pointers about where to be looking in your codebase to build it. I'm a python dev but not a C dev so I might scratch my head a bit if it involves much C.

kovidgoyal:
Well, this will definitely require a fair bit of C code. You will need to load the PNG image usingthe same code that does it for kitty's exisint image display capabilities (see grahoics.c) and then display it in the kitty graphics shader in shaders.c. This will all be C code, I'm afraid.

jason81474:
I need coins

dagadbm:
will this ever be done?

disruptek:
[...]

PR #2271, opened 2020-01-11
Allow drawing images below cells w/background

ctrlcctrlv:
This is part 1 of solving #163.

It allows users to draw an image below cells that might have a background, thus:

2020-01-10-200813_2560x1401_scrot

The image is .25 opacity, while the window is .8 opacity, with a video below.

The next steps, after this is accepted, is to add a configuration option for background images, whether to tile or scale, and perhaps even an opacity for convenience (we might expect users are too lazy to just use GIMP to add the appropriate alpha channel). Now that the rendering is all figured out, all I need to do for that is add a new slot to `GraphicsManager` (graphics.h), and add the appropriate calls in the appropriate places, below CELL_DEFAULT_BG_PROGRAM but above CELL_BG_PROGRAM, similar to what I've done in this commit.

### DIFF
--- a/docs/graphics-protocol.rst
+++ b/docs/graphics-protocol.rst
@@ -335,7 +335,8 @@ placed in the same location with different z-index values will be blended if
 they are semi-transparent. You can specify z-index values using the ``z`` key.
 Negative z-index values mean that the images will be drawn under the text. This
 allows rendering of text on top of images. Negative z-index values below
-INT32_MIN/2 will be drawn under the background.
+INT32_MIN/2 (-1073741824) will be drawn under cells with non-default background
+colors.
 
 Deleting images
 ---------------------
--- a/docs/graphics-protocol.rst
+++ b/docs/graphics-protocol.rst
@@ -334,7 +334,8 @@ Finally, you can specify the image *z-index*, i.e. the vertical stacking order. 
 placed in the same location with different z-index values will be blended if
 they are semi-transparent. You can specify z-index values using the ``z`` key.
 Negative z-index values mean that the images will be drawn under the text. This
-allows rendering of text on top of images.
+allows rendering of text on top of images. Negative z-index values below
+INT32_MIN/2 will be drawn under the background.
 
 Deleting images
 ---------------------
@@ -417,7 +418,7 @@ Key      Value                 Default    Description
 ``Y``    Positive integer      ``0``      The y-offset within the first cell at which to start displaying the image
 ``c``    Positive integer      ``0``      The number of columns to display the image over
 ``r``    Positive integer      ``0``      The number of rows to display the image over
-``z``    Integer               ``0``      The *z-index* vertical stacking order of the image
+``z``    32-bit integer        ``0``      The *z-index* vertical stacking order of the image
 **Keys for deleting images**
 -----------------------------------------------------------
 ``d``    Single character.     ``a``      What to delete.
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -129,18 +129,19 @@ void main() {
 #endif
 
 #if (defined(DEFAULTBG) || defined(BACKGROUND))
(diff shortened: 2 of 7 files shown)


kovidgoyal:
Some comments:

1) Change the docs text 
```diff
diff --git a/docs/graphics-protocol.rst b/docs/graphics-protocol.rst
index 4d3314af..5cadada6 100644
--- a/docs/graphics-protocol.rst
+++ b/docs/graphics-protocol.rst
@@ -335,7 +335,8 @@ placed in the same location with different z-index values will be blended if
 they are semi-transparent. You can specify z-index values using the ``z`` key.
 Negative z-index values mean that the images will be drawn under the text. This
 allows rendering of text on top of images. Negative z-index values below
-INT32_MIN/2 will be drawn under the background.
+-2,147,483,648 will be drawn under the background. Note that this applies only
+for cells with non-default background colors.
 
 Deleting images
 ---------------------
```

2) Dont use `if` in shaders it has terrible performance implications. Instead use the same techniques as used for bg_alpha

ctrlcctrlv:
Just so you know, it's actually -1073741824

I've updated the doc to reflect that, and also wrote the shader in a more "clever" way

kovidgoyal:
I'm not sure I understand why you need to 1) draw the default bg, then draw the below images and then draw the non default background. Cant you just 1) draw the below images 2) draw the background, setting bg_alpha to 0 for cells with default background??

kovidgoyal:
Is it because of semi-transparent images that need to blend with th edefault bg?

kovidgoyal:
In fact, you could avoid the extra draw call completely by simply changing step 2a.3 to output the background color with bg_alpha = 1 for all non default bg cells and 0 otherwise.

kovidgoyal:
Also, dont calculate bgfac per pixel, i.e. in the fragment shader. Instead output it from the vertex shader, that will be approx 100 times faster. Indeed you dont need to output defaultbg from the vertex shader at all, just output bgfac.

kovidgoyal:
I think I understand why you need the extra DEFAULT_BG draw call, in that case, the shader should do it only if there actually are below images, which mostly there will not be, so that in the common case there is no extra draw call and blend. You could use a uniform to toggle the behavior of CELL_BG_PROGRAM to either draw the default bg or not.

kovidgoyal:
With those two optimizations, namely moving bgfac calculation into vertex shader and only doing the extra DEFAULTBG draw call if there are actually below images, this looks good to merge.

ctrlcctrlv:
I'm not sure how to use a uniform like you say without an if statement inside the shader. Would you mind doing that part?

I have however moved bgfac into the vertex shader, good idea.

kovidgoyal:
OK I'll take care of it when I have a minute.

kovidgoyal:
I have committed code for it, could you please test it and see it does what is expected. In particular in the case of transparent windows, as I am not currently at a system that has that capability.

ctrlcctrlv:
Will do Kovid thank you for your help as always. I will be home in an hour or so. I'm a Christian and went to church today. It's not far but the traffic is intense. As you're in Mumbai I'm sure you understand

kovidgoyal:
Hope church was nice. There is an issue here in the semi-transparent background case. drawing the background twice is going to result in correct alpha values. So probably the interleaved_premult case will need to be redesigned, to draw the default and non-default backgrounds separately as you had done it, when below bg images are present.

kovidgoyal:
Or alternately, just set background_opacity to zero with another uniform on default cells.

kovidgoyal:
Actually hold off I think I have a way to do this nicely. I'll update you when I am done tinkering.

ctrlcctrlv:
👍

kovidgoyal:
Am done tinkering check it out. It draws background for all cells and then if below images a re present, draws the below images and then background for non-default cells. Double draw of non default bg cells is fine, since they are opaque anyway.

Issue #2255, opened 2020-01-05
X11 crash 


rien333:
The [newest  commit](https://github.com/kovidgoyal/kitty/commit/0d4a58e9bfb6f505906be180fbc9cb5c51783744) related to X11 completely broke kitty for me. Previously, I was on the commit just before that, which worked fine (I still have a working instance group from that commit open). 

@Luflosi 

## Behavior 

A window briefly flashes, but disappears immediately, and the following error appears:
```
$ kitty
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  40 (X_TranslateCoords)
  Resource id in failed request:  0x0
  Serial number of failed request:  419
  Current serial number in output stream:  419
```

## System info

```console
$ uname -mor
5.4.7-arch1-1 x86_64 GNU/Linux
$ Xorg -version

X.Org X Server 1.20.6
X Protocol Version 11, Revision 0
Build Operating System: Linux Arch Linux
```

kovidgoyal:
Well, I just dropped that commit, since kitty does not care about window
positions. I cant replicate the issue and without that trying to fix
glfw's code is too much work. @Luflosi feel free to actually fix the
issue if you like.

Luflosi:
I don't care enough to actually fix the issue. I also couldn't replicate the issue with xwayland.
You reset to 58688dda38957f664051583f64095c4b87870663 but I think you meant to reset to 70071fe1f679f8fbd514fb061df02b55628e7c9f.

kovidgoyal:
Rebasing drops merge commits, the merge commit in question is not
important anyway.

Luflosi:
It is important to me because I made a few commits on 70071fe1f679f8fbd514fb061df02b55628e7c9f. Rebasing those would be inconvenient. Could you please do `git reset 70071fe1f679f8fbd514fb061df02b55628e7c9f && git push`?

kovidgoyal:
sure

Luflosi:
Thank you.

Luflosi:
According to https://github.com/glfw/glfw/issues/1620, this only happens without window decorations. I presume that a fix is being worked on, so I can just try to apply this commit again when a fix is available.
@rien333 can you confirm that you've turned window decorations off?

rien333:
Yes, I have `hide_window_decorations yes` in my `kitty.conf`, and I don't think my WM even supports them.

PR #2285, opened 2020-01-15
X11: Fix no window position events during resize, Attempt 2

Luflosi:
From upstream: glfw/glfw@fe57e3c and glfw/glfw@9372ba9.

This is #2251 again with a fix for the bug reported in #2255.

@rien333 please test that the fix works properly.


rien333:
Can't find any problems!

PR #2286, opened 2020-01-15
macOS: Properly implement the macos_hide_titlebar option

Luflosi:
Don't merge this yet, I haven't written the documentation yet, I just wanted to get some feedback first.

This commit un-deprecates the `macos_hide_titlebar` config option and implements it how it was supposed to function in the first place.

Thanks to @ntruong for writing https://github.com/ntruong/kitty/commit/cb8279ec54cd81926e94ef0f94b4ebd05880258e and  https://github.com/ntruong/kitty/commit/68074113652f283b5aef4cefd4bb35bded07dd13, the code was very helpful.

Is it ok to un-deprecate this config option and change what it does?
Is it ok to call `cocoa_hide_window_title()` in `cocoa_hide_titlebar()` or does that cause any problems with the `autoreleasepool` blocks?

`hide_window_decorations yes`:
hide_window_decorations

`macos_hide_titlebar yes`:
macos_hide_titlebar


kovidgoyal:
Rather than undeprecate the option make hide_window_decorations take a titlebar-only keyword. It should be fine as far as I know autorelease pools can be arbitrarily nested.

Luflosi:
But this only works on macOS, which is why I think it is not a good idea to use `hide_window_decorations` for this.

kovidgoyal:
That option already has a disclaimer saying actual effects are OS/window
manager dependant. Simply add to that disclaimer saying titlebar-only
works only on macOS.

Luflosi:
Like this? Keep in mind that I still did not change the documentation yet.

kovidgoyal:
Looks fine.

Luflosi:
Please review.

PR #2290, opened 2020-01-16
Use python context managers to open files

Luflosi:
`with open(os.ctermid()) as sys.stdin:` should work as expected, right? I tested it and it seems to work fine.


kovidgoyal:
You dont want to leave sys.stdin in an invalid state. It breaks
assumptions all over since it is a global variable. If you realy want to
use a context manager (though I dont see the point, then you need to
re-assign sys.stdin in the __exit__ clause of the context manager.

Luflosi:
I'll just undo that change then.

PR #2293, opened 2020-01-17
Add `force_ltr`

### LINKED ISSUE #2252: Feature request: Disabling bidi support
TL;DR: Some applications (e.g. vim) have their own bidi support, which conflicts
with kitty's support. Please offer a way of turning off kitty's bidi support.

Kitty's bidi seems to cut the line into words, shape each word, then place them
in one line left-to-right, whereas vim's bidi just places letters one after the
other in the specified direction.
(Side note: I can't find kitty's behaviour here documented anywhere)

Hence, kitty's bidi works reasonably well for interactive shell use and the
like, where there is a prevailing directionality and the reversed text is
relatively rare. Conversely, vim's bidi works well in longform text where
changes in direction occur in multi-word streaks.

(Hypothetically, if a good bidi algorithm were implemented (e.g.
[UAX9](https://www.unicode.org/reports/tr9/)), it would deprecate vim's bidi in
_some_ scenarios, but not all – in cases where extra markup exists, such
algorithms would probably get confused by the markup and give wrong results)

However, kitty+vim is broken in this regard – kitty lies to vim about what's
under the cursor, editing longform text written in this zigzag layout is tiring,
and I don't even have the usual vim workaround of opening the file and its
reflection, since then the words are also reflected.

ctrlcctrlv:
Close #2252.

This allows users to forcibly disable RTL; since Kitty doesn't support BIDI anyway, this makes selecting RTL substrings more tolerable, but many users may prefer the current behavior, so it's off by default. 

@kovidgoyal called this a "bandaid", and I understand why, but I hope he merges it because I find it useful, and I know that @hseg will too

2020-01-17-161718_2398x1185_scrot

2020-01-17-162116_2883x1732_scrot

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -234,7 +234,11 @@ appears to be ם actually writes into the selection buffer the character י.
 
 Kitty's default behavior is useful in conjunction with a filter to reverse the
 word order, however, if you wish to manipulate RTL glyphs, it can be very
-challenging to work with, so this option is provided to turn it off."""))
+challenging to work with, so this option is provided to turn it off.
+Furthermore, this option can be used with the command line program
+:link:`GNU FriBidi <https://github.com/fribidi/fribidi#executable>` to get BIDI
+support, because it will force HarfBuzz to always treat the text as LTR, which
+FriBidi expects for terminals."""))
 
 
 def adjust_line_height(x):
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -223,6 +223,19 @@ def to_font_size(x):
 
 o('font_size', 11.0, long_text=_('Font size (in pts)'), option_type=to_font_size)
 
+o('force_ltr', False, long_text=_("""
+Kitty does not yet support BIDI (bidirectional text), however, for RTL scripts,
+HarfBuzz will "helpfully" shape text as if we were laying it out RTL. That is
+to say, in an RTL script, the words "HELLO WORLD" display in Kitty as "WORLD
+HELLO", and if you try to select a substring of an RTL-shaped string, you will
+get the character that _would_ be there were the string LTR. For example,
+assuming the Hebrew word ירושלים, selecting the character that on the screen
+appears to be ם actually writes into the selection buffer the character י.
+
+Kitty's default behavior is useful in conjunction with a filter to reverse the
+word order, however, if you wish to manipulate RTL glyphs, it can be very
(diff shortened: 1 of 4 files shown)


hseg:
Thanks! I agreed it's a bandaid, but at least it makes sane bidi editing with vim possible.
Sometimes, it pays for a program not to be too smart...

On 17 January 2020 10:24:43 GMT+02:00, Fredrick Brennan <notifications@github.com> wrote:
>Close #2252.
>
>This allows users to forcibly disable RTL; since Kitty doesn&#39;t
>support BIDI anyway, this makes selecting RTL substrings more
>tolerable, but many users may prefer the current behavior, so it&#39;s
>off by default. 
>
>@kovidgoyal called this a &quot;bandaid&quot;, and I understand why,
>but I hope he merges it because I find it useful, and I know that @hseg
>will too
>
>2020-01-17-161718_2398x1185_scrot
>
>2020-01-17-162116_2883x1732_scrot
>
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/kovidgoyal/kitty/pull/2293
>
>-- Commit Summary --
>
>  * Add `force_ltr`
>
>-- File Changes --
>
>    M kitty/config_data.py (19)
>    M kitty/fonts.c (1)
>    M kitty/state.c (1)
>    M kitty/state.h (1)
>
>-- Patch Links --
>
>https://github.com/kovidgoyal/kitty/pull/2293.patch
>https://github.com/kovidgoyal/kitty/pull/2293.diff
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/kovidgoyal/kitty/pull/2293

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ctrlcctrlv:
I agree. I come into contact with Hebrew text surprisingly frequently, as I make fonts and some of my build scripts and tests include Hebrew text. So I need to be able to select Hebrew substrings, which is basically impossible with the current way things work.

I hope this will get merged even though Kovid closed your issue. It can always be taken out if full BIDI support is added, and it's optional

kovidgoyal:
I said I dont want to spend the time on it, I have no objections to
someone else spending the time. Will review the PR when I have a moment.

ctrlcctrlv:
The push I just made was a documentation-only change. I realized that this commit is much more useful than I originally thought&mdash;it allows GNU FriBidi to work with Kitty! 

2020-01-17-175228_3147x1732_scrot

PR #2302, opened 2020-01-20
Remove useless code from shader

Luflosi:
My very limited knowledge of OpenGL led me to believe, that this line is unnecessary. Is that correct?
What is this shader used for?

### DIFF
--- a/kitty/blit_fragment.glsl
+++ b/kitty/blit_fragment.glsl
@@ -7,5 +7,4 @@ out vec4 color;
 
 void main() {
     color = texture(image, texcoord);
-    color = vec4(color.rgb, color.a);
 }


ctrlcctrlv:
This code is not useless. Did you compile? This ought to be closed.

Luflosi:
I did in fact compile kitty without error, the CI also compiled it just fine. On my system kitty even launches fine but since I don't know what this shader is used for, I don't know how to make sure that everything still works. Can you please explain what breaks? As I said, I don't know much about OpenGL.

ctrlcctrlv:
Sorry, you are right. Actually though, these problems are only detected at run time. I meant compile the shader, which happens yes at run time.

kovidgoyal:
the shader is used for drawing on semi-transparent windows with images,
and as far as I can see you are right in that the line in unnecessary.
Probably left over from when the shader was more complex.

PR #2326, opened 2020-01-30
Add `background_image`

### LINKED ISSUE #163: background image

ctrlcctrlv:
This PR closes #163, and builds off my prior work in #2311 and 8020657d8c0a42ce79057b1d9a5db6801c4e4698, and also work by @kovidgoyal to correct my mistakes, which made this possible.

This PR has been tested both on Nvidia's proprietary OpenGL driver, and Mesa, both on Arch Linux. Sorry, I have **no ability** to test on OS X, (which is also an impediment to my work on FontForge, but my friends there test it for me,) but think it should work.

@kovidgoyal asked if I enjoyed the journey. I really did! Never did anything with OpenGL before this. It perhaps folly to take this on as my first OpenGL project, but hopefully users enjoy the result of much trial and error.

# Scaled examples

2020-01-30-171704_2434x1856_scrot
2020-01-30-171915_2434x1856_scrot
2020-01-30-172148_2434x1856_scrot

# Tiling examples

2020-01-30-162821_2434x1640_scrot
2020-01-30-140514_2420x1640_scrot
2020-01-30-135746_2420x1640_scrot

### DIFF
--- a/kitty/bgimage_fragment.glsl
+++ b/kitty/bgimage_fragment.glsl
@@ -1,0 +1,26 @@
+#version GLSL_VERSION
+#define LAYOUT_TYPE
+
+uniform sampler2D image;
+uniform float bgimage_opacity;
+#ifdef TILED
+uniform float bgimage_scale;
+
+// These are of the window, not the screen.
+uniform float width;
+uniform float height;
+#endif
+in vec2 texcoord;
+out vec4 color;
+
+void main() {
+#ifdef TILED
+    vec2 txsz = vec2(width,height) / textureSize(image,0);
+    txsz /= bgimage_scale;
+    color = texture(image, texcoord * txsz);
+#endif
+#ifdef SIMPLE
+    color = texture(image, texcoord);
+#endif
+    color = vec4(color.rgb, color.a * bgimage_opacity);
+}
--- a/kitty/bgimage_vertex.glsl
+++ b/kitty/bgimage_vertex.glsl
@@ -1,0 +1,9 @@
+#version GLSL_VERSION
+
+layout(location=0) in vec4 src;
+out vec2 texcoord;
+
+void main() {
+    texcoord = clamp(vec2(src[0], src[1]*-1), 0, 1);
+    gl_Position = src;
+}
--- a/kitty/borders.py
+++ b/kitty/borders.py
@@ -60,11 +60,13 @@ class Borders:
         active_window,
         current_layout,
         extra_blank_rects,
-        draw_window_borders=True
+        draw_window_borders=True,
+        draw_blank_rects=False
     ):
         add_borders_rect(self.os_window_id, self.tab_id, 0, 0, 0, 0, BorderColor.default_bg)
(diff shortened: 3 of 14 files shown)


ctrlcctrlv:
Sorry @kovidgoyal, I don't really understand how the Python test framework is supposed to work, it seems that by changing the available options I broke something, perhaps I failed to properly declare it in Python

kovidgoyal:
Dont worry about it I will fix it when I merge.

kovidgoyal:
FYI, I'll be reviewing (and probably merging) this over the next few days.

ctrlcctrlv:
Thank you :-) I will in that case not make any more commits to the branch until your first review, to give you a stable base to review.

kovidgoyal:
I have merged, with various fixes, please test to see that I have not broken anything. And if you have questions about my changes feel free to ask.

kovidgoyal:
@Luflosi I'd appreciate if you could test that background iamges work on macOS since I dont currently have access to my mac.

Luflosi:
Will do during the weekend.

Luflosi:
Sorry for the delay but I finally tested it and it works mostly as expected:
screenshot
Making this work with more image formats would be nice. What would be needed to make this work with JPEG images?
I noticed that part of the tab bar (in the not active tab) is transparent. I guess this is not intended and should probably be fixed. The right of the tab bar, where there are no tabs is currently filled with the background color. It may be desirable to show the background image there.
I also tested transparency and it works as expected.
When I tested with `background_tint 0.95`, I noticed one pixel wide strips on the right and bottom where the `background_tint` is not applied. I know that the tint is not applied to the margin/borders but I expected to have no margins since my window size is defined with cells:
`initial_window_width 80c` and `initial_window_height 28c`.
screenshot
Here the same screenshot again with `hide_window_decorations yes`:
screenshot

There must be an off-by-one error somewhere, either in the window size calculation or in the calculation for the tint. How can I help to debug this?

ctrlcctrlv:
It might also be due to the blank rectangles Kitty draws for borders, in anticipation of having multiple visible Kitty windows in an OS window. 

In re JPEG, it's not that much work to support, but would require a new dependency. That might not be a problem, libjpeg is extremely common, but I figured it better to keep the code simpler by choosing one image format which implements everything supported by Kitty (alpha).

kovidgoyal:
On Sat, Feb 15, 2020 at 05:15:57PM -0800, Fredrick Brennan wrote:
> It might also be due to the blank rectangles Kitty draws for borders, in anticipation of having multiple visible Kitty windows in an OS window. 

It almost certainly is the borders. @Luflosi you can check by comenting
out various bits of code in borders.py

> 
> In re JPEG, it's not that much work to support, but would require a new dependency. That might not be a problem, libjpeg is extremely common, but I figured it better to keep the code simpler by choosing one image format which implements everything supported by Kitty (alpha).
> 

I'd really rather not start depending on multiple libraries for multiple
image formats. It's not that much of a hardship to ask users to convert
images to PNG before using them as backgrounds.

Issue #2337, opened 2020-02-05
Change the thickness of the beam cursor


noomly:
I wasn't able to find a parameter that allow the configuration of the thickness of the beam cursor. Is there a way to make it thinner? If not would it be easy to implement?

kovidgoyal:
There is no such configuration, you are welcome to send a PR to add one,
I personally have no interest in doing so. Relevant code is the
render_cursor() function in fonts/render.py

PR #2340, opened 2020-02-06
Make cursor beam thickness configurable

noomly:
Following #2337 I am making this PR to make the cursor beam thickness configurable. This is my first contribution to Kitty and I had some difficulties understanding how accessing the configuration datas is done so I hope what I did is idiomatic. If it isn't please point me in the right direction and I'll fix it!


kitty/config_data.py
```diff
@@ -394,6 +394,8 @@ def cursor_text_color(x):
 background color of the cell underneath instead, use the special keyword: background'''))
 o('cursor_shape', 'block', option_type=to_cursor_shape, long_text=_(
     'The cursor shape can be one of (block, beam, underline)'))
+o('cursor_beam_thickness', 1.5, option_type=float, long_text=_(
```
ctrlcctrlv: It should be `positive_float`.

PR #2370, opened 2020-02-17
X11: Fix setting the clipboard string to itself

Luflosi:
From upstream: https://github.com/glfw/glfw/commit/0c27ed1d0ef36672f0ef28d86bf8e108cc6e379d.


ctrlcctrlv:
Just wondering, why don't we use git submodules so we always have the latest glfw? Do we apply patches on top?

kovidgoyal:
the glfw in kitty has diverged from upstream. Two of the biggest
divergences are keyboard handling and event loop handling. And then
there is runtime sleection of x11/wayland backends.

ctrlcctrlv:
Thanks for answering, sorry to bother you :)

PR #2371, opened 2020-02-17
Ensure valid UTF-8 when cutting string

Luflosi:
This is my attempt at fixing the issue mentioned in the comment.
Where would be a good place to put `utf_8_strndup()`?
Can `title` be NULL? There was a check for it but I don't think `_glfw_strdup()` could handle this situation.
While developing this function, I wrote a little test suite for it and it would be nice to have it in the kitty tests. Maybe not to run it every time by the CI as that seems a little wasteful but as a sort of "extended" test suite to be run manually. There are a few other tests for similar functions that could be written in the future.

### DIFF
--- a/glfw/backend_utils.c
+++ b/glfw/backend_utils.c
@@ -332,3 +332,21 @@ pollForEvents(EventLoopData *eld, monotonic_t timeout, watch_callback_func displ
     }
     return read_ok;
 }
+
+// Duplicate a UTF-8 encoded string
+// but cut it so that it has at most max_length bytes plus the null byte.
+// This does not take combining characters into account.
+GLFWAPI char* utf_8_strndup(const char* source, size_t max_length) {
+    if (!source) return NULL;
+    size_t length = strnlen(source, max_length);
+    if (length >= max_length) {
+        for (length = max_length; length > 0; length--) {
+            if ((source[length] & 0xC0) != 0x80) break;
+        }
+    }
+
+    char* result = malloc(length + 1);
+    memcpy(result, source, length);
+    result[length] = 0;
+    return result;
+}
--- a/glfw/backend_utils.h
+++ b/glfw/backend_utils.h
@@ -95,3 +95,4 @@ unsigned dispatchTimers(EventLoopData *eld);
 void finalizePollData(EventLoopData *eld);
 bool initPollData(EventLoopData *eld, int display_fd);
 void wakeupEventLoop(EventLoopData *eld);
+char* utf_8_strndup(const char* source, size_t max_length);
--- a/glfw/wl_window.c
+++ b/glfw/wl_window.c
@@ -957,12 +957,10 @@ void _glfwPlatformSetWindowTitle(_GLFWwindow* window, const char* title)
 {
     if (window->wl.title)
         free(window->wl.title);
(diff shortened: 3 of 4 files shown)


kovidgoyal:
Looks ok to me. Sadly kitty's testing infrastructure is not setup to test glfw functions. Not sure if title can be null in general but kitty does not pass null titles to glfw. Still I think it is good to have a null check in there.

If you wish to add extra tests you are welcome to do so.

Luflosi:
`glfw_tests.c` seems to test parts of GLFW. Would this be a good place to add the tests for the new function?

kovidgoyal:
Not really that was used during development of the new glfw main loop.
The problem with testing glfw is one has to use the null backend, which
is currently not built, or have a running X/Wayland server. If you want
to test utility functions, such as ut8_strdup the best way to do it
would be to put them in a separate file like monotonic.c and include
them in glfw. Then you can test the actual implementation in kitty like
any other kitty function. 

Or alternately, export the function in glfw and add some boilerplate in
kitty to dlopen it and test it without needing to initialize glfw. See
detect_if_wayland_ok for n example.

Luflosi:
Should I move `utf_8_strndup()` into its own file or into an existing file like `backend_utils.c`?

kovidgoyal:
backend_utils is fine with me.

Luflosi:
I created a second function `static PyObject* py_utf_8_strndup(PyObject *self UNUSED, PyObject *args)` in `backend_utils.c` that wraps `utf_8_strndup()`. How can I now import and call that function from python?

Luflosi:
I added a `module_methods` array to the end of `backend_utils.c` and an `init_backend_utils()` function. Now I think I just need to call that function from `init_glfw()`. Does that work and how can I call that function?

kovidgoyal:
You cant have python wrapped functions in glfw. It doesnt link against
python. To call functions in the glfw module you would need to load it
with ctypes, see for example detect_if_wayland_ok()

Luflosi:
Which `.so` do I need to load? The symbol isn't in `glfw-cocoa.so`. Is it in a different `.so` or do I need to export the symbol somehow?

kovidgoyal:
You have to mark the symbol for export, look at how the symbol used in
the example function I posted earlier is marked.

Luflosi:
`glfwWaylandCheckForServerSideDecorations()` has `GLFWAPI` at the beginning of the function definition. I already tried doing the same with `utf_8_strndup()` but the symbol can still not be found. I pushed the latest code, which I expected to work when compiled on macOS. Could you please take a look?

kovidgoyal:
backend utils is not used on cocoa, so you wil need to load glfw-wayland or glfw-x11 for the testing, and the testing wont run on macOS.

Luflosi:
Ok, thanks.

Luflosi:
Is there a good way to get the platform kitty was built with in the tests, so that I can exclude macOS but run it on X11 and/or Wayland? And is there a way I can indicate that a test was skipped instead of passed without actually doing anything?

kovidgoyal:
Use sys.platform to determine the current OS. And use unittest.skipIf to
skip tests.

Luflosi:
Please review.

PR #2382, opened 2020-02-20
Fix boolean value assigned to pointer

Luflosi:
Found with Cppcheck.

Do you know what could have happed in practice?


kovidgoyal:
It's harmless since that pointer is not used in the function after that
point (beyond the logic error of course). I'm actually fairly surprised
compilers have not caught this.

PR #2455, opened 2020-03-21
Add support for png2icns as an alternative to iconutil

Luflosi:
png2icns is used when building kitty with nix because iconutil seems to be closed-source. They currently patch the build system to use png2icns: https://github.com/NixOS/nixpkgs/blob/569d6d0fb11b0d3a26e565c61f3dbef15a5c1348/pkgs/applications/misc/kitty/png2icns.patch. It would be nice if kitty would support using png2icns without patching.
libicns also has an iconutil clone called icnsutil in the source tree but the last release is from 2012, which does not include this utility yet.
I'm printing a warning message because png2icns does not provide the same functionality as iconutil (no retina support). Is this a good way of printing the warning message or is there a better way to do this?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -861,7 +861,7 @@ def create_macos_app_icon(where: str = 'Resources') -> None:
     iconset_dir = os.path.abspath(os.path.join('logo', appname + '.iconset'))
     subprocess.check_call([
         'iconutil', '-c', 'icns', iconset_dir, '-o',
-        os.path.join(where, os.path.basename(iconset_dir).partition('.')[0] + '.icns')
+        os.path.join(where, appname + '.icns')
     ])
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Why does it care if iconutil is closed source. It comes with macOS. By that logic everything on macOS is closed source, and it should just refuse to run there.

Luflosi:
Since iconutil is closed source, it cannot be included as a nix package and would need to be called from the system at build time. This causes problems because there could be different versions of iconutil on different systems that behave differently, which would make the build non-deterministic. Such so-called impure derivations are usually undesirable. With png2icns, the exact version of the utility is known and identified with a hash, making the build deterministic.

PR #2476, opened 2020-03-24
fix(window): inactive window border fix

### LINKED ISSUE #2474: `inactive_border_color` setting not respected in 0.17.1
Hey !

The `inactive_border_color` no longer works in the latest release. (witnessed on macOS Catalina)

0.17.1:
Screenshot 2020-03-24 at 19 31 10

0.16.0:
Screenshot 2020-03-24 at 19 35 37

I'll try to bisect later today when I find time to narrow down to the commit which introduced the regression.

Cheers.

pyrho:
I'm really not sure of what I'm doing here so please double check :)
But it seems that in commit f7eef3f456f7ca38044355b88cd6ffe45cd91061 the `if` statement got "inverted" and in doing so a typo cause the condition to fail.

For reference this is the change (from f7eef3f456f7ca38044355b88cd6ffe45cd91061):
```diff
-            if w is active_window and not draw_active_borders and not w.needs_attention:
-                yield no_borders
-            else:
+            if (w is active_window and draw_active_borders) or w.needs_attention:
                 yield all_borders
+            else:
+                yield no_borders
```

The `if` initially being `A && !B && !C` it should have been translated to `!A || B || C` when inverted. That's what this fix does.

Fixes #2474.

### DIFF
--- a/kitty/layout.py
+++ b/kitty/layout.py
@@ -549,7 +549,7 @@ class Layout:  # {{{
 
     def minimal_borders(self, windows: WindowList, active_window: Optional[WindowType], needs_borders_map: Dict[int, bool]) -> Generator[Borders, None, None]:
         for w in windows:
-            if (w is active_window and draw_active_borders) or w.needs_attention:
+            if w is not active_window or draw_active_borders or w.needs_attention:
                 yield all_borders
             else:
                 yield no_borders


Luflosi:
I think
```python
if w is not active_window or draw_active_borders or w.needs_attention:
```
will fix the CI warning.

pyrho:
And it looks nicer too :) Will try.

Issue #1196, opened 2018-12-02
Unable to compile kitty on macOS


Luflosi:
The compilation fails with
```
In file included from /path/to/kitty/kitty/core_text.m:14:
In file included from /usr/local/Cellar/harfbuzz/2.2.0/include/harfbuzz/hb-ot.h:33:
In file included from /usr/local/Cellar/harfbuzz/2.2.0/include/harfbuzz/hb-ot-color.h:37:
/usr/local/Cellar/harfbuzz/2.2.0/include/harfbuzz/hb-ot-name.h:79:22: error: 
      redefinition of typedef 'hb_ot_name_id_t' is a C11 feature
      [-Werror,-Wtypedef-redefinition]
typedef unsigned int hb_ot_name_id_t;
                     ^
/usr/local/Cellar/harfbuzz/2.2.0/include/harfbuzz/hb-deprecated.h:244:22: note: 
      previous definition is here
typedef unsigned int hb_ot_name_id_t; /* Since is in hb-ot.h */
                     ^
1 error generated.
make: *** [app] Error 1
```
This might be an issue with harfbuzz.

kovidgoyal:
that is an issue with harfbuzz, not kitty. Stick to older harfbuzz releases until they fix it.

Luflosi:
I case anyone else has this problem,
```Bash
brew switch harfbuzz 2.1.3
```
worked for me.

Luflosi:
@kovidgoyal should I file a bug report for harfbuzz or do they know about this?

kovidgoyal:
@Luflosi no idea, I was planning to do it myself, but am totally swamped at the moment, so I just disabled the kitty travis build with brew for the moment.

Luflosi:
I filed the bug report.

jacwah:
Another solution is to compile in c11 mode: `CFLAGS="-std=c11" make`.

Luflosi:
It's fixed in harfbuzz 2.3.0, which is now available via homebrew.
You may (or may not, I don't really care) revert your commit if you like.

kovidgoyal:
I'm ok with keeping it, unless someone is harmed by it.

PR #2477, opened 2020-03-25
Use c99 instead of c11

ryandesign:
Kitty builds fine with c99. The switch to c11 was due to a bug in Harfbuzz 2 that was fixed in Harfbuzz 2.3.0. See #1196. This reverts f098e6c973c5b514edb6047546bdd90c0705332b.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -256,7 +256,7 @@ def init_env(
         cppflags.append('-DDEBUG_{}'.format(el.upper().replace('-', '_')))
     cflags_ = os.environ.get(
         'OVERRIDE_CFLAGS', (
-            '-Wextra {} -Wno-missing-field-initializers -Wall -Wstrict-prototypes -std=c11'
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
There's no real reason to downgrade. We might want to use some C11 feature some day. Even MSVC can compile C11 now

kovidgoyal:
Is there some reason to use C99?

ryandesign:
It's up to you of course, but in my opinion there's no real reason *not* to downgrade. Since kitty builds fine without c11, it seems illogical to me for kitty's build system to specify that c11 is required. If at some point in the future you want to start using newer language features, then you would adjust the `-std` flag at that time.

kovidgoyal:
It was changed to fix a bug, it breaks nothing. At some point in the
future it might become needed. Changing it back is just churn.

PR #2480, opened 2020-03-25
Fix tab_bar_background not being respected

### LINKED ISSUE #2464: Bottom tab border issue with 0.17.0
After updating today, when I create a new tab, I see a gray bar at the bottom where additional tabs would be shown.

It used to be black before and I would like to keep that look, instead of a visible horizontal bottom bar.

I looked at the changelog and I'm unsure if `:opt: background_image` needs to be set here now to show a blank/black bottom bar?

Screenshot from 2020-03-24 14-16-00

Relevant tab config:
```
bell_on_tab             yes
tab_bar_edge            bottom
tab_title_template      {index}: {title}
focus_follows_mouse     yes
active_tab_foreground   #000
active_tab_background   #73D216
active_tab_font_style   bold-italic
inactive_tab_foreground #444
inactive_tab_background #999
inactive_tab_font_style normal
```

dmitmel:
Hi! After update to v0.17.0 my tab bar changed from:

image

to:

image

By looking through the changelog I found that it was commit 0d87b8fac5168171493cb0d8cef85d60e564c0b0 which broke my configuration. Looks like while that commit did fix #2464, it broke `powerline` and `separator` tab bar styles. Fortunately, the fix was simple enough and now `tab_bar_background` is treated as specified in the documentation.

P. S. I fixed my specific configuration by copying the color from `inactive_tab_background` to `tab_bar_background`. Since my changes to tab bar rendering were insignificant (in other words: part of my config for tab bar isn't much different from the default configuration) I think that this fix is worthwhile including in the default configuration.

### DIFF
--- a/kitty/tab_bar.py
+++ b/kitty/tab_bar.py
@@ -179,12 +179,8 @@ class TabBar:
         self.data_buffer_size = 0
         self.laid_out_once = False
         self.dirty = True
-        self.screen = s = Screen(None, 1, 10, 0, self.cell_width, cell_height)
-        s.color_profile.update_ansi_color_table(build_ansi_color_table(opts))
-        s.color_profile.set_configured_colors(
-            color_as_int(opts.inactive_tab_foreground),
-            color_as_int(opts.background)
-        )
+        self.screen = Screen(None, 1, 10, 0, self.cell_width, cell_height)
+        self.screen.color_profile.update_ansi_color_table(build_ansi_color_table(opts))
         self.blank_rects: Tuple[Rect, ...] = ()
         sep = opts.tab_separator
         self.trailing_spaces = self.leading_spaces = 0
@@ -208,6 +204,10 @@ class TabBar:
             self.opts.tab_bar_background or self.opts.background, self.opts.tab_title_template,
             self.opts.active_tab_title_template
         )
+        self.screen.color_profile.set_configured_colors(
+            color_as_int(self.draw_data.inactive_fg),
+            color_as_int(self.draw_data.default_bg)
+        )
         if self.opts.tab_bar_style == 'separator':
             self.draw_func = draw_tab_with_separator
         elif self.opts.tab_bar_style == 'powerline':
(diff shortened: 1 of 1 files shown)


dmitmel:
Your commit doesn't fix the problem. Now the tab bar looks like (with and without `tab_bar_background` because now it makes no difference whatsoever):

image

dmitmel:
A few more screenshots:

- `tab_bar_style = powerline` + `tab_bar_background #ff0000`
  image
- `tab_bar_style = fade` + `tab_bar_background #ff0000`
  image

Though the powerline screenshot shows undoubtedly incorrect behavior, you might consider the fade one somewhat correct.

Is there anything wrong with my fix?

dmitmel:
I also have to mention that I tested my changes on various `tab_bar_style` and `tab_bar_background` combinations, I can send screenshots of actual/expected behavior if necessary.

kovidgoyal:
Your patch broke changing the colors, via patch_colors. See my latest commit

dmitmel:
I can revert changes to `patch_colors` - they aren't necessary. The point is - your commit doesn't fix the problems (see screenshots).

kovidgoyal:
No, see https://github.com/kovidgoyal/kitty/commit/d9d419991cf9645d7bf326774943ee5842ee1b29

dmitmel:
This one fixes powerline and fade, but separator doesn't look quite right yet:

image

I'll try to fix it as well.

dmitmel:
Here you go:

```patch
diff --git a/kitty/tab_bar.py b/kitty/tab_bar.py
index be2866ba..c5eb6068 100644
--- a/kitty/tab_bar.py
+++ b/kitty/tab_bar.py
@@ -82,6 +82,7 @@ def draw_tab_with_separator(draw_data: DrawData, screen: Screen, tab: TabBarData
     end = screen.cursor.x
     screen.cursor.bold = screen.cursor.italic = False
     screen.cursor.fg = 0
+    screen.cursor.bg = as_rgb(color_as_int(draw_data.inactive_bg))
     screen.draw(draw_data.sep)
     screen.cursor.bg = 0
     return end
```

kovidgoyal:
That was deliberate, I prefer the separator for the active tab to use the active background. Having one cell of inactive bg after the last tab if the last tab is active looks weird.

dmitmel:
image

Kooks weird when some tab in the middle is selected.

dmitmel:
Nevermind. I switched to powerline mode and I hope get used to it soon.

dmitmel:
Having thought about it, as a compromise I can either remove the last separator, similarly to powerline:

image

kovidgoyal:
No last separator looks bad when the tab bar bbackground is set to something other than inactive background. Like I said, I am OK with unising inactive background for all separators except the last one.

PR #2498, opened 2020-03-30
Read strikethrough info from font when using FreeType

trygveaa:
This reads the strikethrough position and thickness from the font so it
is rendered correctly.

This is only implemented for FreeType, not Core Text, because I didn't
find any way to get the info from Core Text, and I don't have a Mac to
test it on either. When using Core Text or when the font doesn't provide
the info, the same approximation as before is used.

### DIFF
--- a/kitty/core_text.m
+++ b/kitty/core_text.m
@@ -312,7 +312,7 @@ harfbuzz_font_for_face(PyObject* s) {
 }
 
 void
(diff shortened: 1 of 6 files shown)


trygveaa:
The test failures seems to be caused by timeouts against azure.archive.ubuntu.com. The [build on the commit](https://github.com/trygveaa/kitty/actions/runs/66741875) ran fine.

PR #2527, opened 2020-04-10
Ignore keys for switching keyboard layout

### LINKED ISSUE #2519: Strange reaction on "menu" button
I have layout switcher on "menu" button `grp:menu_toggle`

When I start vim in kitty and press  "menu" it makes something wierd.

It changes the capitalization of the next few words.

trygveaa:
XKB has various options for using keys to switch the keyboard layout.
When these are set, XKB will return keysyms which are unknown to GLFW,
so kitty will fallback to using a keymap without the options set, which
causes the keys to be interpreted as the original keysyms.

However, when these options are set, kitty shouldn't interpret the keys.
Therefore, check for some specific keysyms and return before translating
to GLFW keysyms.

There may be more keysyms which should be ignored, but as far as I can
see from https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/xkeyboard-config-2.29/symbols/group
these are the ones which can be triggered by XKB options.

Fixes #2519

### DIFF
--- a/glfw/xkb_glfw.c
+++ b/glfw/xkb_glfw.c
@@ -620,6 +620,9 @@ glfw_xkb_handle_key_event(_GLFWwindow *window, _GLFWXKBData *xkb, xkb_keycode_t 
         }
         if (key_text[0]) { debug("%s: %s ", text_type, key_text); }
     }
+    if (xkb_sym == XKB_KEY_ISO_First_Group || xkb_sym == XKB_KEY_ISO_Last_Group || xkb_sym == XKB_KEY_ISO_Next_Group || xkb_sym == XKB_KEY_ISO_Prev_Group || xkb_sym == XKB_KEY_Mode_switch) {
+      return;
+    }
     int glfw_sym = glfw_key_for_sym(xkb_sym);
     bool is_fallback = false;
     if (glfw_sym == GLFW_KEY_UNKNOWN && !key_text[0]) {


imyxh:
Hi, sorry to bother you.

It seems like this commit does not work for me, or else I've done something wrong myself. This is the result of me pressing super+space, my key to change input method on swaywm:

```
$ cat | xxd
^[[32;9u
```

This causes issues e.g. in REPL programs when I am trying to switch to another character, but it types "9u" instead. I cannot reproduce this when using gnome-terminal.

Here's the output from --debug-keyboard:
```
Press xkb_keycode: 0x7d clean_sym: Super_L composed_sym: Super_L mods: none glfw_key: 57444 (LEFT_SUPER) xkb_key: 65515 (Super_L)
on_key_input: glfw key: 0xe064 native_code: 0xffeb action: PRESS mods: none text: '' state: 0 ignoring as keyboard mode does not support encoding this event
Press xkb_keycode: 0x39 clean_sym: space composed_sym: ISO_Next_Group mods: super glfw_key: 32 ( ) xkb_key: 32 (space)
on_key_input: glfw key: 0x20 native_code: 0x20 action: PRESS mods: super text: '' state: 0 sent key to child
[120 20:35:16.792436] [glfw error 65549]: Wayland: Window attention request not implemented yet
Release xkb_keycode: 0x39 clean_sym: space mods: super glfw_key: 32 ( ) xkb_key: 32 (space)
on_key_input: glfw key: 0x20 native_code: 0x20 action: RELEASE mods: super text: '' state: 0 ignoring as keyboard mode does not support encoding this event
Release xkb_keycode: 0x7d clean_sym: Super_L mods: super glfw_key: 57444 (LEFT_SUPER) xkb_key: 65515 (Super_L)
on_key_input: glfw key: 0xe064 native_code: 0xffeb action: RELEASE mods: super text: '' state: 0 ignoring as keyboard mode does not support encoding this event
```

I suspected this commit should fix my issue because the composed_sym is showing up as ISO_Next_Group, but no luck.

kovidgoyal:
This should take care of it: https://github.com/kovidgoyal/kitty/commit/6b9b47849265dee370f7881cf974d6797a177600

Issue #2531, opened 2020-04-11
window zoom refinements


franco:
I'm using the [`zoom_toggle.py`](https://sw.kovidgoyal.net/kitty/kittens/custom.html#using-kittens-to-script-kitty-without-any-terminal-ui) kitten suggested in #870 as a workaround for zooming a window. It works fine for me with two small issues:

### Indication for zoomed window
Sometimes, with multiple tabs and zoomed windows, I'm losing track if I'm looking at a zoomed window or a tab with only one window. A visual indicator to distinguish between the two cases would be nice. I was thinking of drawing a border around a zoomed window. Is it possible to extend [`zoom_toggle.py`](https://sw.kovidgoyal.net/kitty/kittens/custom.html#using-kittens-to-script-kitty-without-any-terminal-ui) to draw a border? Or would that need changes to kitty itself?

### Maximize displayed scrollback buffer in zoomed window
Let's imagine a small window with already scrolled up output gets zoomed. I would expect the zoomed window to show as much of the scrollback buffer as possible (same as in tmux). Currently it does not show any additional output form the scrollback history. Any idea how to handle that with a kitten?

kovidgoyal:
Neither of those are addressable in kittens, I'm afraid. For your first request the
easiest way would be to allow tab title templates to accept
{num_of_windows} and {num_of_visible_windows}

For, the second the way scrollback is handled when resizing is my
personal preference, if you want to change it, you will need to add an
option to kitty to control it.

kovidgoyal:
Or for (1) you could add a must draw border option to the stack layout and have your kitten toggle it.

PR #2533, opened 2020-04-11
Add flag to Layout to enforce drawing a border

franco:
Layout.must_draw_borders = True enforces drawing of a border even
though only a single window is visible. See #2531.

Here how it could be used in the `zoom_toggle.py` kitten in order to indicate that window is zoomed:
```python3
def main(args):
    pass

from kitty.layout import Stack
from kittens.tui.handler import result_handler
@result_handler(no_ui=True)
def handle_result(args, answer, target_window_id, boss):
    tab = boss.active_tab
    if tab is not None:
        if tab.current_layout.name == 'stack':
            tab.last_used_layout()
        else:
            try:
                Stack.must_draw_borders = True
                tab.goto_layout('stack')
            finally:
                Stack.must_draw_borders = False
```

### DIFF
--- a/kitty/layout.py
+++ b/kitty/layout.py
@@ -208,6 +208,7 @@ class Layout:  # {{{
 
     name: Optional[str] = None
     needs_window_borders = True
+    must_draw_borders = False
     needs_all_windows = False
     layout_opts = LayoutOpts({})
     only_active_window_visible = False
--- a/kitty/tabs.py
+++ b/kitty/tabs.py
@@ -231,7 +231,7 @@ class Tab:  # {{{
                 windows=visible_windows, active_window=w,
                 current_layout=self.current_layout, extra_blank_rects=tm.blank_rects,
                 padding_width=self.padding_width, border_width=self.border_width,
-                draw_window_borders=self.current_layout.needs_window_borders and len(visible_windows) > 1
+                draw_window_borders=self.current_layout.needs_window_borders and len(visible_windows) > 1 or self.current_layout.must_draw_borders
             )
             if w is not None:
                 w.change_titlebar_color()


kovidgoyal:
You'd use it as tab.current_layout.must_draw_borders = True

i.e. change the instance variable not the class variable.

And if you reset it immediately the next time the layout is
recalculated, for instance if the window is resized, the border will
disappear.

thisisrandy:
Has something changed since this was written? Both of the changes in 75c8ffd appear to still be in the codebase, but using a `zoom_toggle.py` like the following (imports + `main` elided) doesn't draw any border around windows in the `stack` layout:

```python
@result_handler(no_ui=True)
def handle_result(args, answer, target_window_id, boss):
    tab = boss.active_tab
    if tab is not None:
        if tab.current_layout.name == "stack":
            tab.last_used_layout()
        else:
            try:
                tab.current_layout.must_draw_borders = True
                tab.goto_layout("stack")
            finally:
                tab.current_layout.must_draw_borders = False
```

Note that I'm using `draw_minimal_borders no` in my `kitty.conf`, which seems kind of essential for the feature to work (How else would it decide which edge to draw a border on with just a single window displayed? Maybe there is some logic in place, but I haven't dug in enough to find it).

thisisrandy:
To anyone wanting to use this, here is a correct version of the border zoom kitten as of v0.36.4:

```python
from kittens.tui.handler import result_handler


def main(args):
    pass


@result_handler(no_ui=True)
def handle_result(args, answer, target_window_id, boss):
    tab = boss.active_tab
    if tab is not None:
        if tab.current_layout.name == "stack":
            tab.last_used_layout()
            tab.current_layout.must_draw_borders = False
        else:
            try:
                tab.goto_layout("stack")
                tab.current_layout.must_draw_borders = True
            except Exception:
                pass
        tab.relayout()
```

PR #2567, opened 2020-04-18
Add strikethrough capability to terminfo

wladh:
Add strikethrough capability to terminfo so programs (such as vim/neovim) can find it.


kovidgoyal:
Is there some reference to these codes somewhere? All I can find is that
they are tmux extensions. Do any applications actually use them?

trygveaa:
They are mentioned here at least: https://invisible-island.net/xterm/terminfo-contents.html#tic-ecma_strikeout

```
# The rmxx/smxx capabilities are an ncurses extension
ecma+strikeout|ECMA-48 strikeout/crossed-out,
        rmxx=\E[29m, smxx=\E[9m,
```

wladh:
In addition to what @trygveaa said, the terminfo capabilities are used by neovim (that's how I ended up making this change), and according to vim's doc, it's using the termcap attributes I added there (but I haven't tested it).

PR #2590, opened 2020-04-25
Add --background option to run kitty on the Wayland desktop

quantum5:
This PR adds the `--background` argument to kitty. When specified, the kitty window runs behind all other windows, effectively making it the desktop background. This uses the [`zwlr_layer_shell_v1` Wayland protocol](https://github.com/swaywm/wlroots/blob/master/protocol/wlr-layer-shell-unstable-v1.xml).

This is useful for using CLI applications as the desktop wallpaper. For example, `kitty --background htop` can be used to show `htop` as the desktop background.

It can be used to create interesting effects, like running [`cava`](https://github.com/karlstav/cava) on the desktop to visualize music in the background:

cava

### DIFF
--- a/glfw/glfw.py
+++ b/glfw/glfw.py
@@ -31,6 +31,7 @@ class Env:
     wayland_scanner: str = ''
     wayland_scanner_code: str = ''
     wayland_protocols: Tuple[str, ...] = ()
+    custom_wayland_protocols: Tuple[str, ...] = ()
 
     def __init__(
         self, cc: str = '', cppflags: List[str] = [], cflags: List[str] = [], ldflags: List[str] = [],
@@ -99,7 +100,8 @@ def init_env(env: Env, pkg_config: Callable, pkg_version: Callable, at_least_ver
         scanner_version = tuple(map(int, pkg_config('wayland-scanner', '--modversion')[0].strip().split('.')))
         ans.wayland_scanner_code = 'private-code' if scanner_version >= (1, 14, 91) else 'code'
         ans.wayland_protocols = tuple(sinfo[module]['protocols'])
-        for p in ans.wayland_protocols:
+        ans.custom_wayland_protocols = tuple(sinfo[module]['custom_protocols'])
+        for p in ans.wayland_protocols + ans.custom_wayland_protocols:
             ans.sources.append(wayland_protocol_file_name(p))
             ans.all_headers.append(wayland_protocol_file_name(p, 'h'))
         for dep in 'wayland-client wayland-cursor xkbcommon dbus-1'.split():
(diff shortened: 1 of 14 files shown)


kovidgoyal:
It's a cool idea, I really like it, however, I dont want to add unstable Wayland protocols to kitty. If and when this protocol makes it to wayland-protocols so that kitty does not need to carry around definitions of the protocol in its source code, I will be happy to merge.

Dark-Kernel:
Please finish this thing, we are waiting.

kovidgoyal:
This is now implemented in the layer branch. Only tested with sway so far. Run

```
kitty +kitten panel --edge=background htop
```

to have kitty render the background. Remember to disable any bg in your sway config otherwise it hides the panel background.

jpmhouston:
Can panel be brought to macOS? I'd like htop or btop running on my desktop.

kovidgoyal:
No, macOS doesnt support this kind of thing as far as I know.

jpmhouston:
https://stackoverflow.com/a/4982619/592739

kovidgoyal:
On Sat, Apr 13, 2024 at 03:29:57PM -0700, Pierre Houston wrote:
> https://stackoverflow.com/a/4982619/592739

That's not the same thing, for example the window can still grab
keyboard focus. It will participate in alt-tab window switching, etc.

jpmhouston:
Are you sure you aren’t confusing platforms? I ask because you mention alt-tab window switching whereas of course macOS uses command-tab app switching.

I spoke with the author of the Mac app Particulars, which, in one mode, puts a transparent chromeless window of information on the desktop like it seems this feature currently does on Linux.

FWIW, in its configuration option I use, this app does remove itself from command-tab switching, although I’m not saying kitty needs to do that as, again, this OS feature switches between apps rather than windows. A running app being frontmost with no active aka key window is a-ok.

Here’s the information he sent me about how his app does this using above-board system APIs, over and above the tips in the stack overflow answer I linked to above:

>You’ve correctly identified that the window’s level must be set programmatically. Before we can set the window’s level we need a window. Particulars creates the desktop window with NSWindow’s initWithContentRect:styleMask:backing:defer: with the styleMask set to [NSWindowStyleMaskBorderless](https://developer.apple.com/documentation/appkit/nswindowstylemask/nswindowstylemaskborderless?language=objc). As stated in the linked documentation, a window with this style mask can’t be become key (have keyboard focus) or be the main window.
>
>Other considerations:
>
>* text fields or views in these windows should have selecting and editing disabled
>* the app should respond to the NSApplicationDidChangeScreenParametersNotification posted when displays change and update the window size accordingly

In any case, looking more into this kitten and how background is implemented (with no sizing or positioning on the desktop supported), I think I’ll instead search for another open source terminal app to contribute to and try to get this feature in. Thanks anyway,
sorry for hijacking this issue instead of making a new one.

kovidgoyal:
This is now implemented in master for macOS as well, and you can size and position the background window anywhere using the --margin-* flags to the kitten.

PR #2595, opened 2020-04-26
Make foreground opacity configurable

quantum5:
This PR adds the configuration option `foreground_opacity`. It works exactly like the current option `background_opacity` except it applies to the foreground.


kovidgoyal:
Whats the use case for this? Doesn't making text dim make it harder to read?

quantum5:
My personal use case for this is with #2590 to show transparent console block art in the background, but I am sure there are other uses for this.

kovidgoyal:
On Sun, Apr 26, 2020 at 09:38:41PM -0700, Guanzhong Chen wrote:
> My personal use case for this is with #2590 to show transparent console block art in the background, but I am sure there are other uses for this.

You can use the SGR dim attribute to do this already and control the
dimming via dim_opacity.

PR #2597, opened 2020-04-27
Add note clarifying that icat won't work with tmux

jinnko:
When using icat with tmux we get the following:

```
$ kitty icat ~/Pictures/fc.jpg 
Terminal does not support reporting screen sizes via the TIOCGWINSZ ioctl
```

I found a suggestion[1] this is due to limitations between kitty and tmux, so I'm raising this PR to either confirm this is the case and make it clear in the documentation, or to have it rejected so an issue can be raised to address the feature.

1: https://unix.stackexchange.com/a/484776/41996

### DIFF
--- a/docs/kittens/icat.rst
+++ b/docs/kittens/icat.rst
@@ -21,6 +21,11 @@ Then you can simply use ``icat image.png`` to view images.
     `ImageMagick <https://www.imagemagick.org>`_ must be installed for ``icat`` to
     work.
 
+.. note::
+
+    ``icat`` uses terminal features that are incompatible with screen multiplexers
+    such as `tmux` and `screen`.
+
 
 .. program:: kitty +kitten icat
 


kovidgoyal:
Terminal multiplexers are an extremely bad idea, they waste resources,
they act as drag on the entire terminal ecosystem and they break random
things. I'm afraid I am not interested in listing every terminal
multiplexer under the sun that breaks something in kitty.

jinnko:
I appreciate all the effort that has gone into kitty, and that it may be unrealistic to add support for terminal multiplexers.  This is why I raised this PR with the option of simply accepting the fact it doesn't work and won't be supported.

I lost a good 20-30 mins discovering the limitation.  I would have really appreciated some indication in the docs that it shouldn't be expected to work.

If you'd like this addressed in the docs differently I'd be happy to raise the PR, with your guidance.

kovidgoyal:
A simple note saying that icat might not work when used within a
terminal multiplxer such as tmux or screen is fine. It depends on
whether the individual multiplexer adds support for the image escape
codes or not.

kovidgoyal:
So something like this:

Note: kitty's image display protocol may not work when used withing a terminal multiplxer such as screen or tmux, depending on whether the multiplxer has added support for it or not

PR #2598, opened 2020-04-27
Add note clarifying icat won't work with tmux [take 2]

jinnko:
When using icat with tmux we get the following:

```
$ kitty icat ~/Pictures/fc.jpg 
Terminal does not support reporting screen sizes via the TIOCGWINSZ ioctl
```

In #2597 this was confirmed to generally be a limitation of terminal multiplexers and adding support in kitty would add too much maintenance overhead.  We document this for users.

### DIFF
--- a/docs/kittens/icat.rst
+++ b/docs/kittens/icat.rst
@@ -23,8 +23,7 @@ Then you can simply use ``icat image.png`` to view images.
 
 .. note::
 
-    ``icat`` uses terminal features that are incompatible with screen multiplexers
-    such as `tmux` and `screen`.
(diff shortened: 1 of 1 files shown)


docs/kittens/icat.rst
```diff
@@ -21,6 +21,10 @@ Then you can simply use ``icat image.png`` to view images.
     `ImageMagick <https://www.imagemagick.org>`_ must be installed for ``icat`` to
     work.
 
+.. note::
+
+    kitty's image display protocol may not work when used within a terminal multiplxer such as ``screen`` or ``tmux``, depending on whether the multiplxer has added support for it or not
```
trygveaa, line 26: Typo in "multiplxer".

PR #2613, opened 2020-04-30
Make paths to libraries configurable at build time

Luflosi:
Don't merge this yet.
I'm trying to make the paths to `libEGL.so.1`, `libstartup-notification-1.so` and `libcanberra.so` configurable during build time. This is useful for nix since all paths to libraries are hardcoded in the binary for reproducibility and look like `/nix/store/rcli75fxa9x8raklf19kp0hfcyjc0kfn-libGL-1.2.0/lib/libEGL.so.1`. See https://github.com/NixOS/nixpkgs/blob/b16a5e2847b4e78a29be9fb2d157706f4ca84083/pkgs/applications/misc/kitty/default.nix and the parent directory for how this is done currently.
To change the paths, I want to define preprocessor variables containing a string with the path. I'd like to use the `SPECIAL_SOURCES` dict. Can I move the definition of `SPECIAL_SOURCES` into the function where it is used? I could then pass in the paths as arguments to the function. Do you see a better alternative?

### DIFF
--- a/glfw/glfw.py
+++ b/glfw/glfw.py
@@ -6,7 +6,7 @@ import json
 import os
 import re
 import sys
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 _plat = sys.platform.lower()
 is_linux = 'linux' in _plat
@@ -19,6 +19,7 @@ class Env:
     cppflags: List[str] = []
     cflags: List[str] = []
     ldflags: List[str] = []
+    library_paths: Dict[str, List[str]] = {}
     ldpaths: List[str] = []
     ccver: Tuple[int, int]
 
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Have the init_env() functions store the library_paths on the env object and let SPECIAL_SOURCES use a callback to define the defines, th callback can read the env object to set the defines.

Luflosi:
Please review.
I made the function for `SPECIAL_SOURCES` return an `Optional[List[str]]` instead of a `List[str]` so I can optionally omit a `#define`.
Are the texts in `option_parser()` good or can you think of a better description?

kovidgoyal:
Why are the defines with a leading underscore?

Also the library paths dont have to be actual paths, they can be names
as well, so the help text for the options should reflect that.

Luflosi:
The defines have a leading underscore because I just followed the example from
https://github.com/kovidgoyal/kitty/blob/7bf83603c8ad6a8837d8d03f1d6063d7733e25a2/glfw/egl_context.c#L308-L309
Is "The filename argument passed to dlopen for libEGL. This can be used to change the name of the loaded library or specify an absolute path." better?
I also renamed `libstartup_notification_library` to `startup_notification_library`.

Issue #2616, opened 2020-05-01
Tab Separator Highlight Inconsistency 


rigelrozanski:
Very minor, but I just noticed that the separator _is_ highlighted for the final tab but _is not_ highlighted during in between tabs (see images) 
![Screen Shot 2020-05-01 at 2 19 39 AM](https://user-images.githubusercontent.com/20132176/80787084-7930a900-8b53-11ea-8d44-7a39d79515c7.png)
![Screen Shot 2020-05-01 at 2 19 08 AM](https://user-images.githubusercontent.com/20132176/80787085-79c93f80-8b53-11ea-85d5-11bc4c278bc8.png)

. My preference would be to see the separator consistently highlighted. (just for reference the I have `tab_bar_edge top` set and `tab_separator " ⎤ "` in my conf)

kovidgoyal:
This is deliberate, its the way I like it. I fond having the trailing
separator highlighted in internal tabs looks jarring.

trygveaa:
What about only drawing the separators between the tabs? In my opinion, that looks nicer:
![image](https://user-images.githubusercontent.com/601966/80791406-d77c7c80-8b91-11ea-864c-70f977712ecb.png)

Using a space on both sides in `tab_separator` looks even nicer imo, since it's symmetrical:
![image](https://user-images.githubusercontent.com/601966/80791379-bd429e80-8b91-11ea-98cf-fb4c320a8abf.png)

kovidgoyal:
I'm fine with omitting the final separator, should be an easy PR for
you. Relevant function is draw_tab_with_separator

PR #2620, opened 2020-05-01
Use one space on each side in tab_separator

trygveaa:
This makes the separator in the tab bar symmetrical with equal spacing on both sides.

Do you agree that it looks better with equal spacing, or do you prefer keeping the default as is?

With this change:
image

Relates to #2616


kovidgoyal:
I prefer the default as is, dual spaces look worse to me.

PR #2643, opened 2020-05-11
Graphics protocol: note best detection method

ctrlcctrlv:
Per https://github.com/kovidgoyal/kitty/issues/2636#issuecomment-626467852


kovidgoyal:
I changed it slightly to use DA1 as the fallback query to send.

PR #2644, opened 2020-05-11
Set a default socket timeout when retrieving remote items in icat.

lamby:
Otherwise we can easily hang when "icat"-ing resources from unreliable websites/webcams, etc.


Luflosi:
I think `def socket_timeout(seconds: int) -> None:` instead of `def socket_timeout(seconds):` should fix the CI error.

lamby:
> I think def socket_timeout(seconds: int) -> None: instead of def socket_timeout(seconds): should fix the CI error

Ahh, thanks. Force-pushed for clarity. :+1:

PR #2652, opened 2020-05-13
Use f-string instead of format for tab_title templates

chriszarate:
The documentation for v0.17.4 states this about the `tab_title_template`:

> Note that formatting is done by Python’s string formatting machinery, so you can use, for instance, {layout_name[:2].upper()} to show only the first two letters of the layout name, upper-cased.

However, because the templating uses String `.format`, this generates an error:

```
Invalid tab title template: "{layout_name[:2].upper()}" with error: string indices must be integers
```

If instead we used an f-string, we could fulfill the promise of the docs.

I'm fairly new to Python, so not sure if there is a more direct / safer way to pass a variable to an f-string.

### DIFF
--- a/kitty/tab_bar.py
+++ b/kitty/tab_bar.py
@@ -54,14 +54,18 @@ def draw_title(draw_data: DrawData, screen: Screen, tab: TabBarData, index: int)
         screen.cursor.fg = draw_data.bell_fg
         screen.draw('🔔 ')
         screen.cursor.fg = fg
-    layout_name = tab.layout_name
-    num_windows = tab.num_windows
     template = draw_data.title_template
-    title = tab.title
     if tab.is_active and draw_data.active_title_template is not None:
         template = draw_data.active_title_template
     try:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Hmm, I forgot that str.format() doesn't allow arbitrary expressions. You can make this a bit safer using compile() in 'single' mode and eval that and also use a dict for globals wen calling eval that has only the required variables.

chriszarate:
Ah ok, thanks! I will update the PR later today.

chriszarate:
Ok, updated with your suggestion. However, I found that I had to `compile` in `'eval'` mode. When I used `'single'` the string was simply printed and the return value was `None`. Unless I'm missing something (which is likely!), this seems to be what the docs say as well:

> The mode argument specifies what kind of code must be compiled; it can be 'exec' if source consists of a sequence of statements, 'eval' if it consists of a single expression, or 'single' if it consists of a single interactive statement (in the latter case, expression statements that evaluate to something other than None will be printed).

PR #2663, opened 2020-05-18
Fixed marks grouping and coloring for simultaneous multi marks definition

ahlinc:
This PR fixes an issue with multiple simultaneous mark groups defined. The issue exists on a latest Kitty version and on top of the master branch. If I try to define several marks simultaneously in any available approach I got a text colored just by one color and that color just corresponds to a first mark group in the making definition.

**Steps to reproduce:**
1. `echo -e 'WARN\nERROR\nFAIL'`
2. `kitty @ create-marker regex 1 FAIL 2 ERR  3 WARN`

All three words should have different colors but all of them just colored as the `FAIL` word by the color of the first group in the applied marking definition.

**Bellow screenshots demonstrate the issue:**
Screenshot from 2020-05-18 06-52-20
Screenshot from 2020-05-18 06-51-21

**With the PR fix:**
Screenshot from 2020-05-18 07-08-22

**P.S: And thanks for the great project!** :rocket:

### DIFF
--- a/kitty/config.py
+++ b/kitty/config.py
@@ -311,7 +311,7 @@ def parse_marker_spec(ftype: str, parts: Sequence[str]) -> Tuple[str, Union[str,
 def toggle_marker(func: str, rest: str) -> FuncArgsType:
     parts = rest.split(maxsplit=1)
     if len(parts) != 2:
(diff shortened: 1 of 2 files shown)


ahlinc:
Super fast merge :rocket: :smiley_cat: Thanks!

PR #2674, opened 2020-05-20
Create .editorconfig

TheDaemoness:
[Most major editors](https://editorconfig.org/#download) support EditorConfig for declaring per-file per-project indentation rules. It costs almost nothing to maintain, and it makes it easier to follow the coding styles already used in kitty.


Luflosi:
Here are two more files indented with tabs that I think you missed: `docs/Makefile` and `kittens/unicode_input/names.h`.

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_style = spaces
+indent_size = 4
+end_of_line = lf
+trim_trailing_whitespace = true
+
+[{Makefile,*.terminfo}]
+indent_style = tab
+
+# Autogenerated with tabs.
+[kitty/{unicode-data.c,emoji.wcwidth-std.h}]
```
Luflosi, line 13: You missed the .h and the comma at the end of emoji.

TheDaemoness:
The Makefile selector should match any file named "Makefile" in the project. A very informal test on my end suggests it works correctly.

I also missed all of the Wayland protocol files in glfw, however. Whoops.

Luflosi:
> I also missed all of the Wayland protocol files in glfw, however. Whoops.

Well, I missed those too because they are only generated during compilation of the Wayland backend and I'm using macOS.

Issue #2647, opened 2020-05-11
[Wayland+HiDPI] Mouse cursor half as large as it should be


keynslug:
Hi! Thanks for the quality software.

Under GNOME Wayland session with 2x display scale factor I see that kitty renders much smaller (2x smaller I believe) mouse cursor, hovewer I'm sure that cursor _geometry_ itself is correctly scaled. This manifests in those troubles I have when I'm trying to select some word and missing constantly by a line, selecting a word line below instead.

Screenshots for comparison with GTK-based terminal:

![Screenshot from 2020-05-11 20-57-29](https://user-images.githubusercontent.com/1148366/81604461-93327d00-93d8-11ea-9d20-1330f3b36366.png)

![Screenshot from 2020-05-11 20-57-46](https://user-images.githubusercontent.com/1148366/81604466-9594d700-93d8-11ea-8ab8-e6c3c0fd3370.png)

```
 ~ gsettings get org.gnome.desktop.interface scaling-factor                                                                                                               
uint32 2
 ~ gsettings get org.gnome.desktop.interface cursor-size                                                                                                                  
24
 ~ gsettings get org.gnome.desktop.interface cursor-theme                                                                                                                 
'Adwaita'
```

```
 ~ kitty --debug-gl --debug-config                                                                                                                                        
kitty 0.17.4 created by Kovid Goyal
Linux kenfawks 5.6.8-arch1-1 #1 SMP PREEMPT Wed, 29 Apr 2020 16:22:56 +0000 x86_64
Arch Linux \r (\l)
LSB_VERSION=1.4
DISTRIB_ID=Arch
DISTRIB_RELEASE=rolling
DISTRIB_DESCRIPTION="Arch Linux"
Loaded config files: /home/keynslug/.config/kitty/kitty.conf
Running under: Wayland

Config options different from defaults:
background              Color(red=12, green=17, blue=21)
background_opacity      0.9
color0                  Color(red=46, green=52, blue=60)
color1                  Color(red=189, green=15, blue=47)
color10                 Color(red=73, green=233, blue=152)
color11                 Color(red=253, green=223, blue=110)
color12                 Color(red=42, green=139, blue=193)
color13                 Color(red=234, green=71, blue=39)
color14                 Color(red=160, green=182, blue=211)
color2                  Color(red=53, green=167, blue=112)
color3                  Color(red=251, green=148, blue=53)
color4                  Color(red=31, green=88, blue=114)
[...]

PR #2709, opened 2020-05-30
Implement HiDPI cursors on Wayland

TheDaemoness:
Free time is a fickle thing, isn't it?

These commits implement properly-scaled cursors on Wayland compositors with HiDPI support, addressing #2647/#2669. It implements a reference-counting manager for cursor themes (which can probably be extended to manage `wl_cursor` instances later), and changes some of the Wayland-specific platform code to use the new manager to acquire themes with the required pixel sizes.

I tried to be as surgical as possible in my implementation, which lead to a couple of questionable design choices that I'd prefer to be upfront about:

* The semantics of the `cursor` variable in `_GLFWcursorWayland` have changed. `NULL` no longer means that `buffer` and its associated variables should be used instead. Instead, the `scale` variable stores the relevant information (see the attached comment). I've tried to find the spots where this would be an issue, but won't claim it's impossible that I've overlooked something.
* Actually loading a default cursor is delayed in most cases until `setCursorImage` gets called. Loading can fail, and `setCursorImage` handles this quietly, which in the worst case may result in a null pointer dereference. **If you folks have a preference on what way this failure should be handled instead, please let me know.** As a tangent, setting `XCURSOR_THEME` to an invalid value doesn't produce this issue for me (on sway).
* Managed cursor themes are pointed to by each window instance, not each monitor instance. This is probably the smallest problem of the bunch, especially given the possibilities for per-window custom cursors that this opens up.

### DIFF
--- a/glfw/source-info.json
+++ b/glfw/source-info.json
@@ -78,6 +78,7 @@
       "wl_init.c",
       "wl_monitor.c",
       "wl_window.c",
+      "wl_cursors.c",
       "posix_thread.c",
       "xkb_glfw.c",
       "dbus_glfw.c",
--- a/glfw/wl_cursors.c
+++ b/glfw/wl_cursors.c
@@ -1,0 +1,176 @@
+// Future devs supporting whatever Wayland protocol stabilizes for cursor selection: see _themeAdd.
+
+#include "wl_cursors.h"
+
+#include "internal.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+typedef struct {
+    struct wl_cursor_theme *theme;
+    int                     px;
+    int                     refcount;
+} _themeData;
+
+struct _wlCursorThemeManager {
+    size_t count;
+    /** Pointer to the head of an unsorted array of themes with no sentinel.
+     *
+     * The lack of sort (and thus forcing a linear search) is intentional;
+     * in most cases, users are likely to have 1-2 different cursor sizes loaded.
+     * For those cases, we get no benefit from sorting and added constant overheads.
+     *
+     * Don't change this to a flexible array member becuase that complicates growing/shrinking.
+     */
+    _themeData *themes;
+};
+
+static void
+_themeInit(_themeData *dest, const char *name, int px) {
+    dest->px = px;
+    dest->refcount = 1;
+    if(_glfw.wl.shm) {
+        dest->theme = wl_cursor_theme_load(name, px, _glfw.wl.shm);
+        if(!dest->theme) {
+            _glfwInputError(GLFW_PLATFORM_ERROR,
+                        "Wayland: Unable to load cursor theme");
+        }
+    } else {
+        dest->theme = NULL;
+    }
+}
+
+static struct wl_cursor_theme*
+_themeAdd(int px, _wlCursorThemeManager *manager) {
+   ++manager->count;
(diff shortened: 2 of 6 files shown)


kovidgoyal:
I'll review when I have some of the fickle free time :) But right off the bat, what happens currently if loading a theme fails? To me the ideal behavior would be to fallback to the "hidpi" theme which IIRC is supposed to be always present, and if that also fails either to use a deault image or to abort with an error message. The last ight be problematic given that theme loading no longer happens at init time. Might be best to just store a default cursor image as a C header and use that as final fallback.

TheDaemoness:
So in regards to actual theme loading, that's handled by `wl_cursor_theme_load`, which already has fallbacks for most situations that aren't allocation failures.

Sadly, the fallback for xcursor theme loading doesn't appear to be HiDPI. Moreover, there doesn't seem to be a good way to test if that theme is in use:
* `wl_cursor_theme` pointers are opaque.
* `wl_cursor_theme_load` doesn't set errno or have any way of reporting an error other than returning NULL.
* I can't find a function to extract the theme size from a `wl_cursor_theme` struct, much less confirm if it's the default fallback theme or not.

All well.

kovidgoyal:
I have merged, but I have a couple of questions.

1) When scale is changed, you load a new cursor theme, but you dont invalidate the actual cursor. So how does the rendered cursor change?

2) in _glfwSetPlatforCursor when no cursor is specified, the default arrow cursor is set, however this is done with a stack based variable. If the cursor happens to be animated this will cause the timer to fire calling incrementCursorImage which in turn will do nothing but reset the timer to call itself again and again, without end.

kovidgoyal:
Not to mention that the result of glfwloadcursor leaks in (2) above

kovidgoyal:
Also in setCursorImage cursorWayland never has its scale set to be window->wl.scale

kovidgoyal:
> Not to mention that the result of glfwloadcursor leaks in (2) above

On further reading, it seems the result of wl_cursor_theme_get_cursor() does not need to be destroyed, so no leak there.

TheDaemoness:
Closing the PR as the commits have been merged outside of it, but will still watch for discussion. Thank you!

As you've noted, `wl_cursor_theme_get_cursor` is effectively an accessor for data in `wl_cursor_theme`.

With respect to scale changes, there is a brief period where the cursor pointers might dangle after the scale gets changed. However, `setCursorImage` is called soon after a scale change, if not immediately. The cursor image does change when I change the scale of the display, I've not been able to confirm an invalid read that happens as a result, but it still might be a good idea to call `setCursorImage` in the same functions where `checkWindowScale` returns true anyway.

With respect to the behavior of `_glfwPlatformSetCursor`, the use of a stack-allocated _GLFWcursorWayland [was existing behavior](https://github.com/kovidgoyal/kitty/pull/2709/files#diff-1af3a71b69b12b2d2f3dd87eef9ac582L1459). It could probably be changed, though it's worth examining first how much code outside of `wl_window.c` relies on that particular behavior.

> Also in setCursorImage cursorWayland never has its scale set to be window->wl.scale
I know I had that line at one point in the `if (newCursor != NULL)` statement, but must have deleted it by accident. Sorry!

kovidgoyal:
I have committed code to fix both the stack cursor animation issue and to change the cursor image immediately on scale change. Please test.

TheDaemoness:
The changed code appears to work! Cursors animate properly (including the arrow cursor in grabbed contexts), and scale changes work as before. ASAN builds don't report any new leaks related to the cursor code either.

Issue #2662, opened 2020-05-17
multiple `symbol_map` entries do not override as expected


loveencounterflow:
When I configure in my `kitty.conf` the following:

```
# Note: inline comments only for reference, not in `kitty.conf`
font_family               IosevkaNerdFontCompleteM-Thin
symbol_map U+61-U+7a      Iosevka-Slab-Heavy  # (1) /[a-z]/
symbol_map U+51-U+5a      Iosevka-Slab-Heavy  # (2) /[Q-Z]/
symbol_map U+61           LastResort          # (3) /[a]/
symbol_map U+65           LastResort          # (4) /[e]/
symbol_map U+69           LastResort          # (5) /[i]/
symbol_map U+6F           LastResort          # (6) /[o]/
symbol_map U+75           LastResort          # (7) /[u]/
symbol_map U+51-U+53      LastResort          # (8) /[QRS]/
symbol_map U+59-U+5a      LastResort          # (9) /[YZ]/
```

I'd expect the outcome to be as follows:

* **(1)** Use `IosevkaNerdFontCompleteM-Thin` as default,
* **(2)** except for the lower case `[a-z]` range, for which use `Iosevka-Slab-Heavy`,
* **(3)** except for the upper case `[Q-Z]` range, for which likewise use `Iosevka-Slab-Heavy`; of these,
* **(4-7)** use LastResort for `/[aeiou]/`, overriding **(1)**,
* **(8)** use LastResort for `/[QRS]/` overriding **(2)**,
* **(9)** use LastResort for `/[YZ]/`, likewise overriding **(2)**.

What happens in reality is that Kitty applies the only rules **(1)**, **(2)**, **(3)**, **(8)**, but
discards **(4)**, **(5)**, **(6)**, **(7)** and **(9)**; it so happens that both the ranges / single
codepoints addressed in the shadowing rules that *were* honored (**(3)** and **(8)**), their first
codepoints (`a` = `U+61` and `Q` = `U+51`) coincided with the first codepoints of the rules they shadowed.
Observe that while e.g. codepoint `e` = `U+65` of rule **(4)** is indeed the first (and only) codepoint of
that rule, that is not what matters; what matters is where that shadowing codepoint falls within the rule it
is intended to supplant, namely rule **(1)**. Since there it is the 5th, not the 1st codepoint of the range
given, it does—erroneuosly, one should think—not succeed.

[...]

PR #2728, opened 2020-06-04
Disjunct codepoint ranges

loveencounterflow:
This pull request if accepted will change the behavior of `kitty/fonts/render.py#coalesce_symbol_maps()` such that codepoint ranges configured with the `symbol_map` directive are made free of overlaps (conflicting codepoints). 

Assuming that the input argument `maps: Dict[ Tuple[ int, int, ], str, ]` preserves ordering of its entries as given in the configuration file (which it should as of Python3.6), disjunct codepoint ranges are computed such that entries in the configuration file that come *later* are given *higher* precedence than those that come *earlier*; this is in keeping with current practices in the field (cf. JavaScript's `Object.assign()`; CSS declarations) and desirable because defaults defined by the system and the application come (at least notionally) earlier than the choices defined by the user.

To illustrate, when a user has configured

```
symbol_map      U+3000-U+30ff                  Sun-ExtA
symbol_map      U+300a-U+300b                  NanumMyeongjo
```

this will result in and be indistinguishable from a configuration given as 

```
symbol_map      U+3000-U+3009                  Sun-ExtA
symbol_map      U+300a-U+300b                  NanumMyeongjo
symbol_map      U+300c-U+30ff                  Sun-ExtA
```

where all ranges are disjunct (and are, hence, independent of ordering, which is a nice property). This will work with arbitrary numbers of overlapping segments and is seen as beneficial since it allows users to configure their choices in brad strokes at fist, then refining them as seen fit, without any need to manually compute and remove any overlapping parts.

There are tests for the range arithmetics in https://github.com/loveencounterflow/hengist/blob/master/dev/kitty-font-config-writer-kfcw/python/tests.py; if deemed necessary, I could look into incorporating them into the Kitty testing lineup. Other than that, I have visually tested various configurations and am reasonably sure everything works as expected.

### DIFF
--- a/kitty/fonts/render.py
+++ b/kitty/fonts/render.py
@@ -96,6 +96,14 @@ def merge_ranges(a: Range, b: Range, priority_map: Dict[Tuple[int, int], int]) -
 
 
 def coalesce_symbol_maps(maps: Dict[Tuple[int, int], str]) -> Dict[Tuple[int, int], str]:
+    ### NOTE https://docs.python.org/3/library/stdtypes.html?highlight=dict#mapping-types-dict:
+    # Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an
+    # implementation detail of CPython from 3.6.
+    print( '^22333198^', f"maps             {maps           }" )
+    print( '^22333198^', f"type( maps )     {type( maps )   }" )
+    import sys
+    print( '^22333198^', f"sys.version   {sys.version }" )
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Umm this was already implemented in https://github.com/kovidgoyal/kitty/commit/362e4266bbd7764588f7229d41b25ee8b3b30402  two weeks ago, as noted at the end of #2662

loveencounterflow:
I saw the incoming code but it kept giving me errors. I opened an issue about this: https://github.com/kovidgoyal/kitty/issues/2735

PR #2732, opened 2020-06-04
Fix fake scroll when scrolling in unfocused kitty window

Luflosi:
I'm working on a little bug fix. Unfortunately I can't figure out how to get it to compile because I need to pass a pointer to a `Window` to `fake_scroll()` but I can't do that since `Window` is not defined in `kitty/data-types.h`. I can't import `state.h` in `data-types.h` because `state.h` already imports `data-types.h`. I can't put `fake_scroll()` into `kitty/mouse.c` because `send_key_to_child()` is not defined there. Can you please help me find a way to make this work? Maybe making `send_key_to_child()` available in `kitty/mouse.c`?

### DIFF
--- a/kitty/data-types.h
+++ b/kitty/data-types.h
@@ -303,7 +303,6 @@ void enter_event(void);
 void mouse_event(int, int, int);
 void focus_in_event(void);
 void scroll_event(double, double, int, int);
-void fake_scroll(int, bool);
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Why not move fake_scroll() declaration to state.h ?

Luflosi:
That works, thank you!
Please review.

Issue #2022, opened 2019-10-04
Vertically mis-aligned text


roginfarrer:
Version: 0.14.6
OS: MacOS 10.14.4

Using Fira Code (v2), the text appears to be oddly aligned to the top instead of the center or baseline. This does not happen in iTerm. I was able to reproduce this on two different machines (but same configs).

I don't have anything special in my `kitty.conf`. Just setting the font-family to Fira Code.

Screenshots:

<img width="1414" alt="Screen Shot 2019-10-03 at 10 08 06 PM" src="https://user-images.githubusercontent.com/9063669/66176152-c040d500-e62a-11e9-858a-4bb38b3739ae.png">

<img width="198" alt="Screen Shot 2019-10-03 at 10 07 37 PM" src="https://user-images.githubusercontent.com/9063669/66176160-c8991000-e62a-11e9-9122-c52ea3b358ad.png">

Here is the vim statusbar in iTerm, where the text is aligned normally.

<img width="711" alt="Screen Shot 2019-10-03 at 10 12 18 PM" src="https://user-images.githubusercontent.com/9063669/66176197-e1a1c100-e62a-11e9-956e-c47ca3136b3e.png">

<img width="155" alt="Screen Shot 2019-10-03 at 10 12 47 PM" src="https://user-images.githubusercontent.com/9063669/66176226-f67e5480-e62a-11e9-860e-684b824d9bfd.png">

kovidgoyal:
Post the output of kitty --debug-config

roginfarrer:
Result of `kitty --debug-config`

```sh
kitty 0.14.6 created by Kovid Goyal
Darwin Rogins-MBP.T-mobile.com 18.5.0 Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64
ProductName:	Mac OS X ProductVersion:	10.14.4 BuildVersion:	18E226
Loaded config files: /Users/rfarrer/.config/kitty/kitty.conf

Config options different from defaults:
background            Color(red=0, green=38, blue=53)
color0                Color(red=0, green=56, blue=77)
color1                Color(red=196, green=48, blue=97)
color10               Color(red=156, green=240, blue=135)
color11               Color(red=255, green=204, blue=27)
color12               Color(red=126, green=178, blue=221)
color13               Color(red=251, green=148, blue=255)
color14               Color(red=0, green=255, blue=255)
[...]

PR #2795, opened 2020-06-25
add macOS glyph adjustment options

fenetikm:
I was still seeing issues with FiraCode on macOS so I added in a few more options to allow tweaking it.

Added options:
- macos_adjust_glyph_scale
- macos_adjust_glyph_y
- macos_adjust_glyph_x

For me, setting the scale to 0.1 and setting the glyph_y to -6.0 makes everything look right again (at a font size of 15.0).

I haven't added anything regarding documentation / changes - is that something to leave for you @kovidgoyal if you accept this? happy to do what needs to be done if you can let me know.

My specific issue was reported here: https://github.com/kovidgoyal/kitty/issues/2022

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -67,7 +67,7 @@ Detailed list of changes
 
 - A new option :opt:`undercurl_style` to control the rendering of undercurls (:pull:`5883`)
 
-- Bash integration: Fix ``clone-in-kitty`` not working on bash >= 5.2 if an environment variable values contain newlines or other special characters (:iss:`5629`)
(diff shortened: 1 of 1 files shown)


fenetikm:
actually, there is a problem with this and that is that now the underline isn't positioned correctly. _sigh_ need either an option for that or to adjust it relative to the above.

kovidgoyal:
y adjustment should happen to the underline position as well, and the strikethrough position. But really, the question is why is this necessary at all. And can it be automated. For example, by drawing a selection of glyphs from the font, checking if the bottom of glyphs like g or y is too high and adjusting.

fenetikm:
yeah, I don't like this manual adjustment either - I'll look a bit more into what Core Text can tell me.

Issue #1155, opened 2018-11-19
Reproducible builds


Luflosi:
Does kitty support reproducible builds? If so, I didn't find any mention of it on the website. If not, I think it should be supported. https://reproducible-builds.org might be helpful.

maximbaz:
[It's not reproducible right now](https://tests.reproducible-builds.org/archlinux/community/kitty/kitty-0.12.3-1-x86_64.pkg.tar.xz.html), I'll have a look if I can fix this.

maximbaz:
By the way, this issue should really be brought up with package maintainers, there's almost zero chance that anything will need to be changed in source code. Unless you are requesting provided binary releases to be reproducible, in which case I'd say it won't happen, that's something next to impossible to achieve because you'd have to precisely reproduce Kovid's environment and just doesn't worth the effort.

I'm looking into making Arch Linux package reproducible.

maximbaz:
Turns out no python app can be reproducible today: https://bugs.python.org/issue29708

Luflosi:
Yeah, I meant the binary releases in https://github.com/kovidgoyal/kitty/releases. But the packages of distributions would also be nice.
The python non-reproducibility is a bummer although the first comment in that issue says, that the `.py` files have to be touched for the issue to occur, so maybe it's not as bad as you might have thought. Would it be possible to not have the `.pyc` files in the binary release? Python would create them when kitty is run, right? But I guess there is some reason against doing that since many people don't seem to do that.

martinetd:
If the pyc non-reproductible part is only the timestamp that can be worked around, distros like debian touch all the files to the date of the changelog entry in the package when building so all timestamps should be using that date.

However, looking at @maximbaz's build diff, I think we're falling into that other problem (from his issue link):
```
For what it's worth, in Endless OS we still saw slight variations between builds
in the .pyc files, even with all the source files' mtimes set to the epoch (ie.
equivalent to setting & supporting SOURCE_DATE_EPOCH, I believe). Looking at the
contents of the file suggested it was just reordering of class fields; indeed, we
only saw this on Python versions where hash randomization is enabled by default,
and disabling hash randomization made the output reproducible.
```

Apparently, setting `PYTHONHASHSEED` might help; haven't tested.

maximbaz:
[...]

PR #2843, opened 2020-07-11
Use consistent quotes and dashes in html docs

maximbaz:
I played around with reproducible builds (#1155) again as it became a discussed topic again 🙂 

While doing so, I noticed that when we generate docs, it produces slightly different html files on different locales, in some cases html files would contain fancy Unicode dashes and typographical quotes, and in others not. 

I have a feeling that in practice nobody would care which quotes and dashes we use, so I'd just fix them to always be regular ASCII symbols, if that's okay with you?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -109,6 +109,9 @@ rst_prolog = '''
 
 '''.replace('VERSION', str_version)
 
+# Do not convert quotes and dashes to typographical symbols
+smartquotes = False
+
 
 # -- Options for HTML output -------------------------------------------------
 


kovidgoyal:
Why would one expect output of docs to be identical on different locales? Someday the docs might even be translated/localized. I rather enjoy having smartquotes in my text when reading, would be a shame to lose that.

maximbaz:
Okay, I didn't know you like them! 🙂 I'll try to find how to do the opposite, to force smartquotes to always be applied - please let me know if you can think of it from the top of your mind 🙂 

I think even if there would be multiple translations, a command that is meant to generate docs for all languages should produce the same folder regardless on what environment it was executed...?

kovidgoyal:
Depends on if you localize by building all languages or just a specified
language for a localized build, I guess.

As for making it always on, no idea, I dont really know why it should be
locale dependent at all, since it depends on the language of the
document, not the language of the system. However an easy workaround is
to just set a standard locale in the makefile before calling
sphinx-build.

See https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes

maximbaz:
I couldn't find anything that will force consistent typographical quotes (tried all ways I could think of to set locale to no avail), I think it is unwise to spend more time on this so let's just close this 🙂 

```diff
- $(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+ LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 $(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) -D language=en -D encoding=utf8 $(O)
```

Issue #2870, opened 2020-07-21
Support old alternate screen xterm code


Kody-Quintana:
When exiting a script that uses the Python TUI library [urwid](http://urwid.org/), content of the screen does not disappear until any newer content writes over it.
`[PARSE ERROR] Unsupported screen mode:  47 (private)` is emitted from kitty.
Any of the examples from the [here](https://github.com/urwid/urwid/tree/master/examples) will show this behavior. Would it be possible to get support for this screen mode? As a temporary work around I'm manually calling `clear` after exiting my urwid script.
edit: it looks like it could be as simple as adding the old code as an additional case, I submitted a PR

kovidgoyal:
This is a bug in urwid, it needs to use terminfo to get the correct
escape code, namely the smcup and rmcup capabilities. You should report
it to them.

PR #2871, opened 2020-07-21
Add old xterm code for alternate screen mode

Kody-Quintana:
This fixes [#2870](https://github.com/kovidgoyal/kitty/issues/2870#issue-663362790), according to [this manual](https://www.gnu.org/software/screen/manual/screen.html) screen mode 47 is the old xterm code for alternate screen. (1049 being the new one). This fixes the issue but I'm unsure why 47 was deprecated in the first place.

### DIFF
--- a/kitty/modes.h
+++ b/kitty/modes.h
@@ -69,6 +69,7 @@
 
 // Alternate screen buffer
 #define ALTERNATE_SCREEN  (1049 << 5)
+#define OLD_ALTERNATE_SCREEN  (47 << 5)
 
 // Bracketed paste mode
 // https://cirw.in/blog/bracketed-paste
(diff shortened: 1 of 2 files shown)


kovidgoyal:
You cannot make 47 a synonym for 1049 it is actually a synonym for 1047 which switches screens without saving the cursor position and without clearing the alternate screen. See https://invisible-island.net/xterm/ctlseqs/ctlseqs.html (search the page for 4 7)

Kody-Quintana:
Thanks for the link, I think I better understand how those codes work now. Do you know why there is a 1047 and a 47? Is there any reason to treat these two differently?

kovidgoyal:
On Tue, Jul 21, 2020 at 09:03:52PM -0700, Kody Quintana wrote:
> Thanks for the link, I think I better understand how those codes work now. Do you know why there is a 1047 and a 47? Is there any reason to treat these two differently?

No idea, really poor development practices historically, I'd guess. The
escape code landscape is full of these, because a lot of them pre-date
the internet, and modern software development practices
(specs/testing/etc.) and so people had to guess and re-invent and
reverse engineer things all the time.

PR #2875, opened 2020-07-23
GLFW: Make joystick platform code init on demand

Luflosi:
This PR removes our custom `GLFW_ENABLE_JOYSTICKS` GLFW init hint and replaces it with the new lazy joystick initialisation from upstream.

### DIFF
--- a/glfw/cocoa_init.m
+++ b/glfw/cocoa_init.m
@@ -669,8 +669,6 @@ int _glfwPlatformInit(void)
     if (!initializeTIS())
         return false;
 
-    _glfwInitJoysticksNS();
-
     _glfwPollMonitorsNS();
     return true;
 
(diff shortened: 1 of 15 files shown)


Luflosi:
Looks like I made a mistake, don't merge please until I fixed it.

Luflosi:
The problem is that `_glfwDetectJoystickConnectionLinux()` is still being called in the upstream code. Do you know what this function does and what performance impact it has? You disabled it in 7e36489034bed9a052403301345c8547939aff11.

kovidgoyal:
Look at the source of that function, it does a whole bunch of work including reading from inotify, opening devices, etc. There is absolutely no point in calling it if you dont want joysticks.

Luflosi:
Please review.

Issue #2805, opened 2020-06-26
Doesn't open on macOS Big Sur (Version: 11.0 Beta (20A4299v))


jdbrewer:
**Describe the bug**
Clicking on the application icon in the dock or launching from /Applications or from Spotlight does not open the application.

**To Reproduce**
1. Click on the dock icon, launch from spotlight, or open from /Applications.
2. Application fails to open.

**Expected behavior**
Application opens.

**Screenshots**
Terminal Command: 
```open -a Kitty```
Produces:
```LSOpenURLsWithRole() failed for the application /Applications/kitty.app with error 1.```

**Environment details**
OS:  macOS Big Sur Version: 11.0 Beta (20A4299v)

**Additional context**
Kitty can still be launched by navigating to /Applications, right clicking on kitty.app, selecting "Show Package Contents", navigating to /Contents/MacOS/, and double clicking the Unix Executable File "kitty".

kovidgoyal:
I'm afraid I dont have the resources to support un-released versions of
macOS. Apple manages to break working applications with pretty much
every release of their OS. Hopefully they will fix whatever they have
broken before the final release. Since it seems like its something in
the bundle wrapper they dont like, not the actual program, its likely a
trivial one line fix. If they dont manage to fix it, I will take a look
as soon as the final version is released. In the meantime, patches are
most welcome.

kchibisov:
I'll just cross link it here, since some kind of solution was posted on our tracker. https://github.com/alacritty/alacritty/issues/3880 . It's likely a macOS fault to some point, but it seems like there're some solutions already. I'm not saying that you should do anything, just FYI.

kovidgoyal:
Thanks, more information is always helpful. However if I understand correctly the solution was setting MACOS_DEPLOYMENT_TARGET=10.16 and rebuilding. That's not a workable solution for pre-built binaries since it will break running on anything other than Big Sur. So really need to wait for Apple.

alexanderniebuhr:
@kovidgoyal : I doubt it's just a mistake in the bundle. Some of the used API are deprecated now. For example NSUserNotification

kovidgoyal:
no, kitty doesnt use any currently deprecated apis (apart from opengl) and apis newly deprecated for big sur, will be deprecated not removed. not to mention that the OP is able to run kitty fine, just not via the bundle wrapper.

Luflosi:
[...]

PR #2876, opened 2020-07-23
Try to fix deprecation of NSUserNotification

Luflosi:
Here is an unfinished and not actually working commit that's supposed to fix the deprecation issue mentioned in #2805. Maybe you find it useful to develop actually working code.

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -11,6 +11,7 @@
 #include <Cocoa/Cocoa.h>
 
 #include <AvailabilityMacros.h>
+#include <UserNotifications/UserNotifications.h>
 // Needed for _NSGetProgname
 #include <crt_externs.h>
 #include <objc/runtime.h>
@@ -22,6 +23,12 @@
 #define NSEventModifierFlagCommand NSCommandKeyMask
 #define NSEventModifierFlagControl NSControlKeyMask
 #endif
+
+/*#if (MAC_OS_X_VERSION_MAX_ALLOWED < 101600)
+#define UNNotification NSUserNotification
+#define UNUserNotificationCenter NSUserNotificationCenter
+#define UNUserNotificationCenterDelegate NSUserNotificationCenterDelegate
+#endif*/
 
 typedef int CGSConnectionID;
 typedef int CGSWindowID;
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Oh for fuck's sake more of Apple's silly users need to be protected from
their own software nonsense. Apparently now one needs to ask permission
before daring to display a notification. 

https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications?language=objc

I will look into this someday, when the only alternative is gouging out
my own eyeballs.

alexanderniebuhr:
@Luflosi just to clarify, do you have any problems with this PR?
I just pulled it and voila .. it does work. 👍

Luflosi:
Do you mean that you can get a notification to display? In my testing it didn't work. It does compile without warnings though.

Luflosi:
Do you mean that you can get a notification to display? In my testing it didn't work. It does compile without warnings though.

alexanderniebuhr:
negative. no notification - haven't tested it, cause I am not needing them for terminal ... also don't need update notifications

I had a look, it seems alright. Might be issue with privacy. For notifications to work, you need pop-up asking user for permission, which I am a fan of!
However don't know how to get this implemented right with python / obj-c projects. (haven't worked a lot with them, everything Xcode and currently converting Universal Apps for Apple chips) 

sorry

kovidgoyal:
I tried implementing this, after wasting eight hours of my life on it. I find that at least on Catalina permission for notifications is always denied, so this entire exercise is pointless. Read the commit message for details. If i missed something and someone can get this to work, patches welcome. In the meantime I am just going to turn of Werror for cocoa_window.m

https://github.com/kovidgoyal/kitty/commit/2fafdae60bbf56216553e06f5eea90a39959a47a

Luflosi:
Would you accept a PR that makes it possible to select the old API with a compile-time option?
It would be nice to have notifications for unsigned builds of kitty. Also, I couldn't find a way to make this even compile with Nix since the API seems to only be available since macOS 10.14 and Nix uses the macOS 10.12 SDK.

kovidgoyal:
Since the API is deprecated it will presumably be removed by Apple, at
which point the compile time option will break and have to be removed.
I'd rather avoid such churn. Instead simply use compile time version
detection to choose the API.

Issue #2552, opened 2020-04-14
feature request: add OS window position to kitty @ ls


eraserhd:
Rationale: I'm moving from tmux to kitty+yabai using OS windows only, and one of my normal mappings, `C-w ,`, sends `C-p Enter` to the window in the lowest, rightmost corner.  This is useful for running unit tests or rerunning a shell command, for example.

In order to implement this, I need to discover which window is in the bottom right corner.

I can ask yabai, but this only works if the asker is local, and I'm often SSH'd into a server for development.

An alternate approach would be some kind of terminal escape with a payload that I can handle with a shell command locally.

Thoughts?

kovidgoyal:
This is a bit of a crap shoot. The problem is that not all platforms
make this information available. For instance, on Wayland, applications
are not allowed to query the system for their own window positions.

eraserhd:
This looks nice, thank you!

eraserhd:
I'm just getting to using this, and I can't figure out a way to get an OS window ID or handle.  (This is Mac OS.)

So, I can get any process's Kitty window ID, and I can make the kitten that asks yabai for all the window positions, but I can't match up window positions to Kitty windows:

1. Kitty is the process for all the windows, so I can't match the window by process id.
2. Kitty window ids appear to be unrelated to OS window IDs, so I can't use that.
3. This leaves title, but title is often the same for me for several windows.

The best option I can think of is to add a unique cookie to every window title to make them unique.  I actually want my windows numbered by position (a separate idea), but I need two passes:

1. Make sure all windows have unique titles.
2. Get all window positions.
3. Renumber windows based on positions.

Thoughts?

kovidgoyal:
The output of kitty @ ls will tell you the kitty os window id for any
given kitty window id. Each entry in the ls output is of the form

{
'id': XXX # this is the OS window id
'tabs': [
  'id': XXX # this is the tab id
  'windows': [
     'id': XXX # this is the window id

eraserhd:
It doesn't look like the OS window id on Mac OS:

```
$ kitty @ ls | jq '.[].id'
1
2
3
4
5
6

$ yabai -m query --windows | jq '.[].id'
9628
9592
9480
9555
9458
9516
9051
8333
236
179
187
11704
11701
198
```

kovidgoyal:
This is kitty's internal id for the OS window, kitty has no information
about what ids external programs assign to windows. If you are asking
[...]

PR #2877, opened 2020-07-23
On Mac OS, set $WINDOWID to OS's window number

eraserhd:
The API used is rather plain, not private or suspicious.  (I did research, though, and AFAICT, private APIs will pass the notary, but Apple threatened that they wouldn't when they first introduced the notary, so ... ??.) 

I'm running this locally, and the window ids are stable and agree with yabai's window numbers (apparently in spite of the vague comment in `windowNumber`'s documentation):

```
jfelice@C02X421DJHD4 src 
$ echo $WINDOWID
20273

jfelice@C02X421DJHD4 src 
$ yabai -m query --windows --window | jq '.id'
20273
```

I chose to reuse the `WINDOWID` variable because it seems like the same purpose and it is possible that some scripts might now work on Mac OS.

Let me know about any code quality issues, and I'll fix them up.

(See #2552 and koekeishiya/yabai#625)

### DIFF
--- a/kitty/tabs.py
+++ b/kitty/tabs.py
@@ -17,7 +17,7 @@ from .child import Child
 from .cli_stub import CLIOptions
 from .constants import appname, is_macos, is_wayland
 from .fast_data_types import (
-    add_tab, attach_window, detach_window, get_boss, mark_tab_bar_dirty,
+    add_tab, attach_window, cocoa_window_id, detach_window, get_boss, mark_tab_bar_dirty,
     next_window_id, remove_tab, remove_window, ring_bell, set_active_tab,
     set_active_window, swap_tabs, sync_os_window_title, x11_window_id
 )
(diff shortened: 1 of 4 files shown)


Luflosi:
I think you need to put the definition of `cocoa_window_id()` in an `#ifdef __APPLE__` so it compiles on non-macOS.

eraserhd:
@Luflosi doh, thanks.  Updated.

kitty/tabs.py
```diff
@@ -17,7 +17,7 @@
 from .cli_stub import CLIOptions
 from .constants import appname, is_macos, is_wayland
 from .fast_data_types import (
-    add_tab, attach_window, detach_window, get_boss, mark_tab_bar_dirty,
```
Luflosi, line 20: pyflakes complains that you removed this space.

Luflosi:
The definition of `x11_window_id()` should probably not be in the `#ifdef __APPLE__`.

eraserhd:
> The definition of `x11_window_id()` should probably not be in the `#ifdef __APPLE__`.

😭

Issue #2881, opened 2020-07-28
test_multiprocessing_spawn not passed in 0.18.2


tim77:
**Describe the bug**
`test_multiprocessing_spawn` not passed inf 0.18.2 on some arches:
```
test_multiprocessing_spawn (kitty_tests.tui.TestTUI) ... /usr/lib64/python3.9/multiprocessing/resource_tracker.py:96: UserWarning: resource_tracker: process died unexpectedly, relaunching.  Some resources might leak.
  warnings.warn('resource_tracker: process died unexpectedly, '
ERROR
test_word_trie (kitty_tests.unicode_input.TestUnicodeInput) ... ok
======================================================================
ERROR: test_multiprocessing_spawn (kitty_tests.tui.TestTUI)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/kitty-0.18.2/kitty_tests/tui.py", line 54, in test_multiprocessing_spawn
    test_spawn()
    self = <kitty_tests.tui.TestTUI testMethod=test_multiprocessing_spawn>
    test_spawn = <function test_spawn at 0x7f990d6133a0>
  File "/builddir/build/BUILD/kitty-0.18.2/kitty/multiprocessing.py", line 62, in test_spawn
    x = q.get(timeout=2)
    ctx = <multiprocessing.context.SpawnContext object at 0x7f990d69e730>
    get_context = <bound method DefaultContext.get_context of <multiprocessing.context.DefaultContext object at 0x7f990d6b2e20>>
    p = <SpawnProcess name='SpawnProcess-1' pid=896215 parent=896197 stopped exitcode=255>
    q = <multiprocessing.queues.Queue object at 0x7f990e7946d0>
  File "/usr/lib64/python3.9/multiprocessing/queues.py", line 114, in get
    raise Empty
    block = True
    deadline = 4362487.813879826
    self = <multiprocessing.queues.Queue object at 0x7f990e7946d0>
    timeout = 1.9999742964282632
_queue.Empty
----------------------------------------------------------------------
Ran 77 tests in 2.532s
FAILED (errors=1, skipped=2)
```

**To Reproduce**
Build Fedora RPM:
- [Task](https://koji.fedoraproject.org/koji/taskinfo?taskID=48003528)
- [Log](https://kojipkgs.fedoraproject.org//work/tasks/3577/48003577/build.log)

**Expected behavior**
Successful build as before, 0.18.1.

**Environment details**
[...]

PR #2885, opened 2020-07-30
Let kitty_exe() search in linux-package and kitty.app as well

Luflosi:
Fixes https://github.com/kovidgoyal/kitty/issues/2881.

This allows executing the tests after building only the linux-package or kitty.app.
`kitty_exe()` will also exit now if kitty could not be found. Is that good or could it be a problem?


kovidgoyal:
kitty_exe() is meant to find kitty at run time. It should not know or care about details of kitty's build system. If the nix kitty package wants to run the tests, it should first run ./setup.py build-launcher, then run the tests. Since the launcher is now required for the spawn test.

PR #2935, opened 2020-08-24
Addin OpenBSD compilation support, and executable path checking

spurgelaurels:
Kovid's changes to launcher.c are included in this PR, although they can be discarded if already in master.


kovidgoyal:
Thanks, I made your fixes slightly differently. Let me know if I missed anything.

PR #2937, opened 2020-08-25
Make visual bell flash much more gentle

keynslug:
Right now visual bell makes background flash sharply with bright white (when configured with darkish color theme). This causes eye strain, especially prominent in unlit environments.

This change makes background bounce smoothly between regular bg color and highlight (selection) bg color for the configured visual bell duration. Intensity is animated with cubic easing functions. It currently peaks at 20% of the duration, this is hardcoded.

---

Overall I acknowledge this change as quick and somewhat dirty starting point. For example I chose not to introduce any new configuration options yet. I'm wholly open to your suggestions on how to improve it.

### DIFF
--- a/kitty/cell_vertex.glsl
+++ b/kitty/cell_vertex.glsl
@@ -220,7 +220,10 @@ void main() {
     draw_bg = 1;
 
 #if defined(BACKGROUND)
-    background = bg;
+    background = choose_color(visual_bell_intensity,
+        blend_softlight(color_to_vec(highlight_bg), bg),
+        bg
+    );
     // draw_bg_bitfield has bit 0 set to draw default bg cells and bit 1 set to draw non-default bg cells
     uint draw_bg_mask = uint(2 * cell_has_non_default_bg + (1 - cell_has_non_default_bg));
     draw_bg = step(1, float(draw_bg_bitfield & draw_bg_mask));
--- a/kitty/cell_vertex.glsl
+++ b/kitty/cell_vertex.glsl
@@ -134,10 +134,6 @@ float is_cursor(uint xi, uint y) {
     float x_equal = step(0.5, x1_equal + x2_equal);
     return step(2.0, x_equal + y_equal);
 }
-
-vec3 blend_softlight(vec3 base, vec3 blend) {
-    return (1.0 - 2.0 * blend) * base * base + 2.0 * blend * base;
-}
 // }}}
 
 
@@ -195,10 +191,6 @@ void main() {
     // Selection
     vec3 selection_color = color_to_vec(highlight_fg);
     foreground = choose_color(float(is_selected & ONE), selection_color, foreground);
-    foreground = choose_color(visual_bell_intensity * 0.5,
-        blend_softlight(color_to_vec(highlight_bg), foreground),
-        foreground
-    );
     decoration_fg = choose_color(float(is_selected & ONE), selection_color, decoration_fg);
 #endif
     // Underline and strike through (rendered via sprites)
@@ -247,12 +239,9 @@ void main() {
 
 #if defined(SPECIAL) || defined(SIMPLE)
     // Selection and cursor
+    bg = max(bg, choose_color(visual_bell_intensity, color_to_vec(highlight_bg), bg));
     bg = choose_color(float(is_selected & ONE), color_to_vec(highlight_bg), bg);
-    bg = choose_color(cell_has_block_cursor, color_to_vec(cursor_color), bg);
-    background = choose_color(visual_bell_intensity,
-        blend_softlight(color_to_vec(highlight_bg), bg),
-        bg
-    );
+    background = choose_color(cell_has_block_cursor, color_to_vec(cursor_color), bg);
 #if !defined(TRANSPARENT) && defined(SPECIAL)
     float is_special_cell = cell_has_block_cursor + float(is_selected & ONE);
     bg_alpha = step(0.5, is_special_cell);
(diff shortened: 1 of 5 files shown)


kovidgoyal:
In principle, looks ok. Some comments:

1) How well does this work in inverted text? Or when using a fullscreen program like vim that sets its own colors. With background_opacity?

2) Make the easing and intensity functions static inline. 

3) I dont think this needs config options, lets just pick the numbers for now and config options can always be added later if there is actual demand.

keynslug:
>     1. How well does this work in inverted text? Or when using a fullscreen program like vim that sets its own colors. 

Spot on, not great at all. I guess light theme users are also affected. I'll come up with some robust solution.

> With background_opacity?

That seems fine.

>     2. Make the easing and intensity functions static inline.

Done.

kovidgoyal:
Doesnt seem very smooth to me, see attached screencast
[screencast.zip](https://github.com/kovidgoyal/kitty/files/5137497/screencast.zip)

kovidgoyal:
You know, stepping outside the box, a little. I dont see why we need to
render it via some kind fo reverse video effect at all. Why not just
render a semi-transparent overlay color over the entire window? Might
look much better and have zero performance impact when a visual bell is
not occurring.

keynslug:
> Why not just
> render a semi-transparent overlay color over the entire window? Might
> look much better and have zero performance impact when a visual bell is
> not occurring.

I have no objections. Thought of something similar, had a question though: when a bell fires in an inactive tab kitty does **not** start visual bell duration timer, is that right?

kovidgoyal:
On Thu, Aug 27, 2020 at 09:16:55AM -0700, Andrew Mayorov wrote:
> > Why not just
> > render a semi-transparent overlay color over the entire window? Might
> > look much better and have zero performance impact when a visual bell is
> > not occurring.
> 
> I have no objections. Thought of something similar, had a question though: when a bell fires in an inactive tab kitty does **not** start visual bell duration timer, is that right?

Yes, I think that is the case instead it shows a bell in the tab bar. Try it and see. 

sleep 5s; print '\a'

should work

keynslug:
> Doesnt seem very smooth to me, see attached screencast
> [screencast.zip](https://github.com/kovidgoyal/kitty/files/5137497/screencast.zip)

MacOS? Seems to me this is happening because screen repaints are effectively limited by cursor blink frequency. Is there some way I can request repaints more often? Also wondering why am I not seeing the same under Linux + Wayland.

keynslug:
> Yes, I think that is the case instead it shows a bell in the tab bar. Try it and see.

Great, I'll try to add a seprate pass for the flash then.

kovidgoyal:
On Thu, Aug 27, 2020 at 09:32:25AM -0700, Andrew Mayorov wrote:
> > Doesnt seem very smooth to me, see attached screencast
> > [screencast.zip](https://github.com/kovidgoyal/kitty/files/5137497/screencast.zip)
> 
> MacOS? Seems to me this is happening because screen repaints are effectively limited by cursor blink frequency. Is there some way I can request repaints more often? Also wondering why am I not seeing the same under Linux + Wayland.

No linux + x11, just wake up the main loop with self->dirty = true and
wakeup_main_loop(). This will cause continuous repainting and high cpu
usage so, better to use a timer.

But I think just using a single overlay color with no need for these
complications would be easier. And look fine.

To do it, you will need to add a step to the end of draw_cells() that
checks for a color to overlay and does it if set.
-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

kovidgoyal:
Another idea: rather than an overlay color, maybe just draw a big exclamation mark or bell symbol centered over the window. It should be fairly easy to draw an exclamation mark in GLSL.

kovidgoyal:
To make it less disruptive maybe just a red rectangle over the window edges, so the content is not obsucred. There are lots of options now that we are thinking outside the box :)

keynslug:
> But I think just using a single overlay color with no need for these
complications would be easier. And look fine.

Made an attempt. Looks nice and a bit more consistent.

There's one little issue I found though. No matter [what I try](https://github.com/kovidgoyal/kitty/blob/318936cf5151ec7db8fd7d36e09a82b878e91644/kitty/shaders.c#L465) this effect seems to affect background opacity for some reason. For example if you run `./kitty/launcher/kitty -o background_opacity=0.0 -o visual_bell_duration=1.0 -o selection_background=#a0a0a0` the flash would make window not fully transparent for a flash duration. 

> To make it less disruptive maybe just a red rectangle over the window edges, so the content is not obsucred.

This may be even better, that's right. I would propose to flash color of a _screen border_ though, this way it's much clearer which screen in a multiscreen layout wants your attention. My only doubt is that user might have turned off border completely so this wouldn't work.

I want to give it a try, so we have at least two options to choose from.

kovidgoyal:
Yes, since it is merged.

gshpychka:
@kovidgoyal not sure what you mean. It's closed, not merged.

kovidgoyal:
Read the commit that closed it.

Luflosi:
screenshot
The commit adds a new entry to the changelog. Previous commits add the changes from this PR if you look at the commit history. The pull request is marked as "closed" since it was not merged directly but the commit was rebased and maybe improved/changed.

gshpychka:
>Read the commit that closed it.

I did, and have no idea what it means. It's linking to this PR, which is closed (not merged).

>The pull request is marked as "closed" since it was not merged directly but the commit was rebased and maybe improved/changed.

Oof, interesting.

PR #2984, opened 2020-09-28
Remove unnecessary condition

Luflosi:
According to the python documentation, `Py_CLEAR()` has no effect when the argument is `NULL`, see https://docs.python.org/3/c-api/refcounting.html#c.Py_CLEAR.

In case you were planning to release a new version in the next couple days, could you please wait til this weekend, when I will have made all the PRs that I would like to include in the next version?


kovidgoyal:
kitty is due for a release, but sure I can hold off for a few days.

Issue #2711, opened 2020-05-31
Invoking `launch --copy-env` through a binding doesn't copy the environment 


suvayu:
**Describe the bug**
When bound to a key, `launch --copy-env` doesn't seem to copy the environment correctly.  Using the remote control feature, and calling `kitty @launch --copy-env` works as expected.

**To Reproduce**
1. add a binding like this in the configuration:

       map kitty_mod+c --copy-env

2. in a kitty window, set a dummy variable

       export _TEST=foo

3. activate the binding defined in (1)
4. check in the new window with something like this

       env | grep _TEST

   No match found

**Expected behavior**
Step (4) should find a match.  If instead of activating the binding in (3), I type `kitty @launch --copy-env` in the original window, (4) succeeds.  For some context, I want to bind it to the Python interpreter/applications that run inside a virtual/conda environment. 

Minimal config file needed to reproduce:

    allow_remote_control    yes
    map kitty_mod+c launch --copy-env

**Enviroment details**
OS: Linux - 5.6.14-300.fc32.x86_64 (Fedora 32)
Hardware: AMD Ryzen 7 PRO 3700U w/ Radeon Vega

<details>
<summary><code>$ kitty --debug-config</code></summary>

```
kitty 0.17.4 created by Kovid Goyal
Linux hostname 5.6.14-300.fc32.x86_64 #1 SMP Wed May 20 20:47:32 UTC 2020 x86_64
\S
Kernel \r on an \m (\l)
Loaded config files: /home/username/.config/kitty/kitty.conf
Running under: X11

Config options different from defaults:
allow_remote_control y
background_opacity   0.8
enable_audio_bell    False
enabled_layouts      ['tall', 'fat']
visual_bell_duration 0.01
Added shortcuts:
         f1 KeyAction(func='launch', args=['--stdin-source=@screen_scrollback', '--stdin-add-formatting', 'less', '+G', '-R'])
         shift+control+o > c KeyAction(func='launch', args=['--cwd=current', '--copy-env'])
         shift+control+o > x > b KeyAction(func='launch', args=['--cwd=current', 'bc', '-q', '-l'])
         shift+control+o > x > e KeyAction(func='launch', args=['--cwd=current', 'emacsclient', '-nw'])
         shift+control+o > x > g KeyAction(func='launch', args=['--cwd=current', 'gnuplot'])
         shift+control+o > x > h KeyAction(func='launch', args=['--type=tab', 'htop'])
[...]

PR #2986, opened 2020-09-30
launch documentation improvements

poliorcetics:
As said in #2711, here is a PR that describes the limitation of `launch --copy-env` as well as hint at its workaround.

I also added code blocks for `--stdin-source` in a separate commit so that if it is not wanted, it can be reverted easily.

EDIT: I would like to provide an example of using the remote control feature workaround. I will try to make it work for me but if you read this before I commit the example and you have the solution, could you point me toward it ?

### DIFF
--- a/kitty/launch.py
+++ b/kitty/launch.py
@@ -112,11 +112,11 @@ computers (for example, over ssh) or as other users.
 type=choices
 default=none
 choices=none,@selection,@screen,@screen_scrollback,@alternate,@alternate_scrollback
-Pass the screen contents as :code:`STDIN` to the child process. @selection is
-the currently selected text. @screen is the contents of the currently active
-window. @screen_scrollback is the same as @screen, but includes the scrollback
-buffer as well. @alternate is the secondary screen of the current active
-window. For example if you run a full screen terminal application, the
+Pass the screen contents as :code:`STDIN` to the child process. :code:`@selection` is
+the currently selected text. :code:`@screen` is the contents of the currently active
+window. :code:`@screen_scrollback` is the same as :code:`@screen`, but includes the
+scrollback buffer as well. :code:`@alternate` is the secondary screen of the current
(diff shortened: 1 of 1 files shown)


poliorcetics:
Found an example that should work pretty nicely, I'm done with this if you have no objection to the additions. :)

kitty/launch.py
```diff
@@ -81,7 +81,11 @@ def options_spec() -> str:
 --copy-env
 type=bool-set
 Copy the environment variables from the currently active window into the
-newly launched child process.
+newly launched child process. It will not copy environment variables that have
+been :code:`export`ed since the :code:`export` command does not change the environment
```
trygveaa, line 85: Isn't it a bit misleading to say "variables that have been:`export`ed? It makes it sound like variables you set with export (`export TEST=test`) doesn't work, while variables you just set (`TEST=test`) works, while in reality none of them work, only the variables the shell had when it started?

kitty/launch.py
```diff
@@ -81,7 +81,11 @@ def options_spec() -> str:
...
+of the shell, only of child processes launched in the shell. You can use remote control
+instead to have the current environment copied (:code:`allow_remote_control` must be on): 
+:code:`map kitty_mod+x remote_control launch --copy-env vim`.
```
trygveaa, line 88: Hm, this isn't working for me, while running `kitty @launch --copy-env` in the shell works.

poliorcetics: It works for me though it fails 1 time out of 3 or each time I create a new tab. This is strange.

poliorcetics: I haven't managed to make it work each time and in every situation, I'll remove the text.

kitty/launch.py
poliorcetics: This is a given that `A=B` variables that are not exported are not available to child processes.

PR #2996, opened 2020-10-03
macOS: add ability to show subtitles in notifications

Luflosi:
This used to be implemented before 4e3c6e52aad41b9deb98d052c806f4a84846b782, when the now deprecated notifications framework was still being used.
Implement it again for feature parity, since my next PR will re-add the old notifications framework.

Note that I don't know how to actually test this, since it apparently requires signing and notarization, which I can't do.
Since this isn't actually used anywhere yet, do you think we should just leave this feature out?
Also, should I implement it in `kitty/notify.py` and in `send_test_notification()` as well?
How do you call `send_test_notification()` when testing? Which command do you use?

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -1569,7 +1569,7 @@ class Boss:
         from .notify import notify
         now = monotonic()
         ident = f'test-notify-{now}'
-        notify(f'Test {now}', f'At: {now}', identifier=ident)
+        notify(f'Test {now}', f'At: {now}', identifier=ident, subtitle=f'Test subtitle {now}')
 
     def notification_activated(self, identifier: str, window_id: int, focus: bool, report: bool) -> None:
         w = self.window_id_map.get(window_id)
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -156,13 +156,14 @@ static PyObject *notification_activated_callback = NULL;
 
 
 static void
-schedule_notification(const char *identifier, const char *title, const char *body) {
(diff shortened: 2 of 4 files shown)


Luflosi:
I'll fix the CI error in a minute.

kovidgoyal:
I believe you can sign things you build locally with some kind of null key or something and have it work. Dont recall the details. As for how to use send_test_notification, 

kitty -o "map f1 send_test_notification"

kovidgoyal:
And I am fine with having subtitle there, though as you noticed it isnt actually used.

Luflosi:
Thanks for the tip with signing things. I found https://stackoverflow.com/questions/27474751/how-can-i-codesign-an-app-without-being-in-the-mac-developer-program/27474942, which was very helpful. Now I can see notifications in the macOS notification center but I still don't get a notification. But this at least allows me to test the functionality.

I fixed the CI error by adding `subtitle` to `kitty/notify.py` and I made `send_test_notification()` show the subtitle too.

kovidgoyal:
On Sat, Oct 03, 2020 at 07:11:53AM -0700, Luflosi wrote:
> Thanks for the tip with signing things. I found https://stackoverflow.com/questions/27474751/how-can-i-codesign-an-app-without-being-in-the-mac-developer-program/27474942, which was very helpful. Now I can see notifications in the macOS notification center but I still don't get a notification. But this at least allows me to test the functionality.

You're welcome :) And yes I cant get it to actually show the
notification either, just pops up in notification center.

PR #2997, opened 2020-10-03
macOS: re-add deprecated notification API

Luflosi:
Please don't merge this yet.

This PR adds the deprecated macOS notification API again, so that notifications work when compiling from source without code signing. It was removed in 4e3c6e52aad41b9deb98d052c806f4a84846b782.

What did you mean by "compile time version detection" in https://github.com/kovidgoyal/kitty/pull/2876#issuecomment-697109685? Simply check if the macOS version is 10.15 (Catalina) or lower and then use the old API? But I think it would be useful to be able to use this API as long as it's still only deprecated and not removed. Do you have an idea, how this could be done best?

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -9,7 +9,9 @@
 #include "state.h"
 #include "monotonic.h"
 #include <Cocoa/Cocoa.h>
+#if !defined(KITTY_USE_DEPRECATED_MACOS_NOTIFICATION_API)
 #include <UserNotifications/UserNotifications.h>
+#endif
 
 #include <AvailabilityMacros.h>
 // Needed for _NSGetProgname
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I dont see the point in relying on an API we know is going to be
removed. I was mainly talking about your issue of not being able to
compile on older macOS. So at compile time detect that the new API is
not available and if so, compile against the old one. Probably can do so
with Availability.h macros.

Luflosi:
Ok, I'm fine with not having this in Big Sur. Is it possible to query `MAC_OS_X_VERSION_MAX_ALLOWED` from `setup.py`? I need to do this to not add the `UserNotifications` framework to `platform_libs`.

kovidgoyal:
You could just try to compile a small dynamically generated .m file and link with th framework, if it succeeds, use it, if not, use the older API. This is how configure traditionally works.

Luflosi:
Please review.
I changed `test_compile()` to pipe stdout and stderr to /dev/null, so the error message is not visible. Does this cause any problems?

Luflosi:
These were all the commits I wanted to add before the next release. Thank you for waiting a couple days with the release, it will make it much easier for me to update the kitty package for nix. You may now release the next version if you like.

PR #3030, opened 2020-10-14
GLFW: Wayland: Assume 96 DPI if physical size is <= 0

Luflosi:
From upstream: https://github.com/glfw/glfw/commit/0ef149c8f2451fbc8e866834675a075cfc295b6c.


Luflosi:
The CI failed. It doesn't look like I broke it. Can you investigate please?

PR #3043, opened 2020-10-19
macOS: Fix detection for notification API

### LINKED ISSUE #3042: Build system help
I'm on macOS 10.13.6 because of my hardware.
I'm getting an error when building Kitty.app about `UserNotifications.h` not being found.
The changelog for 0.19 mentions the new notification framework which seems like the likely culprit.
A little perusal of the source shows there's [an ifdef for people in a similar situation](https://github.com/kovidgoyal/kitty/blob/master/kitty/cocoa_window.m#L12).

Here comes the question. I can't find any documentation on how to specify this value. Do I set an env var or do I need to pass it to setup.py?

Luflosi:
For some reason it can happen that the `UserNotifications` framework exists but the `UserNotifications/UserNotifications.h` header doesn't.
The temporary directory is needed because clang would try to write to `/dev/null.d`, which would obviously fail.
Closes #3042.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -343,7 +343,10 @@ def kitty_env() -> Env:
         platform_libs = [
             '-framework', 'CoreText', '-framework', 'CoreGraphics',
         ]
-        user_notifications_framework = first_successful_compile(ans.cc, '-framework UserNotifications')
+        test_program_src = '''#include <UserNotifications/UserNotifications.h>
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I merged it making some changes to avoid unnecessary disk IO, please check that it still works

Luflosi:
I made a mistake, see #3044.

PR #3054, opened 2020-10-23
Nointense

jaeheum:
￼For hints kitten --hints-text-color. Prefer "standard" text colors for better general usability


kovidgoyal:
I'm afraid the use of intense colors for the hints kitten is not going to
change.

jaeheum:
How about a new option of text-background-color? Would that be more acceptable?

kovidgoyal:
No, I'm afraid I simply dont see a good reason for this.

PR #3058, opened 2020-10-25
Fix memory leak in unlikely error case

Luflosi:
Found with the Clang Static Analyzer.

Please double-check, that this is the correct way to free this piece of memory.


kovidgoyal:
Given that matches is not released anywhere in this function, why does it need tobe released there?

Luflosi:
When running the analyzer without this PR, it only finds this memory leak. With this PR, it finds the next memory leak at the end of the function. With that leak fixed, it finds the third leak due to the remaining `return`. I was just going to make three separate PRs but I can combine them into one if you prefer that. I don't care either way.

kovidgoyal:
please do combine it into one.

Luflosi:
I created a new commit with the three changes and pushed it to this branch. Is it possible to reuse this PR or should I open a new one?

Issue #3051, opened 2020-10-22
Kitty quits if "sub-os-window" exits on Wayland


LeonardKoenig:
**Describe the bug**

If I create a new os-type window on Wayland, and this exits, the whole kitty instance stops.

**To Reproduce**
1. Spawn a kitty instance with wayland forced enable, as well as remote control active
   ```
   $ kitty -c /dev/null -o allow_remote_control=true -o linux_display_server=wayland
   ```
2. Spawn a new OS window from the fresh kitty instance
   ```
   $ kitty @ new-window --window-type=os
   ```
3. Exit the newly created "sub" window
4. Observe the other window exiting as well, with the following log:
   ```
   dri_create_context: requested glthread but driver is missing backgroundCallable V2 extension
   [296 18:02:24.589116] [glfw error 65549]: Wayland: Window attention request not implemented yet
   dri_create_context: requested glthread but driver is missing backgroundCallable V2 extension
   wl_display@1: error 0: invalid object 51
   [296 18:02:29.440750] [glfw error 65544]: Wayland: fatal display error: Invalid argument
   ```

**Expected behavior**
Only the sub-window exits.

**Screenshots**
N/A

**Environment details**
OS: ArchLinux on GNOME on Wayland

```
kitty 0.19.1 created by Kovid Goyal
Linux Eddie 5.9.1-zen2-1-zen #1 ZEN SMP PREEMPT Sun, 18 Oct 2020 02:45:20 +0000 x86_64
Arch Linux \r (\l)
LSB_VERSION=1.4
DISTRIB_ID=Arch
DISTRIB_RELEASE=rolling
DISTRIB_DESCRIPTION="Arch Linux"
Loaded config files: /dev/null
Running under: Wayland

Config options different from defaults:
allow_remote_control y
linux_display_server wayland
```

kovidgoyal:
Works fine for me on an intel gpu, with both sway and weston. From your output it looks
like an issue with whatever gpu drivers and/or wayland compositor you
are using.

dri_create_context: requested glthread but driver is missing backgroundCallable V2 extension

Incidentally, you dont need to use remote control to create a new OS
window, just press ctrl+shift+n.

LeonardKoenig:
The backgroundCallable extension seems unrelated, it happens with virtually any program and is logged at startup, the actual issue seems to be:

```
wl_display@1: error 0: invalid object 51
[...]

PR #3066, opened 2020-10-28
wl: Fix destroying CSDs in correct order (#3051)

LeonardKoenig:
On Wayland, some compositors refuse to implement SSDs in which case we draw
primitive CSDs.  However, the destructors were called in the wrong order,
leading to a dereference of an already freed object.

### DIFF
--- a/glfw/wl_window.c
+++ b/glfw/wl_window.c
@@ -413,10 +413,10 @@ static void createDecorations(_GLFWwindow* window)
 
 static void destroyDecoration(_GLFWdecorationWayland* decoration)
 {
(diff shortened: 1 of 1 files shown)


Luflosi:
Could you please also make a similar PR to [GLFW](https://github.com/glfw/glfw/pulls)?

Luflosi:
The code is here: https://github.com/glfw/glfw/blob/0ef149c8f2451fbc8e866834675a075cfc295b6c/src/wl_window.c#L247-L258

LeonardKoenig:
ah, sure!

PR #3068, opened 2020-10-29
Add shell.nix

Luflosi:
This makes it possible to execute `nix-shell` to create the correct environment to build kitty. Use `nix-shell --pure` to eliminate most of the influence of the outside system, e.g. globally installed packages.
This works on NixOS and any Linux or macOS system with the Nix package manager installed.
The build inputs are split into `buildInputs`, `nativeBuildInputs`, `propagatedBuildInputs` and `checkInputs` so it closely resembles https://github.com/NixOS/nixpkgs/blob/2bb3a9da24ca60d9f5bed69f679a1ec50dbdf997/pkgs/applications/terminal-emulators/kitty/default.nix. This makes it easy to port changes between the two files.

I tested that it works correctly with NixOS (both X11 and Wayland), Arch Linux with Nix installed (just X11) and macOS of course.
I first created this file a couple months back. It has been the main way I develop kitty on my machine for the past couple weeks but it's been a little annoying since I had to checkout this branch, then do `nix-shell` and then switch back to whatever I wanted to work on.
I think this will also be very useful for other contributors using Nix.
I hope you don't mind the extra file in the root directory of kitty.

### DIFF
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -78,6 +78,17 @@ you might have to rebuild the app.
    a self signed certificate, see for example, `here
    <https://stackoverflow.com/questions/27474751/how-can-i-codesign-an-app-without-being-in-the-mac-developer-program/27474942>`_.
 
+Build and run from source with Nix
+-------------------------------------------
+
+On NixOS or any other Linux or macOS system with the Nix package manager installed,
+execute ``nix-shell`` to create the correct environment to build kitty or use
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I'm OK with adding another file, however, if you want this, then please
document it in build.rst.

Luflosi:
Good idea. I added a little comment to the documentation. Please review.

PR #3086, opened 2020-11-10
GLFW: X11: Make libX11 dynamically loaded

Luflosi:
From upstream: https://github.com/glfw/glfw/commit/a0a5cc57dff620f97428168a2580714bebc22381.

I moved some of the `typedef`s to `glfw/xkb_glfw.h` because that's where a certain `struct` is located in kitty (but not in GLFW). GLFW has the `typedef`s in `src/x11_platform.h`. Is putting the typedefs in that file a good solution or do you see a better one? I tried including `src/x11_platform.h` but that didn't work out well.
Please also check that I removed the correct library from the list of libraries to link against and that no other libraries should be removed.

### DIFF
--- a/glfw/glfw.py
+++ b/glfw/glfw.py
@@ -79,7 +79,7 @@ def init_env(env: Env, pkg_config: Callable, at_least_version: Callable, test_co
         at_least_version('xkbcommon', 0, 5)
 
     if module == 'x11':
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Why do we want libx11 loaded dynamically? Dynamic loading is slow, fragile and adds code complexity for no benefit, since we already have runtime loading of the backend, by dynamically loading the correct glfw module, which is a much nicer solution than dynamically loading every needed library. For instance, kitty's fork of glfw links against a whole bunch of libraries, see line 82 of glfw.py

Issue #3126, opened 2020-11-28
Any plans to make an macOS universal app?


screendriver:
**Is your feature request related to a problem? Please describe.**

Are there any plans to make an macOS universal app that includes both binaries for Intel as well as binaries for the new ARM64-based Apple M1 silicon chip?

kovidgoyal:
Someday, sure.

Luflosi:
@screendriver do you have a Mac with an Apple silicon chip?

screendriver:
Yes 🙂

kovidgoyal:
kitty is pretty trivial to build from source, do that any you will have
an ARM version. https://sw.kovidgoyal.net/kitty/build.html

yingzhu146:
I've just tried to build it from source on a Macbook Air with M1.

I'm using the following flags:

```bash
CFLAGS='-arch arm64' CC="$(which clang)"
```

and running either

```bash
make
```

or 

```bash
python setup.py kitty.app
```

I'm running into 

```bash

❯ make
python3 setup.py 
CC: /usr/bin/clang (12, 0)
[1/1] Building launcher... done
[4/4] Linking kittens/choose/subseq_matcher ... done
Linking kittens/choose/subseq_matcher ...
/usr/bin/clang -Wextra -Wfloat-conversion -Wno-missing-field-initializers -Wall -Wstrict-prototypes -std=c11 -O3 -fwrapv -fstack-protector-strong -pipe -fvisibility=hidden -D_FORTIFY_SOURCE=2 -arch arm64 -flto -Ikitty -I/Users/<user>/miniconda3/include/python3.9 -Wall -O3 -shared -flto build/subseq_matcher-charsets.c.o build/subseq_matcher-main.c.o build/subseq_matcher-output.c.o build/subseq_matcher-score.c.o build/subseq_matcher-unix_compat.c.o -ldl -framework CoreFoundation -L/Users/<user>/miniconda3/lib -lpython3.9 -Wl,-stack_size,1000000 -framework CoreFoundation -o build/kittens/choose/subseq_matcher.so
ld: -stack_size option can only be used when linking a main executable
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [all] Error 1
```
Any idea how to get around here @kovidgoyal?  Not too much experience with compiling.

Happy to open separate issue too.

kovidgoyal:
that is caused by a bug in miniconda. It has added -stack_size to the
[...]

PR #3142, opened 2020-12-04
do not use clang option -march=native on Apple Silicon

runchard:
Apple Silicon arm64-apple-darwin clang do not support `-march` option.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ import shutil
 import subprocess
 import sys
 import sysconfig
+import platform
 import time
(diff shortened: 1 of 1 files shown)


runchard:
The good thing is, it just works in my own M1 laptop.

But, it is hard to verify the correctness, for CI do not have such environment.

kovidgoyal:
According to #3126 there is not need for this.

kovidgoyal:
Sorry, ignore my previous comment.

PR #3158, opened 2020-12-09
Add headless mode

alin23:
This would allow users to run commands in the context of a running kitty instance without showing a window.

Use cases:
1. Prepare an entire session with windows/tabs and show it all at once
```shell
kitty -o allow_remote_control=yes --listen-on unix:/tmp/kitty --start-as headless
kitty @ --to unix:/tmp/kitty launch htop
kitty @ --to unix:/tmp/kitty launch webserver
kitty @ --to unix:/tmp/kitty focus-window --match title:htop  # this will show the window
```
2. Extract information about kitty configuration using kittens
```python
# /tmp/kitten.py

from kittens.tui.handler import result_handler
from pprint import pprint

def main():
    return ""

@result_handler(no_ui=True)
def handle_result(args, res, target_window_id, boss):
    pprint(boss.opts.keymap)
```
```shell
kitty -o allow_remote_control=yes --listen-on unix:/tmp/kitty --start-as headless
kitty @ --to unix:/tmp/kitty kitten /tmp/kitten.py
```
3. Quake style terminal started as hidden at boot

*Have something like this run at boot (crontab/launchd)*
```shell
kitty --title=quake \
  -o initial_window_width=$(system_profiler SPDisplaysDataType | grep 'Resolution:' | head -n 1 | awk '{ print $2 }') \
  -o initial_window_height=400 \
  -o allow_remote_control=yes \
  --listen-on unix:/tmp/quake_kitty \
  --start-as headless
```
-----
*Bind this to a hotkey through something like skhd/karabiner/BetterTouchTool*
```shell
kitty @ --to unix:/tmp/quake_kitty focus-window
```

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -204,7 +204,8 @@ class Boss:
                 os_window_id = create_os_window(
                         initial_window_size_func(size_data, self.cached_values),
                         pre_show_callback,
-                        self.args.title or appname, wname or self.args.name or wclass, wclass)
+                        self.args.title or appname, wname or self.args.name or wclass, wclass,
+                        self.args.start_as == 'headless')
         tm = TabManager(os_window_id, self.opts, self.args, startup_session)
         self.os_window_map[os_window_id] = tm
         return os_window_id
--- a/kitty/cli.py
+++ b/kitty/cli.py
@@ -655,7 +655,7 @@ can also be specified in :file:`kitty.conf`.
 --start-as
 type=choices
 default=normal
-choices=normal,fullscreen,maximized,minimized
+choices=normal,fullscreen,maximized,minimized,headless
 Control how the initial kitty window is created.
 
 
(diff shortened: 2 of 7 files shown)


kovidgoyal:
(1) and (3) are already covered by the sessions support. Not sure what (2) is about?

alin23:
I've been working on a way to extract the already parsed keymap from kitty so that I can fuzzy search through it because I can't remember all the mappings.

Best way I found is to launch a kitty instance with `allow_remote_control=yes` and print `boss.opts.keymap` and `boss.opts.sequence_map`. 
But this always shows a kitty window so I figured this headless feature would help and would probably be useful for others too.

kovidgoyal:
kitty +runpy "from kitty.cli import parse_args, create_opts; cli_opts = parse_args()[0]; opts = create_opts(cli_opts); print(opts.keymap)"

for parsed keymaps, no need to invoke the GUI at all.

alin23:
Thank you! This is much better than my solution!

PR #3174, opened 2020-12-16
Support default font_features from fontconfig

Diaoul:
As discussed here: https://github.com/kovidgoyal/kitty/pull/2248

### DIFF
--- a/kitty/fontconfig.c
+++ b/kitty/fontconfig.c
@@ -37,7 +37,21 @@ pattern_as_dict(FcPattern *pat) {
         if (PyDict_SetItemString(ans, #name, p) != 0) { Py_CLEAR(p); Py_CLEAR(ans); return NULL; } \
         Py_CLEAR(p); \
     }}
+
+#define L(type, get, which, conv, name) { \
+    type out; PyObject *p; int n = 0; \
+    PyObject *l = PyList_New(0); \
+    while (get(pat, which, n, &out) == FcResultMatch) { \
+        p = conv(out); if (p == NULL) { Py_CLEAR(l); Py_CLEAR(ans); return NULL; } \
+        if (PyList_Append(l, p) != 0) { Py_CLEAR(p); Py_CLEAR(l); Py_CLEAR(ans); return NULL; } \
+        Py_CLEAR(p); \
+        n++; \
+    } \
+    if (PyDict_SetItemString(ans, #name, l) != 0) { Py_CLEAR(l); Py_CLEAR(ans); return NULL; } \
+    Py_CLEAR(l); \
+}
 #define S(which, key) G(FcChar8*, FcPatternGetString, which, PS, key)
+#define LS(which, key) L(FcChar8*, FcPatternGetString, which, PS, key)
 #define I(which, key) G(int, FcPatternGetInteger, which, PyLong_FromLong, key)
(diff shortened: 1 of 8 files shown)


Diaoul:
I'll do a pass on the doc when I get the validation of the implementation :memo: 

For reference:
```bash
cat ~/.config/fontconfig/fonts.conf
```
```xml
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
	<alias>
		<family>monospace</family>
		<prefer>
			<family>Fira Code</family>
		</prefer>
	</alias>
  <match target="font">
    <test name="family">
      <string>Fira Code</string>
    </test>
    <edit name="fontfeatures" mode="append">
      <string>ss01</string>
    </edit>
  </match>
</fontconfig>
```

kovidgoyal:
Looks generally fine with a quick glance over. Add the docs in config_data.py and I will review in detail.

Diaoul:
Actually there is a bug in `pattern_as_dict`: it does not handle multiple values correctly. This applies to `FC_FONT_FEATURES` but maybe others as well. For reference, [here](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/master/src/fcdbg.c#L188) is how multiple values are handled by `FcPatternPrint` using [`FcValueListPrint`](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/master/src/fcdbg.c#L129).

kovidgoyal:
Feel free to send a separate PR to fix it or wait and I will get around
to it when I have a moment.

kovidgoyal:
Also, what's the bug, as far as I know, patter_as_dict doesnt get any non-scalar values currently.

Diaoul:
I just fixed it for font_features. As for the others I don't know if they are lists or not and I could not find a reliable source of information for this. [Here](https://cgit.freedesktop.org/fontconfig/tree/fontconfig/fontconfig.h#n123) FC_FONT_FEATURES is defined as a String...

I've seen references to fonts being able to be part of multiple families somewhere (sorry lost the reference) and despite `family` being singular, [the description](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/master/doc/fontconfig-devel.sgml#L138) reads *Font family name**s***. :man_shrugging:

kitty/fontconfig.c
```diff
@@ -37,7 +37,21 @@ pattern_as_dict(FcPattern *pat) {
...
+    PyObject *l = PyList_New(0); \
+    while (get(pat, which, n, &out) == FcResultMatch) { \
+        p = conv(out); if (p == NULL) { Py_CLEAR(l); Py_CLEAR(ans); return NULL; } \
+        if (PyList_Append(l, p) != 0) { Py_CLEAR(p); Py_CLEAR(l); Py_CLEAR(ans); return NULL; } \
+        Py_CLEAR(p); \
+        n++; \
+    } \
+    if (PyDict_SetItemString(ans, #name, l) != 0) { Py_CLEAR(l); Py_CLEAR(ans); return NULL; } \
+    Py_CLEAR(l); \
+}
```
Diaoul, line 52: Please doublecheck that one @kovidgoyal, I tried to make sure I've cleaned up properly but I might have missed something.

PR #3197, opened 2020-12-29
macOS: Fix change font size keyboard shortcut

### LINKED ISSUE #2839: MacOS cmd++ does not increase font size
When performing keyboard shorcut cmd++ on MacOS the font size is not increased. 

cmd+- and cmd+0 both decreases and reset font sizing correctly per [documentation](https://sw.kovidgoyal.net/kitty/#other-keyboard-shortcuts).

ctrl+shift+= does work as described. 

I've attempted holding command and hitting "=+" key along with holding command and shift while hitting "=+" key. Neither solution worked. Other programs accept the command w/"=+" key (i.e without shift). 

1. Start kitty
2. Hit Cmd++

OS: macOS Catalina; V10.15.5
Hardware: Macbook Pro 13"; Early 2015

Thanks!

russelldavies:
Fixes #2839.

I can't find any definitive MacOS keyboard shortcuts for changing application font size but from most applications it seems that `cmd+equal` and `cmd+minus` are mapped to increasing and decreasing, respectively. Some applications, like Chrome, also allow you to use the shift key as well but this doesn't seem to be across the board and without shift seems more common. Considering that some people might want to map `cmd+shift+minus|equal` to other actions I think it makes sense to not include these as a default mapping without being surprising.

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -1475,8 +1475,7 @@ k('next_layout', 'kitty_mod+l', 'next_layout', _('Next layout'))
 g('shortcuts.fonts')  # {{{
 k('increase_font_size', 'kitty_mod+equal', 'change_font_size all +2.0', _('Increase font size'))
 if is_macos:
-    k('increase_font_size', 'cmd+plus', 'change_font_size all +2.0', _('Increase font size'), add_to_docs=False)
-    k('increase_font_size', 'cmd+shift+equal', 'change_font_size all +2.0', _('Increase font size'), add_to_docs=False)
+    k('increase_font_size', 'cmd+equal', 'change_font_size all +2.0', _('Increase font size'), add_to_docs=False)
 k('decrease_font_size', 'kitty_mod+minus', 'change_font_size all -2.0', _('Decrease font size'))
 if is_macos:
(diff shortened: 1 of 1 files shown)


Luflosi:
At least on my Mac with macOS 10.15.7 and a German keyboard layout, <kbd>cmd</kbd >+<kbd>+</kbd> seems to increase the font size or zoom in all applications i tested: Terminal.app, iTerm2, Safari, Firefox, Chrome, TextEdit, Pages and Notes. <kbd>cmd</kbd >+<kbd>=</kbd>, which I need to type as <kbd>cmd</kbd >+<kbd>shift</kbd>+<kbd>0</kbd> on my keyboard layout does nothing except in iTerm2, Firefox and Notes. In iTerm it maximises the window, in Firefox it zooms and in Notes it resets the font size to the original one.

kovidgoyal:
If you wish to map a default keyboard shortcut to other things, you can always do so in kitty.conf. I think it is better to have all variants work out of the box.

russelldavies:
Yeah, good point. I've amended the commit to include all variants but possibly because this PR is closed it's not showing the [amended commit](https://github.com/russelldavies/kitty/commit/83c53effb47c088068f6f790494c7a4711366dcb). Could you please merge this yourself as currently `cmd+equal` is not configured at the moment which is quite surprising.

kovidgoyal:
sure

PR #3240, opened 2021-01-13
Add MacOS menu option to close window

quentin-fox:
I often find that I accidentally hit Cmd-W when coding, and accidentally close the window, sometimes losing changes, and often just having to deal with the pain of recovering swap files. MacOS allows you to overwrite the shortcuts for an app, but it seems like it's only if the command is explicitly in one of the menus.

For example, I am able to create a new shortcut for "Quit kitty" so that I don't accidentally quit the entire application, but I am not able to do it for "Close" since there is not a dedicated menu option for it.

Other applications I have looked at have it in the "Window" menu tab and call it "Close", so the UX will be similar enough to other MacOS applications.

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -395,6 +395,10 @@ cocoa_create_global_menu(void) {
     NSMenu* windowMenu = [[NSMenu alloc] initWithTitle:@"Window"];
     [windowMenuItem setSubmenu:windowMenu];
 
+    [windowMenu addItemWithTitle:@"Close"
+                          action:@selector(performClose:)
+                   keyEquivalent:@"w"];
+
     [windowMenu addItemWithTitle:@"Minimize"
                           action:@selector(performMiniaturize:)
                    keyEquivalent:@"m"];


kovidgoyal:
You might be interested in: https://sw.kovidgoyal.net/kitty/conf.html#opt-kitty.confirm_os_window_close

quentin-fox:
Thanks! Very useful option, will look into it.

sheriferson:
I see this is already merged, but is it too late to make an argument for why the default shortcut should be different?

This violates how all other Mac apps treat that shortcut (<kbd>⌘w</kbd>). <kbd>⌘w</kbd> closes the window when the window doesn't have tabs, but when the window has tabs it closes only the current tab (this is the case in Safari, Finder, any other tabbed app I know of).

It seems the problem can be solved if the default shortcut (if a default is needed) in the commit is changed to <kbd>⇧⌘w</kbd>, which would still allow the user to change it in System Preferences, but would be more in line with Mac app conventions and would be less surprising to new users.

sheriferson:
Exhibits A and B to support my case:

Screen Shot 2021-01-16 at 2 50 55 AM
Screen Shot 2021-01-16 at 2 50 43 AM

PR #3264, opened 2021-01-27
Update Nix package link

davidak:


Luflosi:
May I suggest changing it to https://search.nixos.org/packages?channel=unstable&show=kitty&sort=relevance&query=kitty instead?

davidak:
I tried to make the link simple, but i can do that too. I added a commit. You can squash when merging.

PR #3419, opened 2021-03-25
Add *.d to .gitignore

elebow:
These are dependency files created by the compiler during the build process.


kovidgoyal:
kitty's build system creates these inside the build directory only which is already gitignored.

elebow:
Ah, I must have had them left over for some reason. After `git clean -f && make`, I do not see them in git status.

Issue #3409, opened 2021-03-19
Box drawing rounded corners with extra line-height produces stretched radius


jesseleite:
**Describe the bug**

I notice box drawing of rounded corners doesn't produce a proper corner radius when Kitty is configured with a larger line-height.  The corners look stretched...

**Expected behavior**

I would expect the rounded corner radius to render properly in the center of a character cell, across any line-height setting.

**Screenshots**

![image](https://user-images.githubusercontent.com/5187394/111794841-06c8d480-889d-11eb-84f6-91b5556b6a39.png)

**Environment details**
OS: MacOS Big Sur 11.2.1

```
kitty 0.19.3 created by Kovid Goyal
Darwin Jesses-MacBook-Pro.local 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 x86_64
ProductName:	macOS ProductVersion:	11.2.1 BuildVersion:	20D74
Loaded config files: /Users/jesseleite/.config/kitty/kitty.conf

Config options different from defaults:
active_border_color        Color(red=255, green=255, blue=255)
adjust_line_height         2.0
bold_font                  Menlo
bold_italic_font           Menlo
color0                     Color(red=33, green=33, blue=33)
color1                     Color(red=255, green=63, blue=97)
color10                    Color(red=134, green=232, blue=106)
color11                    Color(red=255, green=227, blue=95)
color12                    Color(red=62, green=180, blue=232)
color13                    Color(red=255, green=123, blue=52)
color14                    Color(red=62, green=180, blue=232)
color2                     Color(red=134, green=232, blue=106)
color3                     Color(red=255, green=227, blue=95)
color4                     Color(red=62, green=180, blue=232)
color5                     Color(red=255, green=123, blue=52)
color6                     Color(red=62, green=180, blue=232)
color7                     Color(red=255, green=255, blue=255)
color8                     Color(red=66, green=66, blue=66)
color9                     Color(red=255, green=63, blue=97)
cursor                     Color(red=255, green=255, blue=255)
cursor_blink_interval      0.0
font_family                Menlo
font_size                  16.0
foreground                 Color(red=255, green=255, blue=255)
hide_window_decorations    2
italic_font                Menlo
macos_option_as_alt        3
macos_show_window_title_in none
macos_thicken_font         0.65
mouse_hide_wait            1.0
[...]

PR #3420, opened 2021-03-25
Prevent subpixel rendering on box drawn corners

jesseleite:
Firstly, I want to thank you @kovidgoyal for your patience and help; I really appreciate all the work you do on Kitty! ❤️

Given the discussion at https://github.com/kovidgoyal/kitty/issues/3409#issuecomment-803767623...

>You basically want to change the values returned by the lower x() function by a delta. Play with it and see what happens, though in my testing, the current code is optimal across all cell sizes.

I agree that the rectircle equation is correct/perfect, but I found that the cause of the seemingly-misaligned corners (especially on the right side) is due to subpixel rendering when the `x()` functions return certain floats / decimals on certain displays.  I found that applying `math.floor()` to the left quadrant, and `math.ceil()` to the right quadrant prevents subpixel rendering issues at every font size and cell_width I tried.  Here are screenshot examples, before and after my change...

### Before

This can be extremely subtle, but if you look at where the corner meets the vertical side, there is a subpixel rendering issue causing it to look ever-so-slightly misaligned (at least on both of my displays in MacOS here):

CleanShot 2021-03-25 at 00 08 42

### After

After this `math.floor()` and `math.ceil()` change, I cannot reproduce this subpixel rendering error at any font size or cell width:

CleanShot 2021-03-25 at 00 09 49

### DIFF
--- a/kitty/fonts/box_drawing.py
+++ b/kitty/fonts/box_drawing.py
@@ -478,11 +478,11 @@ def rectircle_equations(
     if left_quadrants:
         def x(t: float) -> float:
             xterm = 1 - pow(t, yexp)
-            return cell_width - abs(a * pow(xterm, xexp)) - adjust_left_quadrant
+            return math.floor(cell_width - abs(a * pow(xterm, xexp)) - adjust_left_quadrant)
     else:
         def x(t: float) -> float:
             xterm = 1 - pow(t, yexp)
-            return abs(a * pow(xterm, xexp))
+            return math.ceil(abs(a * pow(xterm, xexp)))
 
     return x, y
 


jesseleite:
Rad! ❤️

No rush at all, but curious when the next tagged release is slated to drop?

Anyway, thank you for your time once again!

kovidgoyal:
On Wed, Mar 24, 2021 at 09:29:53PM -0700, Jesse Leite wrote:
> No rush at all, but curious when the next tagged release is slated to drop?

It's waiting on the two RFCs to mature for animation and keyboard, see
the issues list.

xzbdmw:
> Rad! ❤️
> 
> No rush at all, but curious when the next tagged release is slated to drop?
> 
> Anyway, thank you for your time once again!

Hello what are the unicodes you used for box drawing? My looks like this                
 borderchars = { "─", "│", "─", "│", "╭", "╮", "╯", "╰" },
截屏2024-02-18 08 37 59
截屏2024-02-18 08 38 09

PR #3430, opened 2021-03-30
Implementation of hyper and meta

orki:
Implementation of modifiers: hyper and meta as first class citizens, similar to alt and super

I'm using this as my daily driver for emacs with 6 modifier keys: Shift, Alt, Control, Meta, Hyper, Super. (There are emacs lisp changes, which are not public yet; they will be once some variant of this pull request is merged.) Tested with an NKRO keyboard with the following XKB specification on Linux X11:

``key <LFSH> { [ Shift_L   ] };
key <RTSH> { [ Shift_R   ] };

key <LCTL> { [ Control_L ] };
key <RCTL> { [ Control_R ] };

key <LWIN> { [ Meta_L    ] };
key <RWIN> { [ Super_L   ] };

key <LALT> { [ Alt_L     ] };
key <RALT> { [ Hyper_L   ] };

modifier_map Shift  { Shift_L, Shift_R };
modifier_map Lock   { Caps_Lock };
modifier_map Control{ Control_L, Control_R };
modifier_map Mod1   { Meta_L };
modifier_map Mod2   { Alt_L };
modifier_map Mod3   { Hyper_L };
modifier_map Mod4   { Super_L, Super_R };``

It fixes a couple of bugs in the current XKB implementation in `glfw_xkb.c`:

1. XKB modifiers are not looked up correctly; a new function `glfw_xkb_update_masks` now adds the correct modifiers under X11.
2. The return value of `xkb_state_mod_index_is_active` did not handle error returns correctly.

The following additions have been made:

1. Key codes for HYPER and META. This extends the modifier protocol in a mostly backwards-compatible way; the main exception is that modifier-only keycodes after HYPER_LEFT have changed to make room for META_LEFT.
2. Two additional modifier bits for hyper and meta in the keyboard protocol.
3. _GLFWXKBData has acquired mask/idx members for the two new modifiers.
4. Implicit in the above, but worth stating: some modifiers may not be present at all in a user's configuration; the bug-fixes above allow for the absence of modifiers in the XKB map.

What is left to do:

1. Support for detecting the new mods on ibus, OS X, and evdev/wayland. I do not have any systems on which I add/test such support. To the best of my knowledge, nothing is broken.
2. Document the additional bits, and the small changes in keycode indices; these are trivial, but I'd wait until the patch is merged before adding this documentation.


glfw/ibus_glfw.c
```diff
@@ -388,7 +388,7 @@ ibus_key_state(unsigned int glfw_modifiers, int action) {
     M(CONTROL, IBUS_CONTROL_MASK);
     M(ALT, IBUS_MOD1_MASK);
     M(NUM_LOCK, IBUS_MOD2_MASK);
-    M(SUPER, IBUS_MOD4_MASK);
```
kovidgoyal, line 391: Why is IBUS_MOD4_MASK no longer mapped to super?

glfw/x11_window.c
```diff
@@ -185,6 +185,7 @@ static int translateState(int state)
 {
     int mods = 0;
 
+    /* Need some way to expose hyper and meta without xkbcommon-x11 */
```
kovidgoyal, line 188: Might be better to just take the current modifier state from xkb. However X.h has Mod5Mask as well, might be worth looking into.

glfw/xkb_glfw.c
```diff
@@ -308,11 +315,84 @@ glfw_xkb_update_x11_keyboard_id(_GLFWXKBData *xkb) {
...
+}
+
+
 #else
 
 #define xkb_glfw_load_keymap(keymap, map_str) keymap = xkb_keymap_new_from_string(xkb->context, map_str, XKB_KEYMAP_FORMAT_TEXT_V1, 0);
 #define xkb_glfw_load_state(keymap, state) state = xkb_state_new(keymap);
 
+static void
+glfw_xkb_update_masks(_GLFWXKBData *xkb) {
```
kovidgoyal, line 380: So hyper and meta work only under X11, correct?

glfw/ibus_glfw.c
orki: That is an embarrassing bug. Thank you for catching it. Fix pushed.

glfw/x11_window.c
orki: Mod5Mask is generally mapped to a level shift of some sort, in my understanding, which you rarely want to touch. Users of non-ASCII-like encodings may know better than I do. It's possible to check whether any keys are mapped to modifier bits, by going over all keys in the core X11 keyboard, and then using XLookupString to identify candidates for each of the modifiers. However, I'd be surprised if there exist non-XKB users of X11 who actually care about these modifiers. In other words, I'd wait for a bug report from at least one such user before implementing this.

glfw/xkb_glfw.c
orki: Yes, X11 with XKB. I can provide a patch that I think Should Work (TM) for non-X11 XKB by looking at modifier names provided by libxkbcommon (using the same technique as in the patch), but I have no way of validating such a patch.

kovidgoyal:
On Wed, Mar 31, 2021 at 07:19:53AM -0700, orki wrote:
> @orki commented on this pull request.
> 
> 
> 
> >  #else
>  
>  #define xkb_glfw_load_keymap(keymap, map_str) keymap = xkb_keymap_new_from_string(xkb->context, map_str, XKB_KEYMAP_FORMAT_TEXT_V1, 0);
>  #define xkb_glfw_load_state(keymap, state) state = xkb_state_new(keymap);
>  
> +static void
> +glfw_xkb_update_masks(_GLFWXKBData *xkb) {
> 
> Yes, X11 with XKB. I can provide a patch that I think Should Work (TM) for non-X11 XKB by looking at modifier names provided by libxkbcommon (using the same technique as in the patch), but I have no way of validating such a patch.

Yes, please add that, even if it doesnt work, it can form the basis of
future efforts.

orki:
Here's code to find modifiers in the pure X11 case; it's not clear to me where masks can be stored, since I don't see a place in GLFW without XKB where modifiers can be stored:

```
#include <X11/keysymdef.h>

static
int choose_one_modifier( Display *disp, KeyCode *modifiermap, int max_keypermod,
                         KeySym left_ks, KeySym right_ks )
{
  int indx;
  int found_row = -1;
  KeyCode left = XKeysymToKeycode( disp, left_ks );
  KeyCode right = XKeysymToKeycode( disp, right_ks );
  for ( indx = 0; indx < 8*max_keypermod && found_row == -1; ++indx )
  {
    /* Don't handle the nasty case of left mapped to one modifier and the right to another. */
    if ( ( left && modifiermap[indx] == left ) || ( right && modifiermap[indx] == right ) )
      found_row = indx/max_keypermod;
  }
  return found_row;
}

int discover_X_modifiers( Display *disp )
{
  unsigned used_bits = 0;       // Must be alteast 8 bits for X11
  int mod_index = -1;
  XModifierKeymap *mk = XGetModifierMapping( disp );
  if ( !mk )
    return 0;

#define S( a, left, right ) \
  mod_index = choose_one_modifier( disp, mk->modifiermap, mk->max_keypermod, left, right ); \
  if ( mod_index >= 0 && !( used_bits & ( 1u<<mod_index ) ) )           \
    used_bits |= 1U << mod_index, debug( "%s modifier:  %d\n", #a, mod_index )

  S( alt, XK_Alt_L, XK_Alt_R );
  S( super, XK_Super_L, XK_Super_R );
  S( hyper, XK_Hyper_L, XK_Hyper_R );
  S( meta, XK_Meta_L, XK_Meta_R );
  XFreeModifiermap( mk );

  return 1;
}
```
Currently, the code above prints the found modifier, but does nothing else otherwise.

kovidgoyal:
Not sure I understand what you are asking. If you need to store some extra data add it to _GLFWXkbData in xkb_glfw.h this is accessible under both wayland and X11. If you want it to be present only for wayland, ifdef it under _GLFW_WAYLAND

orki:
In the pure X11 case, in `x11_window.c`, what is the equivalent of the persistent storage (`_GLFWXkbData` in xkb_glfw.c) so that these modifiers can be stored somewhere? If XKBData is present in the pure X11 case, how do I get a handle to it where `translateState` is called?

orki:
Untested, probably not working patch for XKB under evdev/wayland:

```
 diff --git a/glfw/xkb_glfw.c b/glfw/xkb_glfw.c
index f774e724..a142d619 100644
--- a/glfw/xkb_glfw.c
+++ b/glfw/xkb_glfw.c
@@ -378,18 +378,35 @@ glfw_xkb_update_masks(_GLFWXKBData *xkb) {
 
 static void
 glfw_xkb_update_masks(_GLFWXKBData *xkb) {
-    /* Should find better solution under evdev or wayland */
+    bool succeeded = false;
+    unsigned used_bits = 0; /* To avoid using the same bit twice */
 
 #define S( a ) xkb->a##Idx = XKB_MOD_INVALID; xkb->a##Mask = 0
-    S(hyper); S(meta);
+    S(alt); S(super); S(hyper); S(meta); S(numLock);
+#undef S
+    for (xkb_mod_index_t i = 0, last = i < xkb_keymap_num_mods(xkb->keymap); i < last; i++) {
+        unsigned mask_rtn = 1 << i;
+        /* Hopefully this works under evdev or wayland */
+        const char *name = xkb_keymap_mod_get_name(xkb->keymap, i);
+#define S( a, s ) if ( !(used_bits & mask_rtn) && strcmp( name, #s ) == 0 ) xkb->a##Mask = mask_rtn, used_bits |= mask_rtn
+        /* Note that the order matters here; earlier is higher priority. */
+        S( alt, Alt );
+        S( super, Super );
+        S( numLock, NumLock );
+        S( meta, Meta );
+        S( hyper, Hyper );
 #undef S
+    }
+    unsigned indx, shifted;
+    for ( indx = 0, shifted = 1; used_bits; ++indx, shifted <<= 1, used_bits >>= 1 ) {
+#define S( a ) if ( xkb->a##Mask == shifted ) xkb->a##Idx = indx
+        S(alt); S(super); S(hyper); S(meta); S(numLock);
+#undef S
+    }
 #define S(a, n) xkb->a##Idx = xkb_keymap_mod_get_index(xkb->keymap, n); xkb->a##Mask = 1 << xkb->a##Idx;
     S(control, XKB_MOD_NAME_CTRL);
     S(shift, XKB_MOD_NAME_SHIFT);
     S(capsLock, XKB_MOD_NAME_CAPS);
-    S(numLock, XKB_MOD_NAME_NUM);
-    S(alt, XKB_MOD_NAME_ALT);
-    S(super, XKB_MOD_NAME_LOGO);
 #undef S
 }
 
```

kovidgoyal:
On Wed, Mar 31, 2021 at 10:14:13AM -0700, orki wrote:
> In the pure X11 case, in `x11_window.c`, what is the equivalent of the persistent storage (`_GLFWXkbData` in xkb_glfw.c) so that these modifiers can be stored somewhere? If XKBData is present in the pure X11 case, how do I get a handle to it where `translateState` is called?

_glfw.x11.xkb

and in wayland

_glfw.wl.xkb

orki:
In `initExtensions`, we have:

```    _glfw.x11.xkb.available = XkbQueryExtension(_glfw.x11.display,
            &_glfw.x11.xkb.majorOpcode,
            &_glfw.x11.xkb.eventBase,
            &_glfw.x11.xkb.errorBase,
            &_glfw.x11.xkb.major,
            &_glfw.x11.xkb.minor);

    if (!_glfw.x11.xkb.available)
    {
        _glfwInputError(GLFW_PLATFORM_ERROR, "X11: Failed to load Xkb extension");
        return false;
    }
```
I don't see how kitty could work under X11 without XKB, given the preceding. Is this still worth pursuing?

orki:
I see. https://github.com/kovidgoyal/kitty/pull/3430#issuecomment-811263885 is my best attempt right now, but I don't have a way to test wayland. This pull request does not remove any functionality; it adds full functionality to X11, and leaves them in the prior state for the rest. The patch above will definitely break working wayland set up, if it is wrong; I wouldn't make that part of this pull request without testing by those who can.
Is there anything else to do before this PR can be merged?

orki:
Indeed. There's quite some voodoo inside the XKB-X11 interface due to the way X insists on using only 8 modifiers, and those too, only in a particular way. `libxkbcommon` never added full support for such queries; [attempts](https://github.com/xkbcommon/libxkbcommon/issues/88) [to do so](https://github.com/xkbcommon/libxkbcommon/pull/36) have all fizzled out. On my system, with the configuration in the OP, X modifiers are
```
alt: 4
super: 6
hyper: 5
meta: 3
shift: 0
capsLock: 1
control: 2
```
but the (incorrect) results from `xkb_keymap_mod_get_name` are
```
XKB modifier 0 name: Shift
XKB modifier 1 name: Lock
XKB modifier 2 name: Control
XKB modifier 3 name: Mod1
XKB modifier 4 name: Mod2
XKB modifier 5 name: Mod3
XKB modifier 6 name: Mod4
XKB modifier 7 name: Mod5
XKB modifier 8 name: NumLock
XKB modifier 9 name: Alt
XKB modifier 10 name: LevelThree
XKB modifier 11 name: LAlt
XKB modifier 12 name: RAlt
XKB modifier 13 name: RControl
XKB modifier 14 name: LControl
XKB modifier 15 name: ScrollLock
XKB modifier 16 name: LevelFive
XKB modifier 17 name: AltGr
XKB modifier 18 name: Meta
XKB modifier 19 name: Super
XKB modifier 20 name: Hyper
```
For completeness, here's what `XkbGetNames( disp, XkbVirtualModNamesMask, xkb )` returns where the second number is the mask computed using `XkbVirtualModsToReal`:
```
0: "NumLock" - 0
1: "Alt" - 16
2: "LevelThree" - 128
3: "LAlt" - 0
4: "RAlt" - 0
5: "RControl" - 0
6: "LControl" - 0
7: "ScrollLock" - 0
8: "LevelFive" - 0
9: "AltGr" - 128
10: "Meta" - 8
11: "Super" - 64
12: "Hyper" - 32
```
Apologies for the inconsistent formatting; these are outputs of test programs.

The obvious theory is that `xkb_keymap_mod_get_name` simply took the standard X modifiers and tacked on the additional names. We need a way then to go from these names to the masks themselves, which, in X11, is `XkbVirtualModsToReal`, for which there is no analogue in `libxkbcommon`.

orki:
The above is also the reason that I made Shift/CapsLock/Control not modifiable in this PR. I can configure XKB to change those modifiers too under X11, but that goes into a rabbit hole that is not worth exploring.

kovidgoyal:
Ugh, that's a pity. Well this PR looks ready to me. Will do final review and merge when I have some time.

orki:
Thanks. You updated `keyboard-protocol.rst` too; I should have saved you the work.

kovidgoyal:
No worries, its less work to just do it sometimes :)

glfw/ibus_glfw.c
dbivolaru: @kovidgoyal Shouldn't there be also

```
M(META, IBUS_MOD1_MASK);
M(HYPER, IBUS_MOD3_MASK);
```

kovidgoyal: No idea somebody that uses ibus and hyper/meta needs to test and validate.

orki:
What is the desired behavior for ibus?

dbivolaru:
I am also running with a keyboard with separate meta keys and when using `vim` with meta mappings eg `<M-e>` they work under `XTerm` but not under `kitty`.

Any ideas where to start looking? This pull request seemed to be the closest thing.

Issue #3247, opened 2021-01-15
Padding between top of window (or bottom of window) and tab bar


jsphnecclia:
Is there a way to configure a padding/margin between the top of the window and the tab bar (alternatively the bottom of the window and the tab bar, depending on the setting)?

kovidgoyal:
https://sw.kovidgoyal.net/kitty/conf.html#opt-kitty.window_margin_width

jsphnecclia:
The tab bar [ignores](https://u.teknik.io/7CZRC.png) that setting. I mean space between the edge of the window and the tab bar, not the tab bar and the edge of the content of the window (i.e. the text).

kovidgoyal:
Ah, in that case, no, there is no such setting.

jsphnecclia:
Ah. Well do you have any idea where a beginner programmer would look to make that change (making the tab bar be affected by the window_margin_width)?

kovidgoyal:
this will be fairly involved. The tab bar is actually a kitty screen.
Margins around it would need to be rendered in the border OpenGL shader,
and the positioning of the tab bar window would need to be adjusted.

jsphnecclia:
Ah. Then could you point me to where the code for the set tab_bar_min_tabs option lies? And/Or the code for the tab bar in general?

kovidgoyal:
tab_bar.py

jsphnecclia:
Duh, of course. Thank you

jsphnecclia:
How does one add a rc like set-background-opacity, set-window-title, etc? I thought it would be as simple as duplicating one file already in the rc folder, changing it, and then rebuilding -- but that doesn't seem to work. I get the error that the command is an unknown command

kovidgoyal:
Adding a file to the RC folder is all you need to do. See the code in
rc/base.py for how they are loaded.

jsphnecclia:
Thats what I thought, but I can't get even a simple duplicate command recognized. I copied the file detach_window.py to toggle_tab_bar.py, and changed only the name - but kitty @ toggle-tab-bar returns command not found. Here is my toggle_tab_bar.py (detach_window.py duplicate with only the name changed):

```
from typing import TYPE_CHECKING, Optional, Union

from .base import (
    MATCH_TAB_OPTION, MATCH_WINDOW_OPTION, ArgsType, Boss, MatchError,
    PayloadGetType, PayloadType, RCOptions, RemoteCommand, ResponseType,
    Window
)

[...]

PR #3436, opened 2021-04-01
Add padding options for around the tab bar

jsphnecclia:
```
The margin to the top or bottom of the tab bar (in pts)
If tab_bar_style is hidden, this value has no effect.
If the number of tabs is less than tab_bar_min_tabs, this is the distance between the top of the os window and
the point where the padding/margin set through window_margin_width or window_padding_width takes over.
If the number of tabs is greater than tab_bar_min_tabs, there is the top of the os window, a padding of tab_bar_margin_height,
the tab bar (which takes a height of one character cell), the padding/margin set through the methods mentioned above, and
then the shell prompt.
```

Edit tab_bar_margin_height in kitty.conf to see the effect (defaults to 0.0)

(reworked os_window_regions to not evaluate unnecessary conditions in this feature, when the tab bar style is set to hidden)

[See Issue](https://github.com/kovidgoyal/kitty/issues/3247)

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -622,7 +622,6 @@ edge_desc = _(
     'A single value sets all four sides. Two values set the vertical and horizontal sides.'
     ' Three values set top, horizontal and bottom. Four values set top, right, bottom and left.')
 
-
 o('window_margin_width', '0', option_type=edge_width, long_text=_('''
 The window margin (in pts) (blank area outside the border).
 For margins with tab bar, see tab_bar_margin_height.''' + edge_desc))
@@ -706,13 +705,16 @@ o('tab_bar_margin_width', 0.0, option_type=positive_float, long_text=_('''
 The margin to the left and right of the tab bar (in pts)'''))
 
 o('tab_bar_margin_height', 0.0, option_type=positive_float, long_text=('''
-The margin to the top or bottom of the tab bar (in pts)
-If tab_bar_style is hidden, this value has no effect.
-If the number of tabs is less than tab_bar_min_tabs, this is the distance between the top of the os window and
-the point where the padding/margin set through window_margin_width or window_padding_width takes over.
-If the number of tabs is greater than tab_bar_min_tabs, there is the top of the os window, a padding of tab_bar_margin_height,
-the tab bar (which takes a height of one character cell), the padding/margin set through the methods mentioned above, and
-then the shell prompt.'''))
+The margin to the top or bottom of the tab bar (in pts). If tab_bar_style is hidden, this value has no effect.
+If the number of tabs is less than tab_bar_min_tabs and retain_tab_bar_margin_height is set on, this is the distance
+between the top of the os window and the point where the padding/margin set through window_margin_width/window_padding_width takes over. 
+If the number of tabs is greater than tab_bar_min_tabs, there is the top/bottom of the os window, a padding of tab_bar_margin_height,
+the tab bar (which takes a height of one character cell), the padding/margin (set through window_margin_width or window_padding_width), and
+then the shell prompt. (or the bottom of the field if tab_bar_edge is set to bottom)'''))
+
+o('retain_tab_bar_margin_height', False, long_text=('''
+Retain tab_bar_margin_height when the number of tabs is too few to trigger the tab bar.
+'''))
 
 o('tab_bar_style', 'fade', option_type=choices('fade', 'separator', 'powerline', 'hidden'), long_text=_('''
 The tab bar style, can be one of: :code:`fade`, :code:`separator`, :code:`powerline`, or :code:`hidden`.
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -623,16 +623,16 @@ edge_desc = _(
     ' Three values set top, horizontal and bottom. Four values set top, right, bottom and left.')
 
 o('window_margin_width', '0', option_type=edge_width, long_text=_('''
-The window margin (in pts) (blank area outside the border).''' + edge_desc +
-'''For margins with tab bar, see tab_bar_margin_height.'''))
+The window margin (in pts) (blank area outside the border).
+For margins with tab bar, see tab_bar_margin_height.''' + edge_desc))
 
 o('single_window_margin_width', '-1', option_type=optional_edge_width, long_text=_('''
 The window margin (in pts) to use when only a single window is visible.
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Why should this option apply when the tab bar is hidden because there are not enough tabs?

jsphnecclia:
The assumption was that if the tab style was not set to hidden, then the margin when the tab bar is hidden would be no less than the margin of the tab bar when it is shown. 
Screen Shot 2021-04-03 at 12 13 31 PM

Important to note: By default the window padding or margin set in window_margin_width and window_padding_width is the margin or padding below (or above) the tab bar. 
Screen Shot 2021-04-03 at 12 22 12 PM


Setting it this way also has the effect of removing a ghost/duplication of the prompt, seen below:
Screen Shot 2021-04-03 at 12 20 05 PM
Resizing the os_window also gets rid of this duplication. We can get rid of your complaint, add more customization, and fix this bug all at once if there were a way to isolate what part of the resizing functionality we need to replicate and place it in the correct create_tab function. I wasn't able to isolate this function when I tried my hand at it earlier

jsphnecclia:
(PS: And more concisely as an answer to why there would be a difference between tab_style hidden and a tab hidden by being less than min_tabs is precisely that the former setup would never see a tab bar and thus window_padding_width/window_margin_width would have the reasonable effect desired, while the latter setup is designed to allow for a case either with or without the tab bar. This causes problems that weren't there in the tab_style route: specifically that window_margin_width/window_padding_width is interpreted in some cases as distance from the os_window, and in some cases as the distance from the tab_bar.)

kovidgoyal:
You cant just change the central region position/size. When you decrease its size it means nothing is going to be drawn where it was drawn before. Which is the cause of the rendering artifacts you found, whatever was in the back buffer at that location previously, is being drawn. Normally the tab bar will render in the space left for it. If you add extra space you have to have something that will render in that space. 

You will need to call add_borders_rect() in borders.py around line 83 to cover the rectangle created by the extra padding. And please change it so that the behavior is the same regardless of how the tab bar is hidden. Its not really intuitive to have it behave differently in the two cases.

jsphnecclia:
I've tried adding border rect in both tab_bar.py and borders.py, to no effect. I'm not sure how add_borders_rect() (in borders.py) works. I've looked over the code but when I try to implement it by adding `add_borders_rect(self.os_window_id, self.tab_id, 0, 0, 0, 100, BorderColor.default_bg)` after the `if not has_background_image` block, I still get the same rendering artifacts

kovidgoyal:
Oh yeah my mistake, sorry. border_rects is used only for in tab
contents. The tab bar lives outside all tabs. You will have to do it the
hard way by changing line 624 in child-monitor.c to call draw_borders()
for the padding region.

jsphnecclia:
What would the inside of that call to draw_borders() look like? 

I was testing around trying to figure it out and I got this:

`draw_borders*br->vao_idx, br->num_border_rects, br->rect_buf, br->is_dirty, os_window->viewport_width, 100, active_window_bg, num_visible_windows, all_windows_have_same_bg, os_window);`

But all it does is now the entire prompt is flickering in the visual artifacts rather than just the cursor, as before.

kovidgoyal:
draw_borders() draws rectangles represented by the BorderRects and a vertex array object (vao_idx). Currently the vao and rects are stored per tab. You will have to add another pair of these to the OSWindow object and add the actual rectangles for the new padding region. See add_border_rext in state.c for an example.

jsphnecclia:
So I'll be adding a vao and rects to OSWindow for the area around (but not in) the tab bar in the child_monitor file, and a new version of add_borders_rect (say add_tab_bar_rect) in the state.c file?

kovidgoyal:
On Tue, Apr 06, 2021 at 10:38:03AM -0700, jsphnecclia wrote:
> So I'll be adding a vao and rects to OSWindow for the area around (but not in) the tab bar in the child_monitor file, and a new version of add_borders_rect (say add_tab_bar_rect) in the state.c file?

Basically, yes.

jsphnecclia:
I can't seem to find the declaration of OSWindow / where would I put the vao and rects otherwise?

kovidgoyal:
state.h

jsphnecclia:
Ah, thank you. I'm not used to this python/c hybrid.

I've been working some more on the different methods you have mentioned in your replies. I've tried several different paths, but each with no success.

in state.c:
Screen Shot 2021-04-09 at 8 38 15 PM
Later:
`KIIIII(add_tab_bar_padding_rect)`
and later:
`MW(add_tab_bar_padding_rect, METH_VARARGS)`

(brings up the following error: 
Screen Shot 2021-04-09 at 6 35 07 PM)


also, in state.c , I tried this in add_os_window(), to no success:
    `ans->tab_bar_padding_render_data.vao_idx = create_cell_vao();`

I then tried to make a set_tab_bar_padding_render_data, with no success:
Screen Shot 2021-04-09 at 8 51 35 PM

I also gave OSWindow in state.h a border_rects and a ScreenRenderData tab_bar_padding_render_data. I think at least these are right. but I used these in child-monitor.c to no effect. I tried to place set_tab_bar_padding_render_data in the layout() function there (among other places such as the borders.py file), again with no luck :(.

I've reached a dead end. If you could give a simple example of an (even partial) fix for this rendering artifact (partial meaning a fix that covers up only part of the artifact), I should be able to implement the rest of the patch no problem.

(I committed a version that showcases the rendering artifact. Use `tab_bar_margin_height 19.0` and `window_padding_width 8.0 27.0 27.0 ` in kitty.conf to test an instance of the bug)

Luflosi:
You should rebase your commits onto master instead of merging with master.

jsphnecclia:
Ahh sorry, I think I fixed that just now, but I'm not sure.

jsphnecclia:
Fairly certain I botched this PR/rebase :/ I'm going to try another time from a fresh fork

Luflosi:
The commits from before you last force-pushed are still available on GitHub.

jsphnecclia:
I think the issue I'm having/seeing is that config_data.py has been moved somewhere else?

jsphnecclia:
So its trying to send in the whole config_data.py file. The rest of the commits are as they should be, I think

jsphnecclia:
I have it mainly transfered to the new scheme, however using pt_to_px no longer works from inside state.c, breaking my feature. Is there a way to get around this?

PR #3463, opened 2021-04-09
[macos] - Add "Secure Keyboard Entry" menu item

multi:
Hi there 👋 

I was missing the "Secure Keyboard Entry" functionality, so here you go 🚀 
Mostly based/copied from https://github.com/gnachman/iTerm2/blob/master/sources/iTermSecureKeyboardEntryController.m

Toggled by `Command-Option-S`

image

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -561,6 +561,133 @@ translateKeyToModifierFlag(uint32_t key)
 static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 
+// SecureKeyboardEntryController {{{
+@interface SecureKeyboardEntryController : NSObject
+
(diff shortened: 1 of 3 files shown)


Luflosi:
Could you please explain what this actually does? I genuinely don't know.
I also think adding a few sentences to the documentation to describe this feature won't hurt.
I haven't really looked at the code so I can't comment on that.

kovidgoyal:
While I dont particularly mind merging this PR, I'm afraid I dont
understand what it is supposed to do. From googling it seems like
"secure keyboard entry" prevents keyloggers from intercepting key
events sent by the OS to the terminal. However, terminals in turn write
those key events to a tty device. And any process running as the same
user as the terminal can read from tty devices, so not sure what
security this buys you.

PR #3470, opened 2021-04-13
Property 'is_self' to identify source window

orki:
is_self is independent of having focus. Remote control for identifying current window.


orki:
I must be missing something. With 05fb09f, I don't see the parameter `is_self` in the output of `ls`. `serialize_callback` is not part of the code path used by `tab.list_windows`; that code path goes through `Window.as_dict`. I did consider monkeypatching that one, but didn't think that it would be accepted.

kovidgoyal:
My mistake, now fixed.

PR #3544, opened 2021-04-28
Set default update interval to 0

bheadmaster:
I believe that any kind of "opt-out" mechanism that connects to a remote resource is an anti-feature.

Automatic updates (or update detection) is a useful feature, but it is only a *feature* and users should be aware of its existence if it is turned on.
By making it an "opt-out" mechanism, you are effectively misleading users who assume that terminal emulators do not (by default) have a built-in phone-home mechanism.

Please do not contribute to the normalization of surveillance in GNU/Linux software.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,7 +70,7 @@ class Options(argparse.Namespace):
     extra_logging: List[str] = []
     extra_include_dirs: List[str] = []
     link_time_optimization: bool = 'KITTY_NO_LTO' not in os.environ
-    update_check_interval: float = 24
+    update_check_interval: float = 0
     egl_library: Optional[str] = os.getenv('KITTY_EGL_LIBRARY')
     startup_notification_library: Optional[str] = os.getenv('KITTY_STARTUP_NOTIFICATION_LIBRARY')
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I disagree.

kovidgoyal:
Oh and for the future, if you are going to make such pull requests to other projects, kindly refrain from using the terms surveillance and misleading, they are beyond offensive.

narodnik:
Wait, kitty terminal opens network connections without my knowledge? Any info on this?

castrated:
> Oh and for the future, if you are going to make such pull requests to other projects, kindly refrain from using the terms surveillance and misleading, they are beyond offensive.

how are those terms in the least offensive? sounds like you're trying to use other terms to shift focus from any sort of issues in relation to such.

kovidgoyal:
Any sort of issues!! Wonderful, hey you aren't doing me a favor by using kitty. You dont like the fact that it has functionality to notify you of updates that is on by default, go use something else. And stop wasting my time. And next time spend two minutes to google the issue before posting duplicates and wasting hardworking maintainers time.

PR #3546, opened 2021-04-28
Fix tasks shortcut on macOS

kidonng:
It's Command/Meta+Tab (⌘+Tab), not Option+Tab.

### DIFF
--- a/kitty/config_data.py
+++ b/kitty/config_data.py
@@ -1303,7 +1303,7 @@ both Option keys as Alt, instead.
 '''))
 
 o('macos_hide_from_tasks', False, long_text=_('''
-Hide the kitty window from running tasks (:kbd:`Option+Tab`) on macOS.
+Hide the kitty window from running tasks (:kbd:`⌘+Tab`) on macOS.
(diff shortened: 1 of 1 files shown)


kovidgoyal:
As far as I know on macOS cmd+tab is used to switch between
applications, aka, the task list. cmd+` is used to switch between
windows of a single application. I dont know what option+tab is for,
but on my macbook, it doesnt do anything.

kidonng:
You are right, but why close? The current documentation is wrong.

Luflosi:
Yes, I think @kovidgoyal should take a look at the diff again. It looks good to me.

kovidgoyal:
Ah, right sorry, senior moment.

DuckSoft:
I think the patch is correct.

Luflosi:
This PR was merged in 56fe6480ce8e8b69150d96f54304d09367182af4 but that doesn't show up here because the PR was not reopened first.

kidonng:
That's OK. Thanks everyone!

Issue #1846, opened 2019-07-25
how do I map Cmd+s to send Ctrl+s


MagicDuck:
I can't figure out the key code for Ctrl+s, when I use `sed`, nothing shows up when typing `Ctrl+s`...

kovidgoyal:
You use showkey -a to see the codes for keys, unless you are on macOS in which case you are mostly out of luck. IIRC ctrl+s is 0x13

MagicDuck:
Thanks @kovidgoyal 😀
Yeah, i am on mac. For those looking at this in the future, the manual way to figure it out is:
1. Look up your key combo decimal value in the table at the bottom of this page or any ANSI escape sequence table http://ascii-table.com/ansi-escape-sequences.php . Theres different modifiers for ctrl, alt, etc. For ex, for ctrl+s , find the S row and look at the third colun value, 19.
2. Convert the decimal value to hex. You can do this in google for example by typing something like: 19 in hex. This shows the hex value, 13 in this case.
3. Use \x<hexval> in your bind command in kitty. So in the example, \x13

@kovidgoyal maybe something like this could be added to the docs in absence of an automated way to do it.

Btw, loving kitty, awesome name, awesome terminal 😀

kovidgoyal:
Sure PRs are welcome :) Probably add it to docs/faq.rst

MagicDuck:
For those finding this issue in the future, it looks like http://ascii-table.com/ansi-escape-sequences.php is down now, but you can compile & use this:

https://gist.github.com/rprichard/9b05a488439455e6e41fd73e5bfa8c15

PR #3556, opened 2021-04-29
fixed docs on how to use showkey on macOs

MagicDuck:
The method used to map keys on macOs depended on aa site like http://ascii-table.com/ansi-escape-sequences.php , but that is now down and I couldn't find any other alternatives. On the other hand, I was able to find this  one file implementation of `showkey` that works on macOs: https://gist.github.com/rprichard/9b05a488439455e6e41fd73e5bfa8c15
It would be ideal if it could be part of kitty, but in the absence of that, updating the docs is the next best thing.


MagicDuck:
see original issue here: https://github.com/kovidgoyal/kitty/issues/1846

PR #3637, opened 2021-05-17
Improve undercurl rendering for HiDPI displays

disrupted:
Fixed/Improved rendering of curly underlines on HiDPI (i.e. Retina) displays.

I took the existing algorithm for the anti-aliased line (Xiaolin Wu) to calculate an upper & lower bound and applied a solid fill in between. This way the amplitude & weight of the undercurl is correctly scaled for HiDPI displays. 

Feedback welcome :)

### @1x Before

kitty_undercurl_before@1x

### @1x After (unchanged)

kitty_undercurl_after@1x

### @2x Before

kitty_undercurl_before@2x

- Issue: Curly underline is thinner than normal and appears flatter (less curly)

### @2x After

kitty_undercurl_after@2x

- Change: Undercurl has correct thickness and curly-ness :)

### DIFF
--- a/kitty/fonts/render.py
+++ b/kitty/fonts/render.py
@@ -236,7 +236,8 @@ def add_dline(buf: ctypes.Array, cell_width: int, position: int, thickness: int,
 def add_curl(buf: ctypes.Array, cell_width: int, position: int, thickness: int, cell_height: int) -> None:
     max_x, max_y = cell_width - 1, cell_height - 1
     xfactor = 2.0 * pi / max_x
-    half_height = max(thickness // 2, 1)
+    half_height = thickness
+    thickness -= 1
 
     def add_intensity(x: int, y: int, val: int) -> None:
         y += position
@@ -244,20 +245,22 @@ def add_curl(buf: ctypes.Array, cell_width: int, position: int, thickness: int, 
         idx = cell_width * y + x
         buf[idx] = min(255, buf[idx] + val)
 
-    # Ensure all space at bottom of cell is used
-    if position + half_height < max_y:
-        position += max_y - (position + half_height)
+    # Ensure curve doesn't exceed cell boundary at the bottom
+    position += half_height * 2
     if position + half_height > max_y:
-        position -= position + half_height - max_y
+        position = max_y - half_height
 
     # Use the Wu antialias algorithm to draw the curve
     # cosine waves always have slope <= 1 so are never steep
     for x in range(cell_width):
         y = half_height * cos(x * xfactor)
-        y1, y2 = floor(y), ceil(y)
-        i1 = int(255 * abs(y - y1))
-        add_intensity(x, y1, 255 - i1)
-        add_intensity(x, y2, i1)
+        y1, y2 = floor(y - thickness), ceil(y)
+        i1 = int(255 * abs(y - floor(y)))
+        add_intensity(x, y1, 255 - i1)  # upper bound
+        add_intensity(x, y2, i1)  # lower bound
+        # fill between upper and lower bound
+        for t in range(1, thickness + 1):
+            add_intensity(x, y1 + t, 255)
 
 
 def render_special(


kovidgoyal:
If you are changing the height of the curl by removing the // 2, that would make the curl too thick at normal DPI. You would need to make it conditional on the dpi. Furthermore in this case the height of the normal underline should also be scaled. The best place to do that would be in core_text.m line 58.  Multiply underline thickness by  fg->logical_dpi_y / 72.0

That will cause all underlines to be thickened not just the curl. Although I can find no documentation suggesting that CTFOntGetUnerlineThickness returns a value that is DPI independent and needs to be scaled. 

And then there is the question of underline position, if we are scaling thickness, should we scale positiona s well? Can you check regular underline thickness and position in some cocoa native apps on DPI to try to answer this question.

disrupted:
Hi Kovid! Thank you for your feedback. 

> If you are changing the height of the curl by removing the // 2, that would make the curl too thick at normal DPI. You would need to make it conditional on the dpi. Furthermore in this case the height of the normal underline should also be scaled. The best place to do that would be in core_text.m line 58. Multiply underline thickness by fg->logical_dpi_y / 72.0

I am testing this with 2 monitors, one of which is 4K (HiDPI `@2x`) and the other one 1080p (standard `@1x` scaling factor).
I think there was an issue before, that the height of the curly underline wasn't scaled correctly to `@2x`. It appeared slimmer/less compared to standard `@1x` scaling which can be seen in the before screenshot above. I found that removing the `// 2` solved the issue for me, but perhaps this is not the best solution yet. (should test this for possible side effects and different resolutions, `@3x`?)


The `thickness` parameter which gets passed to `add_curl()` already seems to have the correct value. (1 for 1x dpi, 2 for 2x)

When I changed the underline_thickness to `self->underline_thickness = CTFontGetUnderlineThickness(self->ct_font) * fg->logical_dpi_y / 72.0;`

the thickness parameter was calculated as 3 for `@2x` display scaling.

> That will cause all underlines to be thickened not just the curl. Although I can find no documentation suggesting that CTFOntGetUnerlineThickness returns a value that is DPI independent and needs to be scaled.

good point! But straight underlines already seem to be scaled to the correct thickness when I view them on `@2x`. I assume the `CTFOntGetUnderlineThickness` already handles the scaling then.
 
> And then there is the question of underline position, if we are scaling thickness, should we scale positiona s well? Can you check regular underline thickness and position in some cocoa native apps on DPI to try to answer this question.

I will look into this. I think the current logic of positioning it at the the very bottom of the cell is already good in all cases, but I'll try and compare it to some other apps.

kovidgoyal:
Cool once you have checked out the positioning and settled on a fix,
ping me again and I will take another look. Also please squash your
commits into one, easier for me to review that way.

disrupted:
@kovidgoyal this is ready for review now.

After testing I found that always placing the undercurl at the very bottom of the cell isn't always desired, especially when setting larger cells through `adjust_line_height`, because then there could be a big blank space between text and undercurl.

I changed the positioning logic to be more consistent across those cases.

akinsho:
I'm sorry for posting on a closed PR, I was looking around this repo to see if there was anything related to an issue I was having which seems to be the same as the OP (@disrupted) except I'm on a 16-inch MacBook and don't seem to get the benefit of this PR. My current undercurls in kitty look a lot more like the before photo than the after. They are very thin and not particularly curly, making them less visible/impactful

My kitty screenshot:

image

Result of the PR 👇🏿 
image

If this is better opened as a new issue please let me know and sorry if I should have done that to start. It just seemed like this was the fix for me, but doesn't seem to work. It doesn't seem to be gated behind any specific flag or anything, so presumably using a retina macbook should be enough to get this behaviour?

kovidgoyal:
You are running kitty 0.25.2? Press ctrl+shift+f6 and post the output. Also try running kitty with kitty --config NONE and see if you can reproduce.

akinsho:
Yep I'm running 0.25.2. The output is 
```
kitty 0.25.2 created by Kovid Goyal
Darwin Akins-MBP 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64
ProductName:    macOS ProductVersion:   12.4 BuildVersion:      21F79
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /bin/zsh
Loaded config files:
  /Users/akin/.config/kitty/kitty.conf

active_tab_title_template          {sup.index} ❐ {fmt.fg.red}{bell_symbol}{activity_symbol}{fmt.fg.tab}{title} {fmt.fg._5B6268}{fmt.nobold}({num_windows})
adjust_line_height                 3
allow_remote_control               y
bold_font                          Cartograph CF Bold
bold_italic_font                   Cartograph CF Extra Bold Italic
confirm_os_window_close            2
copy_on_select                     clipboard
cursor_stop_blinking_after         10.0
disable_ligatures                  1
enable_audio_bell                  False
enabled_layouts                    ['tall:bias=55;full_size=1', 'stack', 'fat', 'grid', 'horizontal']
env:
{'PATH': '/opt/homebrew/bin/:/usr/local/bin/:/Applications/kitty.app/Contents/MacOS:/usr/bin:/bin:/usr/sbin:/sbin'}
focus_follows_mouse                True
font_family                        Cartograph CF Light
font_size                          14.2
hide_window_decorations            2
inactive_text_alpha                0.8
italic_font                        Cartograph CF Light Italic
listen_on                          unix:/tmp/kitty
macos_option_as_alt                2
macos_quit_when_last_window_closed True
macos_thicken_font                 0.75
mouse_hide_wait                    15.0
placement_strategy                 top-left
pointer_shape_when_dragging        hand
resize_draw_strategy               1
scrollback_lines                   10000
startup_session                    /Users/akin/Dropbox/kitty/startup.conf
tab_bar_align                      center
tab_bar_edge                       1
```

@kovidgoyal that seems much better with `--config NONE`
image

I tried changing the font before posting, so it wasn't that, although I guess some other setting I have enabled might be the root cause

kovidgoyal:
Should be easy to bisect kitty.conf to find the culprit

akinsho:
@kovidgoyal so I managed to boil it down to the main font family, it seems to not be specific to just my particular font, [Cartograph CF](https://connary.com/cartograph.html) since I tried with a patched Fira Code, I also happened to have [Jet brains mono](https://github.com/JetBrains/JetBrainsMono) installed and that worked fine, that and the default font which I believe is Menlo on MacOS.

It's evidently not, but I had assumed this feature i.e. undercurl thickness would be agnostic of the font (speaking completely out of my depth here)

akinsho:
As a quick note, I thought it might relate to the font weight but moving to a thicker font weight i.e. `Light` → `Demi Bold` didn't change the undercurl behaviour

kovidgoyal:
It's not font related, it will be font metrics related. You can probably
workaround it by using the adjust_cell_width/height settings.

akinsho:
@kovidgoyal did a bit of logging, these are the arguments to `add_curl` and got:
```
Cartograph CF - Position: 36, cell width: 18, cell height: 48, thickness: 1
JetBrains Mono Medium - Position: 34, cell width: 18, cell height: 41, thickenss: 2
```
Seems the cell height is *lower* for Jetbrains, but the thickness is greater. Tried following it through a bit but still not sure how thickness is determined

kovidgoyal:
thickness is a property of the font. fonts specify the underline
thickness they want for underlines.

akinsho:
Hmm, that makes sense, but is somewhat unfortunate. I guess I can ask the font author if they'd be open to changing it, but I wonder if this is something that can be overriden in kitty, since I don't know all the reasons and use cases they might have for choosing that thickness.

Also, this seems like it could happen at any point if I change my font, and they choose some thickness that produces this behaviour, for example if I switched to fira code which also behaves like my current font here. If it was a user option, could this then be something I set on my machine that would work for all fonts, granted I'm using the same screen resolution etc.

kovidgoyal:
It's not something I care to make configurable, patches welcome. But
note that if you are going to do this, I will want to make the baseline,
underline position, strikethrough position, underline thickness and
strikethrough thickness configurable.

akinsho:
That seems like it overlaps quite a lot with https://github.com/kovidgoyal/kitty/pull/4723 if I'm not mistaken?

akinsho:
@kovidgoyal not sure if I will fare better than other people that have attempted this, but I'm pretty keen on this functionality. Just to clarify some comments, you made in that issue, i.e.

>    Change the size by a percentage of existing size or by pts
    Change the underline, strikethrough positions, by font units
    Change the thickness of underline/strikethrough

1. There you refer to changing the size (is this of the font) separate to the `font_size`? Here, you mention the baseline instead? Is this different to `adjust_baseline` which already exists
2. You mention a `modify_font` setting similar to `font_features` there are not many other options similar to this I think. Did you mean for this to be specified as a series of space separated strings i.e. how did you imagine an example of this setting looking `modify_font underline_thickness:3` there seems to be a lot of precedent for just numeric options which seems like this would easily fall under with a common prefix for grouping them like `modify_font_underline_thickness` or something?
3. Is this feature essentially all or nothing or is there a more minimal subset here that would be acceptable, in the interest of reducing scope as much as possible to make this easier to implement (smaller PR etc.)

kovidgoyal:
1. Yes. The idea is that font_size sets the cell size, but since actual text might be rendered by many different fallback fonts, you can adjust the size of these to work well with the size of the main font.
2. modify_font * underline_thickness 3pt (* because this applies to all fonts)
modify_font FontPostScriptName adjust_size -5% 
3. Yes, it's all or nothing, if we are modifying font rendering, then I want a general framework for it otherwise in the future I would just have to change things which is much more effort thatn doing it right from the beginning.

PR #3647, opened 2021-05-21
macOS: move menu bar items for opening and closing OS windows and tabs to new menu

Luflosi:
Most apps on macOS have the menu items for opening and closing tabs and windows in the "File" menu. Since this doesn't really apply to Terminal emulators, both Terminal.app and iTerm2 have them in the "Shell" menu instead.
This change makes the menu bar of kitty more consistent with these applications.
I also changed the case of "New OS window" to "New OS Window" for consistency.

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -374,7 +374,6 @@ cocoa_create_global_menu(void) {
                 keyEquivalent:@""];
     [appMenu addItem:[NSMenuItem separatorItem]];
     MENU_ITEM(appMenu, @"Preferences…", edit_config_file);
-    MENU_ITEM(appMenu, @"New OS Window", new_os_window);
 
     [appMenu addItemWithTitle:[NSString stringWithFormat:@"Hide %@", app_name]
                        action:@selector(hide:)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Hmm not sure this is a good idea. For Terminal.app and iterm it might make sense, as that global menubar is wasted space anyway, but kitty displays window titles there, adding a new menu just for a  handful of entries doesnt seem like a economical usage of space. However, I am willing to be convinced otherwise.

I have made the capitalization change myself sincethat should be done regardless of the menu issue.

Luflosi:
I don't personally use the feature that shows the window title in the menu bar but I think there is plenty of space left for a long window title, even on my relatively small 1440x900 display. According to https://sw.kovidgoyal.net/kitty/conf.html#opt-kitty.macos_show_window_title_in, this feature "show[s] the title of the currently active window in the macOS menu-bar, making use of otherwise wasted space". I think organising the menu items into familiar groups should have a higher priority than leaving more space for the window title.
I do plan on adding a few more menu items for common actions as I think this is a great way for new users to discover them and their keyboard shortcuts without the need to look at the manual. I'll add menu bar items for `new_window` and `close_window` first. So hopefully, there will be a few more than the four menu items in "Shell" soon.
I rebased so you don't have to bother with the merge conflict in case you do decide to merge.

PR #3653, opened 2021-05-23
macOS: add menu items for new_window and close_window

Luflosi:
As promised in #3647.
Please double-check that I made the changes in exactly the right places, not in too few places and not in too many. I basically just searched the code base for an already existing shortcut and then replicated the appropriate looking code for the new shortcuts.
Should I sort the shortcuts in the code into a particular order or just append to the end like I did now?
I successfully tested this change.

### DIFF
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -997,6 +997,8 @@ process_global_state(void *data) {
             if (cocoa_pending_actions & NEXT_TAB) { call_boss(next_tab, NULL); }
             if (cocoa_pending_actions & PREVIOUS_TAB) { call_boss(previous_tab, NULL); }
             if (cocoa_pending_actions & DETACH_TAB) { call_boss(detach_tab, NULL); }
+            if (cocoa_pending_actions & NEW_WINDOW) { call_boss(new_window, NULL); }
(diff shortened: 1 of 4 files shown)


kovidgoyal:
The changes look correct but i suggest adding a shortcut for reset terminal as well.

Luflosi:
I don't use reset_terminal. What exact action should occur? `clear_terminal reset active`?

kovidgoyal:
yes.

Luflosi:
I'll try to do that in a future PR as it is not quite as simple as the other two options. I got everything working except for binding to the correct keyboard shortcut. If I have
```conf
#:     # Reset the terminal
map kitty_mod+f9 clear_terminal reset active
#:     # Clear the terminal screen by erasing all contents
map kitty_mod+f10 clear_terminal clear active
#:     # Clear the terminal scrollback by erasing it
map kitty_mod+f11 clear_terminal scrollback active
#:     # Scroll the contents of the screen into the scrollback
map kitty_mod+f12 clear_terminal scroll active
```
in my kitty.conf, the keyboard shortcut will be <kbd>cmd</kbd>+<kbd>F12</kbd> instead of <kbd>cmd</kbd>+<kbd>F9</kbd>. I'll try to fix that in the near future and then make a new PR just for that.

PR #3674, opened 2021-05-30
Downgrade libstartup-notification-1.so non-existence to warning

ctrlcctrlv:
I'm not sure how common this library actually is. On my system, an Arch
Linux system with 2,170 packages installed, including KDE (Plasma), I
didn't have it.

Googling around shows me that its purpose is just to show a loading
cursor when an application is loading. That sounds pretty Windows 98 to
me. I see how it could be useful, but certainly it's not an error not to
have this around.

### DIFF
--- a/kitty/desktop.c
+++ b/kitty/desktop.c
@@ -52,8 +52,8 @@ init_x11_startup_notification(PyObject UNUSED *self, PyObject *args) {
             if (libsn_handle) break;
         }
         if (libsn_handle == NULL) {
-            PyErr_Format(PyExc_OSError, "Failed to load %s with error: %s", libnames[0], dlerror());
-            return NULL;
+            log_error("Failed to load %s with error: %s", libnames[0], dlerror());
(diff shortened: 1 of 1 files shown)


kovidgoyal:
No, its needed on X11 to allow running kitty --single-instance on workspace X to show the new window on the same workspace, when other kitty windows are on another workspace. Without it, the new window will whow up in the same workspace as the original.

ctrlcctrlv:
So? This still logs a warning. I can make the warning more descriptive of what's missing? I just don't think it should throw up a traceback.

kovidgoyal:
In that case suppress it in utils.py where the function is called and log the error message using the log_error function.

ctrlcctrlv:
OK, force pushed.

ctrlcctrlv:
(Changes to the base branch won't appear here unless it's reopened.)

kovidgoyal:
merged

ctrlcctrlv:
Thanks :-)

PR #3684, opened 2021-06-02
Add padding options for around the tab bar - Transfer from config_data.py

jsphnecclia:
See [3436](https://github.com/kovidgoyal/kitty/pull/3436)

Tab bar padding no longer functions after moving to the new python options schema

pt_to_px may have something to do with it, as replacing the call to pt_to_px  on line 478 with a constant number causes that number of pixels padding between the top of the window and the tab bar. (also, `printf("%ld", pt_to_px(OPT(tab_bar_margin_height), os_window->id);` does not print anything

EDIT: Not pt_to_px, but the problem lies in OPT(tab_bar_margin_height)

### DIFF
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -688,6 +688,7 @@ opt('window_margin_width', '0',
 The window margin (in pts) (blank area outside the border). A single value sets
 all four sides. Two values set the vertical and horizontal sides. Three values
 set top, horizontal and bottom. Four values set top, right, bottom and left.
+For margins with tab bar, see tab_bar_margin_height.
 '''
     )
 
@@ -709,6 +710,7 @@ The window padding (in pts) (blank area between the text and the window border).
 A single value sets all four sides. Two values set the vertical and horizontal
 sides. Three values set top, horizontal and bottom. Four values set top, right,
 bottom and left.
+For padding with tab bar, see tab_bar_margin_height.
 '''
     )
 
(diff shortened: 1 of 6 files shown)


jsphnecclia:
For some reason, OPT(tab_bar_margin_height) is always 0.0, and OPT(retain_tab_bar_margin_height) is always false, _even_ when given default values of, say, 19.0 and 'yes' respectively in the options/definition.py file.

kovidgoyal:
To add new options, the procedure now is, add them to options/definition.py, then run ./gen-config.py which will update the auto-generated do not edit files.

jsphnecclia:
Yes, that's what I did, and those automatically generated files I added in the last commit were the result

kovidgoyal:
Can you squash your commits into one please, and also is this ready for review? As I recall, you had some issues with drawing blank rectangles.

jsphnecclia:
How would I go about squashing the commits into one?

This is not ready for review - I had it working on the old config_data.py system (I solved my issue with the blank rectangles)

so I transferred it into the new system but for some reason, OPT(tab_bar_margin_height) is always 0.0, and OPT(retain_tab_bar_margin_height) is always false, even when given default values of, say, 19.0 and 'yes' respectively in the options/definition.py file and other autogenerated files.

this breaks the functionality of the feature completely

kovidgoyal:
You use git rebase to squash commits. And remember to add code to state.c in set_options() to copy the option values.

jsphnecclia:
That was it! I hadnt set the set_options code. Is there a reason there is no "tab_bar_margin_width" in the set_options?
Commiting and rebasing

jsphnecclia:
Is that right for the commit squashing?

kovidgoyal:
Doesnt look like it, maybe will be easier for you to just create a new PR with a single commit containing all the changes.

jsphnecclia:
I'll try that method!

PR #3788, opened 2021-07-02
OS-window resizing

williamspatrick:


williamspatrick:
See #3785 for original PR and design discussion.

Intended work tasks / progress:

- [ ] CSI escape codes
    - [X] Add option for escape codes.
    - [X] Basic CSI 4/8 support for resizing single window (OS window).
    - [ ] CSI 4/8 support for multiple windows (resize + reflow layout).
    - [ ] Ensure minimum size windows are handled for CSI 4/8.
    - [ ] CSI 9/10 support for full screen and maximize.
    - [ ] CSI 88 support for OS window-only resize (using CSI 4/8 support).
    - [ ] CSI 89 support for kitty window-only resize (using CSI 4/8 support).
- [ ] `kitty +kitten @ resize-os-window` implementation

kovidgoyal:
You also need to fix the high dpi issue on macOS, I think.

kovidgoyal:
I have implemented the remote control command for this. That should give you all the lower level pieces to implement the rest of it easily.

TheArcaneBrony:
Concern: how would this be handled in tiling window managers? There *is* a way to disable this, right?

ZanzyTHEbar:
this seems stale - where is this at?

williamspatrick:
I am no longer working on this.  I've moved to use sway and do not care about manual window resizing.

PR #3829, opened 2021-07-13
linux: install more icons

Artturin:


kovidgoyal:
Why is this needed? Is there some desktop env that uses larger than 256x256 logos? Or one that does not scale down the logo's well?

Also, logo/kitty.png needs to exist, your patch no longer renders it.

Artturin:
i was packaging swayr for nixos and noticed that it wasn't finding the kitty icon because it didn't look in 256x256 by default so i thought its easier to patch here than in nixos

setup.py
```diff
@@ -908,9 +908,13 @@ def create_linux_bundle_gunk(ddir: str, libdir_name: str) -> None:
...
-    icdir = os.path.join(ddir, 'share', 'icons', 'hicolor', '256x256', 'apps')
-    safe_makedirs(icdir)
-    shutil.copy2('logo/kitty.png', icdir)
+    for sz in (16, 32, 64, 128, 256, 512, 1024):
+        icdir = os.path.join(ddir, 'share', 'icons', 'hicolor', f'{sz}x{sz}', 'apps')
+        safe_makedirs(icdir)
+        shutil.copy2(f'logo/kitty-{sz}.png', f'{icdir}/kitty.png')
+    icscaldir = os.path.join(ddir, 'share', 'icons', 'hicolor', 'scalable', 'apps')
+    safe_makedirs(icscaldir)
+    shutil.copy2('logo/kitty.svg', icscaldir)
```
Artturin, line 917: if the other changes are too much then just this would suffice

kovidgoyal: Yes, I would prefer just installing the svg icon rather than having so many mostly useless PNG ones. Guess I am a little OCD :)

PR #3836, opened 2021-07-16
Directories should have MIME type inode/directory.

william-gooch:
Currently, one can't assign a custom action to folders in hyperlinks because they aren't associated a MIME type. This is because python's `mimetypes` library doesn't include that mime type. You also can't add this in the `.config/kitty/mime.types` file because folders have no extension.

This way any directories automatically get the right MIME type.

### DIFF
--- a/kitty/guess_mime_type.py
+++ b/kitty/guess_mime_type.py
@@ -28,6 +28,8 @@ def is_rc_file(path: str) -> bool:
     name = os.path.basename(path)
     return '.' not in name and name.endswith('rc')
 
+def is_folder(path: str) -> bool:
+    return os.path.isdir(path)
 
 def initialize_mime_database() -> None:
     if hasattr(initialize_mime_database, 'inited'):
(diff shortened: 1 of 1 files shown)


Luflosi:
Since `os.path.isdir()` and `inode/directory` both call it a directory, I suggest renaming `is_folder()` to `is_directory()`.

kovidgoyal:
Couple of comments:

1) I am not comfortable with having guess_type perform filesystem I/O in general. Add a parameter named check_for_dirs defaulting to False and when guess_type is called for open actions use it.

2) Add some documentation for this mime/type to docs/open_actions.rst

3) Add a changelog entry for this to docs/changelog.rst

kovidgoyal:
Actually maybe name the parameter allow_filesystem_access rather than check_for_dirs

Luflosi:
You can fix the CI failure by having two empty lines above and below the new function.

PR #3851, opened 2021-07-21
remove update checking

binex-dsk:
unnecessary botnet, i prefer my terminal to not connect to the internet


Luflosi:
As discussed in the issue tracker before, the update check will not be removed. It was added because many people asked for a way to be notified of new versions. If you don't like it, disable it in your config file or install kitty via your favorite package manager as most package managers that build kitty themselves disable it with a compile time flag as recommended in the build instructions.

kovidgoyal:
Dont waste my time with duplicate issues.

PR #3852, opened 2021-07-22
Added fcitx5 implementation

alkim0:
Hi. I added an implementation for fcitx5. The current implementation using the ibus wrapper is broken for Korean, so I figured I'd add a native implementation for fcitx5. **NOTE: This only works for fcitx5. fcitx4 and previous versions will not work.**

 Run with:
```
GLFW_IM_MODULE=fcitx5 kitty
```

Unfortunately, I had to touch the core kitty code a little. Let me know if you think any of these changes are problematic. I'll try to make another pull request with whatever suggestions you may have. 

## `glfw`
The `glfw` folder is pretty much what you'd expect. An almost one-for-one copy of `ibus_glfw.[ch]` to `fcitx5_glfw.[ch]`. I've made some changes here and there to try to make a more generalized IME interface (i.e., `_GFLWIMEModule` and `_GLFWIMEData`), but those changes are minor and don't alter any of the glfw API calls.

## `kitty`
Before listing the changes I've made, let me give some background information. For many languages, the typical IME workflow looks like this:
1. Type some characters.
2. The pre-edit text is updated with the typed characters.
3. Press tab/enter to commit.
4. The IME gives us the commit text, which is flushed to the application, and the pre-edit text is cleared.

I think the current implementation was built this model in mind, and it works well for languages that follow that model (e.g., Japanese). Korean doesn't follow this model. In Korean, there is no separate commit stage. Instead, it is expected that as you type (regular non-tab/enter) letters, small groups of letters are committed along the way.

To take a simple example, let us assume that I want to type the word 가지. This word can be divided into the letters 'ㄱ', 'ㅏ', 'ㅈ', 'ㅣ'. After the third letter, the pre-edit text should look like '갖' with nothing committed so far. After the 4th letter, the first two letters ('ㄱ' and 'ㅏ') are committed in the form of '가' and the pre-edit text should show the third and fourth letters in the form of '지'.

The current implementation seems to interact with the IME in a way that expects an explicit tab/enter to flush the commit phase, and it needs some work to handle the cases without it.

### Changes
When `screen_draw` from `screen.c` is called from `parse_input` in the global process loop in `child-monitor.c`, it seems to ignore the overlay text completely. However, as we have shown with the example above, a commit may cause text to be drawn with `screen_draw` while the overlay line is still active. This results in text that looks like `blah<overlay>c`, which is not what we want. Instead:
1. The committed text should go where the overlay text starts.
2. The overlay text should appear after the committed text.

To achieve this, we first deactivate the overlay line (if active) in `screen_draw` in `screen.c` before writing any text. Next, we need a way to signal that the overlay line should be rewritten to the screen. The overlay line text doesn't seem to be stored anywhere, so I created a buffer for it in the `OverlayLine` struct. Then in `do_parse` of `child-monitor.c`, if any input has been read while the overlay line is active, that means we have overwritten the overlay line, so we order another `screen_draw_overlay_text`.

In addition, to ensure the IME is updated of the cursor position after a resize (or some other event which modifies the screen), I added a `update_ime_position` to the end of `render` in `child-monitor.c`.


kovidgoyal:
I'm somewhat confused by this. Some questions:

1) committing is something driven by the IME not the application. If the Korean IME wants to commit at any point it is free to do so, kitty doesnt care about pressing enter or tab or whatever. 
Therefore, I dont quite understand why this is needed at all. What I believe needs to be done is to figure out why fcitx is not committing as you want it to, and adjust the communication accordingly.
If you run kitty --debug input you should get a nice log of the events from the IME and keypresses

2) DBUS interaction is X11 oly. On Wayland kitty uses the text-input protocol. What happens there? As I recall fcitx doesnt yet support wayland

3) There is no need to store the text for overlay line separately, it will already be in OverlayLine->cpu_cells

alkim0:
### 1
This is what happens when I try to run the current implementation with the `GLFW_IM_MODULE=ibus` wrapper.
 
https://user-images.githubusercontent.com/9775404/126601286-e4416119-3a79-41b4-8aa4-201d630d4852.mp4

Here, I am pressing the letters 'ㄱ' and 'ㅏ' over and over again. Despite the IME signalling the commit multiple times, kitty keeps overwriting it, and the cursor position never progresses.

What I expect instead is this:

https://user-images.githubusercontent.com/9775404/126601574-bdc014b3-2988-417c-924d-8d3cdf33e574.mp4

This clip shows the patched implementation (with `GLFW_IM_MODULE=fcitx5`) showing the correct behavior.

I've run kitty with --debug-keyboard, and fcitx is sending the correct commit/pre-edit messages, but upon receiving, them kitty is doing something weird.

I think what may be happening is that on commit, `screen_draw` writes to the cursor position, which is advanced by the length of the overlay text, resulting in text being written after the overlay text like "blah\<overlay\>c". Then, when it receives a pre-edit message, in the first line of `screen_draw_overlay_text`, the overlay line is still active, and `deactivate_overlay_line` is called. This results in the cursor position being reset and the cursor never advances as you can see in the first clip.

The changes in the `kitty` folder are to fix this problem.

Regarding the native fcitx5 implementation. I saw that a few issues in the past had people asking for it, and I figured it would be better having a native implementation than using a wrapper. 

### 2
I've only tested this on X11. I imagine on Wayland, people would continue to use Wayland's text-input protocol. The fcitx IME is not forced. You have to enable it with the environment variable with `GLFW_IM_MODULE=fcitx5` just like you did before with `GLFW_IM_MODULE=ibus`. Also, if dbus doesn't work on wayland, then the fcitx using the ibus wrapper (which uses dbus) would not would not work on wayland either, so I don't think this is a regression.

### 3
Yes, but I couldn't find an interfacing function to dump the contents of the cpu cells on to the screen. It seemed messy to leave logic regarding the IME in `screen.c`, and there were no functions in `screen.h` which took in and wrote cpu cells to the screen. Rather than creating another interfacing function, I figured it would be best to just add the buffer and reuse the existing interface. Unless, I'm missing something? You didn't mean `screen_handle_graphics_command` right?

alkim0:
On a side note, the patched implementation which addresses the problem described in 1 in the previous comment made `GLFW_IM_MODULE=ibus` mostly work, but there's still a weird issue when pressing space. Since I already had a fcitx5 implementation that was working, I didn't really track it down...

kovidgoyal:
I think we should nail down the issue with the ibus interaction first.
Fix that using the wrapper and then we can decide whether a dedicated fcitx5 
module is actually required separately.

So addressing the korean ime issue:

An IME commit is handled in keys.c (line 137)
What a commit does is 

1) send the commit text reported by the IME to the terminal program
(shell, or editor or whatever)
2) cancel the overlay line

and that's all. It does not draw anything. The drawing happens in
screen_draw() which is called when the terminal program sends the text
from (1) above back to kitty. 

I think the issue is caused by the fact that for the Korean IME, the IME
instantly re-sends a pre-edit text event immediately after the commit.
This event thus arrives before the text that was sent to the terminal
program arrives back. As a result screen_draw is called while the new
overlay line is active, causing the issue. 

You basic idea to fix it is correct, but needs some refinement.

1) We absolutely must not call update ime state on every render,
this is a huge overhead.
  - instead you should really only need to call it from inside
    screen_draw and then only if an overlay line is currently active.
  - the function to update ime state already exists in keys.c make it
    non-static and re-use it in screen_draw (you will possibly

2) Dont store a second copy of the text in OverlayLine, instead use
cell_as_unicode to convert the cpu_cells in overlay_line into text.

kovidgoyal:
> On a side note, the patched implementation which addresses the problem described in 1 in the previous comment made `GLFW_IM_MODULE=ibus` mostly work, but there's still a weird issue when pressing space. Since I already had a fcitx5 implementation that was working, I didn't really track it down...

Regressing ibus support is not acceptable, what was the issue?

aca:
Finally I can use kitty and ditch alacritty. Thanks @alkim0.

kovidgoyal:
The ibus + Korean part is taken care of by #4078  If you still feel the need for adding dbus support for fcitx5 open a separate PR for it.

PR #3881, opened 2021-07-29
Fix shell.nix error with stdenv.lib.getLib

egnor:
NOTE - I know very very little about nix. But, when I tried to do `nix shell`, I got an error that `stdenv.lib.getLib` did not exist. It _seems_ like this is the right fix (and it runs without errors!) but please verify.


kovidgoyal:
@Luflosi this is your baby, please check this PR is OK and I will merge.

Luflosi:
Yep, this looks good.
`stdenv.lib` has been a deprecated alias for `lib` for a while now.

PR #3882, opened 2021-07-29
clarify the text for the "editor" config variable slightly

egnor:
The previous wording about "system-wide" slightly confused a friend of mine who's starting with kitty, so I made the description a bit more precise (to my eyes). See what you think.

### DIFF
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -2417,10 +2417,12 @@ starts in interactive mode and reads its startup rc files.
 
 opt('editor', '.',
     long_text='''
-The console editor to use when editing the kitty config file or similar tasks. A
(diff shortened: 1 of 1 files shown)


kovidgoyal:
While the modified text reads fine, I should note that currently, if neither EDITOR nor VISUAL is set, kitty actually runs your shell to read its env. See get_editor_from_env_vars and read_shell_environment. So you actually need to set editor only if running:

```
your-shell -l -c env
```

also doesnt output EDITOR.

egnor:
Eep, this broke the doc build -- I was just fixing that!

kovidgoyal:
I fixed it.

PR #3883, opened 2021-07-29
Updated description for the "editor" config var

egnor:
Conveys a more accurate understanding of kitty's default behavior, thanks to Kovid's comments.


egnor:
(I have _no idea_ what the ubuntu-latest failure is all about....)

kovidgoyal:
It probably happened because I was in the middle of uploading the
updated binaries with the changed ncurses

PR #3956, opened 2021-08-19
Broadcast Ctrl-C into windows

Felixoid:
It's impossible currently but very useful in practice. Here's my attempt to implement broadcasting of ^C.


Felixoid:
Thank you, it's indeed a more elegant solution. But what about https://github.com/kovidgoyal/kitty/pull/3956/files#diff-a05fc3c87aba9d7252443f0a41729338833cb97d1b49440ef49fe831c7af03e2R28, isn't it a typo?

kovidgoyal:
It was fixed in the previous commit.

PR #3995, opened 2021-09-06
Add macos_hide_titlebar to option_names

fladson:
As pointed out in https://github.com/fladson/vim-kitty/pull/12, the `option_names` list is missing the `macos_hide_titlebar` option.

`vim-kitty` is using that for generating the syntax file, just for context.

The `no_op` action is also not being returned by `kitty.actions import get_all_actions`, can you point me out the direction to also add it?

### DIFF
--- a/kitty/options/types.py
+++ b/kitty/options/types.py
@@ -367,6 +367,7 @@ option_names = (  # {{{
  'listen_on',
  'macos_custom_beam_cursor',
  'macos_hide_from_tasks',
+ 'macos_hide_titlebar',
  'macos_option_as_alt',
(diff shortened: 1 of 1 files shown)


kovidgoyal:
macos_hide_titlebar is deprecated. It is an alias for
hide_window_decorations. As such it will be present in options.types
(which is auto-generated). And you should probably not syntax highlight
it anyway, or maybe syntax highlight it as an error.

You can get the list of deprecated names as

from kitty.options.definition import definition

definition.deprecations.values()


And I have added no_op

fladson:
Thank you for the clarification.

PR #4059, opened 2021-09-24
ibus's dbus name owner changed detection

unisgn:
fcitx5 works fine with kitty, except that when you restart fcitx5, kitty stop working with fcitx5, because the dbus_connection_get_is_connected(ibus->conn) in #check_connection is always true when you use fcitx5.
this patch add a dbus name owner changed detection to detect the online status of ibus, in case the ibus backend uses the session bus, like fcitx5.


unisgn:
@wengxt  有空请帮忙审查一下这个提交呢。之前在 fcitx5 上提交了 [issue#272，](https://github.com/fcitx/fcitx5/issues/272)，这个提交就是为了解决那个问题的，帮忙看看有没有问题呢。

PR #4078, opened 2021-09-29
ibus: Add support for ForwardKeyEvent signal

pjm0616:
The ForwardKeyEvent ibus signal is used to maintain ordering between CommitText events and non-IME key events.
The ordering is important for Korean IMEs, where it automatically commits text whenever a character is fully composed. This is different from Chinese/Japanese IMEs, where user manually presses a key to commit each word.

Without this patch, kitty can't handle non-IME key events when used with ibus-hangul, which utilizes ForwardKeyEvent signal.

Without ForwardKeyEvent:
```
key | sequence of events from application's view
r   | UpdatePreeditText 'ㄱ'
k   | UpdatePreeditText '가'
1   | (receives the reply to ProcessKeyEvent call: "native_key: 0x31 release: 0 handled: 0")
    | -> UpdatePreeditText ''
    | -> CommitText '가'
```
Application receives "1가", which is incorrect.

With ForwardKeyEvent:
```
key | sequence of events from application's view
r   | UpdatePreeditText 'ㄱ'
k   | UpdatePreeditText '가'
1   | (receives the reply to ProcessKeyEvent call: "native_key: 0x31 release: 0 handled: 1", and kitty discards the event)
    | -> UpdatePreeditText ''
    | -> CommitText '가'
    | -> ForwardKeyEvent keysym=0x31("1")
```
Application receives "가1", which is correct.

Relevant ibus-hangul github issue to implement ForwardKeyEvent: https://github.com/libhangul/ibus-hangul/issues/42
Relevant changes in Qt to handle ForwardKeyEvent: https://codereview.qt-project.org/c/qt/qtbase/+/212179, https://codereview.qt-project.org/c/qt/qtbase/+/255587

### DIFF
--- a/glfw/ibus_glfw.c
+++ b/glfw/ibus_glfw.c
@@ -49,6 +49,70 @@ enum Capabilities {
     IBUS_CAP_SURROUNDING_TEXT   = 1 << 5
 };
 
+typedef enum
+{
+    IBUS_SHIFT_MASK    = 1 << 0,
+    IBUS_LOCK_MASK     = 1 << 1,
+    IBUS_CONTROL_MASK  = 1 << 2,
+    IBUS_MOD1_MASK     = 1 << 3,
+    IBUS_MOD2_MASK     = 1 << 4,
+    IBUS_MOD3_MASK     = 1 << 5,
+    IBUS_MOD4_MASK     = 1 << 6,
+    IBUS_MOD5_MASK     = 1 << 7,
+    IBUS_BUTTON1_MASK  = 1 << 8,
+    IBUS_BUTTON2_MASK  = 1 << 9,
+    IBUS_BUTTON3_MASK  = 1 << 10,
+    IBUS_BUTTON4_MASK  = 1 << 11,
+    IBUS_BUTTON5_MASK  = 1 << 12,
+
+    /* The next few modifiers are used by XKB, so we skip to the end.
+     * Bits 15 - 23 are currently unused. Bit 29 is used internally.
+     */
+
+    /* ibus mask */
+    IBUS_HANDLED_MASK  = 1 << 24,
+    IBUS_FORWARD_MASK  = 1 << 25,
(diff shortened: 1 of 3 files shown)


pjm0616:
@kovidgoyal

Thanks for the quick response!

I've just spotted the changelog you've just committed that adds "Linux: Add support for Korean IME with ibus": 3728d2cfd9b2186f2fabb4b870528043466afb1a
This patch was my first attempt to fix multiple issues regarding Korean IME in kitty, and there are still more issues with Korean input that makes kitty practically unusable for Korean users. I think we should not include Korean support in the changelog yet, because of those issues below.

This is the list of issues with Korean inputs I've found so far:
1) [ibus] user cannot type numbers&punctuations when ibus-hangul is in use - this patch addresses this issue.
2) [ibus,macos] Issues with cursor positions and preedit overlay rendering, as preedit update event is processed first before the text sent to the terminal arrives back. Exactly this comment: https://github.com/kovidgoyal/kitty/pull/3852#issuecomment-884738697
This issue is also present in macos.
3) [ibus,macos] When preedit text is active, mouse clicks(likely to change cursor position) needs to reset ibus context(for ibus) or commit the preedit text(for macos). This is the expected behavior for Korean inputs.
For ibus, when we send Reset command, ibus commits or clears the preedit text as decided by the input method. MacOS always commits the preedit text when cursor position changes by mouse clicks regardless of input method.
Currently when this happens in kitty, the text in preedit text jumps to the new cursor position. The ordering is also important here: 1) preedit text commit to terminal, and then 2) send mouse event to terminal. If mouse event is sent first, previously committed text will incorrectly appear in the new cursor position. 
4) [macos] First keypress following InputSource change to Korean is always directly sent to the terminal, even if it needs to be in composing state.
Pressing "rk" right after changing input source from English to Korean triggers the issue: It sends "ㄱ"(for r) and only "ㅏ"(k) is left in the preedit text. Expected behavior is "가" in preedit text.

kovidgoyal:
OK removed.

PR #4098, opened 2021-10-05
Properly evaluate /etc/issue placeholders

mattfbacon:
Currently, in the debug_config page, the issue file looks really ugly because escape codes are not evaluated properly. For example, my issue looks like this:

Screenshot_2021-10-04_20-51-15

The double backslashes are particularly bothersome.

Instead, since it's not that complicated to evaluate an issue file, just do that.

Note: this PR introduces another C method that is interfaced with from Python. I would appreciate if you took a look at that and made sure all was well with reference counting, GIL, and that kind of thing as I don't have much experience there.

### DIFF
--- a/kitty/debug_config.py
+++ b/kitty/debug_config.py
@@ -152,6 +152,86 @@ def compare_opts(opts: KittyOpts, print: Callable) -> None:
         print('\n\t'.join(sorted(colors)))
 
 
+is_linux = sys.platform in ('linux', 'linux2')
+
+
+if is_linux:
+    import socket
+    import time
+    import re
+    import termios
+    from .fast_data_types import num_users
+    from typing import IO
+
+
+    class IssueData:
+        uname: os.uname_result
+        hostname: str
+        formatted_date: str
+        formatted_time: str
+        tty_name: str
+        baud_rate: int
+        num_users: int
+        def __init__(self):
+            self.uname = os.uname()
+            self.hostname = socket.gethostname()
+            _time = time.localtime()
+            self.formatted_time = time.strftime('%a %b %d %Y', _time)
+            self.formatted_date = time.strftime('%H:%M:%S', _time)
+            try:
+                self.tty_name = format_tty_name(os.ttyname(sys.stdin.fileno()))
+            except OSError:
+                self.tty_name = '(none)'
+            self.baud_rate = termios.tcgetattr(sys.stdin.fileno())[5]
+            self.num_users = num_users()
+
+
+    # https://kernel.googlesource.com/pub/scm/utils/util-linux/util-linux/+/v2.7.1/login-utils/agetty.c#790
+    issue_mappings = {
+        # ctx = IssueData
+        's': lambda ctx: ctx.uname.sysname,
+        'n': lambda ctx: ctx.uname.nodename,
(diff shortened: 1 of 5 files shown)


mattfbacon:
Ah, BTW, here's what it looks like after the changes:

Screenshot_2021-10-04_20-54-06

kitty/debug_config.py
```diff
@@ -152,6 +152,86 @@ def compare_opts(opts: KittyOpts, print: Callable) -> None:
...
+
+
+    def format_tty_name(raw: str) -> str:
+        return re.sub(r'^/dev/([^/]+)/([^/]+)$', r'\1\2', raw)
+
+
+    def print_issue(issue_file: IO[str], print_fn) -> None:
+        last_char = None
+        issue_data = IssueData()
+        while this_char := issue_file.read(1):
```
kovidgoyal, line 222: kitty currently supports python 3.6 so no walrus operator

kitty/debug_config.py
```diff
@@ -161,9 +241,9 @@ def debug_config(opts: KittyOpts) -> str:
     if is_macos:
         import subprocess
         p(' '.join(subprocess.check_output(['sw_vers']).decode('utf-8').splitlines()).strip())
-    if os.path.exists('/etc/issue'):
+    if is_linux and os.path.exists('/etc/issue'):
```
kovidgoyal, line 244: /etc/issue might exist on other operating systems, so there should be a default print_issue() implementation.

kitty/utmp.c
```diff
@@ -0,0 +1,50 @@
+#if defined(__unix__)
+
+#define PY_SSIZE_T_CLEAN
```
kovidgoyal, line 3: Just use #include "data-types.h" instead of these three lines.

kovidgoyal:
Also do we actually need C code for this? And if we do we should use getutxent which is POSIX and provided by glibc at least on linux. This can be laoded dynamically using ctypes without needing C code, see for example: https://gist.github.com/JnyJny/5e5db87b70917b859d56515149b74ef0

mattfbacon:
Are you going to make those changes? Since you merged the PR?

mattfbacon:
> Also do we actually need C code for this? ... This can be laoded dynamically using ctypes without needing C code, see for example: [gist.github.com/JnyJny/5e5db87b70917b859d56515149b74ef0](https://gist.github.com/JnyJny/5e5db87b70917b859d56515149b74ef0)

To me that code looks more complicated. Especially considering you have to redefine `struct utmp`.

> And if we do we should use getutxent which is POSIX and provided by glibc at least on linux. 

That's fair. I used `getutent` because it was what the agetty source code that I was referencing used, but I guess it makes sense to use utmpx.

FYI, there is a Python module for interacting with UTMP, but it didn't seem worth it just to get the number of user sessions.

kovidgoyal:
It's already merged and I have made what changes I felt were needed. Let
me know if you see any problems. I kept the C code, but dropped the
re-entrant branch since there is no other code in kitty that uses utmp
at a let alne from a different thread, so re-entrancy is not required.

mattfbacon:
Sounds good. I tend to try to avoid the "black magic" non-reentrant methods that use secret static storage, but in this case it's a lot simpler, so I agree with that.

PR #4181, opened 2021-11-02
Add a separate color option for visuall bell

page-down:
Provide independent color configuration option for visual bell, instead of only reusing the highlighted selection background color.

Because visuall bell renders in full window, users can choose to use less glaring dark colors.

It also allows theme customization and remote control.

Here is a custom alert bell effect with alternating red and blue colors.

```shell
kitty --config=NONE -o 'visual_bell_duration=0.3' -o allow_remote_control=yes sh -c 'for i in {1..3}; do kitty @ set-colors visual_bell_color=#ff0000; printf "\a"; sleep 0.3; kitty @ set-colors visual_bell_color=#0000ff; printf "\a"; sleep 0.6; done; kitty @ set-colors visual_bell_color=none; exec sh'
```

In addition, I added `# generated by gen-config.py DO NOT edit` to kitty/rc/set_colors.py `nullable_colors`.

Please check if it is appropriate and if there are any omissions, thank you.


kovidgoyal:
Looks generally fine. Only concern is that is OSC 20 actually meant for this? I can find no reference for this. For instance the xterm ctlseqs page only mentions osc 19 and then 22.

If there is some reference please let me know otherwise leave out the bits that set these colors in tui/operations.py and window.py as its possible OSC 20 is used for something else.

kovidgoyal:
Also i made the list of color keys in collection.py generated by gen-config.py as well.

page-down:
> Looks generally fine. Only concern is that is OSC 20 actually meant for this? I can find no reference for this. ...
> If there is some reference please let me know otherwise leave out the bits that set these colors in tui/operations.py and window.py as its possible OSC 20 is used for something else.

As far as I know OSC 20 is not used, I removed the relevant code. It looks like there is no escape sequence for visual bell color at the moment.

> Also i made the list of color keys in collection.py generated by gen-config.py as well.

Thanks, I have updated it according to this.

kovidgoyal:
There is one remaining issue. The selection bg color can itself be set
to none. So you cant use it as a fallback. You need to have a fallback
chain bell color -> selection bg -> default fg

page-down:
Fixed, now falls back to default foreground color when selection background color is none. Please check if it is appropriate.

PR #4192, opened 2021-11-05
Fix macOS fallback configuration folder location and menu minor tweaks

page-down:
Under macOS, `~/Library/Preferences` only holds property list files, all application custom configurations and data are generally located in `~/Library/Application Support`.

In addition, I capitalized the menu item and slightly adjusted the position of the Service menu item according to the convention.

Make them consistent with other menu items.

### DIFF
--- a/kitty/cli.py
+++ b/kitty/cli.py
@@ -46,7 +46,7 @@ priority) than any user config files. It can be used to specify system-wide
 defaults for all users.
 '''.replace(
     '{macos_confpath}',
-    (':file:`~/Library/Preferences/{appname}/{conf_name}.conf`,' if is_macos else ''), 1
+    (':file:`~/Library/Application Support/{appname}/{conf_name}.conf`,' if is_macos else ''), 1
 )
 
 
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -380,7 +380,16 @@ cocoa_create_global_menu(void) {
                 keyEquivalent:@""];
     [appMenu addItem:[NSMenuItem separatorItem]];
     MENU_ITEM(appMenu, @"Preferences…", edit_config_file);
-    MENU_ITEM(appMenu, @"Reload preferences", reload_config);
+    MENU_ITEM(appMenu, @"Reload Preferences", reload_config);
+    [appMenu addItem:[NSMenuItem separatorItem]];
+
+    NSMenu* servicesMenu = [[NSMenu alloc] init];
+    [NSApp setServicesMenu:servicesMenu];
+    [[appMenu addItemWithTitle:@"Services"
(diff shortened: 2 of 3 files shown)


kovidgoyal:
I'm not convinced changing to Application Support is a good idea. It breaks backwards compat in a hard to debug way for no real benefit. I like to avoid breaking backwards compat unless really needed, and not putting non-plist files in ~/Library/Preferences is not a sufficiently compelling reason, since it has no actual downside that I know of. And note that while apple does expect you to use NSUSerDefaults to store data in ~?Library/Preferences, it does also state that it is the directory where config data is supposed to be stored.

page-down:
Apple's `preferences` has some implicitly specific meaning, referring to its specific plist user config. It means that all `defaults database` - `NSUserDefaults` are stored in that location.

Of course it does no harm to put files under that path, except for making users feel weird.
In macOS ecosystem, the software puts its specific configuration and data into NSApplicationSupportDirectory. It would be more comfortable for users to use the convention location for file directories if it is close to the practice under different platforms.

I agree that backwards compatibility shouldn't be broken, even if almost no one would put it in that path.

Do you think it is acceptable to also check `~/Library/Application Support/kitty/kitty.conf`?

kovidgoyal:
On Thu, Nov 04, 2021 at 09:52:27PM -0700, page-down wrote:
> Apple's `preferences` has some implicitly specific meaning, referring to its specific plist user config. It means that all `defaults database` - `NSUserDefaults` are stored in that location.
> 
> Of course it does no harm to put files under that path, except for making users feel weird.

kitty doesn't use that path unless the user decides to put kitty.conf
there.

> In macOS ecosystem, the software puts its specific configuration and data into NSApplicationSupportDirectory. It would be more comfortable for users to use the convention location for file directories if it is close to the practice under different platforms.
> 
> I agree that backwards compatibility shouldn't be broken, even if almost no one would put it in that path.
> 
> Do you think it is acceptable to also check `~/Library/Application Support/kitty/kitty.conf`?

It's fine, but what would be the point? No one has actually asked for
it, and why add yet another useless stat() call at startup.

page-down:
> It's fine, but what would be the point? No one has actually asked for
> it, and why add yet another useless stat() call at startup.

I just saw a strange path when I checking my config and kittens load path. This path is only shown when executing `kitty --help` on macOS, and does not even appear in the documentation on the official site (docs generates on Linux). I think the impact should be minimal, and it might be worth supporting the use of the regular Application Support directory for GUI software.

Since you don't really agree with it, and it affects the performance, I'll leave it at that. I am sincerely sorry for wasting your time.

kovidgoyal:
No worries, and I am happy to merge the menu changes if you send a separate PR for them.

PR #4193, opened 2021-11-05
macOS menu and docs minor tweaks

page-down:
Capitalize the menu item and slightly adjusted the position of the Service menu item according to the convention.

Make them consistent with other menu items.

Add a note on the priority of bold fonts in the document.

Also I found that in the generated kitty.conf, the valid values in the comments are not easily distinguishable from the normal text. Would you be willing to leave backquotes to specifically indicate the values that can be set?

I also found that ``` :link:`title <url>` ``` is forcing line breaks in kitty.conf if the URL is too long, making the URL not recognized by mouseover and certainly not by kitten hints. I don't have a good idea for now and it doesn't matter, so I'll leave it like that.

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -380,7 +380,16 @@ cocoa_create_global_menu(void) {
                 keyEquivalent:@""];
     [appMenu addItem:[NSMenuItem separatorItem]];
     MENU_ITEM(appMenu, @"Preferences…", edit_config_file);
-    MENU_ITEM(appMenu, @"Reload preferences", reload_config);
+    MENU_ITEM(appMenu, @"Reload Preferences", reload_config);
+    [appMenu addItem:[NSMenuItem separatorItem]];
+
+    NSMenu* servicesMenu = [[NSMenu alloc] init];
+    [NSApp setServicesMenu:servicesMenu];
+    [[appMenu addItemWithTitle:@"Services"
+                        action:NULL
+                 keyEquivalent:@""] setSubmenu:servicesMenu];
+    [servicesMenu release];
+    [appMenu addItem:[NSMenuItem separatorItem]];
 
     [appMenu addItemWithTitle:[NSString stringWithFormat:@"Hide %@", app_name]
                        action:@selector(hide:)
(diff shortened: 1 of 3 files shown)


page-down:
I noticed that SemiBold fonts do not take precedence over Bold in Linux, is there any particular reason why this is different from macOS?

```text
kitty --config=NONE -o 'font_family Fira Code' --debug-font-fallback

# macOS
bold face: /path/to/FiraCode-SemiBold.ttf

# Linux
# font exists: ~/.fonts/f/FiraCode_SemiBold.ttf

bold face: ~/.fonts/f/FiraCode_Bold.ttf:0
```

If kitty does need to prioritize Bold over SemiBold on Linux, then the documentation here needs to be updated too. Or the code needs to be updated to prioritize the use of SemiBold.

kovidgoyal:
On Thu, Dec 02, 2021 at 08:30:36PM -0800, page-down wrote:
> I noticed that SemiBold fonts do not take precedence over Bold in Linux, is there any particular reason why this is different from macOS?

They have entirely different font stacks, so the chances of their being
identical in behavior are close to zero.

> 
> ```text
> kitty --config=NONE -o 'font_family Fira Code' --debug-font-fallback
> 
> # macOS
> bold face: /path/to/FiraCode-SemiBold.ttf
> 
> # Linux
> # font exists: ~/.fonts/f/FiraCode_SemiBold.ttf
> 
> bold face: ~/.fonts/f/FiraCode_Bold.ttf:0
> ```
> 
> If kitty does need to prioritize Bold over SemiBold on Linux, then the documentation here needs to be updated too. Or the code needs to be updated to prioritize the use of SemiBold.

I dont think it "needs" to do anything. It just does whatever happens
"naturally" in each system. Generally speaking I would want it to pick
bold over semi bold when automatically selecting faces. macOS exposes
much less font information so that's probably why it ends up selecting
semi-bold there.

page-down:
> ... Generally speaking I would want it to pick bold over semi bold when automatically selecting faces. ...

Thank you. I get it. It seems to have been a problem before, which is why SemiBold was preferred.

If Bold renders properly with all common mono fonts, kitty can also use this default value. Let the small fraction of fonts that have problems be configured by the user.

However, this is not the case. Most fonts will not work with Bold under macOS, and SemiBold must be used by default for the majority of users.

Issue #4166, opened 2021-10-29
Shell integration: additional features for getting command output


page-down:
**Is your feature request related to a problem? Please describe.**
Unable to use shortcuts to quickly open command output in pager that ctrl+shift+z/x jumped to.
Currently only the output of the last executed command is available.

**Describe the solution you'd like**
Take advantage of the benefits of shell integration by extending the capabilities of command output related features.
- Gets the first command output from the screen. (pager/stdin source/rc)
- Get the command output at any line from the screen. (hints/mouse selection extend)

**Describe alternatives you've considered**
n/a

**Additional context**

This feature was previously discussed in the Shell integration thread.

## What I have accomplished:

- Extend the original `last_cmd_output` to support searching up, around, and down for command output at a specific screen line.
- Get the first command output of the screen.
  - Only the first `OUTPUT_START` in the screen will be fetched. Anything else is returned as an empty string.
  - Keyboard action, stdin input source, remote control, etc. associated with this function.
- Fix `last_cmd_output` stdin-source formatting issue.

However, I'm not sure how the assigning shortcuts section should be handled, so it's currently commented out. (`kitty/options/definition.py`)

This currently works great with ctrl+shift+z/x.

You can use shortcuts to open the command output in a newly created kitty window or os window, after a quick ctrl+shift+z/x jump. As an side-by-side reference.

## Function Parameters

```c
static bool find_cmd_output(Screen *self, OutputOffset *oo, index_type start_y, int direction, bool on_screen_only) {}
```

- `int direction`: `-1` search upwards, `0` search around, `1` search downwards
- `bool on_screen_only`: When searching for the command output downwards, this parameter will limit the search to only within the screen.

## Features still to be done

However, the feature about using the command line output around the line has not yet been implemented. My idea is to use the keyboard and mouse related features.

Is it possible to:
- Keyboard: Use hints to select and quickly select(copy) the command output on the screen, without editor pager.
- Mouse: Extend text selection to whole command output (when clicking with modifier keys). Handy for mouse users to quickly select and copy. No more painful manual text selection. (Mouse operation)

I have yet to find a way to make it happen.

If this feature could be added to kitty, it would be a small addition to the Shell integration and shows its convenience.

Thank you very much.

## Patch

<details>
<summary>Click to expand ...</summary>
<p>

```patch
diff --git a/docs/kittens/custom.rst b/docs/kittens/custom.rst
index 0cadd9ea..cfd69431 100644
[...]

PR #4198, opened 2021-11-06
Make scroll to prompt action perform better positioning

page-down:
Mark different prompts with a binary mark.

This allows kitty to distinguish between different prompts when performing scroll to prompt, and to scroll to the exact position.

The kitty shell, fish and bash integration script has been updated and works well.
However, I don't know how to implement it in zsh and need help.

This issue was previously mentioned in #4166

### DIFF
--- a/docs/shell-integration.rst
+++ b/docs/shell-integration.rst
@@ -154,6 +154,11 @@ Just before starting to draw the PS1 prompt send the escape code::
 
     <OSC>133;A<ST>
 
+Binary marks can be used to distinguish the prompt from each other::
+
+    <OSC>133;A;b=0<ST>
+    <OSC>133;A;b=1<ST>
+
 Just before starting to draw the PS2 prompt send the escape code::
 
     <OSC>133;A;k=s<ST>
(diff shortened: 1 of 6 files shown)


page-down:
Thanks for adding the updated zsh integration script and cleaning up the documentation.

However there was one question that kept coming to my mind.
In the original scrolling code, the reason for jumping to the first PROMPT START at the beginning was to solve the multi-line prompt problem.
Do you have an example of multiple PROMPT STARTs in a single time prompt? For example, linebreaks, window resize soft linebreaks.

If we are talking about distinguishing consecutive prompts, without this special case, in fact each PROMPT START is already a separate mark and can be completely distinguished.

I want this mark to be fully functional and not introduce a meaningless thing in the protocol.

kovidgoyal:
As far as I recall, it happens on resize definitely. It may happen in the output of some shell, but I doubt it. It may be worth removing this from the protocol and instead making the resize code more intelligent about not copying prompt start bits. The resize implementation is in rewrap.h and check line-buf.c and screen.c and history.c for how that common code is used. Basically you would want to not copy the promtp start bits to wrapped lines

kovidgoyal:
I have moved the binary procolo extension commits into the binary
branch, pending further discussion/experimentation.

PR #4201, opened 2021-11-07
Add default XDG_DATA_DIRS for fish shell

NextAlone:
Casued by that fish shell has no env XDG_DATA_DIRS default, and add kitty path will caused lots of apps error with `no gsettings schemas`


NextAlone:
found same issue in #4199

PR #4209, opened 2021-11-08
Add how the fish shell integration works and how to configure it manually

page-down:
Add fish-related instructions to the shell integration documentation.

A brief description of how it works, and how the user should configure it manually.

### DIFF
--- a/docs/shell-integration.rst
+++ b/docs/shell-integration.rst
@@ -73,6 +73,12 @@ or in kitty.conf) is a supported shell. If so, kitty adds a couple of lines to
 the bottom of the shell's rc files (in an atomic manner) to load the shell
 integration code.
 
+For fish, to make it automatically load the configuration file provided by
+kitty, the integration script directory path is prepended to the
+:code:`XDG_DATA_DIRS` environment variable. This is only applied to the fish
+process and will be cleaned up by the integration script after startup. No files
+are added or modified.
+
 Then, when launching the shell, kitty sets the environment variable
 :envvar:`KITTY_SHELL_INTEGRATION` to the value of the :opt:`shell_integration`
 option. The shell integration code reads the environment variable, turns on the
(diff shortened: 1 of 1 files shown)


kovidgoyal:
There is no need to jump through all those hoops to configure it manually. I have added a function to automatically source the correct shell integration script for the current kitty.

page-down:
The overhead of launching a new process just to get this file path every time you start an interactive shell is significant. At least for me.

The audience for this document, the ones who will do manual configuration, are generally the ones who value this more. I will try to provide more optimized advice for the target audience. And for the average user it would be suggested to use the default automatic integration.

Also, the integration code for the fish shell starts the process twice, so why not just output this small piece of code and source it directly? It also covers two paths at once.

After all, it's your project, so please feel free to take what suits you best. And I appreciate your professional opinion. Thank you.

kovidgoyal:
There is no need to exec kitty. Just use the new KITTY_INSTALLATION_DIR
env var.

page-down:
Thank you, I saw that. I don't like the performance impact of exec kitty, but I like the end result of your $(kitty +runpy) approach.

Your latest change made me aware of a problem that should be addressed in the documentation.

```text
... if you want to setup shell integration for a remote system over SSH ...
```

I wondered where the best place for `KITTY_INSTALLATION_DIR` to come from.

kovidgoyal:
it's already done. For remote systems you will need to copy the
integration scripts and set KITTY_INSTALLATION_DIR manually to the
copied location.

page-down:
Since kitten ssh can already install terminfo automatically.
Would you mind installing shell-integration scripts automatically with additional parameters, e.g. to `~/.local/share`? 
Or maybe set up the environment variables as well.
In this case I don't know if it's appropriate to write to the user rc profile, if it needs separate parameters to enable.

I think this addresses the most common tasks related to shell integration in a very clear and well organized way.

kovidgoyal:
There is no good way to transmit arbitrary amounts of data over ssh while still preserving its semantics. The ssh kitten works by basically send the terminfo data and script to install it as a command line argument. There are limits to the max size of command line arguments. It would require a different architecture to be able to transport multiple files over. You would need to send the data on stdin and on the remote side after reading stdin change the file descriptor for stdin to the controlling tty. I dont even know if that's doable in POSIX shell scripts. But if so, it's a possibility. I am not going to spend time on it however, you ar ewelcome to contribute.

page-down:
I found that after configuring manually according to the documentation, ssh connections using other terminals (e.g. from android/iOS, which kitty is unavailable) will behave abnormally for obvious reasons.

I currently guarded with TERM, but using that to detect terminal(feature) is known to be unreliable. Wonder if there is a better way, since the current code will cause problems with other programs.

kovidgoyal:
yes, obviously if you are going to be using multiple terminals you need
to guard it. And no over SSH TERM is the only way, short of doing a
potentially very slow roundtrip.

PR #4211, opened 2021-11-08
Fix unwanted hyperlink opening when hyperlink_id is 0

RafaelLeeImg:
The result of
```python
if 0
```
is false, when the hyperlink_id is 0. The interpreter goes to the line
```python
get_boss().open_url(url, cwd=cwd)
```
This leads to unwanted result.

### DIFF
--- a/kitty/window.py
+++ b/kitty/window.py
@@ -644,7 +644,7 @@ class Window:
 
     def open_url(self, url: str, hyperlink_id: int, cwd: Optional[str] = None) -> None:
         opts = get_options()
-        if hyperlink_id:
+        if isinstance(hyperlink_id, int):
             if not opts.allow_hyperlinks:
                 return
             from urllib.parse import unquote, urlparse, urlunparse


kovidgoyal:
hyperlink ids should never be zero, what steps are youusing to produce a
zero hyperlink id?

RafaelLeeImg:
Using cat to create some valid URL in the kitty terminal emulator, select anything in kitty, then switch to another window (other programs, not kitty), and switch back. In rare case (5%), the mouse cursor select something in the kitty window unwantedly (I didn't press mouse key). In similar condition, it will underscore the URL and open a new browser with the first URL in the termins.

I would like to say that the unwanted mouse cursor select may related to this problem. I'm bearing this problem and haven't spent time on that. That maybe the direction.

```bash
$ cat setup.py |grep http
307:        # see https://github.com/kovidgoyal/kitty/issues/3126
354:        # See https://github.com/google/sanitizers/issues/647
1007:        # see https://github.com/kovidgoyal/kitty/issues/1233
1027:        CFBundleGetInfoString='kitty, an OpenGL based terminal emulator https://sw.kovidgoyal.net/kitty/',
```
By the way, I would say, there are other weary problems about the fonts rendering error). The font rendering is also a problem. It loads the wrong font family when a ttf file has multiple fonts.

kovidgoyal:
That cant be the case. hyperlink_id is used for actual hyperlinks which
cannot be created by catting a plain text file. Use ls --hyperlinks=auto to see
actual hyperlinks. Your patch simply prevents opening of plain urls
without hyperlinks, as they have hyperlink_id = 0. Actual hyperlinks
have non zero ids.

As for your issue, I cannot reproduce. And it sound like a mouse
handling issue in any case. So come up with a reproducer if possible.

RafaelLeeImg:
As for your code, using ```python
if some_var
```
The castted boolean of integer varies in different languages.
For example using Scheme
```scheme
(if 0 #t #f) ;; returns #t
```
Programming with multiple languages shall be careful, what's the reason to leave a buggy line in the project?
Have you ever checked if variables are proper locked in C Python interoperate or in multi-threading project?

I won't say the fundamental cause of this bug is easy to find. The bug is just there.

```python
#kitty/window.py
        if 0 == hyperlink_id:
            print(time.strftime("%Y-%m-%d %H:%M:%S", time.localtime()))
            raise ValueError("invalid hyperlink_id (%d)"% hyperlink_id)
```

```Shell
2021-11-08 18:16:42
  File "/dev/shm/d/kitty/kitty/launcher/../../kitty/window.py", line 650, in open_url
    raise ValueError("invalid hyperlink_id (%d)"% hyperlink_id)
ValueError: invalid hyperlink_id (0)
2021-11-08 18:17:14
  File "/dev/shm/d/kitty/kitty/launcher/../../kitty/window.py", line 650, in open_url
    raise ValueError("invalid hyperlink_id (%d)"% hyperlink_id)
ValueError: invalid hyperlink_id (0)
2021-11-08 18:18:12
  File "/dev/shm/d/kitty/kitty/launcher/../../kitty/window.py", line 650, in open_url
    raise ValueError("invalid hyperlink_id (%d)"% hyperlink_id)
ValueError: invalid hyperlink_id (0)
2021-11-08 18:18:33
  File "/dev/shm/d/kitty/kitty/launcher/../../kitty/window.py", line 650, in open_url
    raise ValueError("invalid hyperlink_id (%d)"% hyperlink_id)
ValueError: invalid hyperlink_id (0)
```

kovidgoyal:
Let me repeat. hyperlinks have non-zero ids. URLs that arent hyperlinks have a zero hyperlink id. In python zero evaluates as false. The code is correct. And just FYI all UI code in kitty runs in a single thread, as does all python code. Your patch is what is incorrect.

PR #4229, opened 2021-11-13
Ensure resize when toggling between maximized window and fullscreen

metayan:
When toggling between maximized window and traditional fullscreen, we
need to manually trigger windowDidResize to ensure that the fullscreen
window is resized correctly.

macOS probably ignores the change because the window size actually
stays the same. However, the disappearance of the title bar causes the
usable height to change, leading to the text being stretched when
switching to fullscreen, and this results in ugly artifacts.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -2476,6 +2476,9 @@ bool _glfwPlatformToggleFullscreen(_GLFWwindow* w, unsigned int flags) {
             }
             [window setStyleMask: sm];
         }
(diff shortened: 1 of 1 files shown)


metayan:
Example of correct fullscreen (for example when toggling from a slightly smaller window to fullscreen):
kitty-nice

When toggling from maximized window to traditional fullscreen, this was the result before:
kitty-stretched

This change solves that.

I can only test on macOS Catalina, so if this has been corrected on later versions of macOS, the added code may be moved inside the block above its current place.

PR #4237, opened 2021-11-14
Get the first command output on screen and the last scrolled to, and mouse selection support

page-down:
Record the last position scrolled to in scroll_to_prompt. When its argument is zero, scroll to the last position visited.
Extend the original `last_cmd_output` to support searching up, around, and down for command output at a specific screen line.
Get the first command output of the screen.
Get the last scrolled command output.
Mouse click to select the command output.

Please review and check if there is anything inappropriate, thank you.

If the recent prompt duplication problem is solved for output start as well, perhaps the find upwards related code can be removed in `find_cmd_output`.

Regarding the mouse selection, I'm not quite sure if I'm missing something.
`screen_selection_range_for_cmd_output` is executed twice because I can't find a good way to pass the value into `screen_update_selection` and need help.
The command output mouse selection has not yet configured its default mouse button binding.


kovidgoyal:
I have merged your commits other than the mouse one and fixed various
issues in them. 

Regarding the mouse, dont do it via mouse.c. Instead create a function
in screen.c that given a y sets the current selection to the appropriate
cmd output for that y. Then in window.py in the mouse actions section
add a mappable mouse action that calls a function named
select_clicked_cmd_output this can be modeled on the click_mouse_url
function.

page-down:
Thank you. I will learn from them.
I'm rethinking how mouse selection needs to be handled.

PR #4246, opened 2021-11-16
Docs: scrolling to the last jumped position

page-down:
Also add additional shortcut binding instructions for jumping back to the last jump position.

Regarding the `this` documented in the documentation. Do you think the mappable action needs to be renamed? The current name is `mouse_select_cmd_output`.

Also I found that when mouse_map is bound to an invalid action, it does not output an error message in stderr when loading and verifying the configuration file.


kovidgoyal:
Yes actions and their arguents are not verified at binding time, since that is too much work. One could easily verify the existence of an action, but checking its arguments is not so easy.

PR #4251, opened 2021-11-17
Allows to open the command output in pager by mouse click

page-down:
Add `mouse_show_command_output`

Name the `command` consistently:
- For the user configuration part, use `command`.
- Use `cmd` in code, in parameters, to shorten the length.

Now mouse click will record the last jump position. So when using the relevant get last visited command output, it will also take effect for mouse actions.

Also there are too many shell integration URLs in the commented configuration files now, do you have any good ideas? Or just leave it as is?

### DIFF
--- a/docs/shell-integration.rst
+++ b/docs/shell-integration.rst
@@ -74,29 +74,40 @@ on the output, define the following in :file:`kitty.conf`:
 
 .. code:: conf
 
-   mouse_map right press ungrabbed mouse_select_command_output
+    mouse_map right press ungrabbed mouse_select_command_output
 
 Now, when you right click on the output, the entire output is selected, ready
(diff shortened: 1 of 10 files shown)


kovidgoyal:
On Wed, Nov 17, 2021 at 06:32:28AM -0800, page-down wrote:
> Also there are too many shell integration URLs in the commented configuration files now, do you have any good ideas? Or just leave it as is?

I think it should be left as is. Someone browsing the conf file
would find it useful to be able to visit the shell integration page
per config option that uses it.

If it bothers you, you can postprocess the output with a regex to
replace the links with just the text "Shell integration" But given there
are 3-4 of them in the entire file I dont think its an issue.

PR #4255, opened 2021-11-19
Allow macOS notification to play sound

page-down:
Allow macOS notification to play system alert sound.

Also solved an issue with macOS notifications not showing up. Now macOS notifications from the cli program will be properly alerted when kitty is in the foreground.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -2439,7 +2439,7 @@ void _glfwPlatformSetCursor(_GLFWwindow* window, _GLFWcursor* cursor UNUSED)
 bool _glfwPlatformIsFullscreen(_GLFWwindow* w, unsigned int flags) {
     NSWindow *window = w->ns.object;
     bool traditional = !(flags & 1);
-    if (traditional) { if(@available(macOS 10.16, *)) return w->ns.in_traditional_fullscreen; }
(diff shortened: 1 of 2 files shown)


page-down:
I found that the html page generated for desktop notification is empty. I don't know how to fix it yet.

https://sw.kovidgoyal.net/kitty/desktop-notifications/

kovidgoyal:
The page looks fine to me
screenshot

page-down:
Found the cause, the following css filtering rule is located in a popular ad filtering list.

```text
#desktop-notifications
```

Do you think it is possible to rename the css class name to kitty-desktop-notifications ?

The name itself is not a problem. The purpose of the name change is to allow some users with ad filtering on to visit without being affected. Otherwise they will ask why. Then I had to remind them again to set up their own whitelist.

page-down:
Yes, my first feeling was also that it was inappropriate.

The user should find the source of the problem and have it reported to the list maintainer to fix it. Not here.

page-down:
Regarding the documentation, I also found that `<span class="italic">` was missing the italic style.

But I'm not sure this is something that needs to be added to `custom.css`, it looks like something that upstream needs to deal with. Do you have a good way to handle this?

kovidgoyal:
Fixed.

PR #4260, opened 2021-11-22
Add option action_alias

page-down:
New option action_alias is added to define common actions.
Also restrict kitten_alias to apply only to kitten actions.

---

I found a type mismatch issue. The `KeyAction` object is added to the args during the parsing of the combine action.

```python
class KeyAction(NamedTuple):
    func: str
    args: Tuple[Union[str, float, bool, int, None], ...] = ()

# options/utils.py
@func_with_args('combine')
def combine_parse(func: str, rest: str) -> FuncArgsType:
    # ...
    args = tuple(map(parse_key_action, filter(None, parts))) # parse_key_action() -> KeyAction
    return func, args

```

Should `KeyAction` be added to the Union type of KeyAction args?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -146,6 +146,10 @@ To update |kitty|, :doc:`follow the instructions <binary>`.
 - macOS: Fix keyboard input not working after toggling traditional fullscreen
   till the window is clicked in
 
+- A new option :opt:`action_alias` to create action alias with common options
+  for keyboard shortcuts and mouse actions. Can also be defined in
+  :file:`open-actions.conf`. (:pull:`4260`)
+
 
 0.23.1 [2021-08-17]
 ----------------------
--- a/docs/launch.rst
+++ b/docs/launch.rst
@@ -38,6 +38,14 @@ To pass the contents of the current screen and scrollback to the started process
 
 There are many more powerful options, refer to the complete list below.
 
+.. note::
+
+    To avoid duplicating launch actions with frequently used parameters, you can
+    use :opt:`action_alias` to define launch action aliases. For example::
+
+        action_alias launch_cwd launch --cwd=current
+
+
 The piping environment
 --------------------------
 
--- a/docs/open_actions.rst
+++ b/docs/open_actions.rst
@@ -29,16 +29,20 @@ action which can be used to run external programs. Entries are separated by
 blank lines.
 
 Actions are very powerful, anything that you can map to a key combination in
-`kitty.conf` can be used as an action. You can specify more than one action per
-entry if you like, for example:
+`kitty.conf` can be used as an action. You can use ``action_alias`` to define
+action aliases that can be reused in each entries. You can specify more than one
+action per entry if you like. For example:
 
 
 .. code:: conf
 
+    # Action aliases for open actions
+    action_alias launch_os_window launch --title ${FILE} --type=os-window
+
     # Tail a log file (*.log) in a new OS Window and reduce its font size
     protocol file
     ext log
-    action launch --title ${FILE} --type=os-window tail -f ${FILE_PATH}
+    action launch_os_window tail -f ${FILE_PATH}
     action change_font_size current -2
 
 
--- a/kitty/conf/types.py
+++ b/kitty/conf/types.py
@@ -185,11 +185,11 @@ class Option:
             return ans
         if option_group:
             sz = max(len(self.name), max(len(o.name) for o in option_group))
-            a(f'{self.name.ljust(sz)} {self.defval_as_string}')
+            a(f'{self.name.ljust(sz)} {self.defval_as_string}'.rstrip())
             for o in option_group:
-                a(f'{o.name.ljust(sz)} {o.defval_as_string}')
+                a(f'{o.name.ljust(sz)} {o.defval_as_string}'.rstrip())
         else:
-            a(f'{self.name} {self.defval_as_string}')
+            a(f'{self.name} {self.defval_as_string}'.rstrip())
         if self.long_text:
             a('')
             a(render_block(self.long_text))
@@ -248,8 +248,10 @@ class MultiOption:
         a = ans.append
         for k in self.items:
             if k.documented:
-                prefix = '' if k.add_to_default else '# '
-                a(f'{prefix}{self.name} {k.defval_as_str}')
+                a(f'{self.name} {k.defval_as_str if k.add_to_default else ""}'.rstrip())
+                if not k.add_to_default and k.defval_as_str:
+                    a('')
+                    a(f'#: E.g. {self.name} {k.defval_as_str}'.rstrip())
         if self.long_text:
             a('')
             a(render_block(self.long_text))
@@ -264,7 +266,7 @@ class MultiOption:
         a('')
         for k in self.items:
             if k.documented:
-                a(f'    {self.name:s} {k.defval_as_str}')
+                a(f'    {self.name:s} {k.defval_as_str}'.rstrip())
         a('')
         if self.long_text:
             a(expand_opt_references(conf_name, self.long_text))
@@ -297,8 +299,9 @@ class Mapping:
         if self.short_text:
             a(render_block(self.short_text.strip())), a('')
         for sc in [self] + action_group:
-            if sc.documented and sc.add_to_default:
-                a(sc.setting_name + ' ' + sc.parseable_text)
+            if sc.documented:
+                prefix = '' if sc.add_to_default else '#::  E.g. '
+                a(f'{prefix}{sc.setting_name} {sc.parseable_text}')
         if self.long_text:
             a(''), a(render_block(self.long_text.strip(), '#::  '))
         a('')
--- a/kitty/conf/utils.py
+++ b/kitty/conf/utils.py
@@ -118,6 +118,29 @@ def choices(*choices: str) -> Choice:
     return Choice(choices)
 
 
+def resolve_action_alias(key: str, val: str, aliases: Dict[str, Any]) -> Optional[str]:
+    if key == 'map':
+        # map: key <action> [args]
+        idx = 1
+    elif key == 'mouse_map':
+        # mouse_map: btn event modes <action> [args]
+        idx = 3
+    elif key == 'open_actions':
+        # open_actions: <action> [args]
+        idx = 0
+    else:
+        return None
+    parts = val.split(maxsplit=idx+1)
+    if len(parts) > idx and parts[idx] in aliases:
+        action = aliases.get(parts[idx])
+        if action:
+            parts[idx] = action[0]
+            if len(action) > 1 and action[1]:
+                parts.insert(idx + 1, action[1])
+            return ' '.join(parts)
+    return None
+
+
 def parse_line(
     line: str,
     parse_conf_item: ItemParser,
@@ -151,6 +174,11 @@ def parse_line(
                 format(val)
             )
         return
+    aliases = ans.get('action_alias')
+    if key in ('map', 'mouse_map') and aliases:
+        resolved_val = resolve_action_alias(key, val, aliases)
+        if resolved_val is not None:
+            val = resolved_val
     if not parse_conf_item(key, val, ans):
         log_error(f'Ignoring unknown config key: {key}')
 
--- a/kitty/conf/utils.py
+++ b/kitty/conf/utils.py
@@ -118,7 +118,7 @@ def choices(*choices: str) -> Choice:
     return Choice(choices)
 
 
-def resolve_action_alias(key: str, val: str, aliases: Dict[str, Any]) -> Optional[str]:
+def resolve_action_alias(key: str, val: str, aliases: Dict[str, Any]) -> str:
     if key == 'map':
         # map: key <action> [args]
         idx = 1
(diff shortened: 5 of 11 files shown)


page-down:
I left the `kitten_alias` implementation untouched because I need to replace the alias before the string is converted to a KeyAction, and parse the entire string with the alias replaced. Otherwise it would be discarded when converting to KeyAction because the parser detects insufficient parameters or a problem with the position.

There is a problem with aliases not taking effect within the combine action. I tried to resolve it in `combine_parse` (`parse_key_action(resolve_action_alias(...))`), but the global action aliases are not easy to pass into these parser functions and I need help. Thank you.

kovidgoyal:
Make a separate PR for your conf formatting changes please.

page-down:
Thank you. I found that there seems to be a problem in validating the action.

```shell
kitty --config=NONE -o 'action_alias ac1 launch --type=tab' -o 'map f1 ac1 sh -c read' -o 'map f2 ac1 --title=debug2'
```

```text
Ignoring unknown map action: ac1 sh -c read
Ignoring unknown map action: ac1 --title=debug2
```

kovidgoyal:
should be fine now

page-down:
How to configure with the new action_alias for setting default parameters for kittens?
(This is one reason why I left kitten_alias untouched before.)
This feature is very convenient.

The following only works for the shortcuts configured afterwards, and not for the default `ctrl+shift+e` etc.

```text
kitty --config=NONE -o 'kitten_alias hints hints --hints-offset=3' -o 'map f1 kitten hints'
```

Not related to action_alias:

```text
kitty --config=NONE -o 'kitten_alias hints --invalid' -o 'map f1 kitten hints'
ModuleNotFoundError: No module named 'kittens.__invalid'
```

It would be more convenient to be able to configure the following as well.
Action with fixed position parameters. Validate when finalizing the alias resolution.

```text
kitty --config=NONE -o 'action_alias ac1 change_font_size current' -o 'map f1 ac1 +2' -o 'map f2 ac1 -2'
TypeError: change_font_size() missing 1 required positional argument: 'amt'
```

Is it possible to do parameter validation when finalizing the alias resolution? Instead of throwing an exception on execution.
Users can also know that there is a configuration problem that needs to be fixed.

```text
kitty --config=NONE -o 'action_alias ac1 toggle_marker --invalid' -o 'map f1 ac1'
TypeError: toggle_marker() missing 2 required positional arguments: 'spec' and 'flags'
```

Also I remember that if the option is configured incorrectly, an error message is displayed in the foreground (overlay), when the kitty config is finished parsing.

kovidgoyal:
action_alias mykitten kitten hints --whatever
map f1 mykitten --extra

page-down:
> action_alias mykitten kitten hints --whatever
> map f1 mykitten --extra

Yes, this is obviously working. So that means that `kitten_alias hints hints --hints-offset=0` in previous versions modified all `kitten hints` parameters by default, including the default shortcuts. However, with the new `action_alias`, all `kitten hints` parameters will no longer be changed. This should be considered a breaking change?

If this change has to be introduced, does the default value for kitten_alias in definition.py need to be removed? Is it possible to not include deprecated configuration items in the commented configuration files generated for new users?

I like the new `action_alias` approach, just need to reconfigure all kitten default shortcuts.

kovidgoyal:
On Mon, Nov 22, 2021 at 09:13:18AM -0800, page-down wrote:
> > action_alias mykitten kitten hints --whatever
> > map f1 mykitten --extra
> 
> Yes, this is obviously working. So that means that `kitten_alias hints hints --hints-offset=0` in previous versions modified all `kitten hints` parameters by default, including the default shortcuts. However, with the new `action_alias`, all `kitten hints` parameters will no longer be changed. This should be considered a breaking change?

The same remains true if you use kitten_alias now. If you use
action_alias, yes you have to re-write.

> 
> If this change has to be introduced, does the default value for kitten_alias in definition.py need to be removed? Is it possible to not include deprecated configuration items in the commented configuration files generated for new users?
> 

What change? kitten_alias should continue to work as before, unless i
made a mistake in its implementation.

page-down:
> The same remains true if you use kitten_alias now.

The following issues may need to be addressed, latest master branch.

```text
kitty --config=NONE -o 'kitten_alias hints hints --hints-offset=3'
RecursionError: maximum recursion depth exceeded while calling a Python object
```

> If you use action_alias, yes you have to re-write.

I understand that `action_alias myhints hints` needs to be called with myhints as the action.

However, in the definition.py documentation, you mention that

> Similarly, to alias kitten invocation::
>
>    action_alias hints kitten hints --hints-offset=0

How is this interpreted? Is it the same as `kitten hints hints`? According to the literal meaning, a new alias named `hints` is defined and called with `map f1 hints`.

> What change? kitten_alias should continue to work as before

I fully understand your want backward compatibility, so the issues raised so far are to ensure that the current implementation is consistent with the previous behavior.

Since it is worthwhile to keep the previous features, it would be nice to have new ways of defining the previous behavior (set default parameters for kitten) without the need for the previous `kitten_alias`.

page-down:
The action alias combine does not seem to be working properly.

```text
kitty --config=NONE -o 'action_alias ac1 combine : send_text normal 1 : send_text normal 2' -o 'map f1 ac1'

combine : send_text normal 1 : send_text normal 2
combine() takes from 2 to 4 positional arguments but 9 were given
```

And this one below.

```text
kitty --config=NONE -o 'action_alias ac1 change_font_size current' -o 'map f1 ac1 +2' -o 'map f2 ac1 -2'

change_font_size current +2
change_font_size() missing 1 required positional argument: 'amt'
```

Everything else looks fine for now.

kovidgoyal:
On Mon, Nov 22, 2021 at 09:38:34AM -0800, page-down wrote:
> > The same remains true if you use kitten_alias now.
> 
> The following issues may need to be addressed, latest master branch.
> 
> ```text
> kitty --config=NONE -o 'kitten_alias hints hints --hints-offset=3'
> RecursionError: maximum recursion depth exceeded while calling a Python object
> ```

Fixed.

> 
> > If you use action_alias, yes you have to re-write.
> 
> I understand that `action_alias myhints hints` needs to be called with myhints as the action.
> 
> However, in the definition.py documentation, you mention that
> 
> > Similarly, to alias kitten invocation::
> >
> >    action_alias hints kitten hints --hints-offset=0
> 
> How is this interpreted? Is it the same as `kitten hints hints`? According to the literal meaning, a new alias named `hints` is defined and called with `map f1 hints`.

Correct, it needs map f1 hints it is *not* the same as kitten_alias hints
hints.

> 
> > What change? kitten_alias should continue to work as before
> 
> I fully understand your want backward compatibility, so the issues raised so far are to ensure that the current implementation is consistent with the previous behavior.

It is exactly consistent as far as I know (apart from now allowing
recursion in kitten_alias). action_alias didnt exist before so there is
nothing to be consistent with.

> 
> Since it is worthwhile to keep the previous features, it would be nice to have new ways of defining the previous behavior (set default parameters for kitten) without the need for the previous `kitten_alias`.

kitten_alias still exists if you prefer using it.

kovidgoyal:
As is noted in the docs for action_alias it only works for some actions,
not arbitrary ones. Making it work for arbitrary ones, is *much* harder
and also not needed so much.


On Mon, Nov 22, 2021 at 09:52:14AM -0800, page-down wrote:
> The action alias combine does not seem to be working properly.
> 
> ```text
> kitty --config=NONE -o 'action_alias ac1 combine : send_text normal 1 : send_text normal 2' -o 'map f1 ac1'
> 
> combine : send_text normal 1 : send_text normal 2
> combine() takes from 2 to 4 positional arguments but 9 were given
> ```
> 
> And this one below.
> 
> ```text
> kitty --config=NONE -o 'action_alias ac1 change_font_size current' -o 'map f1 ac1 +2' -o 'map f2 ac1 -2'
> 
> change_font_size current +2
> change_font_size() missing 1 required positional argument: 'amt'
> ```
> 
> Everything else looks fine for now.
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/4260#issuecomment-975775108

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

page-down:
So combine and other actions are not included, at least for now.

For users with such niche needs, it is recommended to use templates or external tools such as sed to generate configuration files with string replacement.

I found that in `class MultiOption: as_conf(); as_rst();`, the `long_text` of the option is not affected by `documented`. This causes the long_text to be output alone.

It might be worthwhile to keep information about the deprecated option in the documentation so that users can be informed, yet not make this option available to new users (generated config).

Thank you for implementing and making this feature available.

page-down:
In the latest change, `kitten_alias` fails to apply to the default configured shortcuts.

```text
kitty --config=NONE -o 'kitten_alias hints hints --hints-offset=3'
```

press ctrl+shift+p>w

It worked fine in earlier revisions.

Also error is reported when a non-existent consecutive shortcut key is pressed.

press ctrl+shift+p>p

```text
"kitty/boss.py", line 1137, in combine
    if self.dispatch_action(actions[0], window_for_dispatch, dispatch_type):
IndexError: tuple index out of range

During handling of the above exception, another exception occurred:
"kitty/boss.py", line 1142, in combine
    self.show_error('Key action failed', f'{actions[0].pretty()}\n{e}')
IndexError: tuple index out of range
```

kovidgoyal:
On Tue, Nov 23, 2021 at 02:16:20AM -0800, page-down wrote:
> In the latest change, `kitten_alias` fails to apply to the default configured shortcuts.
> 
> ```text
> kitty --config=NONE -o 'kitten_alias hints hints --hints-offset=3'
> ```

Yes, that's by design. It's a performance optimization, otherwise all
the default shortcut actions have to be parsed on every startup, to
check for aliases. So to override them you need to remap them in your 
kitty.conf. I suppose I could reparse them only if aliases are defined. 

> 
> press ctrl+shift+p>w
> 
> It worked fine in earlier revisions.
> 
> Also error is reported when a non-existent consecutive shortcut key is pressed.
> 
> press ctrl+shift+p>p
> 
> ```text
> "kitty/boss.py", line 1137, in combine
>     if self.dispatch_action(actions[0], window_for_dispatch, dispatch_type):
> IndexError: tuple index out of range
> 
> During handling of the above exception, another exception occurred:
> "kitty/boss.py", line 1142, in combine
>     self.show_error('Key action failed', f'{actions[0].pretty()}\n{e}')
> IndexError: tuple index out of range
> ```

Fixed.

page-down:
>> ... kitten_alias fails to apply to the default ...
>
> Yes, that's by design. It's a performance optimization, ... I suppose I could reparse them only if aliases are defined.

I appreciate the good design. As long as the documentation is consistent with the behavior, it should be fine.

>> ... Also error is reported when a non-existent consecutive shortcut key is pressed. 
>
> Fixed.

It works fine now.

Regarding the empty alias, is it possible to reset (remove the definition)?

```text
kitty --config=NONE -o 'action_alias change_font_size change_font_size current' -o 'action_alias change_font_size' -o 'map f1 change_font_size all +2'
```

Currently it needs to be defined again using `action_alias change_font_size change_font_size`.

kovidgoyal:
On Tue, Nov 23, 2021 at 03:15:22AM -0800, page-down wrote:
> >> ... kitten_alias fails to apply to the default ...
> >
> > Yes, that's by design. It's a performance optimization, ... I suppose I could reparse them only if aliases are defined.
> 
> I appreciate the good design. As long as the documentation is consistent with the behavior, it should be fine.

I decided to do the re-parse if aliases are defined, it only costs 0.5ms
on my laptop.

> 
> >> ... Also error is reported when a non-existent consecutive shortcut key is pressed. 
> >
> > Fixed.
> 
> It works fine now.
> 
> Regarding the empty alias, is it possible to reset (remove the definition)?
> 
> ```text
> kitty --config=NONE -o 'action_alias change_font_size change_font_size current' -o 'action_alias change_font_size' -o 'map f1 change_font_size all +2'
> ```
> 
> Currently it needs to be defined again using `action_alias change_font_size change_font_size`.

What would be the point of removing an alias? Remember alias mapping is
done after all alias are processed. So removing an alias would be the
same as not setting it in the first place.

page-down:
> What would be the point of removing an alias?

Since alias will not take effect for the mapping configured before then, resetting to the original definition is still somewhat usable.

But now that re-parse has been implemented, it makes no sense at all.

Thank you for the improvements to action_alias, it now behaves the same as the previous kitten_alias.

page-down:
After the recent `Parse actions on demand [0c274a9]` update, I noticed that one of my mouse_map was acting abnormally.

This can be reproduced with the following command, related to `combine`, `launch`.

```shell
kitty --config=NONE -o 'map f1 combine : launch sh -c read : send_text normal : no_op'
```

Before the recent changes, everything was fine and one kitty window would be created. Now it opens three windows (with the wrong text if stdin-source + pager is used).

kovidgoyal:
https://github.com/kovidgoyal/kitty/commit/af7a104f5b2d82f8ab48d3fcf0524b0281723eb1

page-down:
I found a very common scenario in daily use.
Both launch and program have the need to specify their parameters.
```conf
action_alias mypager launch --type os-window --stdin-source={action_args} less {program_args:-default_value_here}
```

I am not sure if there is a good solution. Can slightly cover this simple and efficient use case and is easy to maintain.
Allow users to define such action aliases to reduce configuration duplication.

action_alias should not be overly complex, and any additional parsing needs should be written in their own kitten.

PR #4262, opened 2021-11-22
Allows focus to the last window

page-down:
Focuses to the last window when the nth_window parameter exceeds the number of windows.
The description of the visual window selection has also been added.

I found that nth_window jumps to not nth, but nth+1. (window index starts from 0)
Would you mind fixing this? It will brings minimal breaking changes and the user just needs to update their configuration.

### DIFF
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -3222,7 +3222,15 @@ last two active windows, use the value -1.
 '''
     )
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
No, I dont think nth_window should use 1 based indexing. Its backward
compatible and fairly unusual to use 1 based indexing. visual select
uses 1 based indexing because 1 is the first key, not zero.

Also, the mapping to select the previous window is already documented in
basic.rst no need to add it to the conf file as well.

PR #4265, opened 2021-11-22
Formatting config options that are not added to default

page-down:
There are some options that are not added to the default configuration.


page-down:
It looks like the recent changes have caused flake8 to fail.

PR #4269, opened 2021-11-23
Append long_text only if at least one documented option exists

page-down:
For MultiOption, if `documented=False` is set and long_text is also present, then the commented configuration file will be generated with only the long_text, without any name above.

The recent `kitten_alias` changes is affected by this.


page-down:
I just submitted it and noticed that you made another change, please review if this is still appropriate, thank you.

Prevents outputting long_text alone without a documented item.

PR #4275, opened 2021-11-25
Slightly tweaked macOS modifier key documentation

page-down:
Added command key full name.
Added modifier key Unicode character.
Replaced ASCII `^` with the control modifier key Unicode character representation.

I'm not sure about using :ref: `Modifier key definitions <modifiers>` in the documentation.
https://sw.kovidgoyal.net/kitty/conf/#keyboard-shortcuts
` For a list of modifier names, see: :link: GLFW mods ... `

```python
ref_map = { ..., 'modifiers': f'{website_url("keyboard-protocol")}#modifiers', ... }
```

I also found some :kbd: conversion issues in the generated HTML documents.

Execute make html in the master branch.

Comma in the key name. Space-separated words are parsed as standalone keys, e.g. `and`.

```html
the <kbd>Enter,</kbd> <kbd>Tab</kbd> <kbd>and</kbd> <kbd>Backspace</kbd> keys
```

In previous versions, such as the online version :kbd: was rendered as a whole < kbd >.

https://sw.kovidgoyal.net/kitty/keyboard-protocol/

### DIFF
--- a/docs/basic.rst
+++ b/docs/basic.rst
@@ -34,8 +34,8 @@ Action                      Shortcut
 ========================    =======================
 New tab                     :sc:`new_tab` (also :kbd:`⌘+t` on macOS)
 Close tab                   :sc:`close_tab` (also :kbd:`⌘+w` on macOS)
-Next tab                    :sc:`next_tab` (also :kbd:`^+⇥` and :kbd:`⇧+⌘+]` on macOS)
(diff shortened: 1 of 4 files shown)


kovidgoyal:
kbd is implemented in sphinx, not kitty. Presumably your sphinx version has changed.

page-down:
Yes, this is obviously related to sphinx and the theme. I installed the deps via `docs/requirements.txt`. It should be the same as in CI. Maybe it's due to the furo theme update.

PR #4298, opened 2021-11-29
Add option to configure shell integration enabled by environment var

page-down:
After the recent changes, only bash actually needs to modify the rc file to enable shell integration.
A new option `no-env` has been added, along with `no-rc` to control both cases independently.

Also added definition type check to the shell_integration option to filter out invalid values.

### DIFF
--- a/docs/shell-integration.rst
+++ b/docs/shell-integration.rst
@@ -47,25 +47,34 @@ default, all shell integration is enabled. Individual features can be turned
 off or it can be disabled entirely as well. The :opt:`shell_integration` option
 takes a space separated list of keywords:
 
+enabled
+    Enable shell integration. This is the default.
+
 disabled
-    turn off all shell integration
+    Turn off all shell integration.
 
 no-rc
-    dont modify the shell's launch environment to enable integration. Useful if you prefer
-    to :ref:`manually enable integration <manual_shell_integration>`.
+    Don't modify the shell's rc files to enable integration. Useful if you prefer
+    to :ref:`manually enable integration <manual_shell_integration>`. Turn off bash
+    automatic integration.
+
+no-env
+    Don't modify the shell's launch environment to enable integration. Useful if you prefer
+    to :ref:`manually enable integration <manual_shell_integration>`. Turn off zsh and fish
+    automatic integration.
 
 no-cursor
-    turn off changing of the text cursor to a bar when editing text
+    Turn off changing of the text cursor to a bar when editing text.
 
 no-title
-    turn off setting the kitty window/tab title based on shell state
+    Turn off setting the kitty window/tab title based on shell state.
 
 no-prompt-mark
-    turn off marking of prompts. This disables jumping to prompt, browsing
+    Turn off marking of prompts. This disables jumping to prompt, browsing
     output of last command and click to move cursor functionality.
 
 no-complete
-    turn off completion for the kitty command.
+    Turn off completion for the kitty command.
 
 
 More ways to browse command output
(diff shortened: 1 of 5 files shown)


kovidgoyal:
Why do we need a separate option for this? Either you want automatic
integration code loading or you dont for your shell of choice. no-rc
covers this. I dont think wanting to disable code loading for bashrc but
not for zsh/fish is such a common use case as to warrant an extra option
for it.

page-down:
Maybe someone wants to enable automatic shell integration, but does not want to automatically modify the rc files.

This allows the user to use `shell_integration no-rc` in the configuration file. In this case the user only needs to configure bash manually. zsh and fish do not need to be configured.

Since you think it's useless, forget about it.

Also is it necessary to use KITTY_INSTALLATION_DIR in bash integration scripts? If not configured manually, executing different versions of the instance will load incorrect scripts. Although this basically has no impact on normal users.

```shell
version-a/kitty --config=NONE -o shell=bash
# window-a-1: version-a kitty.bash loaded
verseion-b/kitty --config=NONE -o shell=bash
# window-b-1: version-b kitty.bash loaded

# Open a new kitty window in A.
# window-a-2: version-b kitty.bash loaded
```

I also found that configuring ZDOTDIR via `env` does not work.

```shell
mkdir /tmp/zdotdir
touch /tmp/zdotdir/.zshrc
unset ZDOTDIR
kitty --config=NONE -o shell=zsh -o env=ZDOTDIR=/tmp/zdotdir
```

It looks like this requires checking `env ZDOTDIR` in opts at boot time. Instead of just checking the kitty process `os.environ['ZDOTDIR']`.

page-down:
After pulling in the latest version and configuring ZDOTDIR via env it works fine.

The only drawback of using KITTY_INSTALLATION_DIR is that it does not work after ssh connection. However zsh, fish will not work either, so it is the same.

If bash can use KITTY_INSTALLATION_DIR, there is no reason for me to configure it manually.

kovidgoyal:
On Sun, Nov 28, 2021 at 10:44:32PM -0800, page-down wrote:
> Also is it necessary to use KITTY_INSTALLATION_DIR in bash integration scripts? If not configured manually, executing different versions of the instance will load incorrect scripts. Although this basically has no impact on normal users.
> 
> ```shell
> version-a/kitty --config=NONE -o shell=bash
> # window-a-1: version-a kitty.bash loaded
> verseion-b/kitty --config=NONE -o shell=bash
> # window-b-1: version-b kitty.bash loaded
> 
> # Open a new kitty window in A.
> # window-a-2: version-b kitty.bash loaded

OK changed.

> ```
> 
> I also found that configuring ZDOTDIR via `env` does not work.
> 
> ```shell
> mkdir /tmp/zdotdir
> touch /tmp/zdotdir/.zshrc
> unset ZDOTDIR
> kitty --config=NONE -o shell=zsh -o env=ZDOTDIR=/tmp/zdotdir
> ```
> 
> It looks like this requires checking `env ZDOTDIR` in opts at boot time. Instead of just checking the kitty process `os.environ['ZDOTDIR']`.

No just needed to use env rather than os.environ in the setup_zsh_env
function.

kovidgoyal:
On Sun, Nov 28, 2021 at 11:01:51PM -0800, page-down wrote:
> After pulling in the latest version and configuring ZDOTDIR via env it works fine.
> 
> The only drawback of using KITTY_INSTALLATION_DIR is that it does not work after ssh connection. However zsh, fish will not work either, so it is the same.

After ssh connection, it wont work anyway, since KITTY_SHELL_INTEGRATION
wont be set and the integration scripts wont be available on the remote
machine.

PR #4300, opened 2021-11-30
Fix remote control detach-tab detach-window

page-down:
Fix the issue that detach remote control command cannot match the target.

### DIFF
--- a/kitty/rc/detach_tab.py
+++ b/kitty/rc/detach_tab.py
@@ -16,7 +16,7 @@ class DetachTab(RemoteCommand):
 
     '''
     match: Which tab to detach
-    target: Which OS Window to move the detached tab to
+    target_tab: Which OS Window that containing the target tab to move the detached tab to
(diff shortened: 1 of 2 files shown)


page-down:
I found that after using `detach-tab` to move a tab, the focus of the window activated in the moved tab was not restored and the focus could not be switched by mouse click.

```shell
kitty --config=NONE -o allow_remote_control=yes sh -c 'kitty @ launch; kitty @ launch --type=tab --title=move-this; kitty @ launch; kitty @ send-text "# last focused"; kitty @ launch --type=os-window --title=target; kitty @ detach-tab -m title:^move-this -t title:^target;'
```

I think it's related to the code in boss.py `_move_tab_to()`, tabs.py `take_over_from()`.

kovidgoyal:
https://github.com/kovidgoyal/kitty/commit/d372b27ccd683e33c293b63a49738fb9af66b1a2

PR #4304, opened 2021-12-01
Docs: Add valid values for background image related options

page-down:
Add documentation for options that are missing valid values.

Since indented lines are considered commands without line break, so manually wrap these document lines.

### DIFF
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -969,7 +969,12 @@ The tab bar style, can be one of:
 '''
     )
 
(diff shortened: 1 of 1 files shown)


page-down:
I found that the background image renders strangely when testing with the google logo png.

https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png

```shell
kitty --config=NONE -o 'background_image /tmp/google.png' -o 'background_image_layout clamped' -o 'background_image_anchor center'
```

kitty_bg_img_debug

kovidgoyal:
You should report that in #4167

PR #4337, opened 2021-12-10
Docs: List kitten names in the kitten default shortcuts list

page-down:
Improve the list of default shortcuts for kitten in the documentation by listing the kitten names.

https://sw.kovidgoyal.net/kitty/actions/#kitten

Also use action names as link text in some places.

### DIFF
--- a/docs/actions.rst
+++ b/docs/actions.rst
@@ -4,6 +4,7 @@ Mappable actions
 .. highlight:: conf
 
 The actions described below can be mapped to any key press or mouse action
-using the ``map`` and ``mouse_map`` directives in :file:`kitty.conf`.
+using the ``map`` and ``mouse_map`` directives in :file:`kitty.conf`. For
+configuration examples, see the default shortcut links for each action.
 
 .. include:: /generated/actions.rst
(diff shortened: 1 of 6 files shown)


page-down:
Strange, I did not make any changes there.
There is also a small change.

Thank you.

```patch
diff --git a/kitty/actions.py b/kitty/actions.py
index a42f50a8..4080a56a 100644
--- a/kitty/actions.py
+++ b/kitty/actions.py
@@ -112,9 +112,9 @@ def kitten_link(text: str) -> str:
                 a('')
                 a('Default shortcuts using this action:')
                 if action.name == 'kitten':
+                    a('')
                     scs = {(kitten_link(m.parseable_text), m.short_text, f':sc:`kitty.{m.name}`') for m in maps[action.name]}
                     for s in sorted(scs):
-                        a('')
                         a(f'- {s[0]} - {s[2]} {s[1]}')
                 else:
                     scs = {f':sc:`kitty.{m.name}`' for m in maps[action.name]}
```

page-down:
Sorry, I found one more problem. In the docs, only `unicode-input.rst` uses hyphen. Causing this needs to be handled differently than the rest of the kittens.

There is a problem with this code: `f':doc:`kittens/{x[2]}`'.replace('_', '-')` 

Do you accept the use of hyphen in filenames in docs/kittens consistently? (Introducing URL changes, hopefully with minimal impact.)
Or special string substitution in the code for `unicode-input`?
I have the patch ready, but would like to discuss it first. Thanks.

kovidgoyal:
I am fine with changing unicode-input.rst to unicode_input.rst just add a
redirect for the old location. See add_old_redirects() in publish.py

PR #4339, opened 2021-12-10
Docs: Use underscores in kitten document filenames

page-down:
Use underscores in kitten document filenames.

Add a redirect page for unicode-input.

In the side bar, put custom kitten docs at the top instead of in the middle of the standalone kittens.


page-down:
Is it possible to move the custom kitten to the bottom? It looks weird because it's currently in the middle.

Maybe it's not worth putting so much into it.

kovidgoyal:
There are a bunch of kittens that are very uninteresting, such as the ask kitten. While the custom kittens page is fairly useful. I dont think it should be at the bottom. If it really bothers you figure out how to get sphix to output a line separator in the middle of the toc so the unimportant kittens can be placed after.

PR #4343, opened 2021-12-11
Fix remote control scroll-window argument parsing

page-down:
When executing the remote control `scroll-window 1u` (or end with `+`), an exception is thrown.

```text
ValueError: invalid literal for int() with base 10: '1u'
```

I also found that when "unscroll" is executed with escape code or this remote control command. The cursor position does not seem to be updated. For example, after generating some empty lines under the current prompt cursor, execute unscroll. `printf "\e[3+T"`
Although the scrollback content is moved down. But the cursor did not move down together.

Would you mind taking a look? Thanks.

Perhaps the cursor position needs to be updated even when scrollback is scrolled, to keep it at the shell prompt. (when not unscrolling past the cursor position)

---

Also, for "unscroll", the common meaning of this word seems to be:
"To open or unfold progressively, as a scroll does."

I felt some confusion when I learned the name of this feature.
I don't have any ideas, just letting you know.

### DIFF
--- a/kitty/rc/scroll_window.py
+++ b/kitty/rc/scroll_window.py
@@ -40,9 +40,8 @@ class ScrollWindow(RemoteCommand):
         if amt not in ('start', 'end'):
             pages = 'p' in amt
             unscroll = 'u' in amt
-            amt = amt.replace('p', '')
             mult = -1 if amt.endswith('-') and not unscroll else 1
-            q = int(amt.replace('-', ''))
+            q = int(amt.rstrip('+-pu'))
             amount = q * mult, 'p' if pages else ('u' if unscroll else 'l')
 
         return {'match': opts.match, 'amount': amount}


kovidgoyal:
unscroll is meant to be used mostly by the shell itself. So the shell
would issue an unscroll for the number of lines it wants unscrolled, and
then similarly move the cursor. If you are doing it yourself, then it's
up to you to move the cursor yourself too. 

If you wish you can add an option to the unscroll remote control command that also moves
the cursor by the specified amount.

page-down:
Thanks for the explanation. One command to accomplish a single task.

After the remote control unscroll, it seems that send-text can be executed to move the cursor.

page-down:
Wait, send-text is sent to the program inside the window. So is it possible to send escape code to the terminal (kitty window) via remote control?

If it needs to be implemented, does this make sense?

Maybe I should just add the option, move the cursor, and call it a day.
Yes, I know it's better to issue the escape codes from the shell, easy and simple.

kovidgoyal:
On Sat, Dec 11, 2021 at 12:14:07AM -0800, page-down wrote:
> Wait, send-text is sent to the program inside the window. So is it possible to send escape code to the terminal (kitty window) via remote control?

yes, IIRC send text works with escape codes.

PR #4344, opened 2021-12-11
Docs: Enable kbd transform when building dirhtml with sphinx

page-down:
I found that when building the docs, make html and make dirhtml end up with different kbd styles.

This PR enables sphinx KeyboardTransform builtin extension. The final result will be the same.

related:

https://github.com/sphinx-doc/sphinx/issues/7530

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd#representing_keystrokes_within_an_input

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -22,7 +22,10 @@ from pygments.token import (  # type: ignore
     Comment, Keyword, Literal, Name, Number, String, Whitespace
 )
 from sphinx import addnodes, version_info  # type: ignore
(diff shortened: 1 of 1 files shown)


kovidgoyal:
You should perhaps report the issue upstream to sphinx, they are fairly
responsive, as monkey patching their code is likely to break in the
future.

page-down:
Of course. This would be best fixed in upstream. Hopefully this monkey patch can be dropped soon.

Issue #4349, opened 2021-12-12
lost focus after toggle_fullscreen


shuxiao9058:
**Describe the bug**
lost focus after toggle_fullscreen, can not input text with kitty, must use mouse or switch to kitty again.

**To Reproduce**
1. from none fullscreen toggle_fullscreen to fullscreen or from fullscreen to none fullscreen.

**Environment details**

```
kitty 0.23.1 created by Kovid Goyal
Darwin jiyadeM1.local 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000 arm64
ProductName:	macOS ProductVersion:	12.0.1 BuildVersion:	21A559
Loaded config files:
  /Users/jiya/.config/kitty/kitty.conf

Config options different from defaults:
active_tab_font_style              (True, False)
bell_on_tab                        False
bold_font                          Cascadia Mono PL Bold
bold_italic_font                   Cascadia Mono PL Bold Italic
click_interval                     0.5
copy_on_select                     clipboard
enable_audio_bell                  False
font_family                        Cascadia Mono PL
font_size                          13.0
hide_window_decorations            1
input_delay                        2
italic_font                        Cascadia Mono PL Italic
kitty_mod                          8
macos_option_as_alt                3
macos_quit_when_last_window_closed True
mouse_hide_wait                    0.5
repaint_delay                      20
scrollback_lines                   10000
startup_session                    /Users/jiya/.config/kitty/startup_session
strip_trailing_spaces              smart
symbol_map:
	U+e0a0 - U+e0a3 → PowerlineSymbols
	U+e0c0 - U+e0c7 → PowerlineSymbols
sync_to_monitor                    False
tab_bar_edge                       1
tab_bar_style                      powerline
tab_powerline_style                slanted
tab_title_template                 {index}:{title}
url_style                          2
window_border_width                (0, 'pt')
Added shortcuts:
	alt+enter KeyAction('toggle_fullscreen')
	cmd+. KeyAction('move_tab_forward')
	cmd+[ KeyAction('previous_window')
	cmd+] KeyAction('next_window')
	cmd+` KeyAction('move_window_to_top')
	cmd+a > 1 KeyAction('set_background_opacity', ('1',))
	cmd+a > d KeyAction('set_background_opacity', ('default',))
	cmd+a > l KeyAction('set_background_opacity', ('-0.1',))
[...]

PR #4350, opened 2021-12-12
macOS: Fix keyboard input not working after toggling full screen

page-down:
```shell
kitty --config=NONE -o hide_window_decorations=yes
```

Issue:
https://github.com/kovidgoyal/kitty/issues/4349

I am not sure if this is the best way to handle this, please review, thanks.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1644,6 +1644,9 @@ void _glfwPlatformUpdateIMEState(_GLFWwindow *w, const GLFWIMEUpdateEvent *ev) {
     [self setResizeIncrements:NSMakeSize(1.0, 1.0)];
     [super toggleFullScreen:sender];
     [self setResizeIncrements:original];
+    // When the window decoration is hidden, toggling fullscreen causes the style mask to be changed,
+    // and causes the first responder to be cleared.
+    if (glfw_window && !glfw_window->decorated && glfw_window->ns.view) [self makeFirstResponder:glfw_window->ns.view];
 }
 
 @end


kovidgoyal:
You should probably only do this if the window doesnt have decorations as it may have other side effects.

kovidgoyal:
And add a comment explaining why it is neccessary

page-down:
I found that there is still a problem.

1. `kitty --config=NONE -o hide_window_decorations=yes`
2. Press `ctrl+up` to bring up Mission Control.
3. Move your mouse to the top window list thumbnail, hold down `Option` and click the `><` button in the top left corner of the window to restore the full screen window to a window.
4. Mouse click on the window that changes from full screen under Mission Control.
5. The keyboard input is not responsive at this point.

This issue does not occur when `hide_window_decorations` is not used, everything works fine.
Does it need to be handled elsewhere?

kovidgoyal:
No idea, does it not call toggleFullscreen?

page-down:
All of the following are toggling full screen by dragging the window in the mission control.

1) `kitty --config=NONE -o resize_in_steps=yes ...`
- `toggleFullScreen()` will not be called.
- Keyboard input is normal.
- Window size is missing a few pixels causing a missing line.

2) `kitty --config=NONE -o hide_window_decorations=yes -o resize_in_steps=yes ...`
- `toggleFullScreen()` will not be called.
- Keyboard input not working.
- The number of lines is normal.

kovidgoyal:
Well if cocoa is not calling toggleFullscreen I dont see what we can do about it. The bug needs to be fixed by Apple.

page-down:
Hopefully, most users use the keyboard shortcuts rather than dragging windows around.

Personally, I don't worry about that. Time is too limited to make everything perfect.

PR #4360, opened 2021-12-15
Fix missing `fmt.bg.normal` in tab title template

thanhph111:
I'm currently using the stable version and realized that `{fmt.bg.normal}` was missing, which is one of [documentation example](https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.tab_title_template). I think it's there to reset any background color changes to match the current tab background color. Just close the PR if you've already implemented this.

Here the result (the top windows is the fixed version):

Demo

I injected the `normal` field in the `draw_title` instead of passing some conditions to the lower code, which I think is too over engineering, but too many special cases like this in the future may pollute the code. Let me know if that's not your intention.

### DIFF
--- a/kitty/tab_bar.py
+++ b/kitty/tab_bar.py
@@ -179,6 +179,8 @@ def draw_title(draw_data: DrawData, screen: Screen, tab: TabBarData, index: int)
             'sup': SupSub(data),
             'sub': SupSub(data, True),
         }
+        normal_bg = draw_data.active_bg if tab.is_active else draw_data.inactive_bg
+        eval_locals['fmt'].bg.normal = f'\x1b[4{"8" + color_as_sgr(normal_bg)}m'
         title = eval(compile_template(template), {'__builtins__': {}}, eval_locals)
     except Exception as e:
         report_template_failure(template, str(e))


page-down:
This looks like it needs to be handled in the Formatter / ColorFormatter.
There is also `active_fg` and `inactive_fg` that can be reset (e.g. `fmt.fg.normal`).

thanhph111:
Yeah, but I think we have to send `tab.is_active` down, which is too much for one field. If we have a way to take that as one shot, I'm glad to implement that. I really haven't had a deep dig to the code so correct me if I'm wrong.

kovidgoyal:
Yes, that is correct. But, there is no "normal". If you want there to be
a way to switch the background back to the current tab_bg support for
that should be added to ColorFormatter as a new field named "tab" to
supplement the existing "default"

kovidgoyal:
I have committed an implementation, without testing it. Let me know if
it works.

page-down:
```python
class ColorFormatter:
    tab: TabBarData

    def __getattr__(self, name: str) -> str:
        # ...
        elif q == 'tab':
        # ...
```

It looks like a different name is needed. Otherwise, the TabBarData repr is output directly when getting the `tab` property.

kovidgoyal:
right, fixed.

thanhph111:
Work perfectly now, thanks.

PR #4377, opened 2021-12-20
Minor robustness improvements to zsh shell integration

romkatv:
This fixes a few minor issues in zsh shell integration. If this is something you are interested in, I can have a pass over `kitty.zsh`. That one can have more substantial improvements.


romkatv:
Forgot to mention: I sent this PR because I thought you might find it useful. It's not something that I have a personal need for. So if you don't find these changes useful, it's totally fine to close or ignore the PR.

PR #4378, opened 2021-12-21
Add light and dark to the macOS title bar option

page-down:
Support for explicitly setting the window title bar style for macOS.

When the system appearance of macOS is set to Auto, you can set the appearance of kitty to a fixed Light or Dark.

### DIFF
--- a/gen-config.py
+++ b/gen-config.py
@@ -31,7 +31,7 @@ def main() -> None:
             if opt.parser_func.__name__ in ('to_color_or_none', 'cursor_text_color'):
                 nullable_colors.append(opt.name)
                 all_colors.append(opt.name)
-            elif opt.parser_func.__name__ in ('to_color', 'macos_titlebar_color'):
+            elif opt.parser_func.__name__ in ('to_color', 'titlebar_color', 'macos_titlebar_color'):
                 all_colors.append(opt.name)
     patch_color_list('kitty/rc/set_colors.py', nullable_colors, 'NULLABLE')
     patch_color_list('kittens/themes/collection.py', all_colors, 'ALL', ' ' * 8)
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -620,6 +620,16 @@ cocoa_set_activation_policy(bool hide_from_tasks) {
 }
 
 void
+cocoa_set_titlebar_appearance(void *w, unsigned int theme)
+{
+    if (!theme) return;
+    @autoreleasepool {
+        NSWindow *window = (NSWindow*)w;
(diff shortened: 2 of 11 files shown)


kovidgoyal:
What's the use case for this? Why would you want kitty to use a different color scheme than the rest of the system?

page-down:
Because I generally use dark themes in kitty.
This will keep the title bar of the window dark as well, but it doesn't need to be set to the background color.
It is consistent with users's kitty theme (dark, or light) and does not need to be consistent with the whole system.

page-down:
Talking about themes, do you have more considerations for compatibility between version upgrades and themes?

For example using a higher version of a theme on a lower version. Since themes don't have version numbers, kitty will report an error on version mismatch if a new configuration option is used in it.

As far as I know, kitten themes will be pulling the latest kitty-themes from GitHub, and after the new version is released, older versions of kitty will probably pull new themes that use the new options.

I'm still holding `visual_bell_color` and may add it to the kitty-themes template when the new version is released.

kovidgoyal:
On Tue, Dec 21, 2021 at 12:34:17AM -0800, page-down wrote:
> Talking about themes, do you have more considerations for compatibility between version upgrades and themes?
> 
> For example using a higher version of a theme on a lower version. Since themes don't have version numbers, kitty will report an error on version mismatch if a new configuration option is used in it.

Not something I care enough about. Errors in the conf file are reported
to STDERR IIRC and can simply be ignored.

> 
> As far as I know, kitten themes will be pulling the latest kitty-themes from GitHub, and after the new version is released, older versions of kitty will probably pull new themes that use the new options.
> 
> I'm still holding `visual_bell_color` and may add it to the kitty-themes template when the new version is released.

Feel free to add it now. Unrecognized keys in the conf file are a very
small problem indeed.

page-down:
> Not something I care enough about. Errors in the conf file are reported
> to STDERR IIRC and can simply be ignored.

That's what I thought before, it's not worth putting too much effort into error handling for options.

However this option is a bit different, the error is shown in the overlay. Maybe I should do something about that and have it reported in stderr.

```text
Errors parsing configuration

In file <override>:
1:Invalid color name: dark in line: macos_titlebar_color dark

Press Enter to close.
```

> Feel free to add it now. Unrecognized keys in the conf file are a very
> small problem indeed.

Thanks. Then I'll stop worrying about that.

kovidgoyal:
On Tue, Dec 21, 2021 at 12:44:39AM -0800, page-down wrote:
> > Not something I care enough about. Errors in the conf file are reported
> > to STDERR IIRC and can simply be ignored.
> 
> That's what I thought before, it's not worth putting too much effort into error handling for options.
> 
> However this option is a bit different, the error is shown in the overlay. Maybe I should do something about that and have it reported in stderr.

Yes just change the parse function to catch the error and report to
stderr and return default value.

PR #4386, opened 2021-12-22
Rewrite zsh integration

romkatv:
Note:

- Very lightly tested.
- Docs not updated.

### DIFF
--- a/shell-integration/zsh/.zshenv
+++ b/shell-integration/zsh/.zshenv
@@ -1,11 +1,15 @@
+# This file can get sourced with aliases enabled. To avoid alias expansion
+# we quote everything that can be quoted. Some aliases will still break us
+# though.
+
 # Don't use [[ -v ... ]] because it doesn't work in zsh < 5.4.
-if (( ${+KITTY_ORIG_ZDOTDIR} )); then
+if [[ -n "${KITTY_ORIG_ZDOTDIR+X}" ]]; then
     # Normally ZDOTDIR shouldn't be exported but it was in the environment
     # of Kitty, so we export it.
-    export ZDOTDIR=$KITTY_ORIG_ZDOTDIR
-    unset KITTY_ORIG_ZDOTDIR
+    'builtin' 'export' ZDOTDIR="$KITTY_ORIG_ZDOTDIR"
+    'builtin' 'unset' 'KITTY_ORIG_ZDOTDIR'
 else
-    unset ZDOTDIR
+    'builtin' 'unset' 'ZDOTDIR'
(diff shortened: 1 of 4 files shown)


romkatv:
The same disclaimer as in #4377 applies here. I personally don't need these changes. If you don't like them, it won't upset me if you close or ignore this PR.

romkatv:
I've benchmarked the new and the current integration code on Linux x86 and Darwin arm64.

The impact on startup lag (the delay when you start a new shell) went down about by 2x on Linux and by at least 100x on Mac (the latter basically went down to zero). The impact on command lag (the delay every time you press enter) went down by 5x on Linux and 20x on Mac.

I think at the point performance impact is not a serious concern.

kovidgoyal:
I have merged, but I do have one concern. Namely that this leaves us
with a "kitty-integration" function defined. This is annoying because
tab completing kitty now gives kitty-integration as an option. So can we
either hide it which would be ideal (but I dont know if that's possible)
or name it something less likely to conflict when completing.

romkatv:
Have you tested it? I tested it very little.

By the way, does Kitty have features that depend on extracting the command between OSC 133 B and C? They won't work with zsh because of right prompt and zle redisplay. As far as zsh goes, OSC 133 B and OSC A with k=s aren't useful.

romkatv:
How about `zsh-kitty-integration`?

kovidgoyal:
On Thu, Dec 23, 2021 at 06:54:44AM -0800, Roman Perepelitsa wrote:
> Have you tested it? I tested it very little.

Yes, it works fine with my setup.

> 
> By the way, does Kitty have features that depend on extracting the command between OSC 133 B and C? They won't work with zsh because of right prompt and zle redisplay. As far as zsh goes, OSC 133 B and OSC A with k=s aren't useful.

No, kitty uses only A and C, see the notes for shell developers section
of shell-integration.rst.

As for zsh-kitty-integration its fine by me personally, but wont that
annoy people that are tab completing zsh? I dont knwo what zsh's
restrictions on autoloaded function names are, but if we can start it
with a period or underscore or hyphen or something like that, it would
be best I think.

romkatv:
We can unfunction it. I'll send you a PR later today.

romkatv:
> We can unfunction it. I'll send you a PR later today.

Sent https://github.com/kovidgoyal/kitty/pull/4388.

page-down:
Unrelated to the zsh integration:
I noticed the initial capitalized `Kitty`, should it be a uniform `kitty`? I made the change in the new PR.

PR #4390, opened 2021-12-24
Docs: Add zsh unfunction kitty-integration, recent changes and minor tweaks

page-down:
- Add unfunction kitty-integration to the shell integration documentation.
- Unified use of lowercase names for `kitty`.
- Add some recent changes to the changelog.
- I previously introduced  Unicode left single quote in the docs and now replaced it with ASCII.
- Other minor changes.


kovidgoyal:
merged

Issue #4352, opened 2021-12-13
Kitty fails to build through Nix on M1Max/macOS 12


alexghr:
**Describe the bug**
Kitty fails to build through Nix on 2021 Macbook Pro (M1Max) running macOS 12.

**To Reproduce**
1. clone repo
2. `nix-shell` or `nix-shell --pure`
3. `make app` inside the new shell

(alternatively try installed via `nix-env` or `darwin-configuration.nix`)

<details>
  <summary>Compile time error: `'NSUserNotificationCenter' is deprecated`</summary>

```
[10:37:51] ~/code/pers/kitty (master)$ nix-shell --pure

[nix-shell:~/code/pers/kitty]$ make app
python3 setup.py kitty.app
CC: ['clang'] (11, 1)
[23/32] Compiling [cocoa] glfw/posix_thread.c ... done
Compiling kitty/cocoa_window.m ...
clang -MMD -DNDEBUG -DPRIMARY_VERSION=4000 -DSECONDARY_VERSION=23 -DXT_VERSION="0.23.1" -DKITTY_USE_DEPRECATED_MACOS_NOTIFICATION_API -DGL_SILENCE_DEPRECATION -Wextra -Wfloat-conversion -Wno-missing-field-initializers -Wall -Wstrict-prototypes -std=c11 -pedantic-errors -Werror -O3 -fwrapv -fstack-protector-strong -pipe -fvisibility=hidden -D_FORTIFY_SOURCE=2 -DKITTY_HAS_RS_SIG_ARGS -pthread -I/nix/store/kncv7vlvpirwcb3s6ys4kfaxdyaq8m6c-zlib-1.2.11-dev/include -I/nix/store/gr768cjnb8kj6l0rn4nm9rf8ad77j96i-libpng-apng-1.6.37-dev/include/libpng16 -I/nix/store/ay8y8r4n7w9p7sx47ndaklsp5m3z63ba-lcms2-2.12-dev/include -I/nix/store/fxz7zs8qnwc409jrxd0a4gdrmdlhcj9y-harfbuzz-3.0.0-dev/include/harfbuzz -I/nix/store/cdq84bgzsvy2k991qjrj0vzprn9f9b46-python3-3.9.6/include/python3.9 -c kitty/cocoa_window.m -o build/fast_data_types-cocoa_window.m.o
[...]

PR #4399, opened 2021-12-28
shell.nix: add dependency needed on Apple silicon

Luflosi:
shell.nix is missing a depencency on the UserNotifications framework on macOS with Apple silicon. Unfortunately we can't just unconditionally include this dependency because Nixpkgs uses a different macOS SDK version for macOS on an Intel processor compared to macOS on Apple silicon. The older SDK version 10.12 on Intel does not have the UserNotifications framework, which would lead to an "attribute missing" error when trying to execute `nix-shell`. The condition can be dropped when the macOS SDK version for Intel processors is finally updated. See https://github.com/NixOS/nixpkgs/issues/101229 for progress on that.
See https://github.com/NixOS/nixpkgs/pull/137512 for more context.
Closes https://github.com/kovidgoyal/kitty/issues/4352.

@kovidgoyal Please don't merge this until someone with an M1 Mac has actually tested this.
@alexghr could you please test this and confirm that it fixes the problem? I don't have an M1 Mac to test with myself.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -24,6 +24,8 @@ mkShell rec {
     libpng
     python3
     zlib
+  ] ++ lib.optionals (stdenv.isDarwin && (builtins.hasAttr "UserNotifications" darwin.apple_sdk.frameworks)) [
+    darwin.apple_sdk.frameworks.UserNotifications
   ] ++ optionals stdenv.isLinux [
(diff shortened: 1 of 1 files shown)


alexghr:
Can confirm it builds on a M1Max Macbook 🙌 
kitty-build

PR #4402, opened 2021-12-29
Fix cannot use colon in command strings with `combine`

thanhph111:
I have this config and it fails due to the `:` in `:edit`:
```vim
map ctrl+shift+k>v>c combine : launch --cwd=current --type=tab vim -c ':edit $MYVIMRC': remote_control set-spacing padding=0
```
This would be an easy fix with a new regex to ignore `:` in string.


kovidgoyal:
If you wish to use colons in the arguments to combined commands, use a
different separator character, combine can use any character as a
separator. For example ; or | or * or & or whatever.

thanhph111:
My bad, should have looked closer.

Issue #4367, opened 2021-12-17
Error activation tab bar


zzhaolei:
**Describe the bug**
When you click on the border at the bottom of the window, the top corresponding tab is activated.

**To Reproduce**
...

**Screenshots**
![image](https://user-images.githubusercontent.com/17561793/146502172-518ef25c-0138-4f23-a4da-d6e4fd153901.png)

**Environment details**
```
kitty --config=NONE -o tab_bar_edge top
```

**Additional context**
...

PS: Sorry, My English is terrible

kovidgoyal:
I cannot reproduce running from master. I suggest you try the kitty
nightly build.

page-down:
Oh, good catch. I can reproduce the issue under macOS.

```shell
kitty --config=NONE -o tab_bar_edge=top
```

- Create a new tab and a new kitty window.
- Switch to the tab with the two kitty windows.
- Click on the bottom edge of the window shown at the screenshot.

There is about 1px ~ 2px, and clicking the mouse will switch the tab on the top tab bar corresponding to the x-coordinate.

zzhaolei:
Night builds have the same problem. I made a video

https://user-images.githubusercontent.com/17561793/146511876-f14c77c2-3bb8-4226-9535-06f91ef6df33.mp4

zzhaolei:
I found another problem. In the full screen, place the mouse in the bottom left corner of the screen and double-click the left mouse button. This creates a new tab:

https://user-images.githubusercontent.com/17561793/146512847-26e5ea9d-1fec-4136-9cb8-43d787be1caa.mp4

zzhaolei:
Both problems are more likely to recur in full screen mode

kovidgoyal:
@page-down since you can repro it it should be an easy fix, relevant code is in mouse.c window_for_event

Luflosi:
The problem was introduced in caa44b3d4a0c531cd85d8d865dc33827638bd26d, specifically the change to `window_for_event()`. Another side-effect of this change is that scrolling with the mouse wheel doesn't work in kitty when a different application is focused on macOS.

kovidgoyal:
Note, I have not tested the fix, so please test.

page-down:
I can confirm that the latest commit fixes this issue.
When another window gets focus, scrolling the mouse wheel in the window (with cursor focused) also responds properly.

[...]

PR #4422, opened 2022-01-04
macOS: Fix mouse scrolling multi-panel program when no focus

page-down:
Allow scrolling of the mouse pointing part of a multi-panel program when not focused under macOS.

https://github.com/kovidgoyal/kitty/issues/4367#issuecomment-1004641539

I'm a little concerned about the performance impact of these redundant update steps. I have not found any other good way to do this.

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -795,6 +795,16 @@ scroll_event(double UNUSED xoffset, double yoffset, int flags, int modifiers) {
         if (t) w = t->windows + t->active_window;
     }
     if (!w) return;
+    // Also update mouse cursor position while kitty OS window is not focused.
+    // Allow scrolling of the mouse pointing part of a multi-panel program when not focused under macOS.
+    if (!osw->is_focused) {
+        unsigned int x = 0, y = 0;
+        bool in_left_half_of_cell;
+        if (cell_for_pos(w, &x, &y, &in_left_half_of_cell, osw)) {
+            w->mouse_pos.cell_x = x; w->mouse_pos.cell_y = y;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm confused, why do we need to send a fake mouse move event to the kitty window on a scroll event? This has lots of side effects, like switching focus to the window, possibly highlighting a URL, sending events to the child in the window (though I assume this is what you want to do?).

page-down:
Yes, this is equivalent to compensating for the lack of mouse movement events under macOS without focus. Do you think it can be guarded in specific scenarios?

- Only effective under macOS.
- If and only if all OS windows have no focus.

kovidgoyal:
I dont understand why we would need to move the mouse on a scroll event though? If you want the scroll event delivered to the child, just deliver it to the child, why also send mouse move events?

page-down:
The mouse cursor position in the kitty window is not updated and the program in it does not scroll the correct part.

Oh, there is also a problem that every scroll will handle this fake move event even if the mouse is not moved. I really don't like this.

Maybe we don't need to think so much in this case, just update mouse_pos cell_x, cell_y and that's it.

Or can I send the scroll to the child with the cell_x cell_y coordinates?

Do you think this is reasonable?

kovidgoyal:
Just updating the cell position should be sufficient, IIRC the code to
send the event to the child will use it automatically.

page-down:
Ok, I have updated the PR, please review if this is appropriate, thank you.

I hope there is no problem with high precision mode or something like that. For unfocused scrolling, there really isn't that much to think about.
The most common use case I think is to scroll through scrollback or text editor.

PR #4442, opened 2022-01-05
Minor improvements in zsh integration

romkatv:
- Replace ff63e58 with a better fix.
- Make writing to the TTY more robust.
- Remove PS2 patching as it has no effect.

I would very much prefer if kitty didn't inject zsh code into the shell (at least by default) but I don't think I can change your mind on it. The next best option is to make the injected code less likely to cause issues. I can help with the latter if you ping me whenever issues with zsh integration arise.

### DIFF
--- a/shell-integration/zsh/kitty-integration
+++ b/shell-integration/zsh/kitty-integration
@@ -35,7 +35,7 @@ builtin typeset -gi _ksi_state
 # Asks kitty to print $@ to its stdout. This is for debugging.
 _ksi_debug_print() {
     builtin local data
-    data=$(command base64 <<<"${(j: :}@}") || builtin return
+    data=$(command base64 <<<"${(j: :)@}") || builtin return
     builtin printf '\eP@kitty-print|%s\e\\' "${data//$'\n'}"
 }
 
--- a/shell-integration/zsh/kitty-integration
+++ b/shell-integration/zsh/kitty-integration
@@ -32,11 +32,42 @@ builtin emulate -L zsh -o no_warn_create_global
 # 2: none of the above.
 builtin typeset -gi _ksi_state
 
-# Asks kitty to print $@ to its stdout. This is for debugging.
+# Attempt to create a writable file descriptor to the TTY so that we can print
+# to the TTY later even when STDOUT is redirected. This code is fairly subtle.
+#
+# - It's tempting to do `[[ -t 1 ]] && exec {_ksi_state}>&1` but we cannot do this
+#   because it'll create a file descriptor >= 10 without O_CLOEXEC. This file
+#   descriptor will leak to child processes.
+# - If we do `exec {3}>&1`, the file descriptor won't leak to the child processes
+#   but it'll still leak if the current process is replaced with another. In
+#   addition, it'll break user code that relies on fd 3 being available.
+# - Zsh doesn't expose dup3, which would have allowed us to copy STDOUT with
+#   O_CLOEXEC. The only way to create a file descriptor with O_CLOEXEC is via
+#   sysopen.
+# - `zmodload zsh/system` and `sysopen -o cloexec -wu _ksi_fd -- /dev/tty` can
+#   fail with an error message to STDERR (the latter can happen even if /dev/tty
+#   is writable), hence the redirection of STDERR. We do it for the whole block
+#   for performance reasons (redirections are slow).
+# - We must open the file descriptor right here rather than in _ksi_deferred_init
+#   because there are broken zsh plugins out there that run `exec {fd}< <(cmd)`
+#   and then close the file descriptor more than once while suppressing errors.
+#   This could end up closing our file descriptor if we opened it in
+#   _ksi_deferred_init.
+typeset -gi _ksi_fd
+{
+    zmodload zsh/system && (( $+builtins[sysopen] )) && {
+        { [[ -w     $TTY ]] && sysopen -o cloexec -wu _ksi_fd --     $TTY } ||
+        { [[ -w /dev/tty ]] && sysopen -o cloexec -wu _ksi_fd -- /dev/tty }
+    }
+} 2>/dev/null || (( _ksi_fd = 1 ))
+
+# Asks kitty to print $@ to its STDOUT. This is for debugging.
 _ksi_debug_print() {
     builtin local data
     data=$(command base64 <<<"${(j: :)@}") || builtin return
-    builtin printf '\eP@kitty-print|%s\e\\' "${data//$'\n'}"
+    # Removing all spaces rather than just \n allows this code to
+    # work on broken systems where base64 outputs \r\n.
+    builtin print -nu "$_ksi_fd" '\eP@kitty-print|'"${data//[[:space:]]}"'\e\\'
 }
 
 # We defer initialization until precmd for several reasons:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Why do you say PS2 is not used??

romkatv:
Maybe I'm missing something. What effect do those marks have? I cannot think of any use of them.

kovidgoyal:
They are used to detect the extent of multi-line prompts for mouse
clicking. Might be used for other things as well, I dont recall, please
restore them.

romkatv:
> please restore them.

Done.

I'm reasonably sure that these marks have no effect. Zle is already marked with `A` at the beginning and `C` at the end. Extra `A;k=s` marks in the middle of zle don't convey any information. If there were `B` marks and no right prompt, then `A;k=s` could be used to separate prompt from the command line (the latter is called `BUFFER` in zsh).

Am I missing something? Can you devise a test via kitty UI that will allow you to infer whether `A;k=s` marks are being printed?

kovidgoyal:
If you want to check if marks are being printed you can do either of the
following

1) Run kitty --dump-commands
this will show what escape codes are sent,with symbolic names

2) Use
kitty -o 'map f1 dump_lines_with_attrs'

and then press f1 to see an overlay that shows what marks are applied to
each line

romkatv:
> They are used to detect the extent of multi-line prompts for mouse clicking.

Can describe a test based on this feature that would allow one to tell whether `A;k=s` are being printed?

page-down:
>> They are used to detect the extent of multi-line prompts for mouse clicking.
>
> Can you describe a test based on this feature that would allow one to tell whether A;k=s are being printed?

Not sure if this is what you need:

kitty/screen.c
```c
if (PyUnicode_CompareWithASCIIString(token, "k=s") == 0) *pk = SECONDARY_PROMPT;
```

- kitty -o 'map f1 dump_lines_with_attrs'
- type: `test '<enter>`, `abc<enter>`, `def<enter>`
- press f1

```text
0: prompt 
zsh> test '
1: secondary_prompt continued
quote> abc
2: secondary_prompt 
quote> def
```

romkatv:
I'm asking which features are affected by these marks.

@kovidgoyal has already removed `B` marks from zsh integration because they have no effect on kitty's functionality and there is no need to pay for things that don't have any effect. I'm saying that by the same logic `A;k=s` marks should also be removed.

If there is any functionality in kitty (I'm not talking about dumping the raw TTY content but a user facing feature) that works differently with and without these marks, could you describe how I can verify this?

kovidgoyal:
grep the code for SECONDARY_PROMPT that will tell you where it is
currently used.

kovidgoyal:
And note that B marks were removed not just because they are unused but also because removing them allows no-cursor to avoid hooking ZLE at all.

romkatv:
I'd really love if you gave me instructions that involve using kitty only through its end-user interface that produce different results when marks are printed vs not printed (or confirmed that the end-user interface is indeed unaffected by these marks). You aren't giving me these instructions, so let me show why I think it's impossible to use `A;k=s` marks for anything in the absence of `B` marks. Consider this TTY output:

```text
A 1 2
3 4
A;k=s 5 6
7 8
C```

Here `A` `A;k=s` and `C` are marks. The rest is just text.

What can kitty do with this that it couldn't do without `A;k=s`? Perhaps it can infer which command is about to run? Nope. Maybe it can infer where the primary prompt ends? Nope. Maybe it can figure out where the second prompt ends? Nope.

Maybe only `5` is a part of `PS2` and the rest is a command (on two lines). Or maybe only `8` is the command and the rest (on two lines) is secondary prompt.

The only thing `A;k=s` tells kitty is that there was secondary prompt. What feature can kitty provide that would need this information?

I should also note that adding `B` marks doesn't make matters any better because of right prompt. It'll still be impossible to separate the command line from prompt based on just marks. If kitty ever implements a feature that requires knowing the command that's about to get executed, this would require sending the actual text of the command in `preexec`.

Edit: Saw your new comment after writing this.

> grep the code for SECONDARY_PROMPT that will tell you where it is currently used.

Looking at the code I'm now absolutely certain that secondary marks have no effect.

kovidgoyal:
Quite simply, when clicking with the mouse on the primary prompt line and the cursor is on the secondary prompt line, kitty can use that information, to send a up arrow keypress to move the cursor since IIRC zsh doesnt allow left arrow to move from PS2 to PS1 lines. This is not currently implemented in kitty, but that is an example of where this information can come in handy.

romkatv:
> Quite simply, when clicking with the mouse on the primary prompt line and the cursor is on the secondary prompt line, kitty can use that information, to send a up arrow keypress to move the cursor since IIRC zsh doesnt allow left arrow to move from PS2 to PS1 lines.

Do I understand it correctly that currently secondary marks have no effect? If so, why are they being printed?

kovidgoyal:
On Wed, Jan 05, 2022 at 08:44:17AM -0800, Roman Perepelitsa wrote:
> > Quite simply, when clicking with the mouse on the primary prompt line and the cursor is on the secondary prompt line, kitty can use that information, to send a up arrow keypress to move the cursor since IIRC zsh doesnt allow left arrow to move from PS2 to PS1 lines.
> 
> Do I understand it correctly that currently secondary marks have no effect? If so, why are they being printed?

Because at some point they will be used and I dont want to have to
modify the crazy mess that is the shell integration scripts at that time
again.

romkatv:
> Because at some point they will be used and I dont want to have to modify the crazy mess that is the shell integration scripts at that time again.

I left the code in there. It's commented out and has an explanation why. It won't be very difficult to uncomment the two lines.

Can you clarify what that future feature would do? Sending the UP arrow key at secondary prompt will not move the cursor up (that's impossible). It'll invoke a zle widget that's bound to UP.  For me that would be `z4h-up-substring-local`, which searches history for a command that contains all tokens that I've already typed.

If you are getting tired of this discussion (to be honest, I am), say so and we'll end it.

kovidgoyal:
On Wed, Jan 05, 2022 at 08:53:16AM -0800, Roman Perepelitsa wrote:
> > Because at some point they will be used and I dont want to have to modify the crazy mess that is the shell integration scripts at that time again.
> 
> I left the code in there. It's commented out and has an explanation why. It won't be very difficult to uncomment the two lines.
> 
> Can you clarify what that future feature would do? Sending the UP arrow key at secondary prompt will not move the cursor up (that's impossible). It'll invoke a zle widget that's bound to UP.  For me that would be `z4h-up-substring-local`, which searches history for a command that contains all tokens that I've already typed.

Presumably there is *some* way to edit text in the primary prompt line
in zsh, whatever that way is kitty can potentially use.

> 
> If you are getting tired of this discussion (to be honest, I am), say so and we'll end it.

Indeed, I am.

romkatv:
> Presumably there is *some* way to edit text in the primary prompt line in zsh.

No, there isn't.

page-down:
@romkatv 
Since the discussion has been long, I also did a little testing.

If I comment out the PS2 `k=s` mark, after resizing the window, there will be empty lines except for the current prompt line. If it has `A;k=s`, everything works fine.



Is it what you need this time?

romkatv:
@page-down Thank you! Yes, that's exactly what I was looking for. I was wrong. In retrospect, this seems obvious (and I suppose it may have been to you) but I totally missed this.

page-down:
I'm glad that the tests I've done have answered your questions.
And thanks for your contribution to the zsh integration.

PR #4478, opened 2022-01-08
macOS: Allow mapping secure keyboard entry keyboard shortcut

page-down:
Allows to customize the key combinations for the `secure keyboard entry` feature.
The original hard-coded key combination can be rebinded by the user.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -607,135 +607,6 @@ translateKeyToModifierFlag(uint32_t key)
 static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 
-// SecureKeyboardEntryController {{{
-@interface SecureKeyboardEntryController : NSObject
-
-@property (nonatomic, readonly) BOOL isDesired;
-@property (nonatomic, readonly, getter=isEnabled) BOOL enabled;
-
-+ (instancetype)sharedInstance;
-
-- (void)toggle;
-- (void)update;
-
-@end
-
-@implementation SecureKeyboardEntryController {
-    int _count;
-    BOOL _desired;
-}
-
-+ (instancetype)sharedInstance {
-    static id instance;
-    static dispatch_once_t onceToken;
-    dispatch_once(&onceToken, ^{
-        instance = [[self alloc] init];
-    });
-    return instance;
-}
-
-- (instancetype)init {
-    self = [super init];
-    if (self) {
-        _desired = false;
-
-        [[NSNotificationCenter defaultCenter] addObserver:self
-                                                 selector:@selector(applicationDidResignActive:)
-                                                     name:NSApplicationDidResignActiveNotification
-                                                   object:nil];
-        [[NSNotificationCenter defaultCenter] addObserver:self
-                                                 selector:@selector(applicationDidBecomeActive:)
-                                                     name:NSApplicationDidBecomeActiveNotification
-                                                   object:nil];
-        if ([NSApp isActive]) {
-            [self update];
-        }
-    }
-    return self;
-}
-
-#pragma mark - API
-
-- (void)toggle {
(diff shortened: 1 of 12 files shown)


kovidgoyal:
convert the action into an actual action, not a no-op and use MENU_ITEM
and PENDING to add it to the global menu bar. Then create a function to
do the toggle in kitty/cocoa_window.m like so

void cocoa_toggle_secure_input(void *w) {
   NSWindow *window = w;
   [window performSelector:toggleSecureKeyboard];
}

then wrap it in glfw.c and expose it to python in state.c (it should
take an os window id and call the glfw function with the OSWindow
pointer).

The implement the action in boss.py and in child-monitor.c use call_boss
to trigger the pending action.

page-down:
Thanks. I have a question about using `MENU_ITEM`.

With `[__mi setTarget:global_menu_target]` set in MENU_ITEM, 
glfw/cocoa_window.m `- (BOOL)validateMenuItem:(NSMenuItem *)item` will not be executed.
How should I handle it properly?

Should I rename the `MENU_ITEM` to `GLOBAL_MENU_ITEM` and define a `MENU_ITEM` with no target set?

kovidgoyal:
I'd say move the code for managing Secure keyboard from
glfw/cocoa_window.m to kitty/coco_window.m and implement 
validatemenutarget on GlobalMenuTarget.

On Sat, Jan 08, 2022 at 06:09:47PM -0800, page-down wrote:
> Thanks. I have a question about using `MENU_ITEM`.
> 
> With `[__mi setTarget:global_menu_target]` set in MENU_ITEM, 
> glfw/cocoa_window.m `- (BOOL)validateMenuItem:(NSMenuItem *)item` will not be executed.
> How should I handle it properly?
> 
> Should I rename the `MENU_ITEM` to `GLOBAL_MENU_ITEM` and define a `MENU_ITEM` with no target set?
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/4478#issuecomment-1008213208
> 
> Message ID: ***@***.***>

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

page-down:
Hmm... All goes well, however, at the end there is a problem when compiling.

After moving `SecureKeyboardEntryController` to kitty/cocoa_window.m

with `#include <Carbon/Carbon.h>`

```text
undef: _DisableSecureEventInput
undef: _EnableSecureEventInput
undef: _IsSecureEventInputEnabled
Undefined symbols for architecture x86_64:
  "_DisableSecureEventInput", referenced from:
      -[SecureKeyboardEntryController update] in lto.o
...
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1
```

Test with the following command, it compiles. Do I need to make some adjustments in setup.py?

```shell
LDFLAGS="-framework Carbon" python3 setup.py --verbose
```

I noticed the following code.
glfw/glfw.py
```python
    elif module == 'cocoa':
        ans.cppflags.append('-DGL_SILENCE_DEPRECATION')
        for f_ in 'Cocoa Carbon IOKit CoreFoundation CoreVideo'.split():
            ans.ldpaths.extend(('-framework', f_))
```

kovidgoyal:
Yes, you will need to link against carbon when building data-types.so on
macos, line 419 in setup.py

page-down:
I noticed that you closed PR and I was unable to push the code. I was under the impression that PRs could not be reopened.
In this case, should I resubmit one?
I would like to understand your intentions before I proceed further, without causing misunderstandings.

kovidgoyal:
It was closed accidentally, sorry.

page-down:
Thanks, I updated the branch.
I'm not quite sure what I'm doing with certain details, and function naming.
Please review if some cleanup needs to be done.

Issue #4491, opened 2022-01-10
How to get hints on compiler errors ? And jump into editor on error file and line ?


akharrou:
Consider this:

```zsh
❯ gcc -Wall main.c
main.c:6:3: error: implicitly declaring library function 'printf' with type 'int (const char *, ...)' [-Werror,-Wimplicit-function-declaration]
  printf("Hello world\n");
  ^
main.c:6:3: note: include the header <stdio.h> or explicitly provide a declaration for 'printf'
main.c:4:7: warning: unused variable 'bla' [-Wunused-variable]
  int bla;
      ^
1 warning and 1 error generated.
```

How can you get [hints](https://sw.kovidgoyal.net/kitty/kittens/hints/) on just `main.c:6` and open `vim` directly on that line ?

---

Bonus: can you additionally include columns in this ? i.e hint selection would include them and `vim` opens cusor directly in that file, on that line, on that column.

kovidgoyal:
https://sw.kovidgoyal.net/kitty/kittens/hints/#cmdoption-kitty-kitten-hints-linenum-action

akharrou:
> https://sw.kovidgoyal.net/kitty/kittens/hints/#cmdoption-kitty-kitten-hints-linenum-action

I tried this:

```properties
# kitty.conf

map super+y kitten hints --type=linenum --linenum-action=tab nvim +{line} {path}
```

...but I keep getting "`No matches found, press Enter to quit.`"

akharrou:
I also tried this:

```properties
# kitty.conf
    map super+y _hints --linenum-action=window --type=regex --regex="\b(?P<filepath>.+):(?P<line>\d+):(?P<column>\d+)" --program="nvim +{line} {filepath}"
```

Which gives me matches on more than just what I need, by matching the command prompt paths, and doesn't seem to do anything once I select a hint... something wrong with my usage of `--program`  ?

![Screen Shot 2022-01-10 at 11 03 39 AM](https://user-images.githubusercontent.com/36056254/148734275-38a083e6-bb07-4866-88bc-62cc048cdca6.png)

So I also tried removing `--program …` like in other examples...

```properties
# kitty.conf
    map super+y _hints --linenum-action=window --type=regex --regex="\b(?P<filepath>.+):(?P<line>\d+):(?P<column>\d+)" nvim +{line} {filepath}
```
...but that doesn't work either and I get: "`Extra command line arguments present: vim + (line} {filepath} Press Enter to quit`"

[...]

PR #4492, opened 2022-01-10
hints: Match single-letter file extensions h c m

page-down:
https://github.com/kovidgoyal/kitty/issues/4491


page-down:
Also I would like to know your opinion on the following file names, thanks.

```text
auth.log.1
syslog.1
```

Currently it will match as `auth.log`. (I don't personally have a need for it, I just found it.)

kovidgoyal:
number suffixes are too rare to care about, atleast for me

Issue #4472, opened 2022-01-07
Using macos_traditional_fullscreen yes fails to hide


npearson72:
**Describe the bug**
Using MacOS version `10.15.7 (19H1615)` (Oddly, it works fine on MacOS version `12.1 (21C52)`)

Make sure no other windows are open except for kitty.

1. Add the following to kitty config: `macos_traditional_fullscreen yes`

2. Enter full screen mode.

3. Try hiding kitty by either using the menu or hitting <kbd>⌘</kbd><kbd>H</kbd>

Expected result is that kitty will be hidden.
Actual result... it appears to go into the background behind other windows (if other windows are open), but it does not get hidden.

**Environment details**
```
kitty 0.24.1 created by Kovid Goyal
Darwin nathan.local 19.6.0 Darwin Kernel Version 19.6.0: Sun Nov 14 19:58:51 PST 2021; root:xnu-6153.141.50~1/RELEASE_X86_64 x86_64
ProductName:	Mac OS X ProductVersion:	10.15.7 BuildVersion:	19H1615
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /usr/local/bin/zsh
Loaded config files:
  /Users/nathan/.config/kitty/kitty.conf

Config options different from defaults:
adjust_column_width          -1
adjust_line_height           3
bold_font                    Hack Nerd Font Mono Bold
bold_italic_font             Hack Nerd Font Mono Bold Italic
enable_audio_bell            False
font_family                  Hack Nerd Font Mono Regular
font_size                    14.0
italic_font                  Hack Nerd Font Mono Italic
macos_thicken_font           1.0
macos_traditional_fullscreen True
shell_integration            frozenset({'no-cursor'})
visual_bell_duration         0.3
Added shortcuts:
	alt+/ send_text all \x1b\x2f
	alt+: send_text all \x1b\x3a
	alt+[ send_text all \x1b\x5b
	alt+] send_text all \x1b\x5d
	alt+_ send_text all \x1b\x5f
	alt+h send_text all \x1b\x68
	alt+j send_text all \x1b\x6a
	alt+k send_text all \x1b\x6b
	alt+l send_text all \x1b\x6c
	alt+o send_text all \x1b\x6f
[...]

PR #4496, opened 2022-01-11
Fixes traditional fullscreen hide for Mac OS >= 10.15

npearson72:
Resolves issue described here:

https://github.com/kovidgoyal/kitty/issues/4472

All credit goes to @page-down


page-down:
I don't think you should remove the original comment.
Additional comments about this change should be appended.

kovidgoyal:
Change the number from 10.15 to 105.whatever the latest minor release of 10.15 is, because this didnt work with 10.15.0

npearson72:
> Change the number from 10.15 to 105.whatever the latest minor release of 10.15 is, because this didnt work with 10.15.0

Good call. I re-built, tested, and confirm it works on `10.15.7`.

Thanks.

npearson72:
🙏 ❤️ @kovidgoyal and @page-down

Issue #4489, opened 2022-01-09
ctrl+/ rings terminal bell even with "enable_audio_bell no"


michaelsbradleyjr:
ctrl+/ rings the terminal bell on macOS even with `enable_audio_bell no` in `kitty.conf`. I have not tried it on Linux or other OS.

```
kitty 0.24.1 created by Kovid Goyal
Darwin elias.local 19.6.0 Darwin Kernel Version 19.6.0: Sun Nov 14 19:58:51 PST 2021; root:xnu-6153.141.50~1/RELEASE_X86_64 x86_64
ProductName:    Mac OS X ProductVersion:        10.15.7 BuildVersion:   19H1615
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /usr/local/bin/bash
Loaded config files:
  /Users/michael/.config/kitty/kitty.conf

Config options different from defaults:
allow_remote_control    y
confirm_os_window_close 1
cursor_blink_interval   0.0
enable_audio_bell       False
font_size               14.0
listen_on               unix:${HOME}/temp/kitty/kitty
macos_option_as_alt     3
visual_bell_duration    0.1
Added shortcuts:
        shift+ctrl+/ launch --allow-remote-control kitty +kitten kitty_search/search.py @active-kitty-window-id
        cmd+k combine : clear_terminal scrollback active : send_text normal \x0c
        cmd+s send_text normal,application \x18@s
```

page-down:
I suspect that Cocoa rings when this key combination is pressed.

I tried to handle it in `glfw/cocoa_init.m` `is_modified_special_key`, but it still makes a sound.

How to remove NSResponder, NSBeep useless responses?

kovidgoyal:
You probably need to implement performKeyEquivalent to get cocoa to not
be a pain. https://stackoverflow.com/questions/8869462/prevent-not-allowed-beep-after-keystroke-in-nsview

Implement it, call the super class method and return YES

page-down:
I handle it in `GLFWContentView`, if returns YES, it will not get the `ctrl`,`cmd` combination key press event. Does it mean that keypress events are not processed at this level and should not be terminated here (return yes)?

```objc
[super performKeyEquivalent:event];
return YES;
```

[...]

PR #4505, opened 2022-01-13
macOS: Suppress the key combinations beep sound

page-down:
Implement our own NSTextInputContext and pass `doCommandBySelector` to the client for processing. There will no longer be a beep sound.

https://github.com/kovidgoyal/kitty/issues/4489

I have done a simple test, modifying the system key binding, the behavior remains the same as before, `doCommandBySelector` does not do anything.

```text
{
  "^/" = (insertText:, "Hello");
} 
```

When insert text is used in the system keybindgs (see above), the text can be inserted normally.

---

It should be noted that there has been a problem with key combinations such as ctrl + xxx for input methods, as I mentioned in a previous issue.
Although the keys are processed by IME, the characters are also output to the child program (e.g. Ctrl+.  -> insert: . (ASCII dot) && toggle IME function). This has nothing to do with this PR change, but maybe it can be handled here somewhere.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -753,6 +753,20 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 
 @end // }}}
 
+// Text input context class for the GLFW content view {{{
+
+@interface GLFWTextInputContext : NSTextInputContext
+@end
+
+@implementation GLFWTextInputContext
+- (void)doCommandBySelector:(SEL)selector
+{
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Are you sure this doesnt break IME? Also rather than passing it to the client, its better to do a debug print right there mentioning it is from the text input context.

kovidgoyal:
No there's no need to call the clients doCommandBySelector. OK I guess we can just implement it and see what happens. Hopefully there will be no new breakage.

PR #4507, opened 2022-01-13
add activity_symbol_color config variable

3nprob:
Allows changing color of `tab_activity_symbol`.

### DIFF
--- a/kittens/themes/collection.py
+++ b/kittens/themes/collection.py
@@ -41,6 +41,7 @@ def patch_conf(raw: str, theme_name: str) -> str:
         'active_border_color',
         'active_tab_background',
         'active_tab_foreground',
+        'activity_symbol_color',
         'background',
         'bell_border_color',
         'color0',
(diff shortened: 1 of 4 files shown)


kovidgoyal:
What's the motivation for this? In general you would use an emoji which is rendered in its own color anyway

3nprob:
The motivation is I want something else than red, and no, no emoji in my terminal plz

kovidgoyal:
On Thu, Jan 13, 2022 at 09:46:05AM -0800, 3nprob wrote:
> The motivation is I want something else than red, and no, no emoji in my terminal plz

Why no emoji in your terminal, are you allergic?

3nprob:
> On Thu, Jan 13, 2022 at 09:46:05AM -0800, 3nprob wrote: The motivation is I want something else than red, and no, no emoji in my terminal plz
> Why no emoji in your terminal, are you allergic?

I could ask you if you're allergic to ASCII - call it a matter of preference

3nprob:
BTW, your solution of allowing templating in `tab_activity_symbol` is a better solution, nice one :+1:

PR #4514, opened 2022-01-14
Improve fish cursor shape integration

page-down:
Only use our cursor shape setting function in fish default keybindgs.
Use `fish_cursor_insert` to modify the insert mode cursor shape to a bar.

### DIFF
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -29,13 +29,29 @@ function _ksi_main
     end
 
     if not contains "no-cursor" $_ksi
-        function _ksi_bar_cursor --on-event fish_prompt
-            printf "\e[5 q"
+        function _ksi_set_cursor --on-variable fish_key_bindings
+            if test "$fish_key_bindings" = fish_default_key_bindings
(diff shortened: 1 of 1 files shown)


page-down:
It should be noted that the cursor does not blink in vi insert mode with `fish_cursor_insert`.
Is this enough? Do you want to completely rewrite `fish_vi_cursor`?

kovidgoyal:
On Fri, Jan 14, 2022 at 07:11:01AM -0800, page-down wrote:
> It should be noted that the cursor does not blink in vi insert mode with `fish_cursor_insert`.
> Is this enough? Do you want to completely rewrite `fish_vi_cursor`?

It's enough for me. It is my fond hope that eventually this
functionality will move into shells and I can remove shell integration
from kitty completely.

You should possibly report the bug about the not
blinking cursor to the fish project.

page-down:
> ... hope that eventually this
> functionality will move into shells and I can remove shell integration
> from kitty completely.

Yes.
I think this is the default design for fish. I can set all cursors to blink by adding `blink`.

```text
set -g fish_cursor_default block blink
set -g fish_cursor_insert line blink
set -g fish_cursor_replace_one underscore blink
```

However there has been controversy about the cursor change.

Issue #4452, opened 2022-01-05
Starting in 0.24 kitty overrides fish autoloaded user-defined functions


mateuszwieloch:
**Describe the bug**
I noticed that something changed between kitty 0.23.1 and 0.24 that causes kitty to override fish user defined functions, e.g. `fish_prompt` and `fish_title`. This is a pretty bad default UX for fish users.

Fish functions autoloading is a feature of fish that loads scripts containing functions from `$fish_function_path` (by default it includes `~/.config/fish/functions` folder among others).

I see that the override that kitty makes happens here https://github.com/kovidgoyal/kitty/blob/master/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish#L42

I'm aware of one solution to this problem which is to modify kitty's `shell_integration` setting, but I wonder if it would make sense to have kitty detect that these functions were modified by a user and not override them.

P.S. Not sure if this issue should be classified as a bug (for existing users) or an enhancement.

**To Reproduce**
1. Run kitty 0.24
2. Run `type --all fish_prompt
3. Notice how it wasn't autoloaded by fish from ~/.config/fish/functions, but is "interactively defined".

**Screenshots**
N/A

**Environment details**
```
kitty 0.24.0 created by Kovid Goyal
Darwin MLC02CW63QMD6T 20.6.0 Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:42 PDT 2021; root:xnu-7195.141.8~1/RELEASE_X86_64 x86_64
ProductName:	macOS ProductVersion:	11.6.1 BuildVersion:	20G224
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /usr/local/bin/fish
Loaded config files:
  /Users/mwieloch/.config/kitty/kitty.conf

Config options different from defaults:
allow_remote_control     y
enabled_layouts          ['fat:bias=75', 'fat:bias=40', 'tall', 'stack']
font_family              FiraMono Nerd Font
font_size                14.0
kitty_mod                8
macos_custom_beam_cursor True
macos_option_as_alt      3
scrollback_lines         20000
tab_bar_min_tabs         1
Added shortcuts:
	cmd+. move_tab_forward
	cmd+[ previous_window
	cmd+] next_window
[...]

PR #4516, opened 2022-01-14
Shell integration: Only override default fish_title

page-down:
https://fishshell.com/docs/current/cmds/functions.html

functions --details
- "/path/to/func.fish"
- "stdin" (defined interactively / from cli / from stdin)
- "-" (created via `source`)
- "n/a" (not available)

`fish_title.fish` defaults to autoload, and fish comes with a function directory path that is always at the last one. All functions that are not read in the last directory location are considered overridden.

Please review if there are any issues, thanks.

https://github.com/kovidgoyal/kitty/issues/4452

The only thing I'm not comfortable with is a path sep "/" in the middle of `"$fish_function_path[-1]/*"`, maybe that should be removed?

### DIFF
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -55,13 +55,21 @@ function _ksi_main
     end
 
     if not contains "no-title" $_ksi
-        function fish_title
-            if set -q argv[1]
-                echo $argv[1]
-            else
-                prompt_pwd
+        function _ksi_function_is_not_overridden -d "Check if the specified function is not overridden"
+            string match -q -- "$__fish_data_dir/functions/*" (functions --details $argv[1])
+        end
+
+        if _ksi_function_is_not_overridden fish_title
+            function fish_title
+                if set -q argv[1]
+                    echo $argv[1]
+                else
+                    prompt_pwd
+                end
             end
         end
+
+        functions --erase _ksi_function_is_not_overridden
     end
 
     if not contains "no-prompt-mark" $_ksi
@@ -131,6 +139,8 @@ function _ksi_main
         # with prompt marking kitty clears the current prompt on resize so we need
         # fish to redraw it
         set --global fish_handle_reflow 1
+
+        functions --erase _ksi_function_is_not_empty
     end
     functions --erase _ksi_main _ksi_schedule
 end


kovidgoyal:
How do you know fish_function_path has the system path as the last item?
As far as I know, users can change the value of fish_function_path, its
not readonly.

It would be more robust to check the path of some builtin fish function
that is unlikely/impossible to override and see if the path is the same
as that of fish_title.

page-down:
Thanks for the reminder. I took a closer look at the source code and docs again.

https://fishshell.com/docs/current/language.html

```text
The functions shipped with fish, usually installed in /usr/share/fish/functions (really $__fish_data_dir/functions).
```

Suppose the user first loads fish-A (/A/functions/fish_title.fish)
and then change to fish-B (/B/functions/), then at this point it is decided that A's is the reloaded one.

However it does not matter for the final result we expect, because it cannot be done precisely. I don't see any read-only variables for now.
Currently it actually uses our fish_title as long as it is turned on, so there is no downside?
We hope that fewer users will change `$__fish_data_dir` and they can choose to use the option `no-title` to turn it off.

So far this is the most accurate variable I could find, but it can still be changed by the user.

page-down:
Potential issue:
A user who has modified `$__fish_data_dir` will not have fish_title enabled.

If this variable has already been modified, I think this user may not need the automatic shell integration.
In principle, I do not want to introduce any potentially faulty results.

kovidgoyal:
On Fri, Jan 14, 2022 at 10:01:36AM -0800, page-down wrote:
> Potential issue:
> A user who has modified `$__fish_data_dir` will not have fish_title enabled.

Yes, ofcourse however there are limits to how much we can do. I am OK
with the current status.

page-down:
Currently our fish_title breaks the autoload feature of fish, there is already feedback from users in the fish github issue, maybe this can be improved.

I think it is possible to use standalone `vendor_functions.d`, but when disabled, all functions in it will not be loaded. I don't think more fish functions will be added in the future, right?
No need to use `functions.d_only_for_fish_title`.
Also it can't be configured by a single script file, additional file (fish_title.fish) is needed (this is due to the autoload feature of fish).

I have a question, what is the meaning of `no-complete` and `_ksi_completions` for fish?
- fish is already lazy loading autocomplete script.
- Even if kitty doesn't come with an autocomplete script, fish has it built in as fallback, so it will have autocomplete anyway (kitty +complete fish), and `no-complete` is meaningless.

I'm thinking of removing that bit of code and maybe moving it to the autocomplete script.

`kitty +complete fish2` What does the number 2 mean here? Does it mean that fish version 2 or above is supported?

kovidgoyal:
On Sat, Jan 15, 2022 at 08:56:57PM -0800, page-down wrote:
> Currently our fish_title breaks the autoload feature of fish, there is already feedback from users in the fish github issue, maybe this can be improved.
> 
> I think it is possible to use standalone `vendor_functions.d`, but when disabled, all functions in it will not be loaded. I don't think more fish functions will be added in the future, right?

No plans to do so.

> No need to use `functions.d_only_for_fish_title`.
> Also it can't be configured by a single script file, additional file (fish_title.fish) is needed (this is due to the autoload feature of fish).

That's OK.

> 
> I have a question, what is the meaning of `no-complete` and `_ksi_completions` for fish?
> - fish is already lazy loading autocomplete script.

Yes but is it the correct one? The last time I checked the builtin fish
one was obsolete.

> - Even if kitty doesn't come with an autocomplete script, fish has it built in as fallback, so it will have autocomplete anyway (kitty +complete fish), and `no-complete` is meaningless.
> 
> I'm thinking of removing that bit of code and maybe moving it to the autocomplete script.
> 
> `kitty +complete fish2` What does the number 2 mean here? Does it mean that fish version 2 or above is supported?

Its a second, more powerful version ofthe completion.

PR #4522, opened 2022-01-15
Fix searching the last command output

page-down:
When searching the output of the last command executed, if scrollback has been scrolled,
then the start point is actually outside (below) the screen and the number of scrolled lines needs to be added.

That's my bad. I don't know how this was slipped through. :(

Please review, thank you.

### DIFF
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -2684,7 +2684,7 @@ find_cmd_output(Screen *self, OutputOffset *oo, index_type start_y, unsigned int
             found_prompt = true;
             // change direction to downwards to find command output
             direction = 1;
-        } else if (line && line->attrs.prompt_kind == OUTPUT_START) {
+        } else if (line && line->attrs.prompt_kind == OUTPUT_START && !line->attrs.continued) {
             found_output = true; start = y1;
             found_prompt = true;
             // keep finding the first output start upwards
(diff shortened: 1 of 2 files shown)


page-down:
In a previous change to fix mark duplication, some code was removed.
But today I tested it and found that the `output continued` is occurring.
I'll try to add it back to see if it fixes the related issue.

https://github.com/kovidgoyal/kitty/commit/2cc39221083b6c8f259b11700dc2b17df9305114

page-down:
After I added the code back in, I could get the command output normally after resizing the window.

I did not submit this part yet. Maybe need to look at the previous fix first.

I found that the prompt mark appears to be copied.

https://github.com/kovidgoyal/kitty/commit/c43637f6ccb6a4958707659251e46897fcfd45d8

page-down:
Not really, it has nothing to do with the previous fix.

I pushed a new commit and checked the continued attribute.
Currently getting the output normal after resizing the window.

Please let me know if there is anything missed.

kovidgoyal:
I wont be able to review this till tomorrow, however I think you should consider adding some tests for it as this is relatively easily testable and fairly complex. See the test_prompt_marking() function

page-down:
Sure, I'll add more tests.
I've always thought kitty's test cases could be covered a bit more.
Apart from system integration (especially macOS), isn't it possible to cover a bit more range? How would it be more appropriate to contribute?

kovidgoyal:
You are welcome to send PRs with more tests, if you need help getting
started, just ask.

PR #4542, opened 2022-01-19
Add missing click and double click names to debug config

page-down:
Adjust the order of modifier key names to match the config and docs.

The modifier key order was adjusted in the previous PR, but the debug config was left out and is now adjusted as well.
(ctrl, shift, alt/opt, super/cmd, ...)

https://github.com/kovidgoyal/kitty/pull/4238

Fix mouse maps are indented with one more space.

Add the missing mouse triggers click (-2) and double-click (-3).

Refactor the print functions to make it easier to use in custom kitten.

These functions, however, all seem to be repeating patterns, and I tried to "DRY" them, but got stuck with mypy.

BTW:
Why does kitty/boss.py have executable permission? In which scenario do it need to be executed directly?


page-down:
One more issue I haven't dealt with yet:
modifier+space is listed as `modifier+ `, is it possible to convert space to the name "space"?

PR #4548, opened 2022-01-22
Add default choice option for ask kitten

page-down:
Add default choice option for ask kitten.
Press Enter to select the default, Esc to select nothing (Ctrl+C).
Make the `No` click range the same as `Yes` for easy clicking.
Makes the default take effect for text input, applied in set_tab_title action.
The tab title can be cleared, and the window title will be used again.
Set reasonable default choice for some features that use ask kitten.

However, my personal preference for quit confirmation is still `No` as the default.

Please review if there are any problems, thank you.

### DIFF
--- a/kittens/ask/main.py
+++ b/kittens/ask/main.py
@@ -300,8 +300,8 @@ def main(args: List[str]) -> Response:
         prompt = '> '
         with suppress(KeyboardInterrupt, EOFError):
             if cli_opts.default:
-                def prefill_text():
-                    readline.insert_text(cli_opts.default)
+                def prefill_text() -> None:
+                    readline.insert_text(cli_opts.default or '')
                     readline.redisplay()
                 readline.set_pre_input_hook(prefill_text)
                 response = input(prompt)
--- a/kittens/ask/main.py
+++ b/kittens/ask/main.py
@@ -298,7 +298,15 @@ def main(args: List[str]) -> Response:
 
         prompt = '> '
         with suppress(KeyboardInterrupt, EOFError):
-            response = input(prompt)
+            if cli_opts.default:
+                def prefill_text():
+                    readline.insert_text(cli_opts.default)
+                    readline.redisplay()
+                readline.set_pre_input_hook(prefill_text)
+                response = input(prompt)
+                readline.set_pre_input_hook()
+            else:
+                response = input(prompt)
     return {'items': items, 'response': response}
 
 
--- a/kittens/ask/main.py
+++ b/kittens/ask/main.py
@@ -241,13 +241,14 @@ class Choose(Handler):
         if text in self.allowed:
             self.response = text
             self.quit_loop(0)
+        elif self.cli_opts.type == 'yesno' and text == 'q':
+            self.on_interrupt()
 
     def on_key(self, key_event: KeyEventType) -> None:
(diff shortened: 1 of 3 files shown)


page-down:
I found some issues with hyperlink's ask.

```shell
kitty --config=NONE -o allow_hyperlinks=ask
```

The position for the url after the line break is odd. There is a very large indent.

```text
What would you like to do with this URL:
                                        file:///...
```

Options cannot be clicked with the mouse.

kovidgoyal:
Hyperlink issue has been fixed and clicking options with the mouse works for me

page-down:
Thanks. Clicking is working fine after fixing the hyperlink multi-line text.

It looks much better now.

---

kitty/boss.py
```python
def on_kitten_finish(
    # ...
        data = deserialize(output)
        if data is None:
            data = default_data
        if data is not None:
```

Is `default_data` working properly? Since Ctrl+C doesn't seem to cause the output to be empty. (Unless killed.)
Also is it necessary to check the kitten exit code?

kovidgoyal:
default_data is for when the window is closed, aka the kitten gets no
chance to respond. It should continue to work, see the confirm()
function.

kovidgoyal:
Or doe you mean for type=choices it might be broken there, havent checked,

kovidgoyal:
This should take care of the choice kitten: https://github.com/kovidgoyal/kitty/commit/32e31a3c6bbebb602ca3eb85853e1a7a4c6d6461

page-down:
Thanks, it's working fine now.

I also thought about whether to select the first one by default, but at the time I thought a truthful answer '<empty>' would be most appropriate, i.e. the user does not make any choice and let subsequent programs choose how to handle it based on that.

But it seems to be unnecessary, just provide a default option.

page-down:
> default_data is for when the window is closed, aka the kitten gets no chance to respond.

You mean Ctrl+C doesn't use default_data, right? (Yes, according to the code.)

If I use `default_data={'response': 'c'}` for testing, hyperlink will not be copied to the clipboard after Ctrl+C. Only after Ctrl+Shift+W closes the window. So it works as expected.

In the recent changes, the default choice are marked using yellow, but I am not sure if green is the default or yellow is the default. Is it appropriate to use underline?

kovidgoyal:
On Fri, Jan 21, 2022 at 11:00:13PM -0800, page-down wrote:
> > default_data is for when the window is closed, aka the kitten gets no chance to respond.
> 
> You mean Ctrl+C doesn't use default_data, right? (Yes, according to the code.)
> 
> If I use `default_data={'response': 'c'}` for testing, hyperlink will not be copied to the clipboard after Ctrl+C. Only after Ctrl+Shift+W closes the window. So it works as expected.

That's not the purpose of default_data, so dont do that. In any case, I
have implemented a choose() method to make using it easier.

> 
> In the recent changes, the default choice are marked using yellow, but I am not sure if green is the default or yellow is the default. Is it appropriate to use underline?

Sure, good idea.

page-down:
I also found a problem where the click is not responsive until the mouse is moved.

For example, if you move the mouse to the No button position first, then press Cmd+Q and click the mouse, you cannot cancel the exit.

Is this something that can be fixed?

PR #4552, opened 2022-01-22
macOS: Fix text insertion by modifier keys when using IME

page-down:
There is another situation when dealing with macOS input method modifier keys.
It is possible for the input method to call insertText and then continue to send the event to the program without intercepting it.

I have tested it and it works fine to insert text in this case too.

There is also a problem with this PR, if the input method inserts text immediately after the initial press of the modifier key when there is no markedText, then it will not work. A separate flag is needed to replace `handled` from `[inpctx handleEvent:event]` (which is meaningless).

But I haven't come across this kind of input method so far.

Please review if there is anything that needs to be adjusted.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1152,14 +1152,14 @@ is_ascii_control_char(char x) {
         debug_key("\x1b[31mPress:\x1b[m native_key: 0x%x (%s) glfw_key: 0x%x %schar_count: %lu deadKeyState: %u repeat: %d ",
                 keycode, safe_name_for_keycode(keycode), key, format_mods(mods), char_count, window->ns.deadKeyState, event.ARepeat);
         marked_text_cleared_by_insert = false;
-        in_key_handler = 1;
         if (process_text) {
+            in_key_handler = 1;
             // this will call insertText which will fill up _glfw.ns.text
             [self interpretKeyEvents:@[event]];
+            in_key_handler = 0;
         } else {
             window->ns.deadKeyState = 0;
         }
-        in_key_handler = 0;
         if (window->ns.deadKeyState && (char_count == 0 || keycode == 0x75)) {
             // 0x75 is the delete key which needs to be ignored during a compose sequence
             debug_key("Sending pre-edit text for dead key (text: %s markedText: %s).\n", format_text(_glfw.ns.text), glfw_keyevent.text);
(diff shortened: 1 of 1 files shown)


page-down:
To clarify, here is the relevant IME code for this case.

```objc
@implementation SquirrelInputController { //... }
- (BOOL)handleEvent:(NSEvent*)event client:(id)sender
{
    BOOL handled = NO; // ...
    switch (event.type) {
        case NSEventTypeFlagsChanged: { // ...
            if (changes & OSX_SHIFT_MASK) { // ...
                [self processKey:XK_Shift_L modifiers:(rime_modifiers | release_mask)];
            }
        } break;
        case NSEventTypeKeyDown: { // ...
            handled = [self processKey:rime_keycode modifiers:rime_modifiers]; // ...
        } break;
        // ...
    } // ...
    return handled;
}
```

The flag `handled = [inpctx handleEvent:event]` does not reflect whether insertText has been called or not.
I think it can be removed (replaced).

kovidgoyal:
On Sat, Jan 22, 2022 at 09:16:08PM -0800, page-down wrote:
> The flag `handled = [inpctx handleEvent:event]` does not reflect whether insertText has been called or not.
> I think it can be removed (replaced).

If handled is false it would imply that the input context ignored the
event completely. It being true does not imply that insetText was
called, but in order for insertext to have been called, it must be true.

page-down:
> If handled is false it would imply that the input context ignored the
> event completely.

I don't understand, why the following log exists when I test it.

Doesn't the value returned represent whether or not it was eventually marked as processed? Then IME can send insertText and return false to let the event flow to the next one, and the input context will return false if no one is marked as true in the end.

```text
-------------- flags changed -----------------
NSEvent: type=FlagsChanged loc=(322,171.016) time=16209.4 flags=0x100 win=0x7fd5ca72ebe0 winNum=933 ctxt=0x0 keyCode=56
flagsChanged: modifer: shift native_key: 0x38 (<noname>) glfw_key: 0xe061 mods: none 
---> [flagsChanged] call handled = [inpctx handleEvent:event]
---> [insertText] start, in_key_handler: 2

	insertText: k replacementRange: (9223372036854775807, 0)
---> [insertText] set marked_text_cleared_by_insert = true
sending text to kitty from insertText called from event loop: k
on_key_input: glfw key: 0x0 native_code: 0x0 action: RELEASE mods: none text: 'k' state: 2 committed pre-edit text: k
---> [flagsChanged] after inpctx handleEvent, handled: 0
Clearing pre-edit text
```

kovidgoyal:
On Sat, Jan 22, 2022 at 10:18:14PM -0800, page-down wrote:
> > If handled is false it would imply that the input context ignored the
> > event completely.
> 
> I don't understand, why the following log exists when I test it.
> 
> Doesn't the value returned represent whether or not it was eventually marked as processed? Then IME can send insertText and return false to let the event flow to the next one, and the input context will return false if no one is marked as true in the end.

If that is the semantics of handled, then yes it should be removed.

Issue #4553, opened 2022-01-23
Entering vi-mode in Fish add a new line


angelus2014:
**Describe the bug**
When I enter (or exit) the vi-mode in Fish, Kitty always adds a new line. It doesn't matter if it's a multiline or single line prompt. I don't see this behaviour in other terminals on my Mac (iTerm, Alacritty).

**To Reproduce**
1. Install Fish and start this shell (or in my case, make it the default)
2. Start typing a command
3. Press ESC to enter (or leave) the vi-mode
4. Kitty prints a new line
<img width="152" alt="Screenshot 2022-01-23 at 9 15 40" src="https://user-images.githubusercontent.com/7578963/150668535-42a00890-5e5a-4996-b897-d6e74beac639.png">
<img width="151" alt="Screenshot 2022-01-23 at 9 15 47" src="https://user-images.githubusercontent.com/7578963/150668537-8c0c9df6-1566-4b0e-b56e-1ebe4943b38b.png">

**Environment details**
kitty 0.24.1 created by Kovid Goyal
Darwin orac.local 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:29:10 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T8101 arm64
ProductName:    macOS ProductVersion:   12.1 BuildVersion:      21C52
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /opt/homebrew/bin/fish
Loaded config files:
  /Users/angelo/.config/kitty/kitty.conf

Config options different from defaults:
active_tab_font_style   (True, False)
bold_font               MesloLGSDZ Nerd Font Mono Bold
enable_audio_bell       False
font_family             MesloLGSDZ Nerd Font Mono Regular
font_size               13.0
inactive_tab_font_style (True, False)
macos_option_as_alt     2
macos_thicken_font      0.25
macos_titlebar_color    370548226
selection_foreground    None
shell                   /opt/homebrew/bin/fish --login
tab_fade                (1.0,)
window_alert_on_bell    False
window_padding_width    FloatEdges(left=5.0, top=5.0, right=5.0, bottom=5.0)
Colors:
        active_border_color     #3d59a1   
        active_tab_background   #16161e   
[...]

PR #4558, opened 2022-01-24
Provide our own workaround for fish vi cursor with multi-line prompt

page-down:
Remove fish_vi_cursor.
Implementing it in our own way, and no need to perform the detection logic for other terminals.
Compatible with fish's cursor config global variables. (Except fish_cursor_unknown)

Since it is implemented in the prompt marking, falling back to the built-in config method when the user explicitly set `no-prompt-mark`, it works fine with the fish default prompt.

While implementing vi mode, two additional fish global variables are also provided for customization default mode cursor. Hope it can bring some help to users.

- kitty_fish_prompt_cursor
- kitty_fish_external_cursor (works for normal mode and vi mode)

The minimal multi-line prompt and starship have been tested and both work fine again.

related:
https://github.com/kovidgoyal/kitty/issues/4553

The following are the test cases used.

---

```text
function fish_prompt
    printf 'fish\n>'
end
```

```shell
# fish default prompt
kitty fish -ilP

# fish vi mode
kitty fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings;"

# fish vi mode with multi-line prompt and default fish_mode_prompt
kitty fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings; function fish_prompt; printf 'fish\n>'; end"

# fish vi mode with multi-line prompt and empty fish_mode_prompt
kitty fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings; functions --erase fish_mode_prompt; function fish_prompt; printf 'fish\n>'; end"

# fish multi-line default mode switch to vi mode (press F1)
kitty -o 'map f1 paste set -g fish_key_bindings fish_vi_key_bindings' fish -ilPC "functions --erase fish_mode_prompt; function fish_prompt; printf 'fish\n>'; end"

# with starship default
kitty fish -ilPC "set -g fish_key_bindings fish_default_key_bindings; starship init --print-full-init fish | source;"

# with starship vi mode
kitty -o 'map f1 paste set -g fish_key_bindings fish_default_key_bindings' -o 'map f2 paste set -g fish_key_bindings fish_vi_key_bindings' fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings; starship init --print-full-init fish | source;"

# with no-cursor (shoud not change cursor and break in vi mode)
kitty -o shell_integration=no-cursor fish -ilPC "starship init --print-full-init fish | source;"

kitty -o shell_integration=no-cursor fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings; starship init --print-full-init fish | source;"

# with no-prompt-mark (should change cursor and break in vi mode)
kitty -o shell_integration=no-prompt-mark -o 'map f1 paste set -g fish_key_bindings fish_vi_key_bindings' fish -ilPC "starship init --print-full-init fish | source;"

kitty -o shell_integration=no-prompt-mark fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings; starship init --print-full-init fish | source;"

# with user defined cursor style
kitty -o env=kitty_fish_prompt_cursor="block" -o env=kitty_fish_external_cursor="underline blink" fish -ilP

kitty \
-o 'map f1 paste set -g fish_key_bindings fish_default_key_bindings' \
-o env=kitty_fish_prompt_cursor="block" \
-o env=kitty_fish_external_cursor="underline blink" \
-o env=fish_default_cursor="block" \
-o env=fish_insert_cursor="bar" \
-o env=fish_replace_one_cursor="underline" \
-o env=fish_visual_cursor="block" \
fish -ilPC "set -g fish_key_bindings fish_vi_key_bindings; starship init --print-full-init fish | source;"
```

### DIFF
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -7,6 +7,8 @@ function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after
     functions --erase __ksi_schedule
     test -n "$KITTY_SHELL_INTEGRATION" || return 0
 
+    # kitty exports the vendored integration script directory to XDG_DATA_DIRS
+    # the original paths needs to be restored here to not affect other programs
     if set -q XDG_DATA_DIRS KITTY_FISH_XDG_DATA_DIR
         set --global --export --path XDG_DATA_DIRS "$XDG_DATA_DIRS"
         if set -l index (contains -i "$KITTY_FISH_XDG_DATA_DIR" $XDG_DATA_DIRS)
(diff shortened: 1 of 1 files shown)


page-down:
The CI test reports errors.

I didn't touch the python-related code, it looks like the dependency has been updated?

kovidgoyal:
Maybe you should propose this upstream, in the fish project? Should we
really be carrying around bug fixes for shell bugs in our code? Not to
mention that if (as I hope) eventually shell integration moves into the
shell itself, we wont anymore have this script at all.

Maybe @faho can comment?

page-down:
Yes, I agree with you.

I don't think I can propose this to upstream, it just happens to be the "ok" solution for our integration script.
I think the root of the problem is that the timing of the output escape sequence conflicts with the repaint.
And this already has some analysis in the upstream issue post from six years ago.

BTW, I could reproduce the issue in Konsole (Linux). Unfortunately I can't provide more meaningful suggestions.

If this issue can be solved in the foreseeable future, then this is indeed unnecessary.

faho:
Why do you believe this fixes it? What does it do differently?

As best as I can tell, this doesn't really do anything *different* from fish's vi-cursor handling. It prints the sequences, but uses the hardcoded kitty versions.

However I can't see it looking at $fish_bind_mode, so presumably it doesn't change when you press escape? That would "fix" the issue by simply breaking the cursor change - essentially removing the feature.

I've never really been able to track down the issue with multiline prompts here - I *think* it's an ordering issue between sequences, which could be worked around by defining this handler at a different point in time. The actual code won't really matter. I don't know, I've never been able to reproduce the problem.

Also: I dislike adding things like this in shell integration, for obvious reasons. It makes my life harder because now I need to send people with kitty-related bugs to the kitty bug tracker.

shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
```diff
@@ -20,66 +22,177 @@ function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after
...
+        # steady:   block 2, underline 4, bar 6
+        function __ksi_init_cursor_var -d "Initialize cursor shape CSI escape sequence"
+            set --local varname kitty_fish_$argv[1]_cursor
+            set --local val 0
+            if contains $argv[1] default insert replace_one visual
+                and test -n "fish_$argv[1]_cursor"
+                set varname fish_$argv[1]_cursor
+            end
+            if test -n "$$varname"
+                set --local shape (string split " " -- "$$varname")
```
faho, line 62: This is super suspect. You would expect these variables to be set as a list, so you can simply use `$$varname`.

shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
```diff
@@ -20,66 +22,177 @@ function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after
...
+            printf "
+                function __ksi_vi_prompt_cursor -d 'Set cursor shape for current vi mode'
+                    switch \$fish_bind_mode
+                        case default; printf \"%s\"
+                        case insert; printf \"%s\"
+                        case replace_one; printf \"%s\"
+                        case visual; printf \"%s\"
+                    end
                 end
+            " $__ksi_cursor_default $__ksi_cursor_insert $__ksi_cursor_replace_one $__ksi_cursor_visual | source
```
faho, line 99: Why is this function defined via `source`?

shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
```diff
@@ -20,66 +22,177 @@ function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after
...
+                set varname fish_$argv[1]_cursor
+            end
+            if test -n "$$varname"
+                set --local shape (string split " " -- "$$varname")
+                switch $shape[1]
+                    case block
+                        set val 2
+                    case underline underscore
+                        set val 4
+                    case bar beam line
```
faho, line 68: Why does this have multiple names? Fish only knows "block", "underscore" and "line".

This means people now go setting their var to "beam" and then come reporting bugs to us if it doesn't work elsewhere. E.g. if they read a blog post of someone who uses kitty.

page-down:
> Why do you believe this fixes it? What does it do differently?

Under macOS and Linux, the prompt redraw position is normal when switching between modes in vi mode. Before that, it will move down one line.

The only thing it does is output an escape code at the beginning of fish_mode_prompt or fish_prompt to change the cursor shape.

Relies on prompt output on redraw, without using `function --on-variable ... --on-event`.

> ... so presumably it doesn't change when you press escape? ...

With my testing, it worked as expected.

> I've never really been able to track down the issue with multiline prompts here ...

If there is anything I can help with, I'd be happy to.

> I dislike adding things like this in shell integration ...

Sorry, these were not supposed to exist.

Thanks for pointing out potential issues in the code. It might be possible to get to the root of this issue, so there is no longer a need for these.

shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
page-down: Yes, my first version was using list, I was worried that someone would write "block blink", that's why I used string. maybe I was overthinking.

shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
page-down: All these variables will be cleaned up after initialization, and no more steps such as switch (for each shape) will be performed to recalculate them at each repaint.
Maybe this is a bad practice?

shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
page-down: The "bar" in kitty's config is called "beam". I'd like to make either name work, sorry, maybe it should be strictly fish naming. These definitely need to be removed.

faho:
>The only thing it does is output an escape code at the beginning of fish_mode_prompt or fish_prompt to change the cursor shape.

Ah, okay. So it changes the ordering of escapes. That does point to an ordering issue.

(I mean that in a general sense of "if fish prints this escape at this point in time it's broken, if it's at that other point it'll work")

>Under macOS and Linux, the prompt redraw position is normal when switching between modes in vi mode. Before that, it will move down one line.

Like I said: I have never been able to reproduce this on linux, and most reports we've gotten have been from mac users. I don't have (or want) a mac, so I've been waiting for someone to figure it out. 

Personally I don't use a multiline prompt or vi-mode, so I've always just picked a simple multiline prompt and enabled vi-mode to test, and that has never caused an issue.

page-down:
> That does point to an ordering issue. ...

Yes, I believe that is the problem.

> Personally I don't use a multiline prompt or vi-mode ...

I don't have any need for multi-line mode or vi mode either.
I just don't want the fish issue to be asked here and then the user doesn't know how to fix it.
I just try to help as much as I can.

Here are the prompt ​and steps I tested.

```fish
function fish_prompt; printf 'fish\n>'; end
```

Ubuntu LTS, kitty/Konsole

- Type some text
- press i, I, a, esc
- Switch between default mode and insert mode.

```text
[I] fish
[N] fish
>sometext
```

faho:
>Here are the prompt ​and steps I tested.

That's roughly what I've always done and it's never reproduced for me. Something is missing.

page-down:
Here is a log from kitty's dump commands comparison, not sure if it helps?

https://bpa.st/5ZUA

faho:
>Here is a log from kitty's dump commands comparison, not sure if it helps?

No, because 

1. I have no idea how to read this
2. It's a ton of stuff and I have no idea what's even relevant
3. I can't figure out what in fish causes what

Okay, I just tested again with both konsole and kitty on archlinux, and I can't reproduce it.

I'm once more at a loss, someone who can reproduce it needs to figure out what causes it and give us the necessary reproduction steps.

Note that I tried fish *3.3.1*. Ubuntu's packages in 20.04 (the current LTS) are still at 3.1.0 (not even 3.1.2, the bugfix release) with some Debian patches. That's rather out of date (about 80% of commits in 3.3.1) and makes tests with it essentially meaningless. You want [the PPA](https://launchpad.net/~fish-shell/+archive/ubuntu/release-3).

> (BTW, the cursor will keep moving forward when switching between insert, default mode in the same place.)

You mean backwards. That's what vi does, so we implemented it as well.

kovidgoyal:
I am closing this as I dont think having workarounds for a fish bug in kitty shell integration is a good idea, but please feel free to continue to discuss and if I can help in any way, am happy to. 

@faho The log @page-down is basically what you can generate with the --dump-commands command line option to kitty, it is a list of *every* escape code that kitty sees from the client program. It can come in handy to isolate problems with incorrect escape codes or ordering or similar issues. But yes it is very verbose, since there are typically a lots escape codes being sent. The technique I use with these logs is to get a log that reproduces an issue (you can replay the log with kitty --replay-commands then in an editor delete large sections of the log till I have just the minimal commands needed to reproduce and then see if I can understand the issue from there. I dont know if this is something generally useful since I doubt many people are as familiar with terminal escape codes as a terminal developer, but...

page-down:
> 3. I can't figure out what in fish causes what

I compared the differences, mainly in that after outputting the cursor shape CSI escape sequence, several cursor forward, one carriage return, and one cursor up are missing.

> ... someone who can reproduce it needs to figure out what causes it and give us the necessary reproduction steps.

It seems more important to `what causes it`, since the reproduction step is nothing special...
Where is the most likely cause of missing cursor movement?

> Note that I tried fish 3.3.1 ...

What I tested is that fish 3.3.1 from PPA.

> You mean backwards. That's what vi does ...

Thanks for the explanation.

---

Since this issue can also be reproduced in konsole, I'm not convinced it has anything to do with kitty. I suggest to continue the discussion back to the fish project, maybe someone else can reproduce this issue.

faho:
>It seems more important to what causes it, since the reproduction step is nothing special...

Well, yeah, but since it doesn't cause it for me there is something missing in the reproduction steps! Something about the environment isn't the same. It might be timing, it might be library versions (ncurses?), it might be locale or config or...

> I suggest to continue the discussion back to the fish project, maybe someone else can reproduce this issue.

https://github.com/fish-shell/fish-shell/issues/3481

PR #4559, opened 2022-01-25
Add comments and minor tweaks to fish integration

page-down:
Set cursor shape to steady block before executing commands in vi mode. I left this out.
Configure the vi mode cursor shape on demand. If the user only uses normal mode, no vi mode global variables will be added.
Since it is unlikely that more features will be added in the future, remove one function call at each marking.
Add a brief comment to the XDG_DATA_DIRS section.

### DIFF
--- a/shell-integration/bash/kitty.bash
+++ b/shell-integration/bash/kitty.bash
@@ -61,7 +61,7 @@ _ksi_main() {
 
     if [[ "${_ksi_prompt[cursor]}" == "y" ]]; then 
         _ksi_prompt[ps1]+="\[\e[5 q\]"  # blinking bar cursor
-        _ksi_prompt[ps0]+="\[\e[5 q\]"  # blinking block cursor
+        _ksi_prompt[ps0]+="\[\e[1 q\]"  # blinking block cursor
     fi
 
     if [[ "${_ksi_prompt[title]}" == "y" ]]; then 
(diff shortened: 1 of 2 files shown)


page-down:
I found that zsh and bash use blinking block for commands by default. However fish was originally written with a steady block cursor. So I adjusted the style in fish.
I also found that bash uses `\e[5 q` for the blinking block cursor, so I changed that as well.

page-down:
I see that the default cursor shape is used in the latest changes. I had a question when I revised it, wouldn't the output of this pre-exec cursor shape make it difficult for the cursor_shape option to take effect? Unless the program is run directly with `launch`.

After trying the latest commit, this concern of mine is resolved.

However I have another question, I see that other terminals, such as Konsole, can select whether the (default) cursor is blinking or not.
It seems that blinking can only be disabled globally in kitty. Is it possible (and does it make sense) to provide the option to do this?

kovidgoyal:
On Tue, Jan 25, 2022 at 01:02:55AM -0800, page-down wrote:
> However I have another question, I see that other terminals, such as Konsole, can select whether the (default) cursor is blinking or not.
> It seems that blinking can only be disabled globally in kitty. Is it possible (and does it make sense) to provide the option to do this?

Doesn't make sense to me. You typically either want blinking cursors or
not. It's not application specific, unlike cursor shape where the cursor
shape can change based on editing mode.

page-down:
Ok, thanks for the reply. I don't have any need for this.
So far, regarding the cursor-related features, it's done and I have no more questions.

PR #4563, opened 2022-01-26
macOS: Explicitly cancel IME composition when losing the focus

page-down:
Some input methods do not cancel the input session when switching focus, resulting in inconsistent states.
Explicitly cancel it in this PR.

Also need to cancel when switching kitty windows (tabs), closing windows. looks like it needs to be called in python.
I don't know if the same issue exists under Linux.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1378,7 +1378,7 @@ is_ascii_control_char(char x) {
         bool had_marked_text = [self hasMarkedText];
         [self unmarkText];
         if (had_marked_text && (!in_key_handler || in_key_handler == 2)) {
-            debug_key("Clearing pre-edit because setMarkedText called from event loop or flagsChanged\n");
+            debug_key("Clearing pre-edit because setMarkedText called from %s\n", in_key_handler ? "flagsChanged" : "event loop");
             GLFWkeyevent glfw_keyevent = {.ime_state = GLFW_IME_PREEDIT_CHANGED};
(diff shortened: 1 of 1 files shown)


page-down:
It looks like a build failure caused by a DNS error in GitHub CI.

kovidgoyal:
If you want to handle  kitty internal focus changes its much more work

page-down:
Yes. In fact, rather than canceling the IME composition, just simply clear the previous pre-edit text after the focus changes.

- If the window is closed, the focus will automatically go to another window.
- If the OS window is closed, then I don't need to bother with it.

So there is no need to actively end the input state.

kitty/state.c `update_ime_position_for_window()`
https://github.com/kovidgoyal/kitty/commit/ffa755c723f465d50efc86bd0b079c0b80192e0e

kitty/window.py `focus_changed()`
https://github.com/kovidgoyal/kitty/commit/7e5cb5092518ef66e9854d2db7a0236d825c4c23

Is it possible to get the window before the focus change? Then just clean up the pre-edit text.

kovidgoyal:
look at line 385 of glfw.c

page-down:
The current problem is that when switching windows (mouse clicks) while typing, (Yes, you've said don't do that.)
the pre-edit text is left on the previous window.
Then typing in the new window will work fine (the pre-edit text is not refreshed in new window yet).

It doesn't seem to be necessary to cancel. Since the focus stays in the same window for the OS.

I'll give it a try.

PR #4571, opened 2022-01-27
Cancel IME for active kitty or OS window only

page-down:
When the inactive window is closed, the input method will no longer be canceled.

I also found that when the window is redrawn (resized), the overlay text is cleared. Is there a good way to redraw it? 

Honestly, the edge case handling requires time investment and is not very useful. The good thing is that it reduces inconsistent state cases and hopefully (if no new bugs are introduced) makes the software more robust.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -713,6 +713,7 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
         [window->ns.view unmarkText];
         GLFWkeyevent dummy = {.action = GLFW_RELEASE, .ime_state = GLFW_IME_PREEDIT_CHANGED};
         _glfwInputKeyboard(window, &dummy);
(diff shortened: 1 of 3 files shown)


kovidgoyal:
On Wed, Jan 26, 2022 at 07:57:49PM -0800, page-down wrote:
> When the inactive window is closed, the input method will no longer be canceled.
> 
> I also found that when the window is redrawn (resized), the overlay text is cleared. Is there a good way to redraw it? 

No, when window is resized the terminal sends a signal to the client
program which redraws the screen, this is asynchronous so there is no
way for the terminal to know when it is complete.

> 
> Honestly, the edge case handling requires time investment and is not very useful. The good thing is that it reduces inconsistent state cases and hopefully (if no new bugs are introduced) makes the software more robust.

Yes, its fairly low return effort, but it definitely does improve
overall robustness in a giant state machine.

PR #4590, opened 2022-01-29
Refactor: More f-string

page-down:
This is the last batch of refactorings for the string f-string.

During my testing, I found that the contents of the generated files are not consistent in different environments (tested in macOS, Linux). So I sorted them to ensure consistency. (`gen-apc-parsers.py`)

I didn't resolve the following warning about building docs under macOS.
```text
changelog.rst:688: WARNING: unknown option: kitty --detach
changelog.rst:2141: WARNING: unknown option: kitty --class
```

Please review if there are any changes that need to be made, thank you.

Is it possible to see the code coverage? For example, which part of the code needs additional test cases?


kovidgoyal:
On Sat, Jan 29, 2022 at 04:45:44AM -0800, page-down wrote:
> Is it possible to see the code coverage? For example, which part of the code needs additional test cases?

No I have never found code coverage metrics to be useful, as code
coverage tools assume all code is equally likely to break/important,
which is never the case. But if you
want to use them feel free to add support for them in the CI pipeline,
it shouldn't be very difficult.

PR #4618, opened 2022-02-03
macOS: Allow opening URLs with kitty

page-down:
Enables kitty to open links such as `ssh://`.

On macOS 10.13 and above, functions such as `openFiles` will no longer be called when `openURLs` is implemented.
The relevant functions can be removed when the minimum supported version is upgraded to 10.13.

I would also like to provide a default configuration for ssh links.

```python
diff --git a/kitty/open_actions.py b/kitty/open_actions.py
index 32ebb84d..a84c7650 100644
--- a/kitty/open_actions.py
+++ b/kitty/open_actions.py
@@ -75,7 +75,8 @@ def parse(lines: Iterable[str]) -> Iterator[OpenAction]:
         entries.append((tuple(match_criteria), tuple(raw_actions)))

     with to_cmdline_implementation.filter_env_vars(
-        'URL', 'FILE_PATH', 'FILE', 'FRAGMENT',
+        'URL', 'URL_SCHEME', 'URL_USERNAME', 'URL_PASSWORD', 'URL_HOSTNAME', 'URL_PORT',
+        'FILE_PATH', 'FILE', 'FRAGMENT',
         EDITOR=shlex.join(get_editor()),
         SHELL=shlex.join(resolved_shell(get_options()))
     ):
@@ -172,6 +173,11 @@ def actions_for_url_from_list(url: str, actions: Iterable[OpenAction]) -> Iterat

     env = {
         'URL': url,
+        'URL_SCHEME': purl.scheme or 'file',
+        'URL_USERNAME': unquote(purl.username or ''),
+        'URL_PASSWORD': unquote(purl.password or ''),
+        'URL_HOSTNAME': unquote(purl.hostname or ''),
+        'URL_PORT': str(purl.port) if purl.port is not None else '',
         'FILE_PATH': path,
         'FILE': posixpath.basename(path),
         'FRAGMENT': unquote(purl.fragment)
```

launch-actions.conf
```text
# Open ssh URLs with ssh command
protocol ssh
action launch --type=os-window sh -c "PORT=$URL_PORT;USER=$URL_USERNAME;HOST=$URL_HOSTNAME;exec ssh ${PORT:+-p$PORT} ${USER:+$USER@}${HOST:-localhost}"
```

Do you think the above approach is appropriate? I would like to provide variables that are generic and can be used on other url schemes. Or is there a better way?

Also do you think it is appropriate to register the `kitty://` url scheme to allow users to call custom launch actions? (The custom url scheme needs to be explicitly defined in the info plist under macOS.)

Finally after the ssh launch action part, I need to add the registration for SSH URL Scheme to finish this PR.
Due to the limitations of macOS, implementing and testing this feature is time consuming and cumbersome...


kovidgoyal:
ssh can handle ssh URLs directly there is no need to translate them. As
for a default action, sure.

kovidgoyal:
On Thu, Feb 03, 2022 at 03:44:51AM -0800, page-down wrote:
> Also do you think it is appropriate to register the `kitty://` url scheme to allow users to call custom launch actions? (The custom url scheme needs to be explicitly defined in the info plist under macOS.)

Seems OK to me.

page-down:
> ssh can handle ssh URLs directly ...

Thanks for the reminder.

>> `kitty://` ...
>
> Seems OK to me.

Do I need to keep the prefix? For example, only allow users to use `kitty://launch/` (or `kitty-launch://`), later we can implement `kitty://open_tab` and so on. If there is no such plan, then no prefix is needed and all are open for use.

kovidgoyal:
I prefer kitty://launch I dont want to pollute the global enviroment with extra URL prefixes.

page-down:
There is another situation to consider regarding kitty app URL scheme.

I want to implement a helper app that sits in the menu bar. (Maybe later)
- Support for configuring `launch at login` to make it easy to run the kitty daemon.
- Windows with `macos_hide_from_tasks` can be managed, as they will no longer be displayed in the dock or task manager.
- Assign execution tasks from URL scheme (or remote control) to daemon background windows, different instance groups, and normal windows.
- Support for binding global shortcut to bring up the quake mode window.

Since it is a menu bar application with no window, it does not take over the foreground when it receives a call.

It will be easier to use if the kitty app URL scheme is tied to this app.
Or use a separate URL scheme on this app.

All of this should be simple to use and should not add too much maintenance complexity.

If we bind `kitty://` to the main App now, it might be a bit tricky to adjust it later. I'd like to discuss it first rather than waste time on directions that aren't really appropriate. (Due to my own misjudgment.)

Or maybe I don't need to overthink it right now.

kovidgoyal:
Not sure I follow. Why would it be better to use kitty:// for this
helper app? 

We can have this app listen on a socket or whatever IPC native macOS
applications use and have kitty forward kitty:// URLs to it if needed.

page-down:
> Why would it be better to use kitty:// for this helper app?
> ... have kitty forward kitty:// URLs to it ...

Yes, various forwarding approaches can be easily implemented.

The biggest problem is that for apps with windows (e.g., kitty), the window will be brought to the foreground when openURLs is called.
This is more problematic if the user wants to perform background tasks (e.g., tasks are executed in other windows within the daemon).

There may be more than one kitty instance, maybe with different KITTY_CONFIG_DIRECTORY, I am not sure which one will be called. And there is only one menu bar app instance.

These are some of my thoughts.

kovidgoyal:
On Thu, Feb 03, 2022 at 10:41:32PM -0800, page-down wrote:
> > Why would it be better to use kitty:// for this helper app?
> > ... have kitty forward kitty:// URLs to it ...
> 
> Yes, various forwarding approaches can be easily implemented.
> 
> The biggest problem is that for apps with windows (e.g., kitty), the window will be brought to the foreground when openURLs is called.
> This is more problematic if the user wants to perform background tasks (e.g., tasks are executed in other windows within the daemon).

You mean if some other application has focus, focus is switched to
kitty? I dont quite understand why that is a problem. Typically when you
click on a URL you expect it to open in its associated application and
that application to be focused. If you want to perform a background task
then use the existing remote control machinery for it.

> 
> There may be more than one kitty instance, maybe with different KITTY_CONFIG_DIRECTORY, I am not sure which one will be called. And there is only one menu bar app instance.
> 
> These are some of my thoughts.

The problem with using kitty:// for a helper app is presumably this
helper app is going to be optional. i.e. not everyone is going to be
running it. So binding kitty:// to it is also problematic since it would
startup the helper app which may not be desired.

page-down:
> ... you expect it to open in its associated application and
> that application to be focused ...

For example, I want kitty to take over all http links, and for URLs with unfriendly heavily CSS/JS, forward them to certain browsers. For download tasks (e.g. GitHub Releases URLs), download them directly with start and finish system notifications.

If I work on another instance or another application and have a default full screen kitty open, it will likely cut across the entire desktop (I have not tested this yet). The remote control cannot receive URL scheme events.

Maybe this reason is not enough to support this.
For these particular needs, a separate custom app should be developed to do the forwarding.

> ... this helper app is going to be optional ...

Thanks, that's what I wanted, I forgot about it.
None of the above needs to be considered. `kitty://` can only be bound to the main app.

page-down:
I've updated the PR and have now restricted it to `kitty:///launch/`. Also follows URL syntax.
The kitty URL can be designed later as needed.

Is there any possibility of command execution when SSH is invoked in this way? I don't want to introduce any potential security holes. Because any program can ask to open the URL .

Please review and let me know if there are more adjustments needed, thank you.

kovidgoyal:
On Fri, Feb 04, 2022 at 09:38:41PM -0800, page-down wrote:
> Is there any possibility of command execution when SSH is invoked in this way? I don't want to introduce any potential security holes. Because any program can ask to open the URL .

Invoking SSH with an URL gives you an interactive shell at that URL
(possibly without authentication depending on how ssh is setup). If the
URL points to localhost, you get an interactive shell in localhost. You
can achieve arbitrary command execution by modifying the shells rc
files, but if you can do that, it's game over anyway. I dont think there
is any pathway to leveraging an interactive shell in an unrelated
process without local filesystem access. And if you have local
filesystem access, there are myriad ways to achieve command execution.

page-down:
> Invoking SSH with an URL gives you an interactive shell at that URL ...
> And if you have local filesystem access, there are myriad ways to ...

Yes, this is not a problem.
I see that the URL is already quoted correctly.
It is impossible to have spaces and then append commands.

kovidgoyal:
On Fri, Feb 04, 2022 at 11:30:07PM -0800, page-down wrote:
> > Invoking SSH with an URL gives you an interactive shell at that URL ...
> > And if you have local filesystem access, there are myriad ways to ...
> 
> Yes, this is not a problem.
> I see that the URL is already quoted correctly.
> It is impossible to have spaces and then append commands.

Ah yes, variables are expanded after the command line is split. So $URL
will always be passed as a single item to ssh.

page-down:
While working on the Open URL feature I noticed that it could be improved for opening multiple URLs at the same time (especially `file:///` files). Currently, you can only open and process them one by one.

Since macOS `Open with` doesn't have any additional menu items, I figured we could use modifier keys to extend the functionality. When opening, check that the corresponding modifier keys in the configured action are pressed and held.

Usage Scenarios:
- Compress multiple files.
- Batch rename files.
- Merge multiple audio or video files.
- Hold down the option key, open an image, select the rotation angle in kitten, save the image and use icat to view it.
- Select multiple file, choose the kitty window by using the visual window select feature, move all the files to cwd.

I would like to work on these for the next step. But I need to think about how to make an action handle multiple URLs first (maybe with modifiers).

kovidgoyal:
seems to be getting too complicated to me. How are you going to assign
actions to arbitrary groups of files? Easier to just open the current
finder directory in kitty and type the command to do what you need.

kovidgoyal:
Indeed, rather than messing around with such hacks, just add scripts to your finder right click menu: https://stackoverflow.com/questions/4215287/adding-a-script-to-macos-finder-contextual-menu

page-down:
> How are you going to assign actions to arbitrary groups of files?

What I can think of is:
- Explicitly configure some actions that receive all URLs as parameters.
- These actions must be configured with modifier keys.

When opening URLs, first check if any `USEFUL_MODS` is pressed and held.
If so, `call_boss(launch_with_single_action, ...) `.

The rest is up to the user to write their own kitten to parse and process as needed.

> Easier to just open the current finder directory in kitty ...

Yes, personally, I just need something like `nnn`.
All I wanted to do is bring more meaningful things to the terminal.
Anything that leads to weird user interactions,
or makes the project more difficult to maintain, is undesirable.

> ... just add scripts to your finder right click menu

Yes, there are many ways to do that. Maybe it could be added to services, better than using the modifier keys.

By the way, there is a problem with the open window function in the services submenu.
Selecting multiple folders only opens the last one. I haven't fixed it yet.
Do you think it's worth dealing with?

kovidgoyal:
On Sat, Feb 05, 2022 at 02:04:09AM -0800, page-down wrote:
> > How are you going to assign actions to arbitrary groups of files?
> 
> What I can think of is:
> - Explicitly configure some actions that receive all URLs as parameters.
> - These actions must be configured with modifier keys.
> 
> When opening URLs, first check if any `USEFUL_MODS` is pressed and held.
> If so, `call_boss(launch_with_single_action, ...) `.
> 
> The rest is up to the user to write their own kitten to parse and process as needed.

Seems to complicated, not worth it, IMO.

> 
> > Easier to just open the current finder directory in kitty ...
> 
> Yes, personally, I just need something like `nnn`.
> All I wanted to do is bring more meaningful things to the terminal.
> Anything that leads to weird user interactions,
> or makes the project more difficult to maintain, is undesirable.
> 
> > ... just add scripts to your finder right click menu
> 
> Yes, there are many ways to do that. Maybe it could be added to services, better than using the modifier keys.
> 
> By the way, there is a problem with the open window function in the services submenu.
> Selecting multiple folders only opens the last one. I haven't fixed it yet.
> Do you think it's worth dealing with?

Sorry I am not a macOS person, what does this refer to exactly?

page-down:
> Seems to complicated, not worth it, IMO.

Ok, I'm not spending time here for now.
All I want to do is to use kitty to compress the files.

> Sorry I am not a macOS person, what does this refer to exactly?

kitty/cocoa_window.m
```objc
- (BOOL)openFilesFromPasteboard ...
for (NSURL *url in filePathArray) {
    ...
    set_cocoa_pending_action(type, [path UTF8String]);
...
```

child-monitor.c
```c
typedef struct {
    char* wd;
    ...
} CocoaPendingActionsData;

static void
process_cocoa_pending_actions(void) {
    if (cocoa_pending_actions_data.wd) {
        if (cocoa_pending_actions[NEW_OS_WINDOW_WITH_WD]) { call_boss(new_os_window_with_wd, "s", cocoa_pending_actions_data.wd); }
        if (cocoa_pending_actions[NEW_TAB_WITH_WD]) { call_boss(new_tab_with_wd, "s", cocoa_pending_actions_data.wd); }
```

kovidgoyal:
Not something I care about, but patches welcome. Should be easy to implement by concatenating the paths with :

page-down:
Do you think it would be appropriate to add the ability to set as default handler? For example, users can set kitty as the default app for `ssh://`.

I have implemented `def cocoa_set_url_handler(...) -> bool`, but I don't know where to put it properly.

```shell
kitty +runpy 'from kitty.fast_data_types import cocoa_set_url_handler; cocoa_set_url_handler("ssh", "net.kovidgoyal.kitty");'
```

Can I add the following to setup.py? As far as I know this should work.
```ini
[Desktop Entry]
...
MimeType=x-scheme-handler/kitty
```

kovidgoyal:
On Sun, Feb 06, 2022 at 12:40:02AM -0800, page-down wrote:
> Do you think it would be appropriate to add the ability to set as default handler? For example, users can set kitty as the default app for `ssh://`.

Sure, though why does this need code in kitty? Isnt that done using an
OS interface?

> 
> I have implemented `def cocoa_set_url_handler(...) -> bool`, but I don't know where to put it properly.
> 
> ```shell
> kitty +runpy 'from kitty.fast_data_types import cocoa_set_url_handler; cocoa_set_url_handler("ssh", "net.kovidgoyal.kitty");'
> ```
> 
> Can I add the following to setup.py? As far as I know this should work.
> ```ini
> [Desktop Entry]
> ...
> MimeType=x-scheme-handler/kitty
> ```

Sure, it would need ot be for kitty-launcher.dekstop or similar I think,
since the Exec key will be kitty +open %U

page-down:
> Sure, though why does this need code in kitty? Isnt that done using an OS interface?

There is no such interface in macOS.
The reason I haven't submitted this is that `LSSetDefaultHandlerForURLScheme` is listed as deprecated, but Apple doesn't offer any alternative. I'm very uncomfortable adding this.

If this is not provided, then the user will need to figure it out themselves.
For example, third-party GUI/CLI programs, modifying plist+killall, using the soon-to-be-dropped `/usr/bin/python2.7` `import LaunchServices`, writing their own Swift, ObjC.

page-down:
I found an issue.
If an action is configured with `launch --type=window`, or an unconfigured file type,
if kitty is running at that moment, but without any OS window, no window will open.

For the error message window, I can open it in the OS window with the following argument.
I have not solved this issue yet. Maybe you have a better idea.

```python
self.launch('--type=os-window', kitty_exe(), '+runpy', #...
```

kovidgoyal:
Running with no os window is macos specific, so simply add a check in
Boss.launch is no os windows are present and create one with an empty
tab, if so and make sure that tab is active.

kovidgoyal:
This should take care of it, though I have not tested: https://github.com/kovidgoyal/kitty/commit/a2269cb66ee99c07d411a9b641ab3c6ca8bef866

page-down:
Thanks, I'm just about to commit a change to Boss.launch.
I tested and it works fine, and it handles better in launch.py.

page-down:
Oh, found another one.

Open the files in the order of one valid and one invalid. The error message is not displayed in a new OS window.

I noticed `with force_window_launch(needs_window_replaced):`, is it appropriate to use `self.launch('--type=os-window', ...`? I think don't do that.

If the second file opened in the window `launch --type=window` (or invalid), then the timing may have caused the program in the first kitty window to fail to receive the window resize event to redraw.
For example, performing an operation in Neovim will result in an incorrect draw (cursor) position.

page-down:
I tried the latest changes and the first problem has been solved.

In the case of opening multiple files in kitty windows at the same time, I think it is up to the user to add the delay.

The following configuration works fine.
```conf
action launch --type=window sh -c 'sleep 0.1 && exec nvim "$FILE_PATH"'
```

PR #4634, opened 2022-02-04
Add mappable action reverse_scroll_to_cursor

page-down:
For some programs that don't work well, they won't clear the blank lines under the cursor line. This action can be used to reverse the scrolling.

Also adjusted the name of the menu item that clears to the cursor `Cmd+K`.
Used initial capitalization and simplified the text (as it also works for the cursor prompt).
Moved Reset to the bottom, as this cleans up the most.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -93,6 +93,9 @@ Detailed list of changes
 
 - Fix a regression in the previous release that broke strikethrough (:disc:`4632`)
 
+- A new action :ac:`scroll_prompt_to_bottom` to moves the non-empty prompt lines
+  to the bottom. (:pull:`4634`)
+
 0.24.2 [2022-02-03]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/kitty/fast_data_types.pyi
+++ b/kitty/fast_data_types.pyi
@@ -1105,6 +1105,9 @@ class Screen:
     def reverse_scroll(self, amt: int, fill_from_scrollback: bool = False) -> bool:
         pass
 
+    def scroll_prompt_to_bottom(self) -> None:
+        pass
+
     def clear_selection(self) -> None:
         pass
 
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -3718,6 +3718,33 @@ reverse_scroll(Screen *self, PyObject *args) {
     Py_RETURN_NONE;
 }
 
+
+static PyObject*
+scroll_prompt_to_bottom(Screen *self, PyObject *args UNUSED) {
+    if (self->linebuf != self->main_linebuf || !self->historybuf->count) Py_RETURN_NONE;
+    int q = screen_cursor_at_a_shell_prompt(self);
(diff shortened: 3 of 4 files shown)


kovidgoyal:
I don't follow what's the use case for this action?

page-down:
For example, sometimes there will be some blank lines below the cursor at the end of autocomplete, so user can reverse scroll to move the cursor line back to the bottom of the screen.

(`scrollback_fill_enlarged_window yes` already configured)

Yes, it would be better if these programs could support unscroll escape sequence.

kovidgoyal:
Maybe name it something better in that case, so it is more clear what it
does. Perhaps "scroll_prompt_to_bottom" also you cant scroll the top
line of the prompt to the bottom of the screen, that will break multiline
prompts. Instead you would need to detect the last line of the prompt
that is not empty and scroll that to the bottom.

page-down:
> Maybe name it something better in that case, so it is more clear what it does.
> Perhaps "scroll_prompt_to_bottom"

Ok, so far I just moved the line where the cursor is to the bottom, maybe I need to check all the empty lines and reverse scroll.
This doesn't seem to require a prompt check, does it?
Starting from the bottom, count all empty lines.

---

changelog.rst

``` :ac:`select_tab` ```

Is this `:ac:` something new? Or is it typo?

Also I noticed that there are errors when generating documents.

```text
rm -rf docs/_build/
make dirhtml

shell-integration.rst:51: ERROR: Unknown interpreted text role "env".
shell-integration.rst:56: ERROR: Unknown interpreted text role "env".
```

kovidgoyal:
On Fri, Feb 04, 2022 at 03:07:29AM -0800, page-down wrote:
> > Maybe name it something better in that case, so it is more clear what it does.
> > Perhaps "scroll_prompt_to_bottom"
> 
> Ok, so far I just moved the line where the cursor is to the bottom, maybe I need to check all the empty lines and reverse scroll.
> This doesn't seem to require a prompt check, does it?

No, it doesnt. Basically check for the last non-empty line after the
cursor and scroll it to the bottom.

> Starting from the bottom, count all empty lines.
> 
> ---
> 
> changelog.rst
> 
> ``` :ac:`select_tab` ```
> 
> Is this `:ac:` something new? Or is it typo?

A typo.

> 
> Also I noticed that there are errors when generating documents.
> 
> ```text
> rm -rf docs/_build/
> make dirhtml
> 
> shell-integration.rst:51: ERROR: Unknown interpreted text role "env".
> shell-integration.rst:56: ERROR: Unknown interpreted text role "env".
> ```

These were fixed a while ago, are you up-to-date?

page-down:
I need to check from the last line, since fish multi-line editing may have empty lines.

Currently there is only one case when the last few lines of a fish multi-line edit are empty lines and will be cleared.
There is no way around this, except not to use it when entering multi-line text.

I saw the latest `:ac:` role update, which I also used.

> These were fixed a while ago, are you up-to-date?

Indeed, I remember reading this `env` related commit.

PR #4639, opened 2022-02-05
Fix issue with launch when running scripts without exec permission

page-down:
When executing a script without execute permission via launch, the arguments are missing.
I didn't think of any reason why the arguments need to be removed, so please let me know if I'm wrong.

Also fix `cmd[0]` that was replaced with `kitty` being executed incorrectly when there was no shebang.

I have a question, why do we use `kitty +hold` here?
If opts.hold is configured, is this a double hold?

### DIFF
--- a/kitty/tabs.py
+++ b/kitty/tabs.py
@@ -389,7 +389,7 @@ class Tab:  # {{{
                                     line = f.read(4096).splitlines()[0]
                                     cmd += shlex.split(line) + [old_exe]
                                 else:
-                                    cmd += [resolved_shell(get_options())[0], cmd[0]]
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Sat, Feb 05, 2022 at 04:47:24AM -0800, page-down wrote:
> I have a question, why do we use `kitty +hold` here?

Because shell scripts typically exit at the end, so running them will
close the window without an opportunity to see the output.

> If opts.hold is configured, is this a double hold?

You mean the --hold command line argument? Yes, it is, but it doesnt
really matter, since if you are editing the command line anyway, you
might as well execute the script by specifying the interpreter
explicitly.

PR #4646, opened 2022-02-06
macOS: Add a way to set kitty as the default handler for the URL schemes

page-down:
Add the ability to set kitty as default URL handler.

Previous discussion:
https://github.com/kovidgoyal/kitty/pull/4618#issuecomment-1030777843

Please review, thank you.

### DIFF
--- a/docs/open_actions.rst
+++ b/docs/open_actions.rst
@@ -120,6 +120,16 @@ These actions can also be executed from the command line by running::
     or
     kitty +open file_or_url ...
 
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I have merged, changing the implementation and calling convention slightly, please test.

page-down:
Thanks, I tested and it works as expected.

However is there a return type issue? There is also a `Py_RETURN_FALSE` left above.

PR #4653, opened 2022-02-07
Docs: Prefer namespaced launch points

page-down:
Replace `kitty list-fonts` with `kitty +list-fonts`.

According to the following commit, it is recommended to use namespaced launch points.

https://github.com/kovidgoyal/kitty/commit/b9f21f285ce800dfd020fd99d5b2fa622438d334


page-down:
I noticed that the namespaced launch point info is missing in the documentation and in the `--help`.
Should these be added? E.g. `+list-fonts`, `+kitten`, ...

Also could the cli page be added to the left navigation menu?
https://sw.kovidgoyal.net/kitty/invocation/

kovidgoyal:
No, the namespaced launch points are essentially separate programs and
they each have their individual documentation. There's not much benefit
to repeating that under a CLI page.

Issue #4694, opened 2022-02-11
Line-based scaling for high precision scrolling


xsrvmy:
There is currently an inconsistency in kitty with regards to how low precision and high precision scrolling is handled. With low precision scrolling, the scale is based on the number of lines. With high precision scrolling, the scale is based on the number of pixel. This leads to a situation where the perceived scrolling speed changes for high precision devices when different display scalings are involved. This is especially annoying on wayland because all devices are reported as high precision, so a mouse tick will scroll different number of lines on different dpi screens.

I am proposing a new settings called something like `scroll_pixel_per_line`. When this setting is set, kitty would ignore `touch_scroll_multiplier`, and instead use this new setting to convert scrolling pixels per line, then apply `wheel_scroll_multiplier`. When the value of this new setting is set correctly, this means that mouse scrolling would work the same on wayland and x11. It would also be a better fix for #4680.

kovidgoyal:
On Fri, Feb 11, 2022 at 12:21:53PM -0800, Rui Ming (Max) Xiong wrote:
> There is currently an inconsistency in kitty with regards to how low precision and high precision scrolling is handled. With low precision scrolling, the scale is based on the number of lines. With high precision scrolling, the scale is based on the number of pixel. This leads to a situation where the perceived scrolling speed changes for high precision devices when different display scalings are involved. This is especially annoying on wayland because all devices are reported as high precision, so a mouse tick will scroll different number of lines on different dpi screens.

That's not an inconsistency. High precision scrolling is supposed to use
pixels. It's the job of the operating system to send the correct number
of pixels, not for every application to make up for deficiencies in the
OS.

As far as dealing with Wayland's brokenness on multiple screens, I think a better approach
is to simply have a boolean setting to divide the number of pixels
reported by the OS by the scale of the OS Window. This can be documented
to be useful for working around OS brokenness and can default to true on
Linux and False on macos, where it isnt needed as far as I know.

xsrvmy:
Actually there are two separate problems:

1. On wayland everything is considered high resolution. This traces all the way back to `glfw/wl_init.c` setting the high resolution flag to `1`. The wayland protocol does expose this information so should be assigned per event like on macOS instead.
2. The scrolling for touchpads is applied as physical pixels, but they should be applied as scaled pixels.

> divide the number of pixels reported by the OS by the scale of the OS Window

[...]

Issue #4680, opened 2022-02-10
Scrollback speed affect scrolling speed in vim (and neovim)


xsrvmy:
**Describe the bug**
When wheel_scroll_multiplier or touch_scroll_multiplier is set, vim scrolls way too fast.

**To Reproduce**
1. Set wheel_scroll_multiplier or touch_scroll_multiplier to 9
2. Open a large file in vim (or just enter a lot of lines and turn on line numbers)
3. Notices that it scrolls 9 lines at once instead of the usual 3

It seems that vim is receiving multiple scroll events per scroll. Does this have anything to do with terminfo?

In other terminals:
- less is affected by scrollback settings
- vim is not affected by scrollback settings

In kitty:
- less is affected by scrollback settings
- vim is affected by scrollback settings

**Environment details**
```
kitty 0.24.2 created by Kovid Goyal
Linux <redacted> 5.16.7-arch1-1 #1 SMP PREEMPT Sun, 06 Feb 2022 16:47:29 +0000 x86_64
Arch Linux 5.16.7-arch1-1 (/dev/tty)

LSB_VERSION=1.4
DISTRIB_ID=Arch
DISTRIB_RELEASE=rolling
DISTRIB_DESCRIPTION="Arch Linux"
Running under: Wayland
Frozen: False
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /usr/bin/zsh
Loaded config files:
  /home/<redacted>/.config/kitty/kitty.conf

Config options different from defaults:
enable_audio_bell       False
font_family             ISVK
font_size               12.0
initial_window_height   (30, 'cells')
initial_window_width    (100, 'cells')
remember_window_size    False
touch_scroll_multiplier 9.0
url_style               1
Colors:
	background              #191c25   
	color0                  #2e313d   
	color1                  #cf8164   
	color10                 #92c47e   
	color11                 #d2b45f   
	color12                 #a0b9d8   
	color13                 #c6aed7   
	color14                 #7dc2c7   
	color15                 #f0ecfe   
	color2                  #76a065   
	color3                  #ab924c   
	color4                  #8296b0   
	color5                  #a18daf   
	color6                  #659ea2   
	color7                  #b5b4c9   
	color8                  #5b5f71   
	color9                  #fe9f7c   
	foreground              #b5b4c9   
	selection_background    #b5b4c9   
	selection_foreground    #191c25   
[...]

PR #4695, opened 2022-02-11
Add touch_scroll_pixel_per_line config option, implements #4694

xsrvmy:
Implements #4694

Adds an option to map touch scrolling to lines instead of pixels for consistency on hidpi screens. Also provides a better fix for #4680.

Note that I defaulted to 0 to not break backwards compatibility. There is the challenge of finding out what to set this value to if your mouse scroll wheel is recognized as high precision and you want to map one scroll to one line. I found mine to be exactly 15 by printing out yoffset in the source code, but I'm not sure if this is different depending on system settings for scroll sensitivity.

Would it make sense to still use the `touch_scroll_multiplier` instead of `wheel_scroll_multiplier` for the number of lines to scroll outside of mouse tracking? I think it makes sense because it would allow reversing the touchpad without reversing the scrollwheel, but also I think it could be confusing.


kovidgoyal:
See discussion in linked issue, once that is resolved you can send a new
PR.

PR #4699, opened 2022-02-12
Seperate discrete (mouse wheel) and continuous (touchpad) scroll events on wayland

xsrvmy:
This bumps seatVersion to 5. The code should still work if the user is using seatVersion 4.
Additionally this means that the code now requires libwayland 1.10 to compile. Even Ubuntu 16.04 had that version so we should be safe here.

Aside: doesn't glfw's license require marking the version as modified?

### DIFF
--- a/glfw/wl_init.c
+++ b/glfw/wl_init.c
@@ -317,10 +317,10 @@ static void pointerHandleButton(void* data UNUSED,
 #undef x
 #undef y
 
-// flags for ignoring axis events following axis_discrete events in the
+// counters for ignoring axis events following axis_discrete events in the
 // same frame along the same axis
-static bool ignoreNextX = false;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Sat, Feb 12, 2022 at 02:59:57PM -0800, Rui Ming (Max) Xiong wrote:
> This bumps seatVersion to 5. The code should still work if the user is using seatVersion 4.
> Additionally this means that the code now requires libwayland 1.10 to compile. Even Ubuntu 16.04 had that version so we should be safe here.
> 
> Aside: doesn&#39;t glfw&#39;s license require marking the version as modified?

I'm not distributing glfw, its a private dependency, so there is no
chance of any confusion. And its pretty obvious it is modified, but if
you wish to add some more explicit marking feel free to send a PR.

PR #4710, opened 2022-02-14
Add wheel_scroll_min_lines to set the minimum scrolling lines

page-down:
Under macOS, some mice have a scroll wheel that starts at yoffset 0.100...1 (with default multiplier it's 0.500..5), resulting in one line per click with a slow scroll.
If a larger multiplier is set, the acceleration is too fast.

Add a new option `wheel_scroll_min_lines` to set the minimum number of scrolling lines.

If the number is positive, the minimum number of lines will be taken. (Obviously acceleration does not take effect when the minimum number is not reached.)
If it is negative, the number of lines is always added and the acceleration remains the same and takes effect immediately.

For me personally, I use the default `wheel_scroll_multiplier 5.0`, and `wheel_scroll_min_lines -2`.

Please review, thank you.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -135,6 +135,8 @@ Detailed list of changes
 
 - Wayland: Fix touchpads and high resolution wheels not scrolling at the same speed on monitors with different scales (:iss:`4703`)
 
+- Add an option :opt:`wheel_scroll_min_lines` to set the minimum number of lines for mouse wheel scrolling (:pull:`4710`)
+
 
 0.24.2 [2022-02-03]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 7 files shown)


page-down:
Is there a chance that yoffset is 0?

```c
if (yoffset != 0) {
    const int min_lines = ...
}
```

I saw that it is already checked for zero and returned.

```c
if (yoffset == 0.0) return;
```

kovidgoyal:
currently, no. but that might change in the future, and I want to be
clear about intent.

page-down:
Nevermind, I missed `SCALE_SCROLL`.

For example, scrolling is disabled when `wheel_scroll_multiplier` is zero. Although this may seem useless.

Thank you. This is always good to check.

page-down:
I noticed that `wheel_scroll_multiplier` only scales when the mouse is not grabbed, does this option need to be the same?

Although I need this, I don't want to introduce effects that do not match the design.

kovidgoyal:
It doesn't matter when grabbed. When the mouse is grabbed yoffset is not
sent to the terminal program anyway.

However, for consistency in case in the future I design a new mouse
protocol that includes this information, it might be worth clamping
min_lines to 1 when grabbed. The idea being that the terminal program
should get unmodified events so that it can best decide what to do.

PR #4714, opened 2022-02-17
macOS: Fix phantom window appears when executing full screen in transition

page-down:
When a window is in the macOS fullscreen transition animation, if you toggle fullscreen again, an identical window (The shell prompt is cleared) appears that cannot be manipulated or closed and you have to quit the app.

Remove hard-coded fullscreen shortcut, the shortcut in the menu can be customized.

During animation, the newly rendered window (the shell prompt has not yet been redrawn) will be used as the animation, and the window will no longer appear to be rescaled up with jaggies.

---

I found a problem. (before this PR, not yet solved)

When the mouse cursor is beam, it gets changed to arrow after switching to full screen.

Before this PR, the cursor changed as follows:
- before fullscreen: beam
- after fullscreen: (After the fullscreen animation has been completed.) 
  - arrow -> beam -> arrow 

After this PR, the cursor changed as follows:
- before fullscreen: beam
- after fullscreen: beam -> arrow

The event handler `cursorUpdate` is executed and `updateCursorImage` is executed.

I also found that occasionally there is a `Bad cursor rect event` error log output by cocoa. It seems likely that there is a problem updating the cursor image (to arrow...) when the window is not fully fullscreen.

```text
---------------- key down -------------------
NSEvent: type=KeyDown loc=(682,750.016) time=821.6 flags=0x140109 win=0x7f7f59933940 winNum=202 ctxt=0x0 chars="" unmodchars="f" repeat=0 keyCode=3
keyDown triggerred global menu bar action ignoring
kitty[1026:16302] Bad cursor rect event, flags = 1310729
-------------- flags changed -----------------
NSEvent: type=FlagsChanged loc=(50,525.016) time=821.8 flags=0x40101 win=0x7f7f59933940 winNum=202 ctxt=0x0 keyCode=55
flagsChanged: modifier: super native_key: 0x37 (<noname>) glfw_key: 0xe064 mods: ctrl
```

Do you have any ideas to fix this?


kovidgoyal:
No clue. Maybe save the cursor state before toggling fullscreen and
restore it after the transition completes.

PR #4716, opened 2022-02-18
macOS: Fix the mouse cursor being set to arrow after switching desktops

page-down:
macOS will set the cursor to arrow after milliseconds after the switch
desktop animation ends.
So the cursor that was updated immediately after the focus will be changed again.
This also affects toggling fullscreen.

Reduce unnecessary cursor updates since we don't use cursor rects.

Please review, thank you.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -144,6 +144,10 @@ Detailed list of changes
 
 - Add an option :opt:`wheel_scroll_min_lines` to set the minimum number of lines for mouse wheel scrolling when using a mouse with a wheel that generates very small offsets when slow scrolling (:pull:`4710`)
 
+- macOS: Allows to configure the toggle fullscreen shortcut in global menu. (:pull:`4714`)
+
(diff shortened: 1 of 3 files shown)


page-down:
This error log still appears (rarely), and I have no idea where this event is coming from.

And I can't find the steps that are reproducible. (ctrl+cmd+f ...)

```text
kitty[4110:173498] Bad cursor rect event, flags = 1376265
```

```text
frame #1: 0x00007ff821b21f58 AppKit`-[NSApplication(NSCursorRects) _handleCursorRectEvent:] + 318
frame #2: 0x00007ff8217376cf AppKit`-[NSApplication(NSEvent) sendEvent:] + 3335
frame #3: 0x00007ff8219f01a7 AppKit`-[NSApplication _handleEvent:] + 65
frame #4: 0x00007ff8215b793e AppKit`-[NSApplication run] + 623
frame #5: 0x0000000105a748ba glfw-cocoa.so`glfwRunMainLoop + 90
```

kovidgoyal:
If the error causes no actual bad effects, I dont think its worth bothering
about.

Issue #1945, opened 2019-09-01
Underline rendering different with iTerm2


ingjieye:
I'don't know if this is a bug or misconfiguration or something. 

Kitty render underline as:
<img width="299" alt="屏幕快照 2019-09-01 下午5 03 49" src="https://user-images.githubusercontent.com/13421169/64074201-e9441200-ccda-11e9-95e2-c4fe5b40cdea.png">
While iTerm2 render underline as:
<img width="296" alt="屏幕快照 2019-09-01 下午5 03 53" src="https://user-images.githubusercontent.com/13421169/64074207-f6f99780-ccda-11e9-8310-e2d11f0d0cd5.png">
 
I've tried to adjust the `adjust_line_height` but won't work.
So how do I make kitty render underline like iTerm2 does? 
Thanks!

kovidgoyal:
I'm afraid that there is no setting to control underline rendering
in kitty. Underline thickness and position comes from the informatin in
the font, specifically CTFontGetUnderLinePosition() see core_text.m for
details.

ingjieye:
> I'm afraid that there is no setting to control underline rendering
> in kitty. Underline thickness and position comes from the informatin in
> the font, specifically CTFontGetUnderLinePosition() see core_text.m for
> details.

In my configuration Kitty and iTerm2 use the same font.
So does this mean iTerm2 use something rather than CTFontGetUnderLinePosition() to get underline positions? Is there a way to optimize this?

kovidgoyal:
No idea, iTerm is open source am sure you can check. Rendering the same
as iTerm is not really a goal for kitty though. If there is some actual
improvement that can be made to the algorithm to place underlines, I'm
open to ideas.

ingjieye:
> No idea, iTerm is open source am sure you can check. Rendering the same
> as iTerm is not really a goal for kitty though. If there is some actual
> improvement that can be made to the algorithm to place underlines, I'm
> open to ideas.

I know that rendering the same as iTerm does is not Kitty's goal. And it shouldn't. 
But underline rendering indeed iTerm do better than Kitty.

So I've checked the iTerm's source code. there's no `CTFontGetUnderLinePosition()` function. And do have some function regarding underline, but I'm not a native developer so it's hard for me to read into iTerm's source code.

Then I decide why don't I just change the return value of `CTFontGetUnderLinePosition` and `CTFontGetUnderlineThickness`. So here's What I do:

[...]

PR #4723, opened 2022-02-19
Add basic underline offset in px

jmederosalvarado:
Hi, so I recently started using macOS and noticed this issue described in #1945. I understand this is not a problem with kitty, but I still would like to have a config option to offset the underline position.

I've never worked with graphic libs before, and I'm just getting around kitty's codebase, so please let me know if there's something that I should do differently. Right now the `underline_offset`  config option is in pixels, but I think maybe it makes sense to use pts, however, I don't know how to convert pts to px within core_text.m.

Thank you.

### DIFF
--- a/kitty/core_text.m
+++ b/kitty/core_text.m
@@ -379,6 +379,7 @@ cell_metrics(PyObject *s, unsigned int* cell_width, unsigned int* cell_height, u
     int baseline_offset = 0;
     if (OPT(adjust_baseline_px) != 0) baseline_offset = OPT(adjust_baseline_px);
     else if (OPT(adjust_baseline_frac) != 0) baseline_offset = (int)(*cell_height * OPT(adjust_baseline_frac));
+    int underline_offset = OPT(underline_offset);
     *baseline = (unsigned int)floor(bounds_ascent + 0.5);
     // Not sure if we should add this to bounds ascent and then round it or add
     // it to already rounded baseline and round again.
@@ -397,6 +398,9 @@ cell_metrics(PyObject *s, unsigned int* cell_width, unsigned int* cell_height, u
         *baseline = adjust_ypos(*baseline, *cell_height, baseline_offset);
         *underline_position = adjust_ypos(*underline_position, *cell_height, baseline_offset);
         *strikethrough_position = adjust_ypos(*strikethrough_position, *cell_height, baseline_offset);
+    }
+    if (underline_offset){
+        *underline_position = adjust_ypos(*underline_position, *cell_height, underline_offset);
     }
 
     CFRelease(test_frame); CFRelease(path); CFRelease(framesetter);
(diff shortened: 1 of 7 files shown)


jmederosalvarado:
I would also like to maybe be able to replicate iTerm2 behavior, pointed out in #1945 as well, of drawing text on top of underlines, I think it looks better(but that's just my personal opinion). However, probably this is not a minor change like offsetting the underline, so I understand if you don't want to accept a patch for it. In case you do consider it a viable option I'd appreciate it if you pointed me in the right direction, I have no idea of how to do this.

kovidgoyal:
If you want to do this then the way to do it is to add configuration to
modify font parameters, just like there is the existing font_features
setting add a modify_font setting that allows the user to specify
various modifications to the font:

1) Change the size by a percentage of existing size or by pts
2) Change the underline, strikethrough positions, by font units
3) Change the thickness of underline/strikethrough

2, 3 will only take effect for the main font, but (1) would apply to all
fonts.

These settings would apply at the time of font face creation.

As for having underlines interrupted by descenders, the way to do it is
to generate per glyph bit masks in the underline band with zero for
pixels to be not drawn. Then store these on the GPU along with the
glyph masks, possibly as a tail on each glyph. Then when drawing
the underline in the cell_fragment shader use that mask.

This is a lot of work and it requires extra GPU memory to store these
bands. And extra GPU compute to apply the masks on every render. I dont
think this is worth it, but I would be willing to consider an
implementation provided it was optional.

nickwynja:
Apologies, I had a typo in an unrelated commit that accidentally referenced this instead of my issue. I force pushed a fix to the commit but it looks like Github has kept around the incorrect reference.

fenetikm:
I'm keen on this one too - the underline can be too close to the base line in some fonts and make it hard to read. Anything else needed in this PR? Happy to pitch in.

kovidgoyal:
What's needed is outlined in my post above.

fenetikm:
Apologies - I lazily read through the thread. So I'll take that as implementing the `modify_font` option, your points 1-3. Cool. Let me know if I have that wrong, otherwise that's what I'll take a look at.

kovidgoyal:
Yes, modify_font is what is needed.

fenetikm:
@kovidgoyal quick question (I didn't think this would be worth opening a separate ticket) I tried to add another argument to `set_font_data` in `fast_data_types` but when I call it in `set_font_family` and pass it in, it says that the argument count is wrong (11 expected but 12 given) - is there something special I need to do re compilation to get the update propagated / recognised? I simply ran `make clean` then `make app` (I'm on macOS)

Similarly if I remove an arg from `set_font_data` in `fast_data_types` it doesn't get picked up either - the error still says it expects 11 args.

kovidgoyal:
Closed in favor of #5265

PR #4733, opened 2022-02-23
Bash integration: Use builtin to call the commands

page-down:
Explicitly use `builtin` to call the commands.
Use Bash's string manipulation facility to remove new lines.
Use the official project name of Bash.

GNU Bash
https://www.gnu.org/software/bash/

---

Also, the fish installed in CI is an outdated version. This version is broken and not worth supporting. I believe it will not work.

https://github.com/kovidgoyal/kitty/commit/081d6a3f166d013e49193838474415e0ded5ebd5

`sudo apt-get install -y ... fish`

```text
https://github.com/kovidgoyal/kitty/pull/4558#issuecomment-1020261408

faho commented on Jan 25

> ... Ubuntu's packages in 20.04 (the current LTS) are still at 3.1.0 (not even 3.1.2, the bugfix release) with some Debian patches.
> That's rather out of date (about 80% of commits in 3.3.1) and makes tests with it essentially meaningless.

https://github.com/kovidgoyal/kitty/issues/4619

>> If you would like to support old versions, please let me know and I will revert them.
>
> It's fine, if fish's maintainer feels that 3.1.2 is not supported, I am
> OK with not supporting it in kitty.
```

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -78,7 +78,7 @@ Detailed list of changes
 0.24.3 [future]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- BASH integration: No longer modify :file:`~/.bashrc` to load :ref:`shell integration <shell_integration>`.
+- Bash integration: No longer modify :file:`~/.bashrc` to load :ref:`shell integration <shell_integration>`.
   It is recommended to remove the lines used to load the shell integration from :file:`~/.bashrc` as they are no-ops.
 
 - macOS: Allow kitty to handle various URL types. Can be configured via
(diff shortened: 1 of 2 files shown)


kovidgoyal:
As for fish, currently there are no tests for it, so it doesnt matter.
If in the future we write tests, we can presumably change the CI script
to installer a newer version.

page-down:
I see CI reporting errors and I haven't found the cause yet.

For fish, I think two months later the version will be updated to 3.3.1 in this year's LTS release.

kovidgoyal:
On Tue, Feb 22, 2022 at 10:55:28PM -0800, page-down wrote:
> I see CI reporting errors and I haven't found the cause yet.

I have already committed the fix for it.

> 
> For fish, I think two months later the version will be updated to 3.3.1 in this year's LTS release.

Good enough, so lets leave it as is for now. If we add tests for fish
we can skip then on CI temporarily.

PR #4735, opened 2022-02-23
Add fish integration tests

page-down:
Add fish integration tests.

Also I noticed that when getting the last command output and comparing the text on the screen it does not match. This is because the last command output is missing the newlines.

`kitty/kitty_tests/__init__.py`
```python
def last_cmd_output( ... ):
    # ...
    return ''.join(lines)
```

Do we need to use `\n` to concatenate lines here?


kovidgoyal:
On Wed, Feb 23, 2022 at 04:03:46AM -0800, page-down wrote:
> Add fish integration tests.
> 
> Also I noticed that when getting the last command output and comparing the text on the screen it does not match. This is because the last command output is missing the newlines.
> 
> `kitty/kitty_tests/__init__.py`
> ```python
> def last_cmd_output( ... ):
>     # ...
>     return &#39;&#39;.join(lines)
> ```
> 
> Do we need to use `\n` to concatenate lines here?

The newlines are supposed to be included already. See the cmd_output()
function in window.py

PR #4745, opened 2022-02-24
Shell integration: Check fish version and exit on outdated versions

page-down:
Efficiently check the version via the new variable added in fish 3.3.0 and exit when not satisfied to avoid breaking outdated fish.

```text
set --show fish_killring 
$fish_killring: set in global scope, unexported, with 0 elements
Variable is read-only
```

I have benchmarked that using glob or regex to check the version string will have performance loss. (The fish that comes with some Linux distributions is broken and has no version number.)

Consistent with the behavior when using lower version of the shell and not outputting warning text.
This is equivalent to not enabling integration.

I have not introduced `\eP@kitty-print|...` or `base64` to minimize the impact.

Please review, thank you.

### DIFF
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -19,6 +19,8 @@ end
 
 status is-interactive || exit 0
 not functions -q __ksi_schedule || exit 0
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I suggest outputting a debug message. You dont need base64 since its a fixed
string you can just include it directly already base64 encoded. And any
performance loss will be only on outdated fish, which will gradually
disappear anyway.

PR #4759, opened 2022-02-27
Try to get ZDOTDIR from /etc/zshenv when all zsh startup files are not present

page-down:
`ZDOTDIR` may be configured in /etc/zshenv.

Check only if the `ZDOTDIR` does not exist and all zsh startup files are not present.

Does not affect the performance for normal users.

Tested with
```zsh
# /etc/zshenv
[[ -o globalrcs ]] || return
[[ -v ZDOTDIR ]] || ZDOTDIR="${XDG_CONFIG_HOME:-$HOME/.config}/zsh"
```

zsh auto integration never worked for me, until this patch. I don't have any zsh files in `$HOME`.

Please review, thank you.

---

kitty/child.py
```python
class Child:
    # ...
    def fork(self) -> Optional[int]:
        # ....
            # bash will only source ~/.bash_profile if ...
```

Is it possible to narrow down the scope of the shells that need to modify `argv[0]`?
For example, limit it to `bash` or any other shell that is known to be necessary.

### DIFF
--- a/kitty/shell_integration.py
+++ b/kitty/shell_integration.py
@@ -3,11 +3,12 @@
 
 
 import os
+import subprocess
 from typing import Optional, Dict, List
 
 from .options.types import Options
 from .constants import shell_integration_dir
-from .utils import log_error
+from .utils import log_error, which
 
 
 def setup_fish_env(env: Dict[str, str], argv: List[str]) -> None:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Sun, Feb 27, 2022 at 07:44:19AM -0800, page-down wrote:
> Is it possible to narrow down the scope of the shells that need to modify `argv[0]`?
> For example, limit it to `bash` or any other shell that is known to be necessary.

Which shell doesnt alter its behavior when it thinks it is a login
shell? If you know of some such sheels, feel free to exclude them from
this.

page-down:
> Which shell doesnt alter its behavior when it thinks it is a login shell?

I personally have not yet encountered a problematic shell,
so if it's not worth whitelisting, wait until someone complains and then exclude.

PR #4764, opened 2022-02-28
Also fall back to check the last known working version of fish

page-down:
The read-only variable fish_killring, added in fish 3.3.0, may become modifiable in the future.
So also fall back to check the last known working version: fish 3.2.0.

```text
https://fishshell.com/docs/3.2/relnotes.html#interactive-improvements

$status_generation: set in global scope, unexported, with 1 elements
Variable is read-only
$status_generation[1]: |0|
```

Finally check that the `$version` string does not match `3.1.*`.
All versions prior to 3.1.0 will not automatically load the integration script,
and if the user loads the script manually, errors will occur before reaching this check.

It will never reach `string match` unless fish breaks backward compatibility by removing the first two variables.
This is the last check to make sure the script will work properly.

Please review, thank you.

### DIFF
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -19,10 +19,10 @@ end
 
 status is-interactive || exit 0
 not functions -q __ksi_schedule || exit 0
-# Check fish version 3.3.0+ efficiently and exit on outdated versions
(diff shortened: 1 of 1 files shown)


kovidgoyal:
why are you using different or operators?

page-down:
The statements after and , or will be run as a whole.
The purpose is to simplify the code without unnecessary `\` line breaks and `if`s.
In this place I think it is acceptable without a large code section.

Do you think there are any readability or maintainability issues?

PR #4779, opened 2022-03-02
wayland: always commit surface on configure

Nefsen402:
A surface commit must happen after the acknowledgement of a configure, even if it is a same size configure.
Just move the surface commit outside of the if statement to make it always run on a xdg configure.


Nefsen402:
Sorry about the PR noise, I forgot about this in my last patch.

PR #4784, opened 2022-03-04
Shell integration: Fix running bash non-interactively

page-down:
In POSIX mode, bash does not execute ENV when it is not an interactive shell.

There may be issues when executing bash commands or script files with `launch` (and perhaps with `--type=background`).

```shell
bash -c "exec sh -i"
# echo $HISTFILE -> ~/.bash_history
```

In this PR, try to parse the arguments and ignore the non-interactive cases.

```text
# non-interactive
bash -c "cmd"
bash file.sh

# interactive
bash -i file.sh
bash -s arg1 --rcfile --posix
bash -ic "cmd" # This one is interactive, but not important so skip it.
```

Please review, thank you.
Hopefully I didn't miss something.

### DIFF
--- a/kitty/shell_integration.py
+++ b/kitty/shell_integration.py
@@ -71,21 +71,66 @@ def setup_bash_env(env: Dict[str, str], argv: List[str]) -> None:
     inject = {'1'}
     posix_env = rcfile = ''
     remove_args = set()
+
+    expecting_multi_chars_opt = True
+    expecting_option_arg = False
+    interactive_opt = False
+    expecting_file_arg = False
+    file_arg_set = False
+
     for i in range(1, len(argv)):
         arg = argv[i]
-        if arg == '--posix':
-            inject.add('posix')
-            posix_env = env.get('ENV', '')
-            remove_args.add(i)
-        elif arg == '--norc':
-            inject.add('no-rc')
-            remove_args.add(i)
-        elif arg == '--noprofile':
-            inject.add('no-profile')
-            remove_args.add(i)
-        elif arg in ('--rcfile', '--init-file') and i + 1 < len(argv):
-            rcfile = argv[i+1]
-            remove_args |= {i, i+1}
+        if expecting_file_arg:
+            file_arg_set = True
+            break
+        if expecting_option_arg:
+            expecting_option_arg = False
+            continue
+        if arg in ('-', '--'):
+            if not expecting_file_arg:
+                expecting_file_arg = True
+            continue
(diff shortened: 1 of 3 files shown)


page-down:
I also found a case of repeated sourcing bash integration script.

- bashrc: Set INST_DIR If empty, set KSI enabled, source. (for SSH)
- Enable automatic integration in kitty. (for other shell, or `kitty --config=NONE` debugging)

I think it might be worth considering putting the code for restoration at the very beginning, and preventing repeated sourcing via `_ksi_in_bash_inject`.
Similar to fish's integration script, maybe also check the version number and exit.

page-down:
I tested the latest commit. The problem with repeated sourcing still exists.

- ENV (with KITTY_SHELL_INTEGRATION)
  - _ksi_inject
    - _ksi_safe_source "$KITTY_BASH_RCFILE";
      - .bashrc set KITTY_INSTALLATION_DIR for SSH
      - .bashrc Manual shell integration (from docs): if test -n "$KITTY_INSTALLATION_DIR"; then KSI=enabled; souce kitty.bash; fi
        - _ksi_inject
          - _ksi_safe_source "$KITTY_BASH_RCFILE";
          - ...

kovidgoyal:
On Fri, Mar 04, 2022 at 05:21:34AM -0800, page-down wrote:
> I also found a case of repeated sourcing bash integration script.
> 
> - bashrc: Set INST_DIR If empty, set KSI enabled, source. (for SSH)
> - Enable automatic integration in kitty. (for other shell, or `kitty --config=NONE` debugging)
> 
> I think it might be worth considering putting the code for restoration at the very beginning, and preventing repeated sourcing via `_ksi_in_bash_inject`.
> Similar to fish's integration script, maybe also check the version number and exit.

This wont actually help on macOS though. Because bash-3.2 doesnt seem to
source ENV in POSIX mode at all.

kovidgoyal:
> I tested the latest commit. The problem with repeated sourcing still exists.
> 
> * ENV (with KITTY_SHELL_INTEGRATION)
>   
>   * _ksi_inject
>     
>     * _ksi_safe_source "$KITTY_BASH_RCFILE";
>       
>       * .bashrc set KITTY_INSTALLATION_DIR for SSH
>       * .bashrc Manual shell integration (from docs): if test -n "$KITTY_INSTALLATION_DIR"; then KSI=enabled; souce kitty.bash; fi
>         
>         * _ksi_inject
>           
>           * _ksi_safe_source "$KITTY_BASH_RCFILE";
>           * ...

SHould be fine now

PR #4794, opened 2022-03-07
Minor tweaks to ssh kitten and shell integration

page-down:
- Avoid unnecessary which and fix typos.
- Adjust the order of exec function definitions: zsh, fish, bash.
- Improve SSH detection without affecting daily performance.
- Remove the fish integration B prompt marking. (The same as zsh, bash.)

For SSH detection, first exclude the case that KITTY_PID exists.
Then check the existence of KITTY_WINDOW_ID.
Call `who` only if sudo is running (environment variables are cleared) and shell integration is manually installed for the switched user.

- When run locally, only one environment variable, KITTY_PID, will be checked.
- In most SSH connections, only KITTY_PID, SSH_TTY will be checked.
- Only when the integration is installed manually and sudo is run, four environment variables will be checked and who will be called once.

Users can add KITTY_WINDOW_ID to sudo env_keep to avoid calling who.
SSH_TTY is not suitable for adding to env_keep, because it is possible that the switched user does not have permission to access it and affecting other programs.

Please review, thank you.

I'm a little concerned about whether `tar x --no-same-owner` will work properly on OpenBSD and am going to give it a try.

### DIFF
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -73,7 +73,7 @@ def make_tarfile(ssh_opts: SSHOptions, base_env: Dict[str, str]) -> bytes:
     def filter_from_globs(*pats: str) -> Callable[[tarfile.TarInfo], Optional[tarfile.TarInfo]]:
         def filter(tarinfo: tarfile.TarInfo) -> Optional[tarfile.TarInfo]:
             for junk_dir in ('.DS_Store', '__pycache__'):
-                for pat in (f'*/{junk_dir}', '*/{junk_dir}/*'):
+                for pat in (f'*/{junk_dir}', f'*/{junk_dir}/*'):
                     if fnmatch.fnmatch(tarinfo.name, pat):
                         return None
             for pat in pats:
--- a/kitty_tests/shell_integration.py
+++ b/kitty_tests/shell_integration.py
@@ -213,16 +213,8 @@ function _set_status_prompt; function fish_prompt; echo -n "$pipestatus $status 
             pty.write_to_child('i')
             pty.wait_till(lambda: pty.screen.cursor.shape == CURSOR_BEAM)
             pty.send_cmd_to_child('_set_key default')
-
-            # pipestatus
-            pty.send_cmd_to_child('clear;false|true|false')
-            pty.send_cmd_to_child('echo $pipestatus $status')
-            pty.wait_till(lambda: pty.screen_contents().count(right_prompt) == 2)
-            self.ae(pty.last_cmd_output(), '1 0 1 1')
-            pty.send_cmd_to_child('_set_status_prompt')
-            pty.send_cmd_to_child('false|true|false')
             pty.wait_till(lambda: pty.screen_contents().count(right_prompt) == 4)
-            self.assertTrue(str(pty.screen.line(pty.screen.cursor.y)).startswith(f'1 0 1 1 {fish_prompt}'))
+            pty.wait_till(lambda: pty.screen.cursor.shape == CURSOR_BEAM)
 
             pty.send_cmd_to_child('exit')
 
--- a/shell-integration/bash/kitty.bash
+++ b/shell-integration/bash/kitty.bash
@@ -155,8 +155,17 @@ _ksi_main() {
     fi
 
     if [[ "${_ksi_prompt[title]}" == "y" ]]; then
-        if [[ -n "$SSH_TTY$SSH2_TTY" ]]; then
+        if [[ -n "$KITTY_PID" ]]; then
+            # kitty running locally
+        elif [[ -n "$SSH_TTY" || -n "$SSH2_TTY$KITTY_WINDOW_ID" ]]; then
+            # connected to most SSH servers
+            # or use ssh kitten to connected to some SSH servers that do not set SSH_TTY
             _ksi_prompt[hostname_prefix]="\h: ";
+        elif [[ -n "$(builtin command -v who)" ]]; then
+            # the shell integration script is installed manually on the remote system
+            # the environment variables are cleared after sudo
+            # OpenSSH's sshd creates entries in utmp for every login so use those
+            [[ "$(builtin command who -m 2> /dev/null)" =~ "\([a-fA-F.:0-9]+\)$" ]] && _ksi_prompt[hostname_prefix]="\h: ";
         fi
         # see https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html#Controlling-the-Prompt
         # we use suffix here because some distros add title setting to their bashrc files by default
--- a/shell-integration/bash/kitty.bash
+++ b/shell-integration/bash/kitty.bash
@@ -155,17 +155,19 @@ _ksi_main() {
     fi
 
     if [[ "${_ksi_prompt[title]}" == "y" ]]; then
-        if [[ -n "$KITTY_PID" ]]; then
-            # kitty running locally
-        elif [[ -n "$SSH_TTY" || -n "$SSH2_TTY$KITTY_WINDOW_ID" ]]; then
-            # connected to most SSH servers
-            # or use ssh kitten to connected to some SSH servers that do not set SSH_TTY
-            _ksi_prompt[hostname_prefix]="\h: ";
-        elif [[ -n "$(builtin command -v who)" ]]; then
-            # the shell integration script is installed manually on the remote system
-            # the environment variables are cleared after sudo
-            # OpenSSH's sshd creates entries in utmp for every login so use those
-            [[ "$(builtin command who -m 2> /dev/null)" =~ "\([a-fA-F.:0-9]+\)$" ]] && _ksi_prompt[hostname_prefix]="\h: ";
+        if [[ -z "$KITTY_PID" ]]; then
+            if [[ -n "$SSH_TTY" || -n "$SSH2_TTY$KITTY_WINDOW_ID" ]]; then
+                # connected to most SSH servers
+                # or use ssh kitten to connected to some SSH servers that do not set SSH_TTY
+                _ksi_prompt[hostname_prefix]="\h: ";
+            else
+                if [[ -n "$(builtin command -v who)" && "$(builtin command who -m 2> /dev/null)" =~ "\([a-fA-F.:0-9]+\)$" ]]; then
+                    # the shell integration script is installed manually on the remote system
+                    # the environment variables are cleared after sudo
+                    # OpenSSH's sshd creates entries in utmp for every login so use those
+                    _ksi_prompt[hostname_prefix]="\h: ";
+                fi
+            fi
         fi
         # see https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html#Controlling-the-Prompt
         # we use suffix here because some distros add title setting to their bashrc files by default
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -19,7 +19,7 @@ end
 
 status is-interactive || exit 0
 not functions -q __ksi_schedule || exit 0
-# Check fish version 3.3.0+ efficiently and fallback to check the last working version 3.2.0, exit on outdated versions.
+# Check fish version 3.3.0+ efficiently and fallback to check the minimum working version 3.2.0, exit on outdated versions.
 # "Warning: Update fish to version 3.3.0+ to enable kitty shell integration.\n"
 set -q fish_killring || set -q status_generation || string match -qnv "3.1.*" "$version"
 or echo -en "\eP@kitty-print|V2FybmluZzogVXBkYXRlIGZpc2ggdG8gdmVyc2lvbiAzLjMuMCsgdG8gZW5hYmxlIGtpdHR5IHNoZWxsIGludGVncmF0aW9uLgo=\e\\" && exit 0 || exit 0
@@ -79,21 +79,6 @@ function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after
             echo -en "\e]133;A\a"
         end
         __ksi_mark_prompt_start
-
-        functions -c fish_prompt __ksi_original_fish_prompt
-        function fish_prompt
-            # fish trims one trailing newline from the output of fish_prompt, so
-            # we need to do the same. See https://github.com/kovidgoyal/kitty/issues/4032
-            # op is a list because fish splits on newlines in command substitution
-            set --local op (__ksi_original_fish_prompt)
-            # print all but last element of the list, each followed by a new line
-            set -q op[2]
-            and printf '%s\n' $op[1..-2]
-            # print the last component without a newline
-            printf '%s' $op[-1]
-            set --global __ksi_prompt_state prompt-end
-            echo -en "\e]133;B\a"
-        end
 
         function __ksi_mark_output_start --on-event fish_preexec
             set --global __ksi_prompt_state pre-exec
--- a/shell-integration/ssh/bootstrap.py
+++ b/shell-integration/ssh/bootstrap.py
@@ -103,21 +103,20 @@ def move(src, base_dest):
 
 
 def compile_terminfo(base):
-    if not shutil.which('tic'):
+    tic = shutil.which('tic')
+    if not tic:
         return
     tname = '.terminfo'
     if os.path.exists('/usr/share/misc/terminfo.cdb'):
         tname += '.cdb'
     os.environ['TERMINFO'] = os.path.join(HOME, tname)
-    tic = shutil.which('tic')
-    if tic:
-        cp = subprocess.run(
-            [tic, '-x', '-o', os.path.join(base, tname), os.path.join(base, '.terminfo', 'kitty.terminfo')],
-            stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-        )
-        if cp.returncode != 0:
-            sys.stderr.buffer.write(cp.stdout)
-            raise SystemExit('Failed to compile the terminfo database')
+    cp = subprocess.run(
+        [tic, '-x', '-o', os.path.join(base, tname), os.path.join(base, '.terminfo', 'kitty.terminfo')],
+        stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+    )
+    if cp.returncode != 0:
+        sys.stderr.buffer.write(cp.stdout)
+        raise SystemExit('Failed to compile the terminfo database')
 
 
 def get_data():
@@ -176,8 +175,8 @@ def exec_zsh_with_integration():
         os.environ.pop('KITTY_ORIG_ZDOTDIR', None)  # ensure this is not propagated
     else:
         os.environ['KITTY_ORIG_ZDOTDIR'] = zdotdir
-    # dont prevent zsh-new-user from running
-    for q in '.zshrc .zshenv .zprofile .zlogin'.split():
+    # dont prevent zsh-newuser-install from running
+    for q in ('.zshrc', '.zshenv', '.zprofile', '.zlogin'):
         if os.path.exists(os.path.join(HOME, q)):
             os.environ['ZDOTDIR'] = shell_integration_dir + '/zsh'
             os.execlp(login_shell, os.path.basename(login_shell), '-l')
--- a/shell-integration/ssh/bootstrap.py
+++ b/shell-integration/ssh/bootstrap.py
@@ -159,15 +159,6 @@ def get_data():
                 move(tdir + '/root', '/')
 
 
-def exec_bash_with_integration():
-    os.environ['ENV'] = os.path.join(shell_integration_dir, 'bash', 'kitty.bash')
-    os.environ['KITTY_BASH_INJECT'] = '1'
-    if not os.environ.get('HISTFILE'):
-        os.environ['HISTFILE'] = os.path.join(HOME, '.bash_history')
-        os.environ['KITTY_BASH_UNEXPORT_HISTFILE'] = '1'
-    os.execlp(login_shell, os.path.basename('login_shell'), '--posix')
-
-
 def exec_zsh_with_integration():
     zdotdir = os.environ.get('ZDOTDIR') or ''
     if not zdotdir:
@@ -192,14 +183,23 @@ def exec_fish_with_integration():
     os.execlp(login_shell, os.path.basename(login_shell), '-l')
 
 
+def exec_bash_with_integration():
+    os.environ['ENV'] = os.path.join(shell_integration_dir, 'bash', 'kitty.bash')
+    os.environ['KITTY_BASH_INJECT'] = '1'
+    if not os.environ.get('HISTFILE'):
+        os.environ['HISTFILE'] = os.path.join(HOME, '.bash_history')
+        os.environ['KITTY_BASH_UNEXPORT_HISTFILE'] = '1'
+    os.execlp(login_shell, os.path.basename('login_shell'), '--posix')
+
+
 def exec_with_shell_integration():
     shell_name = os.path.basename(login_shell).lower()
-    if shell_name == 'bash':
-        exec_bash_with_integration()
     if shell_name == 'zsh':
         exec_zsh_with_integration()
     if shell_name == 'fish':
         exec_fish_with_integration()
+    if shell_name == 'bash':
+        exec_bash_with_integration()
 
 
 def main():
--- a/shell-integration/ssh/bootstrap.sh
+++ b/shell-integration/ssh/bootstrap.sh
@@ -95,7 +95,7 @@ if [ -z "$USER" ]; then USER=$(command whoami 2> /dev/null); fi
 # ask for the SSH data
 leading_data=""
 
-init_tty && trap 'cleanup_on_bootstrap_exit' EXIT
+init_tty && trap "cleanup_on_bootstrap_exit" EXIT
 [ "$tty_ok" = "y" ] && dcs_to_kitty "ssh" "id="REQUEST_ID":hostname="$hostname":pwfile="PASSWORD_FILENAME":user="$USER":pw="DATA_PASSWORD""
 record_separator=$(printf "\036")
 
@@ -287,7 +287,7 @@ exec_zsh_with_integration() {
     else
         export KITTY_ORIG_ZDOTDIR="$zdotdir"
     fi
-    # dont prevent zsh-new-user from running
+    # dont prevent zsh-newuser-install from running
     if [ -f "$zdotdir/.zshrc" -o -f "$zdotdir/.zshenv" -o -f "$zdotdir/.zprofile" -o -f "$zdotdir/.zlogin" ]; then
         export ZDOTDIR="$shell_integration_dir/zsh"
         exec "$login_shell" "-l"
--- a/shell-integration/ssh/bootstrap.sh
+++ b/shell-integration/ssh/bootstrap.sh
@@ -81,7 +81,7 @@ if [ -z "$HOSTNAME" ]; then
     if [ -z "$hostname" ]; then
         hostname=$(command hostnamectl hostname 2> /dev/null)
         if [ -z "$hostname" ]; then
-            hostname="_";
+            hostname="_"
         fi
     fi
 else
@@ -100,12 +100,12 @@ init_tty && trap 'cleanup_on_bootstrap_exit' EXIT
 record_separator=$(printf "\036")
 
 mv_files_and_dirs() {
-    cwd="$PWD";
-    cd "$1";
+    cwd="$PWD"
+    cd "$1"
     command find . -type d -exec mkdir -p "$2/{}" ";"
     command find . -type l -exec sh -c "tgt=\$(command readlink -n \"{}\"); command ln -sf \"\$tgt\" \"$2/{}\"; command rm -f \"{}\"" ";"
     command find . -type f -exec mv "{}" "$2/{}" ";"
-    cd "$cwd";
+    cd "$cwd"
 }
 
 compile_terminfo() {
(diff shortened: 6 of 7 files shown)


kovidgoyal:
According to the man page there is no --no-same-owner in openbsd so I doubt it will. https://man.openbsd.org/tar

page-down:
I saw the latest commit which should fix the tar issue.

I installed an openbsd vm and the first time I connected via ssh kitten it disconnected immediately after enter password.
Try to connect directly using ssh again:
```text
tset: unknown terminal type xterm-kitty
Terminal type? 
```

Try env TERM
```shell
kitty +kitten ssh --kitten="env TERM=xterm-256color" root@openbsd
# ... password: ...
# Connection to openbsd closed.
```

I wonder if other systems may also have this issue.

---
Regarding the hostname configuration, is there a special consideration that does not support matching the hostname set by the user? (Host configured in .ssh/config)
Could you share some thoughts on this?

A service component in a different project may use the same hostname of a common word. The hostname can be changed on the remote.  etc. ...
It would be nice to additionally provide a way to specify the `local name`,
so that the user can explicitly use the specified configuration to send the files to the host when connecting.

kovidgoyal:
On Sun, Mar 06, 2022 at 11:48:08PM -0800, page-down wrote:
> I saw the latest commit which should fix the tar issue.
> 
> I installed an openbsd vm and the first time I connected via ssh kitten it disconnected immediately after enter password.

OpenBSD is an extremely crippled system so I am not surprised. Use the
debug function to see where the script is failing.

> Try to connect directly using ssh again:
> ```text
> tset: unknown terminal type xterm-kitty
> Terminal type? 
> ```
> 
> Try env TERM
> ```shell
> kitty +kitten ssh --kitten="env TERM=xterm-256color" ***@***.***
> # ... password: ...
> # Connection to openbsd closed.
> ```
> 
> I wonder if other systems may also have this issue.
> 
> ---
> Regarding the hostname configuration, is there a special consideration that does not support matching the hostname set by the user? (Host configured in .ssh/config)
> Could you share some thoughts on this?

One can connect to hosts using IP addresses and aliases in /etc/hosts or
in a DNS server or ~/.ssh/config. The remote hostname is one simple
point of configuration. 

That said there is no harm in matching against the name used on the ssh
commandline to connect as well.

page-down:
> ... Use the debug function to see where the script is failing. ...

```text
sh: base64: not found
```

I was wondering if we could check the commands that will be used first, then output the warning log earlier and abort.
This doesn't even compare to the tools that come with the busybox which is a few MB in size.
I don't care about this system for now.

> ... The remote hostname is one simple point of configuration. ...

Ok, I understand that this is the acceptable way to configure it.

page-down:
patches for openbsd:

```text
elif command -v uuencode > /dev/null 2> /dev/null; then
    base64_encode() { command uuencode -m -o /dev/stdout /dev/stdin | command sed '1d;$d' | command tr -d \\n\\r; }
    base64_decode() { printf 'begin-base64 644 -\n%s' $(</dev/stdin) | command uudecode -m -o /dev/stdout 2> /dev/null; }

if [ -n "$ZSH_VERSION" ] && builtin zmodload 

# dcs_to_kitty() { printf "\033P@kitty-$1|%s\033\\\\"
dcs_to_kitty() { printf "\033P@kitty-$1|%s\033\134"
```

Should the following `head -c` fall back to dd?
Is it possible to ask for gz?

```text
head: unknown option -- c
tar: could not exec bzip2: No such file or directory
```

page-down:
I went on to try FreeBSD.

```text
# kitty +kitten ssh root@freebsd
/bin/sh: Event not found.
Missing name for redirect.
saved_tty_settings=: Command not found.
Badly placed ()'s.
...
```

Try modifying bootstrap.sh to see what is happening.
```text
<Here is an empty line.>
ps -f
exit 0
```

```text
Unmatched '''.
 PID TT  STAT    TIME COMMAND
# ...
1493  0  Ss+  0:00.00 csh -c sh -c '\nps -f\nexit 0'
1495  0  R+   0:00.00 ps -f
Unmatched '''.
```

chsh -s /bin/sh
```text
# bootstrap.sh: failed at the last line:
# exec $login_shell "-l"

Illegal option -l
```

Is there any other option?

kovidgoyal:
On Tue, Mar 08, 2022 at 02:34:00AM -0800, page-down wrote:
> I went on to try FreeBSD.

You are very adventurous.

> 
> ```text
> # kitty +kitten ssh ***@***.***
> /bin/sh: Event not found.
> Missing name for redirect.
> saved_tty_settings=: Command not found.
> Badly placed ()'s.
> ...
> ```
> 
> Try modifying bootstrap.sh to see what is happening.
> ```text
> <Here is an empty line.>
> ps -f
> exit 0
> ```
> 
> ```text
> Unmatched '''.
>  PID TT  STAT    TIME COMMAND
> # ...
> 1493  0  Ss+  0:00.00 csh -c sh -c '\nps -f\nexit 0'
> 1495  0  R+   0:00.00 ps -f
> Unmatched '''.
> ```
> 
> chsh -s /bin/sh
> ```text
> # bootstrap.sh: failed at the last line:
> # exec $login_shell "-l"
> 
> Illegal option -l
> ```
> 
> Is there any other option?

Which shell doesnt support -l?? According to the man page on FreeBSD
tcsh does: https://www.freebsd.org/cgi/man.cgi?csh(1)

page-down:
> Which shell doesnt support -l?

`/bin/sh`
https://www.freebsd.org/cgi/man.cgi?sh

The root user uses `/bin/csh` (tcsh) by default, which is supported to run with `-l`, but I haven't been able to run it yet.
It seems to be a quoting problem, and the code is being run with (t)csh.

kovidgoyal:
On Tue, Mar 08, 2022 at 04:17:53AM -0800, page-down wrote:
> > Which shell doesnt support -l?
> 
> `/bin/sh`
> https://www.freebsd.org/cgi/man.cgi?sh

Ah, well no I don't see an alternative. If it doesn't support exec -a and
doesn't support sh -l I don't see what we can do other than tell users to
move to a better system or install python and set the interpreter to
that.

> 
> The root user uses `/bin/csh` (tcsh) by default, which is supported to run with `-l`, but I haven't been able to run it yet.
> It seems to be a quoting problem, and the code is being run with (t)csh.

csh is not POSIX compatible. You wont be able to get bootstrap.sh to
work with it.

kovidgoyal:
Weird are you saying that on freebsd running sh -c "whatever" actually runs csh -c sh -c "whatever"?? If os that's a bug in FreeBSD.

page-down:
Linux

```text
# ssh linux sh -c '"t(){ echo ok; }; t; sleep 10; exit 0"'
ok

# ps
\_ sshd:
   \_ sh -c t(){ echo ok; }; t; sleep 10; exit 0
       \_ sleep 10
```

FreeBSD with login shell /bin/csh

```text
# user login shell: tcsh
# ssh freebsd sh -c '"t(){ echo ok; }; t; sleep 10; exit 0"'
ok

# ps
tcsh -c sh -c "t(){ echo ok; }; t; sleep 10; exit 0"
sh -c t(){ echo ok; }; t; sleep 10; exit 0

# ssh freebsd sh -c "t(){ echo ok; }; t; sleep 10; exit 0"
Badly placed ()'s.
```

FreeBSD with login shell /bin/sh
```text
# user login shell: sh
# ssh freebsd sh -c '"t(){ echo ok; }; t; sleep 10; exit 0"'
ok

# ps
sh -c t(){ echo ok; }; t; sleep 10; exit 0
```

page-down:
For the above reason, setting it to python does not work either.

```text
/usr/bin/env: Event not found.
Missing name for redirect.
import: Command not found.
...
```

I have fish installed on Linux and I would like to confirm that:
The code is actually executed by login_shell, not the explicitly specified sh.

```text
# Linux, login shell: fish
# ssh sh -c '"t(){ echo ok; }; t; sleep 10; exit 0"'
\_ sshd:
    \_ fish -c sh -c "t(){ echo ok; }; t; sleep 10; exit 0"
        \_ sh -c t(){ echo ok; }; t; sleep 10; exit 0
            \_ sleep 10
```

page-down:
To summarize:
The current quoting style, only compatible with the login shell set in the remote system is POSIX compatible.
For example, on Linux the user's login shell is fish, which will not work properly. (`kitty +kitten ssh --kitten="interpreter=python3" user@host`)

I can think of one solution.
The `sh -c` code in the argument sent to ssh without any special quotes,
finds the available base64 decoder in its simplest form, and then decodes the bootstrap script to run.

```shell
ssh host sh -c "echo <BOOTSTRAP_SH_BASE64> | base64 -d | sh"
ssh host python3 -c "eval(compile(standard_b64decode(..."
```

Since the actual code is run by interpreter, it should work on all login shells as long as they don't have any special quotes.

kovidgoyal:
But that will break if base64 is not available.

page-down:
> But that will break if base64 is not available.

Replace the quotes with AAA, BBB, (not sure if Unicode will work) and then replace them back using simple tools like `tr`, do you think it will work?

kovidgoyal:
yes that will likely work, note that awk is part of POSIX so we are not limited to tr

kovidgoyal:
As is sed: https://pubs.opengroup.org/onlinepubs/009695299/utilities/sed.html

https://pubs.opengroup.org/onlinepubs/009604499/utilities/awk.html

page-down:
I don't see string expansion in the bootstrap code, and it looks like it could also be used where needed.

https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_06_02

kovidgoyal:
On Tue, Mar 08, 2022 at 06:58:10AM -0800, page-down wrote:
> I don't see string expansion in the bootstrap code, and it looks like it could also be used where needed.

But that's for POSIX shells, inst the whole problem here fish or other
non posix shells?

page-down:
> But that's for POSIX shells ...

In `ssh host sh -c "HERE"`, POSIX sh is used to run the parsing code.
As long as any login shell configured by the user can parse `sh -c XXX` correctly (without quotes).
Dont: execute by login shell due to quoting errors.

kovidgoyal:
Not sure I follow, feel free to send a PR

page-down:
I have some time and would like to continue working on this.
However I found a problem at startup. (I didn't open a new one but continued the discussion here.)

```text
# My XDG_RUNTIME_DIR -> $TMPDIR (0700, owned by USER)
ControlPath too long ('/var/folders/0k/f7_s3sjj66j7_2c21y5jvms00000gn/T/ssh-kitten-6870-master-9f7d9600667c4d9692bfe7c4262cdd8c9d4756be' >= 104 bytes)
```

Temporarily work around it by configuring the kitty RUNTIME folder environment variable.

```text
macOS:
$TMPDIR: The latest version of macOS will randomly clear the directory and lose files.
/tmp: It will clean up automatically after reboot.
```

Is it not appropriate to use /tmp by default (macOS)? (need to create `/tmp/kitty-USER_ID` folder first)
This will ensure that we don't run into the situation where the $USER username is too long and `/Users/LONG_NAME/Library/Caches/kitty/run` is also too long, making it unusable.

---

```c
#ifdef __APPLE__
// This value is based on experimentation
PyModule_AddIntConstant(m, "SHM_NAME_MAX", 30);
```

Is it the following value? (31 chars+NUL)
https://github.com/apple/darwin-xnu/blob/xnu-4903.270.47/bsd/sys/posix_shm.h
```text
#define PSHMNAMLEN      31      /* maximum name segment length we bother with */
```

---

I started trying bootstrap.py, however, I occasionally encountered the following problem.

```text
# bootstrap.py -> get_data -> move
IsADirectoryError: [Errno 21] Is a directory: '_$HOME_/.kitty-ssh-kitten-untar-RANDOM-STRING/home/.terminfo/78' -> '_$HOME_/.terminfo/78'
```

kovidgoyal:
On Thu, Mar 10, 2022 at 05:02:31AM -0800, page-down wrote:
> I have some time and would like to continue working on this.
> However I found a problem at startup. (I didn't open a new one but continued the discussion here.)
> 
> ```text
> # My XDG_RUNTIME_DIR -> $TMPDIR (0700, owned by USER)
> ControlPath too long ('/var/folders/0k/f7_s3sjj66j7_2c21y5jvms00000gn/T/ssh-kitten-6870-master-9f7d9600667c4d9692bfe7c4262cdd8c9d4756be' >= 104 bytes)
> ```
> 
> Temporarily work around it by configuring the kitty RUNTIME folder environment variable.
> 
> ```text
> macOS:
> $TMPDIR: The latest version of macOS will randomly clear the directory and lose files.
> /tmp: It will clean up automatically after reboot.
> ```
> 
> Is it not appropriate to use /tmp by default (macOS)? (need to create `/tmp/kitty-USER_ID` folder first)
> This will ensure that we don't run into the situation where the $USER username is too long and `/Users/LONG_NAME/Library/Caches/kitty/run` is also too long, making it unusable.

/tmp has its files deleted periodically, not just at reboot, so no, it's not appropriate.
macOS will happily delete our sockets out from under us.

> 
> ---
> 
> ```c
> #ifdef __APPLE__
> // This value is based on experimentation
> PyModule_AddIntConstant(m, "SHM_NAME_MAX", 30);
> ```
> 
> Is it the following value? (31 chars+NUL)
> https://github.com/apple/darwin-xnu/blob/xnu-4903.270.47/bsd/sys/posix_shm.h
> ```text
> #define PSHMNAMLEN      31      /* maximum name segment length we bother with */
> ```

yes but that includes the trailing null, so effectively its 30

> 
> ---
> 
> I started trying bootstrap.py, however, I occasionally encountered the following problem.
> 
> ```text
> # bootstrap.py -> get_data -> move
> IsADirectoryError: [Errno 21] Is a directory: '_$HOME_/.kitty-ssh-kitten-untar-RANDOM-STRING/home/.terminfo/78' -> '_$HOME_/.terminfo/78'
> ```

That looks like its trying to rename a symlink when the dest exists and
is a directory. I wonder why this doesn't fail with bootstrap.sh as well.
I dont think ln -sf deletes directories.
> 
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/4794#issuecomment-1064032194
> 
> Message ID: ***@***.***>

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

page-down:
> macOS will happily delete our sockets out from under us.

That's terrible, it only seems appropriate to use `~/Library/Caches`.

> yes but that includes the trailing null, so effectively its 30

I saw the following and thought it was actually 32, minus NUL which is 31.
But this one character doesn't matter.

```h
struct pshminfo {
    char         pshm_name[PSHMNAMLEN + 1];
```

> I wonder why this doesn't fail with bootstrap.sh as well.

Strange. I did not encounter this when using bootstrap.sh.

Now that kitty no longer writes temporary files locally when connecting, I wonder if it is possible to configure the temporary file directory on the remote host.
For example by being able to use tmpfs instead of HOME. (Yes, I know this is generally small, but it's enough.)
Maybe compare files and then move them.
Or compare file.ext.sha256, if the user knows what he is doing.

If the same file is written repeatedly for each connection,
it is not known whether some error will be caused by various reasons.
Yes, the current practice is the most stable way to ensure that the files are up to date.

page-down:
> That looks like its trying to rename a symlink when the dest exists and is a directory.

I think I know, probably not tested this host.
Some do have the 78 folder in .terminfo. And there are files in it.
That folder should not be deleted or replaced with symlinks.

kovidgoyal:
On Thu, Mar 10, 2022 at 06:27:39AM -0800, page-down wrote:
> > macOS will happily delete our sockets out from under us.
> 
> That's terrible, it only seems appropriate to use `~/Library/Caches`.

Yeah, when I first discovered this behavior many years ago while
debugging issues with calibre it infuriated me. How does Apple think its
ok to delete files out from under running programs. I wil change it to
always use the cache dir on macOS, and make the constant parts of the
name a bit shorter.

> 
> > yes but that includes the trailing null, so effectively its 30
> 
> I saw the following and thought it was actually 32, minus NUL which is 31.
> But this one character doesn't matter.
> 
> ```h
> struct pshminfo {
>     char         pshm_name[PSHMNAMLEN + 1];
> ```
> 
> > I wonder why this doesn't fail with bootstrap.sh as well.
> 
> Strange. I did not encounter this when using bootstrap.sh.

This needs investigation and a proper fix for both the bootstrap
scripts.

> 
> Now that kitty no longer writes temporary files locally when connecting, I wonder if it is possible to configure the temporary file directory on the remote host.

It's possible, but I dont think its a good idea. I chose home explicitly
because most files to be copied are in $HOME and so mv or os.rename()
are atomic and basically free (two directory inode updates).
Putting it on tmpfs means it will no longer be atomic or free. 
Which reminds me I need to handle the case of moving to a different
filesystem in bootstrap.py


> For example by being able to use tmpfs instead of HOME. (Yes, I know this is generally small, but it's enough.)
> Maybe compare files and then move them.

THis requires more rounndtrips which is an absolute killer for latency.

page-down:
> ... so mv or os.rename() are atomic and basically free (two directory inode updates) ...

Yes, the last thing I want is to move across different file systems.

> This requires more rounndtrips which is an absolute killer for latency.

Increased latency is unacceptable.
Here is one way I can think of.

- Create a summary hash based on all the file size and time.
- Add the `last-update-hash` file when sending the tarball.
- The remote bootstrap script sends the contents of the last-update-hash file (a hash string) when requesting data, and ssh kitten decides if it needs to send it all again.

It may not be accurate to identify changes by file modification date and size,
but I will be happy with this opt-in configuration because I know what I'm doing.

kovidgoyal:
On Thu, Mar 10, 2022 at 07:17:03AM -0800, page-down wrote:
> > ... so mv or os.rename() are atomic and basically free (two directory inode updates) ...
> 
> Yes, the last thing I want is to move across different file systems.
> 
> > This requires more rounndtrips which is an absolute killer for latency.
> 
> Increased latency is unacceptable.
> Here is one way I can think of.
> 
> - Create a summary hash based on all the file size and time.
> - Add the `last-update-hash` file when sending the tarball.
> - The remote bootstrap script sends the contents of the last-update-hash file (a hash string) when requesting data, and ssh kitten decides if it needs to send it all again.
> 
> It may not be accurate to identify changes by file modification date and size,
> but I will be happy with this opt-in configuration because I know what I'm doing.

I am currently working on a way to do the copy without any roundtrips at
all. If that works out then it wont be possible to do even this. So lets
table this for now and get back to it later.

kovidgoyal:
why is the hash so long on your system, on mine its 40 chars why is it
67 on yours?

page-down:
```text
# which ssh
/usr/bin/ssh
# ssh -V
OpenSSH_8.6p1, LibreSSL 2.8.3
# ssh -G -o ControlPath=%C test | grep -i path
controlpath c586a5b9927432f0397bed623ab88f818128f848
```

kovidgoyal:
So what command line is generating a hash of

a5bc21fdef2a7e26de907f92602636d6e3c57fff.baqiDKDaIHdFO4Ci

It seems extremely odd that the has format is not constant.

page-down:
I assume you are looking at the latest code, I upgraded to OpenSSH 8.9p1 and the problem still exists.
If that is the case, that `%C` is unusable.

page-down:
I tried the latest commit and now the ssh kitten works fine.
That's a brilliant workaround for the idiotic design.
I looked at the script for clearing tmp, the symlink we created (and the sockets) wouldn't be matched at all.

/etc/periodic/daily/110.clean-tmps
```shell
# ...
find -dx . -fstype local -type f $args -delete $print
find -dx . -fstype local ! -name . -type d $dargs -delete $print
```

page-down:
I updated to the latest commit.

There is a problem with using the `--kitten` option to override the configuration.
kittens/ssh/config.py
```python
def options_for_host(hostname: str, username: str, per_host_opts: Dict[str, SSHOptions], cli_hostname: str = '', cli_uname: str = '') -> SSHOptions:
    # ...
        for name in option_names:
            for opts in rest:
                # ...
                    setattr(ans, name, val)
```

~/.config/kitty/ssh.conf
```conf
shell_integration disabled

hostname debug
shell_integration enabled
```

matches: `'*', 'debug', 'user@debug'`

The login shell and shell integration configured by the following command does not take effect.
`kitty +kitten ssh --kitten='hostname debug' --kitten='login_shell zsh' debug`
```text
hostname -> * -> debug -> user@debug
login_shell -> "" -> "" -> ""
shell_integration -> disabled -> enabled -> inherited
```

The login shell configured by the following command takes effect, but the shell integration is set to the default value from SSHOptions.
`kitty +kitten ssh --kitten='login_shell zsh' debug`
```text
hostname -> * -> debug -> user@debug
login_shell -> "" -> "" -> zsh
shell_integration -> disabled -> enabled -> inherited
```

---

Now the following warning appears, is it possible to suppress it? E.g. `> /dev/null`
I found this a bit annoying, not too concerned about the time difference within one second.
```text
tar: data.sh: time stamp xxx is 0.xxx s in the future
```

---

Is it worth providing a subcommand to disconnect all shared ssh connections? I found it might be useful when I was trying `password login`.

---

Is it appropriate to use the POSIX SHELL environment variable to get the login shell?
(Except that it did not take effect immediately after `chsh` with shared ssh.)
Is it efficient in most cases?

page-down:
When connecting to a non-existent host, the data is written to the local shell after ssh exits.

```text
# non-existent host
kitty +kitten ssh test-not-exists

# ssh not yet started or port not listened to
kitty +kitten ssh user@debug:2222

ssh: Could not resolve hostname test-not-exists: nodename nor servname provided, or not known
# executed as commands
[127]> QZhtQZhtQZhtQZhtQZhtQZhtQZhtQZhtQZht...
```

I guess this also happens when ssh quits in the middle of transmission for any reason. Not yet tested for interruptions due to network connection failure.

kovidgoyal:
On Mon, Mar 14, 2022 at 08:10:13PM -0700, page-down wrote:
> When connecting to a non-existent host, the data is written to the local shell after ssh exits.

Doesnt reproduce for me, check drain_potential_tty_garbage() and see why
it is not working for you.

page-down:
> ... check drain_potential_tty_garbage() ...

After running kitty the first connection works fine and eventually reads `KITTY_DATA_END`.
Connecting again is problematic, and the following condition is not valid.

`while ... and select([tf], [], [], 0.075)[0]:`

kovidgoyal:
On Mon, Mar 14, 2022 at 08:45:53PM -0700, page-down wrote:
> > ... check drain_potential_tty_garbage() ...
> 
> After running kitty the first connection works fine and eventually reads `KITTY_DATA_END`.
> Connecting again is problematic, and the following condition is not valid.
> 
> `while ... and select([tf], [], [], 0.075)[0]:`

I dont follow, are you saying

kitty +kitten ssh badhost

does not spew garbage the first time it is run, but does the second?

page-down:
> does not spew garbage the first time it is run, but does the second

Yes, or run `kitty +kitten ssh localhost` (never connected), even if it's the first time you run it after kitty starts.

macOS, OpenSSH 8.9p1
```text
# I think it's at alt screen.
Are you sure you want to continue connecting (yes/no/[fingerprint])? 
> no
Host key verification failed.
XXXXX base64 ...
```

Linux, OpenSSH 8.2p1

```text
The authenticity of host ...
ECDSA key fingerprint is ...
Are you sure you want to ... ?
Please type 'yes', 'no' or the fingerprint: 
Please type 'yes', 'no' or the fingerprint: 
Please type 'yes', 'no' or the fingerprint: 
Please type 'yes', 'no' or the fingerprint:
```

I haven't tested the latest openssh under Linux yet.

kovidgoyal:
On Mon, Mar 14, 2022 at 07:52:34PM -0700, page-down wrote:
> I updated to the latest commit.
> 
> There is a problem with using the `--kitten` option to override the configuration.
> kittens/ssh/config.py
> ```python
> def options_for_host(hostname: str, username: str, per_host_opts: Dict[str, SSHOptions], cli_hostname: str = '', cli_uname: str = '') -> SSHOptions:
>     # ...
>         for name in option_names:
>             for opts in rest:
>                 # ...
>                     setattr(ans, name, val)
> ```
> 
> ~/.config/kitty/ssh.conf
> ```conf
> shell_integration disabled
> 
> hostname debug
> shell_integration enabled
> ```
> 
> matches: `'*', 'debug', ***@***.***'`
> 
> The login shell and shell integration configured by the following command does not take effect.
> `kitty +kitten ssh --kitten='hostname debug' --kitten='login_shell zsh' debug`
> ```text
> hostname -> * -> debug -> ***@***.***
> login_shell -> "" -> "" -> ""
> shell_integration -> disabled -> enabled -> inherited
> ```
> 
> The login shell configured by the following command takes effect, but the shell integration is set to the default value from SSHOptions.
> `kitty +kitten ssh --kitten='login_shell zsh' debug`
> ```text
> hostname -> * -> debug -> ***@***.***
> login_shell -> "" -> "" -> zsh
> shell_integration -> disabled -> enabled -> inherited
> ```
> 
> ---
> 
> Now the following warning appears, is it possible to suppress it? E.g. `> /dev/null`
> I found this a bit annoying, not too concerned about the time difference within one second.
> ```text
> tar: data.sh: time stamp xxx is 0.xxx s in the future
> ```
> 
> ---

These are both fixed.

> 
> Is it worth providing a subcommand to disconnect all shared ssh connections? I found it might be useful when I was trying `password login`.

I added a mappable action for it.

> 
> ---
> 
> Is it appropriate to use the POSIX SHELL environment variable to get the login shell?
> (Except that it did not take effect immediately after `chsh` with shared ssh.)
> Is it efficient in most cases?

It doesn't reflect chsh, so not really.

kovidgoyal:
On Mon, Mar 14, 2022 at 10:12:24PM -0700, page-down wrote:
> > does not spew garbage the first time it is run, but does the second
> 
> Yes, or run `kitty +kitten ssh localhost` (never connected), even if it's the first time you run it after kitty starts.

macOS is weird. Should be fine now.

page-down:
Thanks, I updated to the latest commit and it's working fine.
Regarding the last issue I mentioned above,
here's what happened on the first and second run on Linux after I cleared `known_hosts`.

```text
$ kitty +kitten ssh debug
The authenticity of host ... can't be established.
ECDSA key fingerprint is ...
Are you sure you want to continue connecting (yes/no/[fingerprint])? 
Host key verification failed.

$ kitty +kitten ssh debug
The authenticity of host ... can't be established.
ECDSA key fingerprint is ...
Are you sure you want to continue connecting (yes/no/[fingerprint])? 
Please type 'yes', 'no' or the fingerprint: 
Please type 'yes', 'no' or the fingerprint: 
Please type 'yes', 'no' or the fingerprint: 
Please type 'yes', 'no' or the fingerprint: 
```

Is this a compatibility issue?

kovidgoyal:
That's an impossible issue :) 

1) ssh is asking for confirmation in the terminal, which means you have
disabled askpass in ssh.conf?

2) The kitten is sending data over the terminal without waiting for
ssh to finish connecting, which can only happen if the kittne thinks ssh
will use askpass, which means you havent changed it in ssh.conf

So why is ssh not using kitty's askpass on your Linux system?

kovidgoyal:
Oh and btw to tets this you dont need to clear th eknown hosts file, you can just do

kitty +kitten ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=ask whatever

page-down:
Thanks for the tip.
OpenSSH 8.2p1 from Ubuntu LTS 20.04 does not support `SSH_ASKPASS_REQUIRE`, so it does not work.
If I remove `os.environ['SSH_ASKPASS_REQUIRE'] = 'force'` then I can reproduce the above issue under Arch Linux (with the latest OpenSSH).
Maybe a fallback is needed.

kovidgoyal:
Sigh, endlessly annoying Linux distros with their ancient software. 

It was introduced in openssh 8.4 so one just needs to check the version
with ssh -V.

page-down:
I tried the latest commit and found that the following doesn't work.

```python
if os.environ.get('DISPLAY') or ssh_version() >= (8, 4):
```

Since the environment variable `DISPLAY` is always present, it is equivalent to not checking the version.
If there must be a fallback by default, perhaps the only way to skip this check is to configure `askpass` explicitly in ssh.conf.

>  ... do the copy without any roundtrips ... it wont be possible to do even this.
> So lets table this for now and get back to it later. ...

It seems that this has landed.
Since we are using a shared connection, I wonder if after the first connection is established, can we assume that the terminfo, shell_integration files, etc. have been updated to the current version and subsequent connections don't need to send these again, and there is no need to tar `data.sh` and extract.

These are files that are mandatory before the login shell runs, and after this is resolved, other files might be able to be synchronized asynchronously after the connection (e.g. explicitly configured to use librsync).

kovidgoyal:
On Tue, Mar 15, 2022 at 04:18:09AM -0700, page-down wrote:
> >  ... do the copy without any roundtrips ... it wont be possible to do even this.
> > So lets table this for now and get back to it later. ...
> 
> It seems that this has landed.
> Since we are using a shared connection, I wonder if after the first connection is established, can we assume that the terminfo, shell_integration files, etc. have been updated to the current version and subsequent connections don't need to send these again, and there is no need to tar `data.sh` and extract.

I dont see much point, it's a very marginal saving for a fair bit of
complication. But you are welcome to try it and see how much exactly the
saving is in terms of latency to prompt. It's a total of ~100K of data
(before compression, probably 40K with compression) to transmit, which
is nothing on modern networks.  Just comment out the bits of code that
add those files to the tarfile and the copy files code in bootstrap.sh
and check.

> 
> These are files that are mandatory before the login shell runs, and after this is resolved, other files might be able to be synchronized asynchronously after the connection (e.g. explicitly configured to use librsync).

Cant do that, other files might be needed to setup the environment for
the shell, for instance.

page-down:
> ... it's a very marginal saving for a fair bit of complication ...
> Cant do that, other files might be needed to ...
> I just checked, it compresses to 15K.

OK, It is not possible to cover everything, and users with special needs will need to write their own tools. This obviously cannot be done with minimal cost to know the status of the target host.

Also this reminds me of a scenario where a user (e.g., root) on a device can log in and run programs or write files, but there is no $HOME.
I think in this environment the user needs to specify the extract base dir and the TERMINFO path themselves.

page-down:
@kovidgoyal 
I tested one more SSH server: Tiny SSH Server.

When connecting with OpenSSH client version v8.9p1, it outputs `KITTY_DATA_START ... KITTY_DATA_END`.
Is there a way to work around this?
For example by specifying `need_to_request_data` in the host configuration. (`stty "-echo"` and `stty "echo"`)

kovidgoyal:
Sounds like a bug in that server, its not cloning the echo state of the
local terminal on the remote side. You can force it by turning off
connection sharing and kitty askpass for that host. There isn't a
dedicated setting for it. Is this server widely used?

page-down:
> ... its not cloning the echo state ...

OK, then this better be fixed in that project.

> ... turning off connection sharing and kitty askpass ...

These two options are sufficient, and the latency is acceptable.

> Is this server widely used?

No. There are actually not many open source SSH servers, and this is the last one I want to try.

PR #4800, opened 2022-03-08
ssh kitten: Fix support for OpenBSD and add Perl in bootstrap.sh

page-down:
Use the more explicit b64encode b64decode.
Add Perl to support more systems.

I tested it on OpenBSD and it works fine now.

Please review, thank you.

I found that if there are errors during execution, the cursor does not move to the beginning of the line after a line break. Is this normal?

```text
ERR: error message line one.
                            ERR: error message line two.
```

### DIFF
--- a/shell-integration/ssh/bootstrap.sh
+++ b/shell-integration/ssh/bootstrap.sh
@@ -28,9 +28,9 @@ detect_python() {
 if command -v base64 > /dev/null 2> /dev/null; then
     base64_encode() { command base64 | command tr -d \\n\\r; }
     base64_decode() { command base64 -d; }
-elif command -v uuencode > /dev/null 2> /dev/null; then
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Mon, Mar 07, 2022 at 08:37:04PM -0800, page-down wrote:
> I found that if there are errors during execution, the cursor does not move to the beginning of the line after a line break. Is this normal?
> 
> ```text
> ERR: error message line one.
>                             ERR: error message line two.

Yes its because the terminal is in raw mode, so \n is not converted to
\r\n. You need to output \r\n instead of just \n explicitly.

page-down:
Yes, I have noticed.
So using `\n\r` here has the same effect as `\r\n`.

```text
die() { printf "\033[31m%s\033[m\n\r" "$*" > /dev/stderr ...
```

PR #4804, opened 2022-03-08
Fix shell integration being disabled with one invalid option

page-down:
In the following case, the invalid option should be ignored and the default shell integration configuration should be used.
(Note that the default option value can be changed with setup.py)

```shell
kitty --config=NONE -o shell_integration=notitle -o shell=zsh
```


page-down:
The PR has been merged and was not recognized due to my force push. Closing now.

PR #4806, opened 2022-03-09
ssh kitten: Add support for FreeBSD tcsh sh default shells

page-down:
- Replacing `\n` and `!` with `\r` and `\241` (Inverted exclamation mark) for tcsh.
- Reimplement the login shell startup loading with POSIX ENV to support sh without the -l option.

FreeBSD 13.0 sh:
https://github.com/freebsd/freebsd-src/blob/release/13.0.0/bin/sh/main.c#L142-L151

Dash 0.5.11:
https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/main.c?h=0.5.11#n149

All the default shells are now working fine under FreeBSD.

Please review, thank you.

---

> ... running FreeBSD, ... relies on termcap rather than terminfo ...
> ... convert the terminfo file ...  on local machine with kitty
>     infocmp -C xterm-kitty
> The output ... which should be appended to `/usr/share/misc/termcap` on the remote server ...

Does this need to be added to bootstrap script (sh and python)?

I see that if it fails in `untar_and_read_env` and exits, the temp folder is not cleared.
Is it possible to use trap and clear the temp folder?

### DIFF
--- a/docs/kittens/ssh.rst
+++ b/docs/kittens/ssh.rst
@@ -114,9 +114,9 @@ the TTY, any other requests are responded to by errors.
 .. note::
 
    When connecting to BSD servers, it is possible the bootstrap script will
-   fail or run slowly, because they are crippled in various ways. Your best bet
-   is to install python on the server, make sure the login shell is something
-   POSIX sh compliant, not csh, and use :code:`python` as the :opt:`interpreter
+   fail or run slowly, because the default shells are crippled in various ways.
+   Your best bet is to install python on the server, make sure the login shell
+   is something POSIX sh compliant, and use :code:`python` as the :opt:`interpreter
    <kitten-ssh.interpreter>` in :file:`ssh.conf`.
 
 .. include:: /generated/conf-kitten-ssh.rst
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -407,10 +407,11 @@ def wrap_bootstrap_script(sh_script: str, interpreter: str) -> List[str]:
         unwrap_script = '''"import base64, sys; eval(compile(base64.standard_b64decode(sys.argv[-1]), 'bootstrap.py', 'exec'))"'''
     else:
         # We cant rely on base64 being available on the remote system, so instead
-        # we quote the bootstrap script by replacing ' and \ with \v and \f and
-        # surrounding with '
-        es = "'" + sh_script.replace("'", '\v').replace('\\', '\f') + "'"
-        unwrap_script = r"""'eval "$(echo "$0" | tr \\\v\\\f \\\047\\\134)"' """
+        # we quote the bootstrap script by replacing ' and \ with \v and \f
+        # also replacing \n and ! with \r and \b for tcsh
+        # finally surrounding with '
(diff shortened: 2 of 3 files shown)


page-down:
Apparently I still need to fix some issues and update the PR.
Can I use other ASCII characters below 31? 
Maybe `\016`, `\017`?

kovidgoyal:
You can use \a or \b experiment and see what works.

kovidgoyal:
On Tue, Mar 08, 2022 at 11:36:31PM -0800, page-down wrote:
> - Replacing `\n` and `!` with `\r` and `\241` (Inverted exclamation mark) for tcsh.

Why not use \b instead of a non-ASCII char. ASCII seems safer :)

> - Reimplement the login shell startup loading with POSIX ENV to support sh without the -l option.
> 
> FreeBSD 13.0 sh:
> https://github.com/freebsd/freebsd-src/blob/release/13.0.0/bin/sh/main.c#L142-L151
> 
> Dash 0.5.11:
> https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/main.c?h=0.5.11#n149
> 
> All the default shells are now working fine under FreeBSD.
> 
> Please review, thank you.
> 
> ---
> 
> &gt; ... running FreeBSD, ... relies on termcap rather than terminfo ...
> &gt; ... convert the terminfo file ...  on local machine with kitty
> &gt;     infocmp -C xterm-kitty
> &gt; The output ... which should be appended to `/usr/share/misc/termcap` on the remote server ...
> 
> Does this need to be added to bootstrap script (sh and python)?

I dount /usr/share/misc/termcap is world writable.

> 
> I see that if it fails in `untar_and_read_env` and exits, the temp folder is not cleared.
> Is it possible to use trap and clear the temp folder?

Fixed.

page-down:
> Why not use \b instead of a non-ASCII char. ASCII seems safer :)

Well, `\b`, that's exactly what I used in my first revision.

> I dount /usr/share/misc/termcap is world writable.

Obviously root permission is required.
So I'll have to ignore it for now, at least I can connect to it.

I've updated the PR and hopefully I haven't missed anything.

page-down:
I saw the latest commit, wouldn't `shell_integration_dir` be empty when `tty_ok=n`?

I also found a problem that the login shell (tcsh, fish, ...) will keep running during the connection.
Is it possible to add exec before `interpreter`? It will work fine in a POSIX-compatible shell.
Or, perhaps it should be Opt-In. or not worth it.

```python
# return [interpreter, '-c', unwrap_script, es]
return ['exec', interpreter, '-c', unwrap_script, es]
```

kovidgoyal:
On Wed, Mar 09, 2022 at 01:45:22AM -0800, page-down wrote:
> I saw the latest commit, wouldn't `shell_integration_dir` be empty when `tty_ok=n`?

If the tty is non-functional, execing a shell is pretty pointless
anyway, but I have added code to give a better error message in such
cases.

> 
> I also found a problem that the login shell (tcsh, fish, ...) will keep running during the connection.
> Is it possible to add exec before `interpreter`? It will work fine in a POSIX-compatible shell.
> Or, perhaps it should be Opt-In. or not worth it.

Added.

PR #4809, opened 2022-03-09
`dd ibs` and `ghead`

k4leg:
`ibs` is ~48.41% faster (on my system) than `bs` for reading first _n_ bytes.  Also see [this](https://unix.stackexchange.com/a/278465).

As hack we could use `dd bs="$n" count=1` instead of `dd ibs=1 count="$n"` if file is small (because `dd bs="$n" count=1` uses more RAM, _~n_ bytes more).

On some OS (like OpenBSD) GNU head is `ghead`, not `head`.

### DIFF
--- a/shell-integration/ssh/bootstrap.sh
+++ b/shell-integration/ssh/bootstrap.sh
@@ -119,6 +119,10 @@ else
         read_n_bytes_from_tty() {
             command head -c "$1" < /dev/tty
         }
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Are you sure that using obs >1 cannot lead to corruption if the write() syscall is interrupted? 

Using ghead is a good idea, but it should be tested for first, before testing head, this is because ghead -c does not overread bytes, unlike head -c from non GNU coreutils, so it should be used preferentially.

PR #4836, opened 2022-03-15
ssh kitten: Reduce bootstrap args to support dropbear, fix NetBSD terminfo, and more

page-down:
I don't use NetBSD, but I decided to install it for testing. I found an error reported in the terminfo preparation part. .terminfo.cdb is a file and cannot be accessed as a folder. I adjusted the order to make sure that both .terminfo (the folder) and .terminfo.cdb are generated.

dropbear has a 9000 bytes ssh args limit. Remove bootstrap script comments and indents to support connecting to OpenWRT or embedded devices.

Use built-in cwd reporting function name from fish. When the feature enabled by default in future fish versions, this will override the built-in one.
I also just added this (hours ago) and am about to suggest fish to enable this.
Do you think it would be acceptable to turn this on for all versions of kitty? (Obviously there would be at most some error logs, no more harm.)

Add `force` option for `askpass` to skip ssh version check.

POSIX SHELL is not suitable to be preferred as login shell, but can fall back to that environment variable.

### DIFF
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -233,7 +233,7 @@ def prepare_script(ans: str, replacements: Dict[str, str], script_type: str) -> 
     if script_type == 'sh':
         # Remove comments and indents. The dropbear SSH server has 9000 bytes limit on ssh arguments length.
         # Needs to be trimmed before replacing EXEC_CMD to avoid affecting the indentation of user commands.
-        ans = re.sub(r'^\s*#.*\n|^\s*', '', ans, flags=re.MULTILINE)
+        ans = re.sub(r'^[ \t]*#.*$|^[ \t]*', '', ans, flags=re.MULTILINE)
     return re.sub('|'.join(fr'\b{k}\b' for k in replacements), sub, ans)
 
 
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -223,13 +223,17 @@ def safe_remove(x: str) -> None:
         os.remove(x)
 
 
-def prepare_script(ans: str, replacements: Dict[str, str]) -> str:
+def prepare_script(ans: str, replacements: Dict[str, str], script_type: str) -> str:
     for k in ('EXEC_CMD',):
         replacements[k] = replacements.get(k, '')
 
     def sub(m: 're.Match[str]') -> str:
         return replacements[m.group()]
 
+    if script_type == 'sh':
+        # Remove comments and indents. The dropbear SSH server has 9000 bytes limit on ssh arguments length.
+        # Needs to be trimmed before replacing EXEC_CMD to avoid affecting the indentation of user commands.
+        ans = re.sub(r'^\s*#.*\n|^\s*', '', ans, flags=re.MULTILINE)
     return re.sub('|'.join(fr'\b{k}\b' for k in replacements), sub, ans)
 
 
@@ -269,7 +273,7 @@ def bootstrap_script(
     if request_data:
         sd.update(sensitive_data)
     replacements.update(sensitive_data)
-    return prepare_script(ans, sd), replacements, shm
+    return prepare_script(ans, sd, script_type), replacements, shm
 
 
 def get_ssh_cli() -> Tuple[Set[str], Set[str]]:
--- a/kitty_tests/shell_integration.py
+++ b/kitty_tests/shell_integration.py
@@ -171,6 +171,8 @@ function _set_status_prompt; function fish_prompt; echo -n "$pipestatus $status 
             pty.send_cmd_to_child('set -q XDG_DATA_DIRS; or echo ok')
             pty.wait_till(lambda: pty.screen_contents().count(right_prompt) == 2)
             self.ae(str(pty.screen.line(1)), 'ok')
+
+            # CWD reporting
             self.assertTrue(pty.screen.last_reported_cwd.endswith(self.home_dir))
             q = os.path.join(self.home_dir, 'testing-cwd-notification-🐱')
             os.mkdir(q)
--- a/kitty_tests/ssh.py
+++ b/kitty_tests/ssh.py
@@ -10,7 +10,8 @@ from functools import lru_cache
 
 from kittens.ssh.config import load_config
 from kittens.ssh.main import (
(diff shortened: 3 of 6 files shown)


kovidgoyal:
On Tue, Mar 15, 2022 at 09:04:02AM -0700, page-down wrote:
> I don&#39;t use NetBSD, but I decided to install it for testing. I found an error reported in the terminfo preparation part. .terminfo.cdb is a file and cannot be accessed as a folder. I adjusted the order to make sure that both .terminfo (the folder) and .terminfo.cdb are generated.
> 
> dropbear has a 9000 bytes ssh args limit. Remove bootstrap script comments and indents to support connecting to OpenWRT or embedded devices.

Maybe add a test to ensure the script remains below the limit in the
future. Also rather than removing comments can we replace then with a
newline so that line numbers are preserved in error messages.

> 
> Use built-in cwd reporting function name from fish. When the feature enabled by default in future fish versions, this will override the built-in one.
> I also just added this (hours ago) and am about to suggest fish to enable this.
> Do you think it would be acceptable to turn this on for all versions of kitty? (Obviously there would be at most some error logs, no more harm.)

Yes, it's fine. I added code to ignore this OSC a while ago to kitty
(the process_cwd_notification function used to be a stub). And even
before that the worst that happens is kitty outputs a parse error to its
stderr.

> 
> Add `force` option for `askpass` to skip ssh version check.

Why do we care about skipping the version check? It wont work with
older OpenSSH so what does skippign the version check achieve?

kovidgoyal:
If it really bothers you, I am fine with adding an *undocumented* env
var to control it. Or maybe a sentinel file in kitty config dir.

page-down:
> ... it can simply be removed in a few years ...
> ... maybe a sentinel file  ...

Thanks, that worked great.

I've updated the PR to preserve line numbers when removing comments and indents.

page-down:
I saw the latest change to split the bootstrap script, and I remember that previously it was possible to fall back without transferring data, but not anymore.

The parameter `script_type: str` in `prepare_script (kittens/ssh/main.py)` can also be removed.

---

Regarding the last reported working directory, I have a long-standing problem:
When running make or other compilers, it may jumps between directories,
and when I want to open a new window in the last shell prompt cwd during compilation, it always goes to some inner directory.

What is the most appropriate design to solve this problem?
For example, add an option to configure which programs (e.g., make) to use `last_reported_cwd` when running.
Or add `last_reported` to `launch --cwd=` and let the user bind another set of shortcuts.

```python
def modify_argv_for_launch_with_cwd(self, argv: List[str]) -> str:
    if argv[0] != shell_path or not self.screen.last_reported_cwd:
        return self.cwd_of_child or ''
    # ...
```

kovidgoyal:
On Tue, Mar 15, 2022 at 07:41:27PM -0700, page-down wrote:
> I saw the latest change to split the bootstrap script, and I remember that previously it was possible to fall back without transferring data, but not anymore.

Yes, that was changed a while ago, since now it is no longer recommended
to use this kitten as an alias for ssh in general but only for
interactive use.

> 
> The parameter `script_type: str` in `prepare_script (kittens/ssh/main.py)` can also be removed.

Yes, I know, but I decided its good to leave it there, it might come in
handy in the future.

> 
> ---
> 
> Regarding the last reported working directory, I have a long-standing problem:
> When running make or other compilers, it may jumps between directories,
> and when I want to open a new window in the last shell prompt cwd during compilation, it always goes to some inner directory.
> 
> What is the most appropriate design to solve this problem?
> For example, add an option to configure which programs (e.g., make) to use `last_reported_cwd` when running.
> Or add `last_reported` to `launch --cwd=` and let the user bind another set of shortcuts.

I prefer this. Add another keyword named, "last_reported" just like the
"current" keyword.

page-down:
> map f1 new_window_with_cwd
> Then, pressing :kbd:`F1` ... same server ... same directory.

I'm not sure what to expect with this one.

```shell
kitty --config=NONE -o shell=fish -o 'map f1 new_window_with_cwd'
```

After connecting with ssh kitten switch to /tmp and press f1, which opens the local /tmp in a new window.

```python
def modify_argv_for_launch_with_cwd( ...
    # argv[0]: "fish"
    # shell_path: "/bin/zsh"
```

Then I found that when the login shell is `/bin/zsh`, the shell is configured as `fish`, and launch fish (or launch bash), fish will not run as a login shell (`status is-login` returns 1).
Does this work exactly the way it was designed to?

kitty/child.py
```python
```

kovidgoyal:
The way it is supposed to work is:

If you try to launch a new child without any program, it will run your
local login shell or whatever you configured in kitty.conf as your
shell. 

However, if you ask it preserve the current working dir, and you are
using the ssh kitten, then it will instead open the new window running
the ssh kitten with the same args as the current window, with the
addition of --kitten=cwd=whatever. This happens *only* if you are trying
to run your login shell, not if you are trying to run anything else.

page-down:
> The way it is supposed to work is:
> ... current working dir ... using the ssh kitten
> ... only if you are trying to run your login shell ...

For this, I don't quite understand.
When executing `new_window_with_cwd` and the configured shell is not the login shell,
why ssh kitten (with `--kitten=cwd=`) will not be used?

```python
# /etc/passwd login shell: /bin/zsh
# kitty -o shell=configured

if argv[0] != shell_path ... # configured != /bin/zsh
    if self.at_prompt and self.screen.last_reported_cwd:
        return path_from_osc7_url(...)
``` 

In this case, shouldn't the `new_window_with_cwd` action open the remote cwd (with ssh kitten)?
At least it shouldn't use remote cwd either.

kovidgoyal:
On Wed, Mar 16, 2022 at 05:24:16AM -0700, page-down wrote:
> > The way it is supposed to work is:
> > ... current working dir ... using the ssh kitten
> > ... only if you are trying to run your login shell ...
> 
> For this, I don't quite understand.
> When executing `new_window_with_cwd` and the configured shell is not the login shell,
> why ssh kitten (with `--kitten=cwd=`) will not be used?

As long as the configured shell is used it is fine. kitty doesn't care
about login shell. If you pass some other program to launch then it wont
run it via ssh since it has no way to know that's safe.

PR #4845, opened 2022-03-17
Allow launch with cwd to work with ssh kitten when the shell is configured

page-down:
The launch with cwd actions (e.g. `new_window_with_cwd`) uses the configured shell by default.
So when checking it is also compared against the configured shell.
Otherwise, if the action is executed on a window using ssh kitten, the configured shell will be used to open the remote cwd path locally.

https://github.com/kovidgoyal/kitty/pull/4836#issuecomment-1069070364

### DIFF
--- a/kitty/window.py
+++ b/kitty/window.py
@@ -20,7 +20,7 @@ from typing import (
 from .child import ProcessDesc
 from .cli_stub import CLIOptions
 from .config import build_ansi_color_table
-from .constants import appname, is_macos, shell_path, wakeup
+from .constants import appname, is_macos, wakeup
 from .fast_data_types import (
     BGIMAGE_PROGRAM, BLIT_PROGRAM, CELL_BG_PROGRAM, CELL_FG_PROGRAM,
     CELL_PROGRAM, CELL_SPECIAL_PROGRAM, CURSOR_BEAM, CURSOR_BLOCK,
@@ -47,7 +47,7 @@ from .typing import BossType, ChildType, EdgeLiteral, TabType, TypedDict
 from .utils import (
     get_primary_selection, kitty_ansi_sanitizer_pat, load_shaders, log_error,
     open_cmd, open_url, parse_color_set, path_from_osc7_url,
-    resolve_custom_file, sanitize_title, set_primary_selection
+    resolve_custom_file, resolved_shell, sanitize_title, set_primary_selection
 )
 
 MatchPatternType = Union[Pattern[str], Tuple[Pattern[str], Optional[Pattern[str]]]]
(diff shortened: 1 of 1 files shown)


page-down:
I also noticed that when the shell is executed with arguments, the reported CWD is also used.
Should we only use the CWD reported by ssh kitten+shell integration when there are no arguments?

- /etc/passwd: `/bin/zsh`
- `kitty -o 'shell /bin/fish --interactive --login' -o 'map f1 launch --cwd=current /bin/bash -c "hostname;pwd;read"`
- Use ssh kitten to connect and press F1.
- At this point, bash opens the remote CWD location in a new kitty window on the local filesystem (or `/` if it doesn't exist locally) and executes the script.

page-down:
I might suggest adding a new default value `-cwd=auto` for commands like `launch`, so that `-cwd=current` always uses the correct local working directory.

Does this work? It probably won't break backwards compatibility with `launch`, right?

kovidgoyal:
On Thu, Mar 17, 2022 at 03:18:17AM -0700, page-down wrote:
> I also noticed that when the shell is executed with arguments, the reported CWD is also used.
> Should we only use the CWD reported by ssh kitten+shell integration when there are no arguments?

No, why? The idea is that if the cursor is sitting at the prompt, then
the reported cwd is more accurate than the one that we get from the
foreground process. This is because there can be more than one process
associated with the tty and there is no way to know which is the
"foreground" one from the users perspective. Currently kitty just
assumes that the one with the higher pid is the foreground one but this
is not alway true. I have seen at least one bug report from a fish shell
user who used some prompt package that spawns long lived processes that
cause kitty to detect the incorrect CWD while at the prompt: https://github.com/IlanCosman/tide/issues/262

We should of course ignore the reported CWD if we detect that it is
remote. That is indeed a bug.

page-down:
> ... the reported cwd is more accurate than the one that we get from the foreground process ...

Sorry, my bad. I understand this, and there may even be some strange programs (shells) that do not change their own cwd and perform the work. (Although I haven't met one yet)

> ... ignore the reported CWD if we detect that it is remote ...

As long as we don't open the remote CWD locally, I think this will all work fine.

page-down:
I tried `launch --cwd=last_reported`, however it did not use the CWD from OSC 7.

```shell
sh -c 'printf "\033]7;file://debug/tmp\033\134";read'
```

- kitty/child.py, class Child, init
  - cwd_from.modify_argv_for_launch_with_cwd()
  - class CwdRequest, modify_argv_for_launch_with_cwd()
    - window.modify_argv_for_launch_with_cwd()
      - class Window, modify_argv_for_launch_with_cwd()
        - self.cwd_of_child() <--- `cwd_of_child` from Window
          - self.at_prompt -> False
          - self.screen.last_reported_cwd -> Yes, Not Empty
          - not self.child_is_remote -> True
          - return self.child.foreground_cwd or self.child.current_cwd

The final path returned is the CWD of the process.

I'm trying some file managers, can `kitty-shell-cwd://` be used by non-shell command line programs?
Is the name generic enough?

kovidgoyal:
On Thu, Mar 17, 2022 at 11:20:23PM -0700, page-down wrote:
> I'm trying some file managers, can `kitty-shell-cwd://` be used by non-shell command line programs?
> Is the name generic enough?


Just use file:// URLs the only reason I created kitty-shell-cwd:// is
because doing URL escaping in shells is not easy, and is not needed for
this usecase anyway.

PR #4848, opened 2022-03-19
macOS: Allow the New kitty Here services to open multiple folders

page-down:
Select multiple folders in Finder, right click, and `New kitty Tab/Window Here` in Services will open them all.

Before this PR only the last one will be opened.

https://github.com/kovidgoyal/kitty/pull/4618#issuecomment-1030595767

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -53,6 +53,8 @@ Detailed list of changes
 
 - Fix continued lines not having their continued status reset on line feed (:iss:`4837`)
 
+- macOS: Allow the New kitty Tab/Window Here services to open multiple selected folders.  (:pull:`4848`)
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I am not comfortable with disallowing : in working directories. Instead
make those functions take optionally a string or a sequence of strings.
And call them with a sequence of strings in the cocoa case, should be
easy to do with PyUnicode_Split()

page-down:
> I am not comfortable with disallowing : in working directories. ...

I thought of this right after committing, and it looks like `new_*_with_wd` is only used under macOS.
However the platform exclusive should be as little as possible in the Python part, while making the Python part more universal.

I will address this.

Issue #4853, opened 2022-03-20
IME is not working properly in kitty under linux


mendaxcat:
**Describe the bug**
After commit  a987fb81a0dd9ec2a4449d196a51d04e5a8162f3, IME is not working properly in kitty under linux. @page-down

**To Reproduce**
1. launch kitty and switch to ibus-libpinyin or ibus-rime input method
2. open a new tab and type
3. IME does not work and only english letters are entered

**Environment details**
```
kitty 0.24.1 (a987fb81a0) created by Kovid Goyal
Linux mendaxca 5.10.0-12-amd64 #1 SMP Debian 5.10.103-1 (2022-03-07) x86_64
Debian GNU/Linux 11 mendaxca /dev/tty

Running under:Wayland
Frozen: False
Paths:
  kitty: /tmp/kitty/linux-package/bin/kitty
  base dir: /tmp/kitty/linux-package/lib/kitty
  extensions dir: /tmp/kitty/linux-package/lib/kitty/kitty
  system shell: /bin/bash

Config options different from defaults:
```

page-down:
It works fine under X11.
I tried the previous kitty versions and could not get ibus to work under wayland. (Arch Linux + KDE)

```shell
GLFW_IM_MODULE=ibus kitty
```

Did I miss something?

kovidgoyal:
Wayland has its own text input protocol, via the compositor so
GLFW_IM_MODULE is not needed.

page-down:
OK, I installed fcitx5 and it works as expected.
No additional configuration is required, it works out of the box.

fcitx5 wayland im frontend -> kwin wayland -> zwp_text_input_v3 -> kitty

@mendaxcat 
I don't know how to configure ibus related so I can't help you.

page-down:
I confirmed that IME works fine with Debian 11 GNOME + Wayland + IBus.

```text
Debian 11.2
IBus 1.5.23
kitty 0.24.4

XDG_SESSION_TYPE=wayland
```

mendaxcat:
Did you follow my steps to reproduce the problem?
The IME does work in kitty, but not after switching tab.
![image](https://user-images.githubusercontent.com/90375986/159193405-dabb1610-a9b7-49c0-8467-9da396e4aef3.png)
kitty under x11 does not have this problem. @page-down

page-down:
> ... but not after switching tab ...

I noticed that after switching windows the IME needs to be switched again to work.
I also found a problem with leftover pre-edit text.

PR #4855, opened 2022-03-21
Wayland: Fix IME is not enabled after switching the focus

page-down:
Enable the IME under wayland after the window gets focus.
Also fixes the issue that pre-edit text cannot be cleared. After pressing backspace, the last character remains.

https://github.com/kovidgoyal/kitty/issues/4853

Please review, thank you.

### DIFF
--- a/glfw/wl_text_input.c
+++ b/glfw/wl_text_input.c
@@ -139,7 +139,12 @@ _glfwPlatformUpdateIMEState(_GLFWwindow *w, const GLFWIMEUpdateEvent *ev) {
     switch(ev->type) {
         case GLFW_IME_UPDATE_FOCUS:
             debug("\ntext-input: updating IME focus state, focused: %d\n", ev->focused);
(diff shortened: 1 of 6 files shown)


ProgramFan:
This is just what happens on my side.

glfw/wl_text_input.c
```diff
@@ -93,6 +93,9 @@ text_input_done(void *data UNUSED, struct zwp_text_input_v3 *txt_input UNUSED, u
     if (pending_pre_edit) {
         send_text(pending_pre_edit, GLFW_IME_PREEDIT_CHANGED);
         free(pending_pre_edit); pending_pre_edit = NULL;
+    } else {
+        // Clear pre-edit text
+        send_text(NULL, GLFW_IME_PREEDIT_CHANGED);
```
fezboy, line 98: This line causes #5105. Removing it fixes the CPU bug, but likely reintroduces the issue this commit addresses, which I am unable to test.

PR #4858, opened 2022-03-21
Add option to set XDG sound theme on Linux

serebit:
Based on #2125. The default sound theme is `__custom`, which is the custom theme created by GNOME and Budgie. If this theme doesn't exist, canberra instead uses the default theme to produce the sound, so it should work without a hitch on all desktop environments.

This is a draft because I was unable to figure out how to regenerate `to-c-generated.h` to get the changes to propagate, and it seems that the code to add a new option changed in the time between #2125 and now. Let me know how to get everything connected and I'll get it ready to merge.

### DIFF
--- a/kitty/data-types.h
+++ b/kitty/data-types.h
@@ -386,7 +386,7 @@ void scroll_event(double, double, int, int);
 void on_key_input(GLFWkeyevent *ev);
 void request_window_attention(id_type, bool);
 #ifndef __APPLE__
-void play_canberra_sound(const char *which_sound, const char *event_id, bool is_path, const char *role);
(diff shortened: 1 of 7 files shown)


kovidgoyal:
You create it with ./gen-config.py  Although is this really needed anymore since we can specify an actual file for the bell sound?

serebit:
I ran gen-config.py, and as far as I c

kovidgoyal:
You need to add ctype to the definition and add the function to-c.h see for example bell_path

serebit:
Rebased, compiled, and tested, and getting a segfault - seems like `OPT(linux_bell_theme_name)` is resolving to null, not sure why. I did figure out that gen-config.py likely isn't making changes because there's no need for a ctype, similar to font_family, since it's just a string value.

PR #4867, opened 2022-03-23
ssh kitten: Fix SSH URL hostname matching and other minor tweaks

page-down:
Fix ssh kitten URL hostname matching.
Fix bash integration cwd reporting missing hostname.
Fix not specifying the executable path when checking less version.

I noticed that if OSC 7 is output in the sub shell, it will remain after exit (since PWD does not change in the shell).
The only solution would be to unconditionally update CWD OSC7 when outputting PROMPT. 
But I think you have noticed this and have chosen not to do so.
In my opinion the current solution is good enough. I personally don't use many programs or sub shells that output OSC 7, so I don't care.

### DIFF
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -466,7 +466,7 @@ def wrap_bootstrap_script(sh_script: str, interpreter: str) -> List[str]:
 
 
 def get_remote_command(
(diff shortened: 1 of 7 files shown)


kovidgoyal:
Yes OSC 7 is not in general a good idea, precisely because it needs to be supported by every program to be workable. That's why kitty only uses it if at a prompt by default.

PR #4875, opened 2022-03-24
ssh kitten: Fix python bootstrap checking zsh rc files with ZDOTDIR and more

page-down:
Fix python bootstrap checking zsh rc files with ZDOTDIR.

Allow to configure HOME environment variable.
Useful if the user does not have a home directory or the directory is not writable.

Clean up environment variables to prevent pollution.

Please review, thank you.

---

The last one on my list (at least for now):

When using `copy` to transfer larger files with less bandwidth,
it is faster to quickly compare file information (e.g. mtime and file size) and request the desired file
than to repeat the transfer and update the same files multiple times.

In this scenario I think it's worth configuring ssh kitten to ask first before transmitting.

```text
metadata; <record sep>; data_and_ask_again.sh (check and build file list); END-SCRIPT;
metadata; <record sep>; optional tarball; END-FILE;
```

### DIFF
--- a/kittens/ssh/main.py
+++ b/kittens/ssh/main.py
@@ -242,7 +242,7 @@ def prepare_exec_cmd(remote_args: Sequence[str], is_python: bool) -> str:
     if is_python:
         return standard_b64encode(' '.join(remote_args).encode('utf-8')).decode('ascii')
     args = ' '.join(c.replace("'", """'"'"'""") for c in remote_args)
-    return f"""exec "$login_shell" -c '{args}'"""
+    return f"""unset KITTY_SHELL_INTEGRATION; exec "$login_shell" -c '{args}'"""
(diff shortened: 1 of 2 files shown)


kovidgoyal:
On Thu, Mar 24, 2022 at 07:29:19AM -0700, page-down wrote:
> The last one on my list (at least for now):
> 
> When using `copy` to transfer larger files with less bandwidth,
> it is faster to quickly compare file information (e.g. mtime and file size) and request the desired file
> than to repeat the transfer and update the same files multiple times.
> 
> In this scenario I think it&#39;s worth configuring ssh kitten to ask first before transmitting.

copy is meant to copy config files not large data. An additional
roundtrip adds a lot of latency unless the remote host is on a local
network.

PR #4880, opened 2022-03-25
Fix hints kitten overlay ready reporting not enabled and eliminate flickering when exits

page-down:
Fix hints kitten overlay ready reporting not enabled.
Report overlay ready in case of no match.
Eliminate flickering when kitten exits in overlay.

Please review, thank you.
There may be cases where robustness needs to be improved.

### DIFF
--- a/kittens/hints/main.py
+++ b/kittens/hints/main.py
@@ -118,6 +118,7 @@ def render(text: str, current_input: str, all_marks: Sequence[Mark], ignore_mark
 
 class Hints(Handler):
 
+    use_alternate_screen = False
     overlay_ready_report_needed = True
 
     def __init__(self, text: str, all_marks: Sequence[Mark], index_map: Dict[int, Mark], args: HintsCLIOptions):
--- a/kittens/hints/main.py
+++ b/kittens/hints/main.py
@@ -27,7 +27,7 @@ from kitty.utils import (
 from ..tui.handler import Handler, result_handler
 from ..tui.loop import Loop
 from ..tui.operations import faint, styled
-from ..tui.utils import report_unhandled_error
+from ..tui.utils import report_error, report_unhandled_error
 
 
 @lru_cache()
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I dont follow the reason for this? Why are you setting pending update in
the Handler exit callbacks? Where is this pending update actually
ended? In any case, the exit handlers don't typically draw much to the
screen and they can be overridden in sub classes.

Also there is no need for pending updates when reporting an error. It
doesnt need to be atomic.

page-down:
> I dont follow the reason for this?

Currently hints kitten will first draw a blank screen and then close the overlay when exiting.
This will cause flickering on exit.

With pending update enabled to exit as the last screen content is maintained.
Maybe there is a better way?

> Where is this pending update actually ended?

If an error occurs, it will end at `with pending_update()` (two enable, one disabled).
If the program exits and the window closes, it does not matter.

> In any case, the exit handlers ... they can be overridden in sub classes.

So this is not good, is there a more appropriate way?

kovidgoyal:
On Fri, Mar 25, 2022 at 04:06:15AM -0700, page-down wrote:
> > I dont follow the reason for this?
> 
> Currently hints kitten will first draw a blank screen and then close the overlay when exiting.
> This will cause flickering on exit.
> 
> With pending update enabled to exit as the last screen content is maintained.
> Maybe there is a better way?

Just dont use the alternate screen. Set use_alternate_screen to False in
the Hints handler.

> 
> > Where is this pending update actually ended?
> 
> If an error occurs, it will end at `with pending_update()` (two enable, one disabled).
> If the program exits and the window closes, it does not matter.
> 
> > In any case, the exit handlers ... they can be overridden in sub classes.
> 
> So this is not good, is there a more appropriate way?

There is no need to do this at all, see above.

page-down:
> Just dont use the alternate screen. ...

Ok, this works well.

However, hints kitten may run in the shell.
The alt screen should only be disabled when run directly by kitty (e.g. overlay).
Do we need to add such an environment variable? (launched by kitty)

kovidgoyal:
On Fri, Mar 25, 2022 at 05:53:37AM -0700, page-down wrote:
> > Just dont use the alternate screen. ...
> 
> Ok, this works well.
> 
> However, hints kitten may run in the shell.
> The alt screen should only be disabled when run directly by kitty (e.g. overlay).
> Do we need to add such an environment variable? (launched by kitty)

Whats the use case for running in the shell? I dont know of any.

PR #4897, opened 2022-03-30
Docs: Add instructions for compiling terminfo under macOS

page-down:
The `tic` that comes with macOS 12.3 is from ancient ncurses 5.7 (2008) and does not seem to support reading from /dev/stdin.

Please review, thank you.


page-down:
Hmm... I don't know how to get around this.
GitHub and CI services have been experiencing instability issues recently.

PR #4945, opened 2022-04-09
Change size_t_min static function in kitty/ringbuf.c to static inline

ChungHsuanChen:
Change static function  size_t_min in kitty/ringbuf.c to static inline function and move to kitty/ringbuf.h for simplifying .c file

### DIFF
--- a/kitty/ringbuf.c
+++ b/kitty/ringbuf.c
@@ -26,10 +26,6 @@
 #include <sys/param.h>
 #include <assert.h>
 
-static size_t
-size_t_min(size_t x, size_t y) {
-    return x > y ? y : x;
-}
 
 /*
  * The code is written for clarity, not cleverness or performance, and
--- a/kitty/ringbuf.h
+++ b/kitty/ringbuf.h
@@ -250,3 +250,8 @@ ringbuf_write(int fd, ringbuf_t rb, size_t count);
  */
 void *
 ringbuf_copy(ringbuf_t dst, ringbuf_t src, size_t count);
+
+static inline size_t
+size_t_min(size_t x, size_t y) {
+    return x > y ? y : x;
+}


kovidgoyal:
I dont understand the motivation of this. Moving the function into a
header file doesnt simplify anything and in fact means that the function
has to now be parsed by the compiler for every translation unit the header
file is included in.

ChungHsuanChen:
The motivation of moving this function to header file is that making c file "only" contain implementation about ringbuf. Other function in ringbuf.c is correlated with struct ringbuf_t.
Making it an inline function is because it will have some overheads, such as read parameter from stack ,read register. Although size_t_min function only calls 10 times in ringbuf.c, it will affect performance if the function is called many times in the future without inline function or a macro.

kovidgoyal:
On Sat, Apr 09, 2022 at 07:08:03PM -0700, ChungHsuanChen wrote:
> The motivation of moving this function to header file is that making c file "only" contain implementation about ringbuf. Other function in ringbuf.c is correlated with struct ringbuf_t.

It's a utility function used by other functions in that .c file. And
only there, as such, it belongs there.

> Making it an inline function is because it will have some overheads, such as read parameter from stack ,read register. Although size_t_min function only calls 10 times in ringbuf.c, it will affect performance if the function is called many times in the future without inline function or a macro. 

inlining in modern compilers is decided by the compiler, the inline
keyword doesn't influence the compiler's decision on this.

ChungHsuanChen:
Yes, I understand. Making this function a macro will it be more resonable?

kovidgoyal:
You could, but it wont make any actual difference, since I am pretty sure every compiler will inline that function call anyway. And using an actual function means nicer ergonomics than a macro. Macros are useful only when you need to run the code for variables of different types, which is not the case here.

ChungHsuanChen:
I see. From my understanding, macro will be handled by preprocessor not compiler. Won't it be beneficial or not?

kovidgoyal:
No, it wont. The actual assembly generated in both cases will be
identical.

PR #4967, opened 2022-04-14
Minor refactoring for fish clone-in-kitty and allow use without env -0

page-down:
The builtin option from the latest version of fish is not used for now. clone-in-kitty has been tested with fish 3.2.x.
BusyBox env does not support the -0 option. Implemented with fish.

Builtins are heavily used in the scripts that come with fish, and any environment that overrides these builtin, I believe fish will not work properly. No need to guard with `builtin`, simplify the code for future maintenance.
If we use `command`, then `base64` or other commands that is not in the path, e.g. alias or function, will not work.
If builtin or commands don't work, I think `clone-in-kitty` is the last thing to worry about, leaving the user to work it out on their own, rather than rejecting the functions from the normal users.

Remove quotation marks and the trailing `# "`, `# ‘` by escaping special characters in the string.

Remove `echo $data_len`.

Please review, thanks.

---

I found that environment variables containing newlines, the `\n` are lost after cloning. This is reproduced in zsh/fish. (I believe the base64-encoded environment variables are correct in fish.)
I need help on this.

Also the zsh and bash versions of clone-in-kitty do not work in environments that do not support env -0.

---

I personally don't like to copy all environment variables, because the shell is not only environment variables, but also many internal variables (and funtions) that are not exported. This feature only copies `half of them`, and my environment is broken and missing something (that I know) after cloning. I prefer to explicitly configure the ones that need to be copied.

Or, maybe it's time to increase the robustness of my environment.

### DIFF
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -6,7 +6,7 @@
 if set -q KITTY_FISH_XDG_DATA_DIR
     if set -q XDG_DATA_DIRS
         set --global --export --path XDG_DATA_DIRS "$XDG_DATA_DIRS"
-        if set -l index (contains -i "$KITTY_FISH_XDG_DATA_DIR" $XDG_DATA_DIRS)
+        if set --local index (contains --index "$KITTY_FISH_XDG_DATA_DIR" $XDG_DATA_DIRS)
             set --erase --global XDG_DATA_DIRS[$index]
             test -n "$XDG_DATA_DIRS" || set --erase --global XDG_DATA_DIRS
         end
@@ -22,7 +22,7 @@ not functions -q __ksi_schedule || exit 0
 # Check fish version 3.3.0+ efficiently and fallback to check the minimum working version 3.2.0, exit on outdated versions.
 # "Warning: Update fish to version 3.3.0+ to enable kitty shell integration.\n"
 set -q fish_killring || set -q status_generation || string match -qnv "3.1.*" "$version"
-or echo -en "\eP@kitty-print|V2FybmluZzogVXBkYXRlIGZpc2ggdG8gdmVyc2lvbiAzLjMuMCsgdG8gZW5hYmxlIGtpdHR5IHNoZWxsIGludGVncmF0aW9uLgo=\e\\" && exit 0 || exit 0 # "
+or echo -en \eP@kitty-print\|V2FybmluZzogVXBkYXRlIGZpc2ggdG8gdmVyc2lvbiAzLjMuMCsgdG8gZW5hYmxlIGtpdHR5IHNoZWxsIGludGVncmF0aW9uLgo=\e\\ && exit 0 || exit 0
 
 function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after other scripts have run, we hope"
     functions --erase __ksi_schedule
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Thu, Apr 14, 2022 at 06:57:33AM -0700, page-down wrote:
> The builtin option from the latest version of fish is not used for now. clone-in-kitty has been tested with fish 3.2.x.
> BusyBox env does not support the -0 option. Implemented with fish.
> 
> Builtins are heavily used in the scripts that come with fish, and any environment that overrides these builtin, I believe fish will not work properly. No need to guard with `builtin`, simplify the code for future maintenance.
> If we use `command`, then `base64` or other commands that is not in the path, e.g. alias or function, will not work.
> If builtin or commands don&#39;t work, I think `clone-in-kitty` is the last thing to worry about, leaving the user to work it out on their own, rather than rejecting the functions from the normal users.
> 
> Remove quotation marks and the trailing `# &quot;`, `# ‘` by escaping special characters in the string.
> 
> Remove `echo $data_len`.
> 
> Please review, thanks.
> 
> ---
> 
> I found that environment variables containing newlines, the `\n` are lost after cloning. This is reproduced in zsh/fish. (I believe the base64-encoded environment variables are correct in fish.)
> I need help on this.

The culprit will probably be serialize_env() one needs to serialize
newlines specially there.

> 
> Also the zsh and bash versions of clone-in-kitty do not work in environments that do not support env -0.

Yes, I know. I dont see a good solution for these. Env vars can have
newlines so env without -0 is pretty pointless.

> 
> ---
> 
> I personally don&#39;t like to copy all environment variables, because the shell is not only environment variables, but also many internal variables (and funtions) that are not exported. This feature only copies `half of them`, and my environment is broken and missing something (that I know) after cloning. I prefer to explicitly configure the ones that need to be copied.

Trying to serialize and restore complete shell state is pretty hairy,
but feel free to attempt it, the mechanism is present, it can work by
setting an env var with the serialized state which th eshell integration
script can read and try to restore.

> 
> Or, maybe it&#39;s time to increase the robustness of my environment.

Yes, clone isnt currently meant to be a full clone just to work for simple use
cases.

kovidgoyal:
Actually I did the following test
export n=$'a\nb'; clone-in-kitty
echo $n (in the cloned window)
and it output
a
b

as expected under zsh. Did you mean over ssh?

page-down:
I tested again and confirmed that local and ssh, zsh clone-in-kitty are working fine.

However, the newline in the environment variable is normal after using local fish clone-in-kitty, but missing with ssh, and the cause has not been found yet.

Also my local zsh clone-in-kitty is throwing an error. (when the macOS user's shell is zsh)
I think it also errors out when the shell in /etc/passwd is changed to bash or fish.

```text
Failed to launch child: -zsh
With error: No such file or directory
Press Enter to exit.
```

page-down:
I noticed that after using zsh clone-in-kitty, it displays a percentage sign with a white background and black text. Where is this coming from? Is it some kind of "partial line"?

page-down:
> > Actually I did the following test ... as expected under zsh. Did you mean over ssh?
>
> I tested again and confirmed that local and ssh, zsh clone-in-kitty are working fine.

I remembered how I tested zsh (with ssh) before.

```shell
export DEBUG=$(printf "abc\ndef")
env
```

```text
DEBUG=abc
def
```

after zsh clone-in-kitty with ssh:

```text
DEBUG=abcdef
```

kovidgoyal:
yes, with ssh its probably not going to work because of the serialize_env problem
I mentioned before. But, locally it should work fine.

page-down:
I removed `\n` from quote_pat and it looks normal.
Do you think that's enough?
I wonder if something could go wrong.

```python
# quote_pat = re.compile('([\\`"\n])')
quote_pat = re.compile('([\\`"])')
```

As for `Failed to launch child: -zsh`

kitty/child.py
```python
```

kitty/macos_process_info.c `cmdline_of_process`

Does it require special case handling? Or should we get process information in some other way?

kovidgoyal:
> kitty/macos_process_info.c `cmdline_of_process`
> 
> Does it require special case handling? Or should we get process information in some other way?

https://github.com/kovidgoyal/kitty/commit/0fcfe104e3850a6d597be342502caac1df02a834

kovidgoyal:
On Thu, Apr 14, 2022 at 10:48:48PM -0700, page-down wrote:
> I noticed that after using zsh clone-in-kitty, it displays a percentage sign with a white background and black text. Where is this coming from? Is it some kind of "partial line"?

I dont see this, that is the zsh no newline at end of output marker.
Where are you seeing it?

page-down:
> ... Where are you seeing it?

This percent sign appears when clone-in-kitty is run again in a zsh or fish session after clone-in-kitty. zshrc is cleared.
I want to be sure that it works properly in various situations.

In the recent commit there is a mention of a change related to macOS, I tried it and still got `Failed to launch child: -zsh`.

kovidgoyal:
I tried running clone-in-kitty, then clone-in-kitty again in the cloned window and saw no %. Also with empty .zshrc

I dont have access to my macos machine to test at the moment, but it should be easy to verify whats going on. See line 590 in launch.py

page-down:
> ... saw no %. Also with empty .zshrc

I specifically tried it in a Linux environment and `kitty -o shell=/bin/zsh` ran clone-in-kitty (first time) and the percent sign came up. I don't know the root cause yet until I dig deeper.

> ... it should be easy to verify whats going on ...

It is, only some minor changes are needed.

`kitty --config=NONE`
```text
cmdline: ['-zsh']
window.child.argv[0]: /bin/zsh
window.child.final_exe: /bin/zsh
```

kovidgoyal:
On Fri, Apr 15, 2022 at 02:46:17AM -0700, page-down wrote:
> > ... saw no %. Also with empty .zshrc
> 
> I specifically tried it in a Linux environment and `kitty -o shell=/bin/zsh` ran clone-in-kitty (first time) and the percent sign came up. I don't know the root cause yet until I dig deeper.

Ah I think I know what this is. You are presumably using a layout that
causes windows to be resized when a new window is created. That resize
occurs during clone-in-kitty execution, which causes the cursor to move,
which breaks whatever heuristic zsh uses to determine if there has been
command output. Use the stacked layout or --type=tab and you wont see
the %

> 
> > ... it should be easy to verify whats going on ...
> 
> It is, only some minor changes are needed.
> 
> `kitty --config=NONE`
> ```text
> cmdline: ['-zsh']
> window.child.argv[0]: /bin/zsh
> window.child.final_exe: /bin/zsh
> ```

https://github.com/kovidgoyal/kitty/commit/716813e38af38de96e1ec1363b719c2e4f352008

PR #4971, opened 2022-04-15
clone-in-kitty: Skip SSH_TTY and other SSH environment variables

page-down:
Use new values for SSH environment variables, making sure SSH_TTY, etc. is correct.

Perhaps a configurable filter list (prefix `+` for whitelist item) would allow users to solve these problems on their own.

I'm wondering if we should check the `USER` environment variables matches current user in ssh kitten bootstrap script or when running it locally, and not use them (all env vars) if they (the user) are different. If SSH connects and switches users and then runs clone-in-kitty, the reported environment variables such as HOME, USER, etc. will most likely cause the environment to be abnormal.

### DIFF
--- a/kitty/launch.py
+++ b/kitty/launch.py
@@ -567,9 +567,13 @@ class CloneCmd:
                 self.args.append(v)
             elif k == 'env':
                 self.env = parse_bash_env(v) if self.envfmt == 'bash' else parse_null_env(v)
-                for filtered in ('PS0', 'PS1', 'PS2', 'PS3', 'PS4', 'RPS1', 'PROMPT_COMMAND'):
+                for filtered in (
                     # some people export these. We want the shell rc files to
                     # recreate them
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Fri, Apr 15, 2022 at 04:55:57AM -0700, page-down wrote:
> Use new values for SSH environment variables, making sure SSH_TTY, etc. is correct.
> 
> Perhaps a configurable filter list (prefix `+` for whitelist item) would allow users to solve these problems on their own.

There is already one

clone-in-kitty --env=WHATEVER --env=WHATEVER2

> 
> I&#39;m wondering if we should check the `USER` environment variables matches current user in ssh kitten bootstrap script or when running it locally, and not use them (all env vars) if they (the user) are different. If SSH connects and switches users and then runs clone-in-kitty, the reported environment variables such as HOME, USER, etc. will most likely cause the environment to be abnormal.

I'm OK with filtering out HOME and USER. Feel free to send a PR.

PR #4979, opened 2022-04-17
clone-in-kitty: Fix activating the conda environment

page-down:
- Use conda activate and avoid duplicate entries in the PATH.
- Avoid directly sets env vars when running clone-in-kitty locally. Fix the value is always the same when comparing conda env var.
- Explicitly specifies the allowed launch types.

Since conda activate still returns zero when activation fails, use environment variables to check for success.
When the activation fails, the PATH still contains the cloned virtual environment path, which cannot be avoided.

I tried zsh and after conda venv activation there are duplicate virtual environment paths in the PATH.
Do you encounter the same in your environment?

Also currently there are some errors after running clone-in-kitty.
```text
Failed to send resize signal to child with id: 2 (children count: 1) (add queue: 1)
Failed to send resize signal to child with id: 4 (children count: 1) (add queue: 1)
```

Thanks to the recent improvements, clone-in-kitty is now working properly in my environment again.

### DIFF
--- a/kitty/launch.py
+++ b/kitty/launch.py
@@ -617,4 +617,4 @@ def clone_and_launch(msg: str, window: Window) -> None:
             cmdline[0] = window.child.final_exe
         if cmdline and cmdline == [window.child.final_exe] + window.child.argv[1:]:
             cmdline = window.child.unmodified_argv
-    launch(get_boss(), c.opts, cmdline, base_env=c.env, active=window, is_clone_launch=is_clone_launch)
+    launch(get_boss(), c.opts, cmdline, active=window, is_clone_launch=is_clone_launch)
--- a/kitty/launch.py
+++ b/kitty/launch.py
@@ -573,10 +573,13 @@ class CloneCmd:
                     # some people export these. We want the shell rc files to recreate them
                     'PS0', 'PS1', 'PS2', 'PS3', 'PS4', 'RPS1', 'PROMPT_COMMAND', 'SHLVL',
                     # conda state env vars
-                    'CONDA_SHLVL', 'CONDA_PREFIX', 'CONDA_EXE', 'CONDA_PROMPT_MODIFIER', 'CONDA_EXE', 'CONDA_PYTHON_EXE',
-                    # skip SSH environment variables
+                    'CONDA_SHLVL', 'CONDA_PREFIX', 'CONDA_PROMPT_MODIFIER', 'CONDA_EXE', 'CONDA_PYTHON_EXE', '_CE_CONDA', '_CE_M',
+                    # SSH environment variables
                     'SSH_CLIENT', 'SSH_CONNECTION', 'SSH_ORIGINAL_COMMAND', 'SSH_TTY', 'SSH2_TTY',
-                }}
+                } and not k.startswith((
+                    # conda state env vars for multi-level virtual environments
+                    'CONDA_PREFIX_',
+                ))}
             elif k == 'cwd':
                 self.cwd = v
 
--- a/kitty/launch.py
+++ b/kitty/launch.py
@@ -589,7 +589,7 @@ def clone_and_launch(msg: str, window: Window) -> None:
         c.opts.cwd = c.cwd
     c.opts.copy_colors = True
     c.opts.copy_env = False
-    if c.opts.type in ('clipboard', 'primary', 'background'):
+    if c.opts.type not in ('os-window', 'tab', 'window', 'overlay'):
         c.opts.type = 'window'
     is_clone_launch = serialize_env(c.shell, c.env or {})
     ssh_kitten_cmdline = window.ssh_kitten_cmdline()
--- a/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
+++ b/shell-integration/fish/vendor_conf.d/kitty-shell-integration.fish
@@ -24,7 +24,6 @@ not functions -q __ksi_schedule || exit 0
 set -q fish_killring || set -q status_generation || string match -qnv "3.1.*" "$version"
 or echo -en \eP@kitty-print\|V2FybmluZzogVXBkYXRlIGZpc2ggdG8gdmVyc2lvbiAzLjMuMCsgdG8gZW5hYmxlIGtpdHR5IHNoZWxsIGludGVncmF0aW9uLgo=\e\\ && exit 0 || exit 0
 
-
 function __ksi_schedule --on-event fish_prompt -d "Setup kitty integration after other scripts have run, we hope"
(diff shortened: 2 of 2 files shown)


kovidgoyal:
I have fixed the issues you pointed out slightly differently.                                                                                      
I dont see how PATH duplication is possible in zsh since it runs                                                                                   
typeset -U path after running conda.                                                                                                               
                                                                                                                                                   
The failed to send resize signal is harmless, it jsut means the window                                                                             
was resized before the child process actually started.

page-down:
> I have fixed the issues you pointed out slightly differently. ...

There are also `CONDA_PREFIX_*` environment variables that are not filtered.
There is still a duplicate PATH under fish, which causes the bin path of venv to be used after the deactivation.
What is the problem you are having with `conda activate`? I tested it under macOS / Linux without any problem. I need this to work properly.

I see the new changes, do I need to submit a new PR?

> ... since it runs typeset -U path after running conda.

Are you sure typeset is case-insensitive? I ran `typset -U PATH` after cloning and it looks fine.

> ... the window was resized before the child process actually started.

It seems to me that this error could just be omitted. Is it completely useless to send a resize signal since the process hasn't started yet? Or put it in the debug log?

kovidgoyal:
On Sun, Apr 17, 2022 at 07:45:52AM -0700, page-down wrote:
> > I have fixed the issues you pointed out slightly differently. ...
> 
> There are also `CONDA_PREFIX_*` environment variables that are not filtered.

Those dont cause any issues but feel free to send a PR to remove them.

> There is still a duplicate PATH under fish, which causes the bin path of venv to be used after the deactivation.

Yes fish has no convenient way to remove duplicates from PATH,
apparently they want everyone to write their own function for it. And I
cant be bothered since I dont use fish and I have to lookup its
non-POSIX syntax every time. Again feel free to send a PR.

> What is the problem you are having with `conda activate`? I tested it under macOS / Linux without any problem. I need this to work properly.

It simply did not work, but the replacement code I used works fine, it's
what conda activate does under the hood anyway. Is it not working for
you?

> 
> I see the new changes, do I need to submit a new PR?
> 
> > ... since it runs typeset -U path after running conda.
> 
> Are you sure typeset is case-insensitive? I ran `typset -U PATH` after cloning and it looks fine.

path is an array variable in zsh that shadows PATH. Try this

export PATH=$PATH:/a:/a
typeset -U path
echo $PATH

it will have only one /a in it.

> 
> > ... the window was resized before the child process actually started.
> 
> It seems to me that this error could just be omitted. Is it completely useless to send a resize signal since the process hasn't started yet? Or put it in the debug log?

I like logging errors when possible. One cannot know in advance that an
error message will never be useful. For this particular error it can
also occur if the signal is sent after the child is dead, which can be
useful to know.

page-down:
>> What is the problem you are having with `conda activate`? ...
>
> It simply did not work ... Is it not working for you?

`conda` is a function, however there are some issues with the official shell script, so I used my own replacement. (same parameters)

I can't reproduce this with an empty shell configuration with official `conda init fish`. That function is so simple that I can't see where it would cause a problem.

If there are any other possibilities to use the official `conda activate ENV_NAME` call form, please let me know.

> ... it can also occur if the signal is sent after the child is dead ...

Thanks for the explanation, that is indeed worth outputting the error log.
It's just the first time I've seen it, and it's particularly strange that it appears every time I run clone-in-kitty, but of course it has no other effect.

page-down:
> path is an array variable in zsh that shadows PATH. ...

I used the following code in `kitty/shell-integration/zsh/zsh-integration` and found that the second `echo $PATH` output is empty.

```zsh
# Ensure PATH has no duplicate entries
export PATH="/a:/a:/usr/local/bin:/usr/bin:/bin"
echo before: $PATH
builtin typeset -U path
echo after: $PATH
```

```text
before: /a:/a:/usr/local/bin:/usr/bin:/bin                                                                                                                      
after:
```

Tested on Linux/macOS, `zsh 5.8`.

kovidgoyal:
On Sun, Apr 17, 2022 at 08:50:45AM -0700, page-down wrote:
> > path is an array variable in zsh that shadows PATH. ...
> 
> I used the following code in `kitty/shell-integration/zsh/zsh-integration` and found that the second `echo $PATH` output is empty.
> 
> ```zsh
> # Ensure PATH has no duplicate entries
> export PATH="/a:/a:/usr/local/bin:/usr/bin:/bin"
> echo before: $PATH
> builtin typeset -U path
> echo after: $PATH
> ```
> 
> ```text
> before: /a:/a:/usr/local/bin:/usr/bin:/bin                                                                                                                      
> after:
> ```
> 
> Tested on Linux/macOS, `zsh 5.8`.


Output for me is:
before: /a:/a:/usr/local/bin:/usr/bin:/bin
after: /a:/usr/local/bin:/usr/bin:/bin

zsh 5.8.1

I will change it to typeset -U path PATH
to make sure.

page-down:
> ... I will change it to typeset -U path PATH to make sure.

Thanks.

I specifically tested zsh 5.8.1 under arch linux (empty .zshrc) and the second echo is still empty.
There is no problem executing it from an interactive prompt.
Executing `zsh /tmp/debug.zsh` (with a standalone zsh script) is also fine.

Really don't know what caused this until I dig further.

page-down:
I tested the last commit (`typeset -U path PATH`) on arch linux zsh 5.8.1 and echo $PATH is still empty.

Finally tried the following to work properly.

```zsh
export -U PATH="$PATH"
```

You may have a better solution.

kovidgoyal:
weird what is going on with your zsh???

Does 

builtin typeset -gxU PATH="$PATH"

work?


On Sun, Apr 17, 2022 at 09:24:21AM -0700, page-down wrote:
> I tested the last commit (`typeset -U path PATH`) on arch linux zsh 5.8.1 and echo $PATH is still empty.
> 
> Finally tried the following to work properly.
> 
> ```zsh
> export -U PATH="$PATH"
> ```
> 
> You may have a better solution.
> 
> -- 
> https://github.com/kovidgoyal/kitty/pull/4979#issuecomment-1100910321
> 
> Message ID: ***@***.***>

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

page-down:
Yes, `builtin typeset -gxU PATH="$PATH"` works fine.

kovidgoyal:
I have fixed the issues you pointed out slightly differently.
I dont see how PATH duplication is possible in zsh since it runs 
typeset -U path after running conda.

The failed to send resize signal is harmless, it jsut means the window
was resized before the child process actually started.

PR #4990, opened 2022-04-19
Docs: Show copy button for installation command snippets

page-down:
The `parsed-literal` in the document does not display the copy button.
Put the install command snippets in `code-block` with string substitution.

I know little about Sphinx. Please review, thanks.


page-down:
Sorry, I see pyflakes reporting code style issues that have not yet had a chance to fix.

Issue #4987, opened 2022-04-18
`tab_switch_strategy` can act weird when closing tabs created with `--keep-focus`


salmankhilji:
**Describe the bug**
If I create tabs using the remote-control command `@ launch --keep-focus`, then closing each tab switches focus in a somewhat confusing order.

**To Reproduce**
1. Launch kitty:
  ```
  kitty --config NONE -o remember_window_size=no -o allow_remote_control=yes -o initial_window_width=90c bash --norc
  ```
2. Create some tabs:
  ```
  for i in {A..F}; do kitty @ launch --type=tab --tab-title=$i --keep-focus; done
  ```
4. Repeatedly close all tabs using `CTRL+SHIFT+w` and observe the order of the focused tab at each step.

**Observed Behavior**

(The tab with focus is **Bold**.)

**~** A B C D E F
A B C D E **F**
**A** B C D E
B C D **E**
**B** C D
C **D**
**C**

It looks like the focus alternates between the first and the last tab as I close each tab.

The issue here is with the meaning of `previous` when you do not really switch to a tab while creating it.

I'd say `--keep-focus` should act as if the user had switched to the tab, so that focused tab should be in the reverse creation order.

**Environment details**
```
kitty 0.25.0 (b569c01b49) created by Kovid Goyal
Linux byteevo 5.13.0-39-generic #44~20.04.1-Ubuntu SMP Thu Mar 24 16:43:35 UTC 2022 x86_64
Ubuntu 20.04.4 LTS byteevo /dev/tty

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.4 LTS"
Running under: X11
Frozen: False
Paths:
  kitty: /home/precor/Developer/kitty/kitty/launcher/kitty
  base dir: /home/precor/Developer/kitty
  extensions dir: /home/precor/Developer/kitty/kitty
  system shell: /bin/bash
Loaded config overrides:
  remember_window_size no
  allow_remote_control yes
  initial_window_width 90c

Config options different from defaults:
allow_remote_control y
initial_window_width (90, 'cells')
remember_window_size False

Important environment variables seen by the kitty process:
[...]

PR #4993, opened 2022-04-21
Avoid adding the last tab to history when creating new tabs

page-down:
When a new tab is created, the last tab id is first recorded in the history, and then the location is checked and focused again.
This causes the right-most tab tab to be focused by default when closing tabs, instead of the previous tab.

https://github.com/kovidgoyal/kitty/issues/4987#issuecomment-1103462759

Please review, thanks.


page-down:
This PR is not ready, I think it breaks `kitty -o tab_switch_strategy=last`.
If you have a better idea, or I need to check again for more.

salmankhilji:
Looks good now.  I can't seem to break it.

PR #5022, opened 2022-04-27
Docs: Improve the documentation for kittens

page-down:
Add more text roles and links.
Add an example that broadcasts only to other windows in the current tab.
Initial capitalization of the key names in the kbd text role.
Add Python type hints for custom kittens.
Note about hyperlink support for ls on macOS.
Add description text for show_key.

I have walked through all of the kittens' docs and have checked this PR several times, but there may be improper changes.

Please review, thanks.

---

May I ask what the current document column width should be kept at? Because I am still using 80, however I see that you seem to be using 100+ and I would like to try to keep the style consistent.

How to keep the cursor blinking in other windows while using broadcast kitten? Currently it stops after a few seconds and I don't know where the cursor is. And it stops blinking after switching tabs.

Is the `adjust_line_height` in `kitty/utils.py` no longer needed?


page-down:
> See TIMEOUT_FOR_SESSION in send_text.py

Thanks, I'm OK with not blinking, but at least showing the last position, otherwise it might be a bit inconvenient.

I see what can be done from here.

kovidgoyal:
On Wed, Apr 27, 2022 at 08:06:00AM -0700, page-down wrote:
> > See TIMEOUT_FOR_SESSION in send_text.py
> 
> Thanks, I'm OK with not blinking, but at least showing the last position, otherwise it might be a bit inconvenient.

Just type a letter and backspace it to resume blinking.

PR #5031, opened 2022-04-28
Docs: Fix the help text of query_terminal and improve rc docs

page-down:
Fix the help text of query_terminal and other minor tweaks
- Use --match-tab in broadcast kitten example.

Improve remote control command documentation
- Improve the --match option description.
- Add more text roles and links.
- Fix typos.

Also fixed the remote control example, because in the new version the shell integration sets the running command to the title, so I changed it to match only the beginning of the title, which would otherwise cause the current window to be matched.

Please review, thanks.

---

Is it possible for the command to return an error code if the `--match` does not match any window, instead of falling back to the current window?
Otherwise the command can be applied to the undesired window (current window).

`goto_layout` and `set_enabled_layouts`:
`You can use special match value all to set the layout in all tabs.`

What is the `all` special value? I'm thinking of adding the `--all` option to goto, but should it be added to `--match`?

### DIFF
--- a/docs/kittens/broadcast.rst
+++ b/docs/kittens/broadcast.rst
@@ -17,7 +17,7 @@ You can use the options described below to control which windows are selected.
 
 For example, only broadcast to other windows in the current tab::
 
-    map f1 launch --allow-remote-control kitty +kitten broadcast --match state:parent_active
+    map f1 launch --allow-remote-control kitty +kitten broadcast --match-tab state:focused
 
 .. program:: kitty +kitten broadcast
 
--- a/docs/remote-control.rst
+++ b/docs/remote-control.rst
@@ -4,8 +4,11 @@ Control kitty from scripts
 .. highlight:: sh
 
 |kitty| can be controlled from scripts or the shell prompt. You can open new
-windows, send arbitrary text input to any window, name windows and tabs, etc.
+windows, send arbitrary text input to any window, change the title of windows
+and tabs, etc.
+
 Let's walk through a few examples of controlling |kitty|.
+
 
 Tutorial
 ------------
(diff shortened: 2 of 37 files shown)


kovidgoyal:
On Thu, Apr 28, 2022 at 10:58:51AM -0700, page-down wrote:
> Fix the help text of query_terminal and other minor tweaks
> - Use --match-tab in broadcast kitten example.
> 
> Improve remote control command documentation
> - Improve the --match option description.
> - Add more text roles and links.
> - Fix typos.
> 
> Also fixed the remote control example, because in the new version the shell integration sets the running command to the title, so I changed it to match only the beginning of the title, which would otherwise cause the current window to be matched.
> 
> Please review, thanks.
> 
> ---
> 
> Is it possible for the command to return an error code if the `--match` does not match any window, instead of falling back to the current window?
> Otherwise the command can be applied to the undesired window (current window).

I dont mind personally, but this is both backwards incompatible and quite a lot of
code changes required. Maybe add an extra option like --only-matched to
change the behavior? Or maybe the backwards incompat is not too bad
since the behavior of matching when --match fails is not currently
specified.

> 
> `goto_layout` and `set_enabled_layouts`:
> `You can use special match value all to set the layout in all tabs.`
> 
> What is the `all` special value? I&#39;m thinking of adding the `--all` option to goto, but should it be added to `--match`?

If you use --match=all it results in all tabs being matched. What is
goto?

page-down:
> ... since the behavior of matching when --match fails is not currently specified

I don't have a very strong need for this, however in the process of checking the remote commands I found it worth doing something for the no-match case.

Also, the previous `--match` was simple, and with the enhanced filtering capabilities, something like `--dry-run` might be useful for debugging (list what is matched).

> If you use --match=all it results in all tabs being matched.

It looks like kitty doesn't accept this value right now.

```text
# 
...
  File ".../kitty/search_query_parser.py", line 259, in base_token
    raise ParseException(f'No location specified before {tt}')
kitty.search_query_parser.ParseException: No location specified before all
```

I see that other commands are using `--all`, maybe it would be good to unify to `--match`.

> What is goto?

Typo. `goto-layout` (and others rc cmds that support matching all windows or tabs)

kovidgoyal:
On Fri, Apr 29, 2022 at 04:10:51AM -0700, page-down wrote:
> > ... since the behavior of matching when --match fails is not currently specified
> 
> I don't have a very strong need for this, however in the process of checking the remote commands I found it worth doing something for the no-match case.

I am OK with changing it, but I have higher priority things to do, PRs
welcome.

> 
> Also, the previous `--match` was simple, and with the enhanced filtering capabilities, something like `--dry-run` might be useful for debugging (list what is matched).

You can just do some harmless operation for example
to test tab matching use select-window
TO test window matching use focus-window or set-window-logo/title, etc.

> 
> > If you use --match=all it results in all tabs being matched.
> 
> It looks like kitty doesn't accept this value right now.
> 
> ```text
> # 
> ...
>   File ".../kitty/search_query_parser.py", line 259, in base_token
>     raise ParseException(f'No location specified before {tt}')
> kitty.search_query_parser.ParseException: No location specified before all
> ```
> 
> I see that other commands are using `--all`, maybe it would be good to unify to `--match`.

yes, one could do that but it would break backwards compat.

kovidgoyal:
And looking at the code, when --match is specified and no matching windows are found an error is raised. For example:

```
kitty @ close-window --match id:11                                                                                                            
No matching windows for expression: id:11
```

page-down:
>> ... return an error code if the --match does not match any window, instead of falling back to the current window ...
>
> ... when --match is specified and no matching windows are found an error is raised ...

Sorry, I recall what this is all about.
`--match title:not-exists` will match the window itself due to shell integration, exactly why I updated the documentation.
Also `--exclude-active` is already available.

So now that leaves one last problem.
Currently `kitty @ send-text --match id:99` returns success when there is no match, and it looks like adding error logs with return codes here does not break backward compatibility.

kovidgoyal:
On Sat, Apr 30, 2022 at 06:54:53AM -0700, page-down wrote:
> >> ... return an error code if the --match does not match any window, instead of falling back to the current window ...
> >
> > ... when --match is specified and no matching windows are found an error is raised ...
> 
> Sorry, I recall what this is all about.
> `--match title:not-exists` will match the window itself due to shell integration, exactly why I updated the documentation.
> Also `--exclude-active` is already available.
> 
> So now that leaves one last problem.
> Currently `kitty @ send-text --match id:99` returns success when there is no match, and it looks like adding error logs with return codes here does not break backward compatibility.

Sure, go ahead.

PR #5042, opened 2022-04-30
Show error message when remote command arguments are invalid

page-down:
I'm not sure how a `send-text` without a match should be handled properly.
If we simply check for empty matches, does that affect the broadcast function?

It seems that broadcast can first match an empty range, e.g. a focused tab with no windows (which should not report an error at this point), and later works for newly opened windows in the focused tab.


kovidgoyal:
On Sat, Apr 30, 2022 at 07:15:43AM -0700, page-down wrote:
> I&#39;m not sure how a `send-text` without a match should be handled properly.
> If we simply check for empty matches, does that affect the broadcast function?
> 
> It seems that broadcast can first match an empty range, e.g. a focused tab with no windows (which should not report an error at this point), and later works for newly opened windows in the focused tab.

IIRC broadcast will update the windows it broadcasts to on every
iteration. You can return an error when nothing matches with send-text
unless in the broadcast case.

PR #5103, opened 2022-05-16
Add option select_by_word_characters_forward

### LINKED ISSUE #4728: Seperate select_by_word_characters for left and right of the cursor
I am used to select a filename and a line number via double-click e.g. in grep results or compiler warnings etc.. However those are very often separated by a colon but also followed by a colon.

Example: 
Screenshot

I than use that to open that file at the specific location in my editor (Vim with [vim-fetch](https://github.com/wsdjeg/vim-fetch)).

@mintty handles left and right characters differently: the colon is included in `select_by_word_characters` on the left of mouse pointer  but not on the right. So that when you double-click on the line number in the first line you'd get `kitty/rc/base.py:238` and if you double-click on the filename you only get `kitty/rc/base.py`

With kitty you would either get `238` (default) or `kitty/rc/base.py:238:def` (with colon in `select_by_word_characters`).

kgraefe:
This adds an option to select separate characters for forward and
backward word extension on double click. If it is empty the old behavior
is preserved. This is the default.

If it is not empty, select_by_word_characters_forward will be used for
extending the selection in forward direction (right) and
select_by_word_characters will be used for extending in backward
direction (left).

Resolves #4728


kgraefe:
My use case for that is so I can quickly select the filename and line number but not the column e.g. in `test.sh:12:45` by double-clicking on the line number. I use the following configuration for that (note the colon):
```
select_by_word_characters @-./_~?&=%+#:
select_by_word_characters_forward @-./_~?&=%+#
```

kovidgoyal:
I ma travelling for a few days I will review this when I am back.

kgraefe:
Great, thank you. :-)

PR #5142, opened 2022-05-28
Paste from finder files / folder selection is to show full path

jameshuynh:
## Description

Kitty does not have the capability to capture the full path of files & folders when a user press Cmd+C on finder and paste, i.e. Kitty only can paste the file name or folder name. On Iterm or Mac terminal, it can paste the escape full path.

This PR is to let Kitty have the same capability as Iterm or Mac terminal where the full path will be pasted instead of just the file name or folder name

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -2498,8 +2498,52 @@ const char* _glfwPlatformGetClipboardString(void)
         return NULL;
     }
 
+    NSArray *classes = [NSArray arrayWithObject:[NSURL class]];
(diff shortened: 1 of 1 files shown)


kovidgoyal:
1) You should use https://developer.apple.com/documentation/foundation/nsurl/1412925-filesystemrepresentation?language=objc  to convert urls to file system paths.

2) You are freeing _glfw.ns.clipboardString but its not guaranteed you are setting it to something else. It should only be freed if being set. Or if you are freeing it, then set it to null.

PR #5149, opened 2022-05-30
Add ability to use custom kitten dir in config dir 

jimbob3806:
### Purpose
This change adds the ability for users to structure their custom kittens in the same way as built in kittens are structured inside a "./kittens" subdirectory, rather than polluting the root directory of their kitty config with multiple custom kitten files.
### Current options
The current resolution of custom kittens implies that users must store an entry point (of one form or another) to the custom kitten in the root directory of their kitty config:
* for simple custom kittens, users must have an entry point at "/CONFIG_DIR/CUSTOM_KITTEN_NAME.py" for example "~/.config/kitty/my_custom_kitten.py"
* for more complex custom kittens, users may adopt some different approaches 
    * pollute the top level root config directory with multiple custom kitten files
    * create a subdirectory for code relating to custom kittens, and use the entry point in the root config directory to import the required files
    * create a subdirectory for custom kittens similar to that found for built-in kittens in the kitty repo, and utilise symlinks to reference the custom kitten entry point from the root config directory
### Proposed addition
Allow users to better organise their custom kittens by storing them in a "./kittens" subdirectory as is used in the kitty repo to define built-in kittens (https://github.com/kovidgoyal/kitty/tree/master/kittens)

To use custom kittens structured in this way, users can:
* create an entry point for the custom kitten at "/CONFIG_DIR/kitten/CUSTOM_KITTEN_NAME/main.py" for example "~/.config/kitty/kittens/my_custom_kitten/main.py"
* call custom kitten by prepending @ symbol to the name of the custom kitten, for example "kitty +kitten @my_custom_kitten"
* customs kittens should still be written as specified by the kitty docs (https://sw.kovidgoyal.net/kitty/kittens/custom/)

For example:
```
# initialise new directory structure for custom kittens
cd ~/.config/kitty
mkdir kittens
cd kittens

# make a subdirectory with the name of your custom kitten
mkdir my_custom_kitten
cd my_custom_kitten

# write custom kitten inside main.py
vim main.py
# alternatively move file(s) for existing custom kittens
mv ~/.config/kitty/my_custom_kitten.py ~/.config/kitty/kittens/my_custom_kitten/main.py

# run custom kitten by prepending @ to the kitten name
kitty +kitten @my_custom_kitten
```
### Remarks
* This change requires users to prepend an @ symbol to the kitten name when calling custom kittens formatted in this way, as such users with custom kittens with entry points at "/CONFIG_DIR/CUSTOM_KITTEN_NAME.py" will not be affected, and may continue calling  their custom kittens in the normal manner, for example "kitty +kitten my_custom_kitten"
* This change allows the creation and running of custom kittens that share their name with existing built-in kittens provided that the custom kitten is formatted using the proposed directory structure, allowing users to potentially add to, change, or override behaviour of built-in kittens by making their own modifications, but maintaining the same name

### DIFF
--- a/kittens/runner.py
+++ b/kittens/runner.py
@@ -28,6 +28,8 @@ def resolved_kitten(k: str) -> str:
 
 
 def path_to_custom_kitten(config_dir: str, kitten: str) -> str:
+    if kitten.startswith('@'):
+        kitten = 'kittens/' + kitten.split('.')[0][1:] + '/main.py'
     path = resolve_abs_or_config_path(kitten, conf_dir=config_dir)
     return os.path.abspath(path)
 


kovidgoyal:
There should be no need for this. You can simply put kittens in
subdirectories of the config directory already. For example:


map f1 kitten mykitten/main.py

and create the kitten in ~/.config/kitty/mykitten/main.py

jimbob3806:
Mapping kittens like this is fine for interactive kittens, but we still have no option for calling kittens with cli arguments, flags etc. For example if I structure my custom kitten into subdirectories, I cannot call it like this "kitty +kitten my_custom_kitten --some-option value". Allowing for an @ prefix (or some other character) lets interactive and non-interactive custom kittens (that have been placed in subdirectories) be called with command line arguments

kovidgoyal:
kitty +kitten mykitten/mykitten.py

works fine

jimbob3806:
Yes, except that it is twice as long as it needs to be, and is not consistent with calling either built-in kittens  nor with calling custom kittens that are in the root of the config directory, for example "kitty +kitten themes" (built-in) "kitty +kitten my_root_kitten" (entry point at "~/.config/kitty/my_root_kitten") however "kitty +kitten kittens/my_custom_kitten/main.py" if we want to use subdirectories

kovidgoyal:
On Mon, May 30, 2022 at 09:24:01AM -0700, James Reid wrote:
> Yes, except that it is twice as long as it needs to be, and is not consistent with calling either built-in kittens  nor with calling custom kittens that are in the root of the config directory, for example "kitty +kitten themes" (built-in) "kitty +kitten my_root_kitten" (entry point at "~/.config/kitty/my_root_kitten") however "kitty +kitten kittens/my_custom_kitten/main.py" if we want to use subdirectories

No you just need

kitty +kitten my_custom_kitten/main.py

unless you want to shove all your kittens into a "kittens" subdirectory
for some reason, and if you want to do that I recommend using an alias.

alias kk="kitty +kitten"
alias kc="kitty +kitten kittens/"

And if typing main.py is too much name it m.py instead, which is an
extra two characters.

kovidgoyal:
Also I should note that you dont really need to use the kittens framework for standalone kittens, you can just do:

kitty +launch mypythonprogram.py

which will run mypythonptogram.py where you can do whatever you like including using kitty's kittens framework but without actually creating a kitten.

PR #5171, opened 2022-06-04
diff kitten: add options to ignore paths when comparing directories

suvayu:
I could not find a way to ignore certain paths while trying to compare directories with the `diff` kitten.  So this is a first pass at implementing it.

This PR adds two config options that are space separated list of patterns:
- `ignore_paths`: patterns that are matched with directory names
- `file_ignores`: patterns that are matched with filenames in a directory

The patterns are constructed as below, and passed on to `fnmatch.fnmatch`.
- directories: `*/<pattern>`
- files: `<pattern>`

I have lightly tested this locally & over SSH:

    $ kitty +kitten diff /local/path /local/path2
    $ kitty +kitten diff /local/path ssh:remote:/path

Example config:
```
file_ignores *~ \#*\#
ignore_paths .git .ipynb* .*
```


kovidgoyal:
Some comments:

1) Why do we need separate file and directory patterns? Why not just have one setting for both. That setting should be applied to the full path of every dir and file.

2) You should remove the directories from dirnames, this prevents them from being scanned.

3) Move walk into a top-level function and add tests for it in kitty_tests/diff.py

4) use shlex.split for pattern_list() otherwise one cant have patterns with spaces in them

5) You shouldnt use mutable containers for class variables. Use a tuple. And why do the default ones have an empty string in them? 

6) removing empty strings from the settings lists should be done only once

3)

suvayu:
Thanks for your comments!

> 1. Why do we need separate file and directory patterns? Why not just have one setting for both. That setting should be applied to the full path of every dir and file.

Someone might want to ignore files and directories differently; for example, I usually want to ignore almost all hidden directories like: `.git`, `.*cache`, but want to see changes to `.gitignore`, or `.gitattribute`.  However, upon more thought, you are right, one config can work with a more explicit set of patterns.

> 2. You should remove the directories from dirnames, this prevents them from being scanned.

Done

> 3. Move walk into a top-level function and add tests for it in kitty_tests/diff.py

Done

> 4. use shlex.split for pattern_list() otherwise one cant have patterns with spaces in them

Done

> 5. You shouldnt use mutable containers for class variables. Use a tuple. And why do the default ones have an empty string in them?

Done, I replaced `['']` with `()`

> 6. removing empty strings from the settings lists should be done only once

Redundant now

PR #5196, opened 2022-06-14
Enable resizing in session files.

sirno:
This PR enables resizing windows in session files.

It is a bit of a hack because semantically I overload the window list in the session object.

Feel free to request changes.

### DIFF
--- a/kitty/session.py
+++ b/kitty/session.py
@@ -3,7 +3,7 @@
 
 import shlex
 import sys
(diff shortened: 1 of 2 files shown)


sirno:
Yes that is indeed much clearer.

I adjusted the code to do this, however currently it will crash when `resize_window` is called before a window is launched. do you think there should be an error message for such a case?

kovidgoyal:
Something provided before the first window should apply to the first window. That's how all the rest of the settings work, IIRC.

PR #5199, opened 2022-06-16
Add `tab_bar_always_update` option.

sirno:
Hi,

I added a `tab_bar_always_update` option, that updates the tab bar whenever the os window is redrawn. This obviously has a negative performance impact when set, but it allows to render things that update independent ot the tab state, such as a clock.

Unfortunately, I was unable to run `gen-config.py` because of a sigsegv when importing `fast_data_types` that I didn't resolve in reasonable time. Instead, I manually edited the option files in line with what would have been generated

### DIFF
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -619,7 +619,7 @@ prepare_to_render_os_window(OSWindow *os_window, monotonic_t now, unsigned int *
     bool needs_render = os_window->needs_render;
     os_window->needs_render = false;
     if (TD.screen && os_window->num_tabs >= OPT(tab_bar_min_tabs)) {
-        if (!os_window->tab_bar_data_updated) {
+        if (OPT(tab_bar_always_update) || !os_window->tab_bar_data_updated) {
             call_boss(update_tab_bar_data, "K", os_window->id);
             os_window->tab_bar_data_updated = true;
         }
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -1231,6 +1231,13 @@ Color for the tab bar margin area. Defaults to using the terminal background
 color.
 '''
     )
+
+opt('tab_bar_always_update', 'no',
+    option_type='to_bool', ctype='bool',
+    long_text='''
+Update tab bar whenever os window is rerendered.
+'''
+    )
 egr()  # }}}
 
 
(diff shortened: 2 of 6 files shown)


kovidgoyal:
If you want a clock why not just use add_timer as described here: https://github.com/kovidgoyal/kitty/discussions/4447#discussioncomment-1940795

sirno:
Because I didn't see that nor think of it. Yet this will effectively cause the same behaviour, because at every update the tab_bar will be marked dirty, won't it?

kovidgoyal:
No not at every update at every timer expiry. Every update can happen at
arbitrary frequencies.

sirno:
@kovidgoyal Surprisingly the callback does not behave the same although in some way. Marking the tab bar dirty will cause the bar to update when there is some input in any window. The option in this PR however will cause the bar to update without any input. This seems quite confusing to me, as both approaches will affect the same conditional in `child-monitor.c`.

kovidgoyal:
I dont see how that's possible. Are you sure? Note that rendering and therefore tab bar updates are limited by the repaint_delay option (see the start of the render() function. The render function will also wait for render frames from the operating system. So the timing may not be too accurate. And finally note that needs_render has to be true for actual rendering to take place. So in your timer callback you might need to set that to true by dirtying one of the visible windows as well.

sirno:
Yes I also don't see how it's possible, but it is what I observe here. I already tried to dirty the os window, however it does not solve the issue. Note in this case that the Opt introduced in this PR has no influence on the state of the os window either, yet frequently rerenders.

It is actually not a big problem for me as of now, but it seems so weird that I wanted to report. If I stay curious I will keep investigating.

sirno:
Ok this happens because on mac the run loop will suspend very deliberately when nothing is going on even when the window is focused. Adding expensive python computations such as tab bar updates keeps the run loop busy and it will not suspend. There are several solutions to this.

1. Call `wakeup_main_loop` when marking things as dirty when compiling for mac os
2. Wake up whenever a timer callback runs
3. Expose `wakeup_main_loop` in the python api

kovidgoyal:
It already is, look at the wakeup() function in constants.py

kovidgoyal:
feel free to send a PR exposing wakup_main_loop as well

PR #5212, opened 2022-06-19
Fix the rc command new-window does not open the shell by default

page-down:
Since `use_shell=False` is hardcoded in tabs.py, Tab, `new_special_window`, when kitty is started with arguments and the remote control command new-window is used, it will default to running the program specified by the arguments at startup instead of the default shell.

```shell
kitty --config=NONE --listen-on=unix:/tmp/mykitty -o allow_remote_control=yes vim
kitty @ --to unix:/tmp/mykitty new-window
# vim opens in a new window
```

Launch is used in this PR to implement the new-window rc function.
Maybe you have a better idea.

I also wonder if we should deprecate `new-window` in the documentation and instead recommend using `launch`.
This rc command is left for backward compatibility.

Please review, thanks.

### DIFF
--- a/kitty/rc/new_window.py
+++ b/kitty/rc/new_window.py
@@ -3,9 +3,6 @@
 
 from typing import TYPE_CHECKING, Optional
 
-from kitty.fast_data_types import focus_os_window
-from kitty.tabs import SpecialWindow
-
 from .base import (
     MATCH_TAB_OPTION, ArgsType, Boss, PayloadGetType, PayloadType, RCOptions,
     RemoteCommand, ResponseType, Window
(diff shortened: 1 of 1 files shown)


page-down:
I see the following errors when running mypy under macOS.
Haven't followed up on the recent `prewarm` things, but noticed that one of the commits mentioned that macOS unfortunately does not have waitid.
Perhaps these errors can be suppressed?

```
kitty/prewarm.py:56: error: Module has no attribute "waitid"; maybe "waitpid" or "wait"?
                    x = os.waitid(os.P_PID, child_pid, os.WEXITED | os.WNOHANG)
                        ^
kitty/prewarm.py:61: error: Returning Any from function declared to return "Optional[int]"
                        return x.si_status
                        ^
```

PR #5231, opened 2022-06-25
MacOS: allow non-1 background opacity with custom titlebar color

nguyenvukhang:
Previously, if the `background_opacity` took any value less than 1:
1. The window will have no shadow
2. The titlebar will be coerced to the default color (`macos_titlebar_color` will be overridden by an internal config)

With this PR, you should be able to set both a custom `macos_titlebar_color` and enjoy a non-1 `background-opacity`, and on top of that have the shadow.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-LDFLAGS=-L/opt/homebrew/lib \
-    python3 setup.py kitty.app \
-    --extra-include-dirs /opt/homebrew/Cellar/librsync/2.3.2/include
-rm -rf /Applications/kitty.app
-cp -r kitty.app /Applications/kitty.app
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I dont follow how this is supposed to work. There is no way to get macos
to render the titlebar with a custom color. So what the existing code
does is get macos to render it transparent and instead  sets the window
background color to the titlebar color. Your patch will as far as I can
see result in the titlebar being the native macos color.

nguyenvukhang:
This was mostly a result of trial-and-error with a few other permutations of the changed configs.

Here's my build result on macOS Monterey 12.4:

Screen Shot 2022-06-25 at 16 33 08

nguyenvukhang:
Though, now that I check again, changing `macos_titlebar_color` has no effect at all.

PR #5239, opened 2022-06-28
Minor fixes, refactorings and docs improvements to recent features

page-down:
- In a recent PR, several diff kitten files were added with executable permissions and are now cleaned up.
- Some documentation improvements for remote commands.
- ssh kitten: Since the path to homebrew is added to the common locations for login shell, the path to MacPorts is also added.
- Fix edit-in-kitty not working when option editor comes with arguments. E.g. `code --wait` (wait until quit)
- fish completion: Avoid overwriting different functions with the same name.
  * After my testing, it doesn't look like there' s a need to specify names like `edit-in-kitty` in `__ksi_completions`. Please let me know if I'm wrong.
- Refactor the fish implementation of edit-in-kitty.
  * Clean up global variable and signal handling function before exit.
  * I think very few users take `head` or `stty` executable names in fish shell. Valid functions are allowed by default. We can't (not worth it) handle all kinds of cases.

Please review, thanks.

I couldn't find a discussion thread for `edit-in-kitty`. So I'm asking an unresolved issue here:
I can't use `edit-in-kitty` with the default configuration under macOS (APFS file system).

After tracing, with vim's default configuration: `nobackup`, `backupcopy=auto`, `writebackup`, the local temporary file inode will change after saving the file, while the original inode modification time will never change and the file contents will not be updated.
Other editors that save files by writing and renaming might have the same problem.

I can solve this by checking the file modification time by path.
Maybe you have a better solution for this.


kovidgoyal:
On Mon, Jun 27, 2022 at 09:29:52PM -0700, page-down wrote:
> I can solve this by checking the file modification time by path.
> Maybe you have a better solution for this.

I dont see how there can be a batter solution. If editors use the rename
trick for atomic writes, then the inode cannot be used for tracking.

PR #5265, opened 2022-07-12
feat(options): add modify_font option

akinsho:
This fixes the issue of not being able to modify a font attributes, namely the `underline_thickness`, `strikethrough_position`, `underline_position` and (eventually) `size`. It supersedes #4723.

# To-do:
- [x] add validation of the arguments for `modify_font`
- [x] Handle specifying postscript font name
- [x] fix handling of `*_position` as neither currently move the line down nor up.
- [ ] Handle size... (not entirely sure where to start with this)
	- [ ] Units are specified as `int` but maybe should accept `%` for `size`
- [x] Fix bug with strikethrough positioning

## Notes for review:
I'm largely new to python and c so any formatting issues or method usages etc. feel free to correct as I've just focused on getting something working and don't know what the best practices are.

### DIFF
--- a/kitty/core_text.m
+++ b/kitty/core_text.m
@@ -332,7 +332,7 @@ adjust_ypos(unsigned int pos, unsigned int cell_height, int adjustment) {
 }
 
 void
-cell_metrics(PyObject *s, unsigned int* cell_width, unsigned int* cell_height, unsigned int* baseline, unsigned int* underline_position, unsigned int* underline_thickness, unsigned int* strikethrough_position, unsigned int* strikethrough_thickness) {
+cell_metrics(PyObject *s, unsigned int* cell_width, unsigned int* cell_height, unsigned int* baseline, unsigned int* underline_position, unsigned int* underline_thickness, unsigned int* strikethrough_position, unsigned int* strikethrough_thickness, unsigned int* modified_thickness, unsigned int* modified_underline_y, unsigned int* modified_strikethrough_y) {
     // See https://developer.apple.com/library/content/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/TypoFeatures/TextSystemFeatures.html
     CTFace *self = (CTFace*)s;
 #define count (128 - 32)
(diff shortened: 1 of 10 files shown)


akinsho:
@kovidgoyal would be good to get an early sense check on this PR i.e. the general approach, code style vs conventions etc. This essentially works (give or take some outstanding bug with unsigned ints and strike-through positioning) locally for the use case I described in the original comment I made, i.e. I'm able to change the thickness and height of the underline etc.

I haven't done anything about the size of fonts yet since I'm less sure about how that functionality should work I guess somewhere in the codebase the cell size of each font including fallbacks is being determined, and I need to get the config value over to there. I appreciate you included this in the required feature set but presumably if a general mechanism for specifying font overrides in the config and them flowing through is done then this could maybe be added without needed a refactor i.e. at that point it's just an additional feature

kovidgoyal:
I'll review it when I have some time. cell sizes are calculated only for
the main font. Size adjustment is intended to apply to
fallback/alternate faces anyway, so it doesnt affect cell size
calculation, for which there are already other options. 

baseline adjustment probably needs to be per font face as well.
underline and strikethrough adjustment apply only to the main font.

akinsho:
@kovidgoyal thanks for clarifying so if I understand correctly based on 
> cell sizes are calculated only for
The main font. Size adjustment is intended to apply to
fallback/alternate faces anyway

Cell size is calculated for only the main ("medium") face here https://github.com/kovidgoyal/kitty/blob/b0666c91801a27d359fe61c2c16f0fe0b99a44dd/kitty/fonts.c#L338
and currently all the other fonts inherit this calculation?

So when you say size, since you've clarified you don't mean cell size what dimension specifically do you mean you've mentioned the baseline but the references I've found for this are for the main font and are acted upon by other options, so do you mean altering the baseline for non-main fonts similar to how it is altered in `cell_metrics`, presumably alternate fonts can be set higher or lower in the cell separate to the main font i.e. that's the desired result.

The understanding I have now is that maybe you mean that rather than have only the medium font have its baseline adjusted, allow adjusting only the baseline for the alternate fonts in a similar way as just the main font is right now?

Apologies if you're having to go over this a few times, I've honestly never dealt with font rendering per se. I don't think I even know what the outcome of setting adjust size is supposed to actually *look like*, are these fonts now supposed to be higher or lower in the cell or actually smaller or larger within the cell (I didn't/don't even know if that's possible)

FWIW the rest of it now works, i.e. strikethrough/underline positions and thickness

kovidgoyal:
adjust_font_size should adjust the actual pt size used to render
fallback/alternate fonts into the cell. The cell's size is fixed by the
point size of the main font, which comes from the font_size setting.

Currently what happens is the alternate font's size is set to the same
size as font_size and it is rendered. If the glyph does not fit in the
cell, the size is reduced until it does and the baseline is changed to
match. So with adjust_size the initial
size that the alternate font is rendered at would be adjusted.

And yes changing of baseline can either be automatic as it is currently
when resizing an alternate font or manual if specified.

akinsho:
Thanks for implementing this @kovidgoyal 🙏🏿 and for your patience with my attempt 👍🏿

kovidgoyal:
You're welcome!

PR #5277, opened 2022-07-14
macOS: Fix unable to open new tab/window when there is no OS window

### LINKED ISSUE #5276: MacOS: Open a window if none are open when opening a new tab.
In iTerm, if I close my last window, then hit cmd+t (new tab), it opens a window for me (with 1 tab in it). In kitty, since no windows are open, hitting cmd+t does nothing (as far as I can tell). Obviously I can open a new window, but I tend to just nuke all of my windows then open a few tabs with a quick bit of cmd+t spam. I am aware of how small and unimportant this is in the big scheme of things, just a small convenience thing.

Ideally, opening a new window when you hit cmd+t with no windows open

I'm happy to _try_ to submit a PR for this, I just wanted to verify that there isn't some different way things work/are viewed in kitty that would in general make this undesirable.

page-down:
Fixes #5276

I hope all cases have been covered, please review, thanks.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -49,6 +49,8 @@ Detailed list of changes
 
 - macOS: Remote control: Fix unable to launch a new OS window or background process when there is no OS window (:iss:`5210`)
 
+- macOS: Fix unable to open new tab or new window when there is no OS window (:iss:`5276`)
+
 - kitty @ set-colors: Fix changing inactive_tab_foreground not working (:iss:`5214`)
 
 - macOS: Fix a regression that caused switching keyboard input using Eisu and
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Why do we need the check in new_tab_with_wd? It just calls _new_tab
which already has the check. 

Also why are we returning None in rc.launch for tab/window launches?

page-down:
> Why do we need the check in new_tab_with_wd? ...

I don't understand, I moved the code that was previously in `new_tab_with_wd` to `_new_tab`.

> Also why are we returning None in rc.launch for tab/window launches?

Am I missing something? Currently `kitty @ launch --type=tab/window` returns id, not None.

kovidgoyal:
ah never mind I accidentally read your PR against a different branch :)

PR #5316, opened 2022-07-26
Convert TabCWD into str so we can use string method on it

nxvxl:


kovidgoyal:
The whole point of using TabCWD is so that the cwd is queries only when it is actually needed in a template, by pre-converting it that optimisation is removed use ```{str(cwd)}``` in your template instead.

kovidgoyal:
Actually I decided to implement it more flexibly https://github.com/kovidgoyal/kitty/commit/352c07987b2a39088a2d010a348bbcd743d8a25a

nxvxl:
Thank you for the feedback.

I tried your suggestions but got an error
```
[207 14:06:09.627658] Invalid tab title template: "{index}: {str(cwd).split('/')[-1]}" with error: name 'str' is not defined
```
I only want last part of the path.
Any other advice?

nxvxl:
Cool, thanks!

Issue #5357, opened 2022-08-06
Speak selection on macOs always reads the first word of the last line


blubber:
**Describe the bug**
macOs features an accessibility option known as 'Speak Selection', it works by selecting a portion of text followed by pressing a hotkey (default is Option-Esc), the macOs tts engine is used to read the selected text aloud. In Kitty this doesn't work properly, it always reads the first word of the current line. For example, my prompt featuers the current working directory as the first item, which results in it always reading the cwd aloud.

It seems that the selection is not properly conveyed to the OS.

**To Reproduce**
1. Go to System Preferences > Accessibility > Spoken content
2. Enable 'Speak selection'.
3. Select some text (other than the first word in your prompt) in Kitty and press the Speak selection hotkey (defaults to Option-Esc).
4. It will speak the first word of the prompt on the current line, instead of the selected text.

**Environment details**
```
kitty 0.25.2 created by Kovid Goyal
Darwin macbookpro.local 21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:22 PDT 2022; root:xnu-8020.140.41~1/RELEASE_ARM64_T6000 arm64
ProductName:	macOS ProductVersion:	12.5 BuildVersion:	21G72
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /bin/zsh
Loaded config files:
  /Users/***/.config/kitty/kitty.conf

Config options different from defaults:

Important environment variables seen by the kitty process:
	PATH                                /Applications/kitty.app/Contents/MacOS:/usr/bin:/bin:/usr/sbin:/sbin
	LANG                                en_US.UTF-8
	SHELL                               /bin/zsh
	USER                                ***

```

**Additional context**
I also tried this with other shells than ZSH, it results in the same issue. For instance in bash, my prompt starts with `bash-3.2`. The speak selection voice will read 'bash` regardless of the actual selection.

kovidgoyal:
I'm afraid I dont care about this, it will require some special purpose
cocoa api or the other and given I dont use apple products the effort
involved is too large for the end result. But patches are welcome. If
you or anyone else wants to implement this feel free to comment and I
will provide any needed guidance.

page-down:
[...]

PR #5359, opened 2022-08-08
Implement macOS accessibility protocol to provide selected text

page-down:
Allow "Speak selection" (Option+Esc) to work properly.
Thanks for the last building block.

Fixes https://github.com/kovidgoyal/kitty/issues/5357

Minor improvements for code style:
Use spaces as indentation consistently.

EDIT:
The improvements about indentation have been picked to the master branch.
https://github.com/kovidgoyal/kitty/commit/06108d66b14c78456268622be285e7e5b11feb12
https://github.com/kovidgoyal/kitty/commit/0d116e6ef0f176bec053907920dd9c3655113562

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1540,6 +1540,36 @@ void _glfwPlatformUpdateIMEState(_GLFWwindow *w, const GLFWIMEUpdateEvent *ev) {
     debug_key("\n\tdoCommandBySelector: (%s)\n", [NSStringFromSelector(selector) UTF8String]);
 }
 
+- (BOOL)isAccessibilityElement
+{
+    return YES;
+}
+
+- (BOOL)isAccessibilitySelectorAllowed:(SEL)selector
+{
+    if (selector == @selector(accessibilityRole) || selector == @selector(accessibilitySelectedText)) return YES;
+    return NO;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Shouldnt isAccessibilityFocused only return true if the OS window is actually focused?

kovidgoyal:
Oh and also, you need to free the string returned by the callback function.

blubber:
I can confirm that this works perfectly on macOs 12. 👍

page-down:
> Shouldnt isAccessibilityFocused only return true if the OS window is actually focused?

I think every windows has its own focused controls, and we only have one control and definitely have keyboard focus, not sure if that's the correct assumption.
However, this is not important for now, so let's leave it to the system.

> ... free the string returned by the callback function.

I have updated the PR.
Please review if this is the correct way.

@blubber
I'm glad it worked for you.
You may want to pull the latest code after this PR is merged.

page-down:
@kovidgoyal 

The latest changes resulted in a compilation error.
```text
glfw/cocoa_window.m:1566:18: error: passing 'const char *' to parameter of type 'void *' discards qualifiers ...
```

kovidgoyal:
Should be fine now.

PR #5362, opened 2022-08-09
Allow completion for show_key kitten

page-down:
Notes for future improvements:
Perhaps a separate documentation description could be added later for each option value (many values are too simple to be worth describing independently and require too much effort) and applied to the autocomplete.

Anyway, a clear list is better than a big block of plain documentation descriptions without paragraphs.

EDIT: The following is for configuration, not command parameters.
Many of the option values are currently listed in the function code of `kitty/options/utils.py`, which also makes it impossible to reuse them.
```python
not in ('left', 'right', 'top', 'bottom')
```

For example, `kitty +kitten icat --align`, pressing tab completes the files in the current folder, not the possible values of `align`.
EDIT: `--align=` This will list the available option values. It needs to recognize parameter value after space. (Syntax in the docs: `--align <ALIGN>`)

`kitty @ launch --logo`, pressing tab does not respond, and the file path cannot be completed.

This can be marked up in code using documentation or custom type hints, etc. (e.g. which parameters receive file paths), and the autocomplete feature can use this information to better process user input.


page-down:
@kovidgoyal 
Is it true that `U+2303` is on a particular keyboard layout?

This alias was added earlier but is missing here. Would you like me to remove all support for this alias?

https://github.com/kovidgoyal/kitty/pull/4275

kovidgoyal:
Ah, I didnt realize that was ⌃ and not ^

Issue #5360, opened 2022-08-08
Infinite wayland recursion between compositor and text-input-v3


wengxt:
**Describe the bug**
Trying to type something in kitty, but sees infinite message.
WAYLAND_DEBUG=1 kitty
Trying to type something with fcitx5

[3940771.262] zwp_text_input_v3@19.done(89081)
[3940771.265]  -> zwp_text_input_v3@19.set_cursor_rectangle(181, 26, 14, 21)
[3940771.268]  -> zwp_text_input_v3@19.commit()
[3940771.294] zwp_text_input_v3@19.preedit_string("ni", 0, 2)
[3940771.298] zwp_text_input_v3@19.done(89082)
[3940771.301]  -> zwp_text_input_v3@19.set_cursor_rectangle(181, 26, 14, 21)
[3940771.304]  -> zwp_text_input_v3@19.commit()
[3940771.330] zwp_text_input_v3@19.preedit_string("ni", 0, 2)
[3940771.334] zwp_text_input_v3@19.done(89083)
[3940771.337]  -> zwp_text_input_v3@19.set_cursor_rectangle(181, 26, 14, 21)
[3940771.341]  -> zwp_text_input_v3@19.commit()
[3940771.367] zwp_text_input_v3@19.preedit_string("ni", 0, 2)
[3940771.371] zwp_text_input_v3@19.done(89084)
[3940771.374]  -> zwp_text_input_v3@19.set_cursor_rectangle(181, 26, 14, 21)
[3940771.378]  -> zwp_text_input_v3@19.commit()
[3940771.405] zwp_text_input_v3@19.preedit_string("ni", 0, 2)
[3940771.409] zwp_text_input_v3@19.done(89085)
[3940771.412]  -> zwp_text_input_v3@19.set_cursor_rectangle(181, 26, 14, 21)
[3940771.415]  -> zwp_text_input_v3@19.commit()
[3940771.442] zwp_text_input_v3@19.preedit_string("ni", 0, 2)
[3940771.446] zwp_text_input_v3@19.done(89086)
[3940771.449]  -> zwp_text_input_v3@19.set_cursor_rectangle(181, 26, 14, 21)
[3940771.452]  -> zwp_text_input_v3@19.commit()
[3940771.479] zwp_text_input_v3@19.preedit_string("ni", 0, 2)
[3940771.483] zwp_text_input_v3@19.done(89087)
[...]

PR #5369, opened 2022-08-11
Fix message recursion between wayland compositor (#5360)

wengxt:
1. Only update preedit if preedit changed.
2. Only send cursor update if cursor position is not changed.

### DIFF
--- a/glfw/wl_text_input.c
+++ b/glfw/wl_text_input.c
@@ -9,12 +9,15 @@
 #include "internal.h"
 #include "wayland-text-input-unstable-v3-client-protocol.h"
 #include <stdlib.h>
+#include <string.h>
 #define debug(...) if (_glfw.hints.init.debugKeyboard) printf(__VA_ARGS__);
 
 static struct zwp_text_input_v3*                  text_input;
 static struct zwp_text_input_manager_v3*          text_input_manager;
 static char *pending_pre_edit = NULL;
+static char *current_pre_edit = NULL;
 static char *pending_commit   = NULL;
+static int last_cursor_left = 0, last_cursor_top = 0, last_cursor_width = 0, last_cursor_height = 0;
 uint32_t commit_serial = 0;
 
 static void commit(void) {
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I dont follow the need for this. kitty already doesnt send anything in response to done events.

wengxt:
> I dont follow the need for this. kitty already doesnt send anything in response to done events.

the reason is that it’s triggered by preedit update when preedit is same. not explicitly reply anything to done.

kovidgoyal:
On Thu, Aug 11, 2022 at 06:52:57PM -0700, Weng Xuetian wrote:
> > I dont follow the need for this. kitty already doesnt send anything in response to done events.
> 
> the reason is that it’s triggered by preedit update when preedit is same. not explicitly reply anything to done.

But does that actually cause an infinite recursion? And if so, why? And
why does it not show up under mutter?

wengxt:
> On Thu, Aug 11, 2022 at 06:52:57PM -0700, Weng Xuetian wrote: > I dont follow the need for this. kitty already doesnt send anything in response to done events. the reason is that it’s triggered by preedit update when preedit is same. not explicitly reply anything to done.
> But does that actually cause an infinite recursion? And if so, why? And why does it not show up under mutter?

Sorry, but I can easily reproduce it under mutter. You need to use an input method that has preedit.

wengxt:
Upload a video as proof.
https://user-images.githubusercontent.com/259684/196316515-e0509b43-b009-4919-9953-248d54ad31a3.mp4

kovidgoyal:
> > On Thu, Aug 11, 2022 at 06:52:57PM -0700, Weng Xuetian wrote: > I dont follow the need for this. kitty already doesnt send anything in response to done events. the reason is that it’s triggered by preedit update when preedit is same. not explicitly reply anything to done.
> > But does that actually cause an infinite recursion? And if so, why? And why does it not show up under mutter?
> 
> Sorry, but I can easily reproduce it under mutter. You need to use an input method that has preedit.

Then please specify how I can reproduce it. Which input method, and what kep presses to trigger it.

kovidgoyal:
Ok I'll take a look at this a little later, thanks.

PR #5404, opened 2022-08-22
Change default config file to put long text before commented-out example

jcs:
I found it confusing to read the default config file since some of the text is above a commented-out option, and some text is below it.  Every time I'd go into the config to search for something I wanted to change, I'd be confused that the option name it's referring to is above it rather than the one below it.  I think most other software's default config files have comments above a commented-out example.

Before:

```
# placement_strategy center

#: When the window size is not an exact multiple of the cell size, the
#: cell area of the terminal window will have some extra padding on
#: the sides. You can control how that padding is distributed with
#: this option. Using a value of center means the cell area will be
#: placed centrally. A value of top-left means the padding will be
#: only at the bottom and right edges.
```

After:
```
#: When the window size is not an exact multiple of the cell size, the
#: cell area of the terminal window will have some extra padding on
#: the sides. You can control how that padding is distributed with
#: this option. Using a value of center means the cell area will be
#: placed centrally. A value of top-left means the padding will be
#: only at the bottom and right edges.

# placement_strategy center
```

### DIFF
--- a/kitty/conf/types.py
+++ b/kitty/conf/types.py
@@ -236,6 +236,9 @@ class Option:
         a = ans.append
         if not self.documented:
             return ans
+        if self.long_text:
+            a(render_block(self.long_text))
+            a('')
         if option_group:
             sz = max(len(self.name), max(len(o.name) for o in option_group))
             a(f'{self.name.ljust(sz)} {self.defval_as_string}'.rstrip())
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This is entirely subjective. I prefer the current style, it puts the
option name front and center.

PR #5409, opened 2022-08-24
Some fixes and improvements to the documentation

page-down:
- Using the lowercase form of the key name consistently.
- Use kitty --help long option name in the document.
- Remove the extra empty line in the CLI help for `--replay-commands`.
- Fix GitHub links in commented configuration.
- Shorten the reference link to the top of the page. (E.g. `https://sw.kovidgoyal.net/kitty/shell-integration/`)
- Hyperlink doc text role.
- Fix doc website URL with duplicate slashes.
- Unify the list style of available values.

---

There are still some unresolved issues.

Text role: The line after the line break fails to be highlighted in color.
For example, see the `--color` description in `kitty @ launch --help`.
Not so sure the same problem exists with italics and other styles.

Can't get the document title, the `:doc:` link is weird.
See `kitty @ launch --help`, `--marker`: `... (see /marks)`

```text
--marker
...
... for the toggle_marker action (see /marks).

```
HTML version:
```text
... for the toggle_marker action (see Mark text on screen).
```

The path to check local documents is wrong under macOS.
`kitty/constants.py`, `local_docs()`
```python
if getattr(sys, 'kitty_run_data').get('from_source'):
    sq = os.path.join(d(base), 'docs', '_build', 'html')
    if os.path.isdir(sq):
        return sq
```

The following path does not exist.
```text
GIT_REPO/kitty/launcher/kitty.app/docs/_build/html
```

It looks like it needs to check if it is running in the source directory `kitty/launcher/kitty`, or run `kitty.app` generated by `make app`.

```text
../../../docs/_build/html/
kitty.app/Contents/Resources/doc/kitty/html/
```

### DIFF
--- a/docs/extract-rst-targets.py
+++ b/docs/extract-rst-targets.py
@@ -14,7 +14,7 @@ def find_explicit_targets(text: str) -> Iterator[str]:
 
 
 def main() -> Dict[str, str]:
-    refs = {'github_discussions': 'https://github.com/kovidgoyal/kitty/discussions'}
+    refs = {}
     base = os.path.dirname(os.path.abspath(__file__))
     for dirpath, dirnames, filenames in os.walk(base):
         if 'generated' in dirnames:
--- a/docs/extract-rst-targets.py
+++ b/docs/extract-rst-targets.py
@@ -25,7 +25,17 @@ def main() -> Dict[str, str]:
                     raw = stream.read()
                 href = os.path.relpath(stream.name, base).replace(os.sep, '/')
                 href = href.rpartition('.')[0] + '/'
+                first_line = raw.lstrip('\n').partition('\n')[0]
+                first_target_added = False
                 for explicit_target in find_explicit_targets(raw):
+                    # Shorten the reference link to the top of the page.
+                    # Note that anchor links should still be used in HTML docs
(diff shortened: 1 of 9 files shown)


page-down:
I confirm that the second and third mentioned issues have been fixed.
Regarding the first issue, you can see that the same affects hyperlink.

```text
    ... (see Mark text on
    screen).  <--- Missing hyperlink underline style
^-- Should not be clickable
```

Also the indented part (spaces) in front of the next line should not be used as hyperlink range.

kovidgoyal:
On Wed, Aug 24, 2022 at 02:04:11AM -0700, page-down wrote:
> I confirm that the second and third mentioned issues have been fixed.
> Regarding the first issue, you can see that the same affects hyperlink.
> 
> ```text
>     ... (see Mark text on
>     screen).  <--- Missing hyperlink underline style
> ^-- Should not be clickable
> ```

That's a bug in less where it resets styles at newlines. 

> 
> Also the indented part (spaces) in front of the next line should not be used as hyperlink range.

I'm afraid I am not interested in changing that, it makes wrapping code
much more complex, as the hyperlink has to be stopped and restarted on
wrap, but patches are welcome.

PR #5428, opened 2022-08-26
Add a --hyperlink-only-matches flag to hyperlinked_grep

groves:
If it's active, only matching lines get links, not file headers or context lines.

I have a kitten that steps through the links it finds to open them like "Find Next" when searching for text in a browser. It's way more useful if the links are only to the lines the search hits, which is what this flag enables.

### DIFF
--- a/docs/kittens/hyperlinked_grep.rst
+++ b/docs/kittens/hyperlinked_grep.rst
@@ -74,6 +74,10 @@ to :program:`rg`. How to do that varies based on the shell:
 To learn more about kitty's powerful framework for customizing URL click
 actions, see :doc:`here </open_actions>`.
 
+By default, this adds hyperlinks for several parts of ripgrep output: the 
+per-file header, match context lines, and match lines. Passing the 
+`--hyperlink-only-matches` flag will only add hyperlinks to the match lines.
+
 Hopefully, someday this functionality will make it into some `upstream grep
 <https://github.com/BurntSushi/ripgrep/issues/665>`__ program directly removing
 the need for this kitten.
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Add some docs for this to hyperlinked_grep.rst

page-down:
I would suggest adding a prefix to the kitten options, just like the `ssh` kitten.
Reduce the possibility of adding options with the same name from upstream.

It would be nice if this option could accept more than one value. (all,file_path,context_line,match_line,match_text)
For example, you can add hyperlinks to file paths and matching lines.
The default is to add hyperlinks to all lines.

```shell
kitty +kitten hyperlinked_grep --kitten hyperlink=file_path --kitten hyperlink=match_text
```

If there are examples in the documentation, then the user will know how to use it with other ripgrep options, such as `rg --context`.

groves:
> Add some docs for this to hyperlinked_grep.rst

Done.

> I would suggest adding a prefix to the kitten options, just like the `ssh` kitten. Reduce the possibility of adding options with the same name from upstream.

I was thinking the `hyperlink` prefix was serving the same purpose of avoiding collisions. I prefer it as it makes parsing and cleanup simpler.

> It would be nice if this option could accept more than one value. (all,file_path,context_line,match_line,match_text) For example, you can add hyperlinks to file paths and matching lines. 

I can't think of a reason why I'd want those other options. Do you have a use in mind?

page-down:
> I was thinking the `hyperlink` prefix was serving the same purpose of avoiding collisions.

I don't think so. What prevents conflicts is the `-only-matches` part.
In the docs:
> Hopefully, someday this functionality will make it into some upstream grep program ...

So someday `--hyperlink` might be introduced upstream like `ls --hyperlink`.

This `boolean set` option is too limited and cannot be extended.

For all kitten wrapping another program, having uniform parameters for configuring additional options would reduce the user's cognitive and ease of use, and would not require each kitten to implement its own argument parsing logic.

So the user clearly knows that it's for kitten, not ripgrep.

kovidgoyal:
Merged, with a few minor changes.

PR #5440, opened 2022-08-29
bash shell integration: Send STDERR of unsetting inherit_errexit to /dev/null

grimm26:
Some bash versions may not have that option and rather than checking if
they even have it or if it is set, just quiet the STDERR it may throw.

```
$ kssh myhost
bash: shopt: inherit_errexit: invalid shell option name
myhost:~$ bash --version
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

### DIFF
--- a/shell-integration/bash/kitty.bash
+++ b/shell-integration/bash/kitty.bash
@@ -23,7 +23,7 @@ if [[ -n "$KITTY_BASH_INJECT" ]]; then
         }
     else
         builtin set +o posix
-        builtin shopt -u inherit_errexit  # resetting posix does not clear this
+        builtin shopt -u inherit_errexit 2>/dev/null  # resetting posix does not clear this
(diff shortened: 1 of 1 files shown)


grimm26:
I can just change it to

```bash
        if builtin shopt -p inherit_errexit 2>/dev/null; then
          builtin shopt -u inherit_errexit  # resetting posix does not clear this
        fi
```

if you think that is cleaner.

kovidgoyal:
No need, redirecting the error message is fine.

Issue #5461, opened 2022-09-02
Custom app icon is lost on update (macOS)


tmillr:
## Describe the bug

For macOS, the [officially recommended way to update](https://sw.kovidgoyal.net/kitty/binary/#binary-install) is via:

```sh
curl -L https://sw.kovidgoyal.net/kitty/installer.sh | sh /dev/stdin
```

> To update kitty, simply re-run the command.

However, running this command overwrites my custom kitty application icon (the icon that is shown in Finder, on the desktop, in the dock, etc.). Am I going to have to manually restore my custom application icon every time I update? That would be very annoying.

## To Reproduce

1. Install kitty on macOS
2. Change the app icon via [the instructions from the kitty website](https://sw.kovidgoyal.net/kitty/faq/?highlight=icon#i-do-not-like-the-kitty-icon)
3. Wait until a new version of kitty is published/released (although this step might not be unnecessary)
4. Update kitty via the officially recommended way to update (see above)

## Additional context

My system is macOS Big Sur (version 11.6.8). I am not sure whether this issue is specific to macOS or not.

I'm hoping this either gets fixed in the sh install script, or that there is another trivial, non-hacky way to keep this from happening every time I update kitty.

kovidgoyal:
All that the install script does is mount the dmg and copy kitty.app
into /Applications for you. If that is removing the custom icon, then
you will need to take it up with Apple.

tmillr:
Darn. Alright. 

It sounds like the app icon is stored within kitty.app (the app package/bundle, or whatever Apple calls it) itself. It appears that dragging and dropping the .icns file, in order to change/customize the app icon, simply creates a file named /Applications/kitty.app/Icon? within kitty.app.

```
/Applications/kitty.app/Icon?
```

I'm guessing that this is my custom icon I just dragged and dropped, and that it gets overwritten when updating kitty.

It sounds like I will have to figure something out myself to keep this from happening every time I update.

tmillr:
> > It sounds like the app icon is stored within ...
> > ... a file named `Icon?` ...
> 
> No, that file is `Icon` followed by a carriage return character.

You're right about it being a carriage return, but I'm 99% sure this file is somehow associated with the app's icon because it only appears after changing the app's icon via the canonical drag & drop method. 

(I finally gave in and did some research on it, and found [this](https://superuser.com/questions/298785/icon-file-on-os-x-desktop).)

[...]

PR #5464, opened 2022-09-02
macOS: Add a way to set a custom application icon for kitty

page-down:
Changing the kitty icon manually as described in the FAQ is troublesome, so a new way to set custom icons via the CLI has been added.

Configure custom icon for the app bundle directly by calling the Cocoa framework's API.

Related discussion:
https://github.com/kovidgoyal/kitty/issues/5461

This PR does not fix the above issue because it does not yet add to the installer script the steps to: detect the presence of custom icon, backup, restore custom icon, etc, as this would add unnecessary complexity to the installation script.

Please review, thanks.

I noticed the following commit, does that mean that using `doc` text roles without the `slash /` prefix is no longer supported?
https://github.com/kovidgoyal/kitty/commit/295743a96e21ffe5fadb67e2336effc140e29798

```text
:doc:`launch` -> :doc:`/launch`
```

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -40,6 +40,8 @@ Detailed list of changes
 
 - X11: Fix a regression in the previous release that caused pasting from GTK based applications to have extra newlines (:iss:`5528`)
 
+- macOS: Allow to set custom app icon automatically (:pull:`5464`)
+
 0.26.3 [2022-09-22]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -287,7 +287,24 @@ homepage:
    :target: https://github.com/samholmes/whiskers
    :width: 256
 
-On macOS you can change the icon by following the steps:
+On macOS you can put :file:`kitty.app.icns` or :file:`kitty.app.png` in the
+:ref:`kitty configuration directory <confloc>`, and this icon will be applied
+automatically at startup if the app bundle has no custom icon. This is
+convenient because app updates under macOS will replace the entire app bundle
+and the custom icon will be removed as well. To automatically update a new icon
+at startup, you need to remove the custom icon on :file:`kitty.app` first.
+
+You can set custom icon via CLI, which can be used in shell scripts:
+
+.. code-block:: sh
+
+    # Set kitty.icns as the icon for currently running kitty
+    kitty +runpy 'from kitty.fast_data_types import cocoa_set_app_icon; import sys; cocoa_set_app_icon(*sys.argv[1:]); print("OK")' kitty.icns
+
+    # Set the icon for app bundle specified by the path
+    kitty +runpy 'from kitty.fast_data_types import cocoa_set_app_icon; import sys; cocoa_set_app_icon(*sys.argv[1:]); print("OK")' /path/to/icon.png /Applications/kitty.app
+
+You can also change the icon manually by following the steps:
 
 #. Find :file:`kitty.app` in the Applications folder, select it and press :kbd:`⌘+I`
 #. Drag :file:`kitty.icns` onto the application icon in the kitty info pane
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -862,6 +862,98 @@ cocoa_set_url_handler(PyObject UNUSED *self, PyObject *args) {
     } // autoreleasepool
 }
 
+static PyObject*
+cocoa_app_has_custom_icon(PyObject UNUSED *self, PyObject *args) {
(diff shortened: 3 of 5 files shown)


kovidgoyal:
Why not just use this method at kitty startup to check for
kitty.app.icns in the kitty config folder and set the icon.
All the user needs to do is place the icon in the config directory.
Ideally one should check if the icon needs updating rather than running
it every time, of course. This can be done by storing a size+timestamp of the icns
file in the cache dir and checking if it has changed or if there is no
custom icon in the bundle.

page-down:
I have never seen another app use this approach and am concerned about potential unknown issues.

The cache information needs to be attached to the app bundle, as each app bundle can have its own custom icon.
Also, if a `kitty.app` with a `dev` label on the custom icon is run, it will be overwritten by `kitty.app.icns`.

I think the easiest way to do this is:
Automatically apply the custom icon if it exists in the configuration folder and the current app bundle does not have a custom icon. (However this will add one more file IO at startup.)
Update the icons by running `rm "/Applications/kitty.app/Icon"$'\r'`, or delete the custom icons via Finder, or overwrite with an unmodified `kitty.app` and run kitty again.
Or just drag and drop your custom icon in Finder.

Automatic updates:
If there are no custom icon, kitty.app.icns in the configuration folder will be used automatically and the path to the icns file (and file size, mtime..) will be recorded via the extended attributes (xattr) on the app bundle folder.
Update the custom icon automatically based on the information in xattr.

I will try the former approach and test it to reduce unnecessary code maintenance in the future (as compared to the second approach).

kovidgoyal:
On Fri, Sep 02, 2022 at 06:25:46AM -0700, page-down wrote:
> I have never seen another app use this approach and am concerned about potential unknown issues.

This will be used by a small fraction of kitty users, and if there are
issues, we can always revert it.

> 
> The cache information needs to be attached to the app bundle, as each app bundle can have its own custom icon.
> Also, if a `kitty.app` with a `dev` label on the custom icon is run, it will be overwritten by `kitty.app.icns`.
> 
> I think the easiest way to do this is:
> Automatically apply the custom icon if it exists in the configuration folder and the current app bundle does not have a custom icon. (However this will add one more file IO at startup.)
> Update the icons by running `rm "/Applications/kitty.app/Icon"$'\r'`, or delete the custom icons via Finder, or overwrite with an unmodified `kitty.app` and run kitty again.
> Or just drag and drop your custom icon in Finder.
> 
> Automatic updates:
> If there are no custom icon, kitty.app.icns in the configuration folder will be used automatically and the path to the icns file (and file size, mtime..) will be recorded via the extended attributes (xattr) on the app bundle folder.
> Update the custom icon automatically based on the information in xattr.
> 
> I will try the former approach and test it to reduce unnecessary code maintenance in the future (as compared to the second approach).

OK.

page-down:
I have updated the PR, please review, thanks.

Is it possible to run mypy tests in a virtual environment?

```shell
python3 -m venv /path/to/venv
# activate venv
make
./test.py mypy
python3 test.py mypy
```

```log
/usr/local/opt/python@3.10/bin/python3.10: No module named mypy
```

https://github.com/kovidgoyal/kitty/commit/57d3d0967940d50c9ca9a8a79a6ec847f8d7ad28

```text
After the virtual environment has been activated:
`PATH` -> /path/to/kitty/launcher:/usr/local/bin:/usr/bin:...
`shutil.which('python')` -> /usr/local/bin/python
```

kovidgoyal:
No idea, never tried as I dont use venvs. Presumably you will need to install mypy in the venv.

page-down:
> ... Presumably you will need to install mypy in the venv.

Yes, mypy is installed.
Everything was working fine before, however, after the above linked commit, `python` outside of venv is called.

kovidgoyal:
Link kitty against the python in the venv.

kovidgoyal:
Actually no, the linked interpreter is not used. See the function type_check() in main.py

kovidgoyal:
This should take care of it: https://github.com/kovidgoyal/kitty/commit/5ea0519f62c8562633f269ce0ea147629ea6334a

kovidgoyal:
I have merged changing the implementation a bit, to allow the icon to be updated easily.

kovidgoyal:
On Fri, Sep 02, 2022 at 05:09:20AM -0700, page-down wrote:
> I noticed the following commit, does that mean that using `doc` text roles without the `slash /` prefix is no longer supported?
> https://github.com/kovidgoyal/kitty/commit/295743a96e21ffe5fadb67e2336effc140e29798

No it works fine. It does not work in sphinx because there doc roles are
relative to the file, not the root of the tree. This has always been the case.

PR #5529, opened 2022-09-23
panel kitten: add override redirect flag

mochaaP:


kovidgoyal:
I'm afraid I am not willing to add a dependency on xdotool for this. It
should be perfectly possible to implement it using xprop. Figure out
what properties xdotool is setting on the window and emulate it with
xprop. Most likely it will be something from here: https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46409511133632

page-down:
When you change the example code (option text role):
```rst
:option:`kitty +kitten panel -o` font_size=20
```
to
```rst
:option:`kitty +kitten panel -o font_size=20`
```

Are you sure you will get the desired result?

```log
generated/cli-kitten-panel.rst:43: WARNING: unknown option: kitty +kitten panel -o font_size=20
```

You can build and check for problems after modifying the document.

mochaaP:
I don't think this can be achieved with `xprop`. Override redirect is a window attribute, not a property. These are two different things in X11.

mochaaP:
> When you change the example code (option text role):
> 
> ```
> :option:`kitty +kitten panel -o` font_size=20
> ```
> 
> to
> 
> ```
> :option:`kitty +kitten panel -o font_size=20`
> ```
> 
> Are you sure you will get the desired result?
> 
> ```
> generated/cli-kitten-panel.rst:43: WARNING: unknown option: kitty +kitten panel -o font_size=20
> ```
> 
> You can build and check for problems after modifying the document.

I misunderstood that. I thought it was like a code block or something.

kovidgoyal:
Then make the dependency optional. Or better implement it in the glfw
x11 backend and use that. It is on my TODO list to get rid of the
dependency on xprop as well, by implementing the needed code in glfw.

mochaaP:
Fine, I will look into that later. Should I split it into another PR?

kovidgoyal:
No, this PR is fine.

kovidgoyal:
I just committed code to drop xprop from the panel kitten. You can use
it as a template.

mochaaP:
`XChangeWindowAttributes` results in a "BadWindow" error. I have no idea why.

kovidgoyal:
Probably because the window has to be mapped before you can change its
attributes. Just a guess, it's been a while since I worked with Xlib.

PR #5567, opened 2022-10-09
add support for out-of-source builds

marksisson:
This is useful for scenarios where source directory cannot be modified, or for building in a different location than source.

Use VPATH to specify source search path, e.g.

`git clone https://github.com/kovidgoyal/kitty.git`
`VPATH=kitty make -f kitty/Makefile`

All build artifacts will be in **build** folder of current directory.

In-source builds can still be executed in the source directory with a simple `make`, but will also place build artifacts in a 'build' folder within the source directory.

PR #5569, opened 2022-10-09
add support for out-of-source builds - redux

marksisson:
Continuation of #5567

This is useful for scenarios where source directory cannot be modified, or for building in a different location than source.

Use VPATH to specify source search path, e.g.

```code
git clone https://github.com/kovidgoyal/kitty.git
VPATH=kitty make -C kitty
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,55 +9,57 @@ ifdef FAIL_WARN
 export FAIL_WARN
 endif
 
-all:
-	python3 setup.py $(VVAL)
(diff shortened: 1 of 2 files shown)


marksisson:
I tried doing an out-of-tree build for the go branch, but could not find a way to integrate the in-tree `tools` go code with the generated go code (in out-of-tree folder).  The go build is pretty particular about where module code is located. I'll need to go deeper into go builds before successfully tackling this one.

kovidgoyal:
I dont know of a way to get the go compiler to read a package split into
multiple locations either.

PR #5572, opened 2022-10-10
support for out of source builds (go branch)

marksisson:
Example of out-of-source build: (copy and paste following blocks into terminal)

```shell
# change ~/src to your directory for git repo clone
cd ~/src
git clone https://github.com/kovidgoyal/kitty.git
```

```shell
# change ~/build/kitty to your directory for kitty build artifacts
mkdir -p ~/build/kitty
cd ~/build/kitty
```

create out-of-source Makefile with the following contents:

```shell
cat <<EOF > Makefile
# change \$(HOME)/src/kitty to your location of kitty source
kitty_source = \$(HOME)/src/kitty
VPATH = \$(kitty_source)
include \$(kitty_source)/Makefile
EOF
```

```shell
make
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,55 +9,57 @@ ifdef FAIL_WARN
 export FAIL_WARN
 endif
 
-all:
-	python3 setup.py $(VVAL)
+THIS_MAKEFILE_DIR=$(dir $(realpath $(lastword $(MAKEFILE_LIST))))
 
-test:
-	python3 setup.py $(VVAL) test
+all: setup.py
+	python3 $< $(VVAL)
 
-clean:
-	python3 setup.py $(VVAL) clean
+test: setup.py
+	python3 $< $(VVAL) test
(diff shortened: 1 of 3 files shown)


kovidgoyal:
If you are just going to be copying the code anyway, just copy the whole source folder. I dont see the point of halfway housing it.

kovidgoyal:
No I meant why bother with out of source builds. Just copy the whole kitty source tree in your nix flake. There isnt much point in copying some part of the source code but not the rest. And adding support for out of tree builds just adds complexity for no real gain.

kovidgoyal:
Yes, I would rather not spend the time reviewing and dealing with
possible breakage for a solution that doesnt avoid copying the go code,
since if one is anyway copying the go code, one might as well copy the
rest of the code. There is no significant advantage to a halfway house
that I can see.

PR #5590, opened 2022-10-15
+KITTY_SHELL_ACTIVE_TAB_ID; echo tab ID on shell open

ctrlcctrlv:


page-down:
Would it be better to combine the previous window id with the tab id in one line?
Capping the number of lines of information after startup to three lines. Reduce unnecessary duplicated text.

PR #5592, opened 2022-10-16
Check correct pair member

ms-test:
I haven't looked deeply, but this looked to be a simple typo unless it's checking one both times because of a type invariant. If that is the case, and the check is correct, then a comment and/or code refactoring to call attention to this would likely be a good idea.


ms-test:
It just occurred to me after submitting that I haven't checked your contribution guidelines, but I hope there are no issues with this (deceptively?) trivial PR. I apologize in advance if there are any issues.

I will take a look when I have a moment for any future contributions and correct this if necessary.

kovidgoyal:
There is no contribution bureaucracy all you need to do is convince me of the wisdom of your PR

PR #5674, opened 2022-11-16
Fix completion for fish

page-down:
I also noticed that `kitty --help` and `kitty-tool --version` etc. are not listed.

For `kitty-tool`, it looks like `kitty-tool update-self` should not take effect under macOS, as it will break the signature of the entire application bundle.

I think it's best to run `kitty +update-self` to update the whole macOS bundle.
If anyone is interested in implementing this, replace the package downloaded via `kitty +update-self` the next time the user launches kitty via launchd, and run kitty again when the upgrade is complete.
Updates will not take effect when running kitty via CLI or `open` to avoid complications with handling envs, stdin, args, etc.


kovidgoyal:
When I finish update-self, it will be restricted to only be present for cross
compiled kitty-tool instances. At the moment its easier to do
development with it present for all.

PR #5676, opened 2022-11-16
Add completion for kitty +list-fonts

page-down:
Hope I didn't miss anything, please review, thanks.

There are still some subcommands for internal use and `+icat` missing completion.
```text
+complete +edit +edit-config +hold +icat +shebang
```

### DIFF
--- a/gen-go-code.py
+++ b/gen-go-code.py
@@ -106,14 +106,22 @@ def generate_completions_for_kitty() -> None:
     # kitty +
     print('plus := k.AddSubCommand(&cli.Command{Name:"+", Group:"Entry points", ShortDescription: "Various special purpose tools and kittens"})')
 
+    # kitty +launch
     print('plus_launch := plus.AddSubCommand(&cli.Command{'
           'Name:"launch", Group:"Entry points", ShortDescription: "Launch Python scripts", ArgCompleter: complete_plus_launch})')
     print('k.AddClone("", plus_launch).Name = "+launch"')
 
+    # kitty +list-fonts
+    print('plus_list_fonts := plus.AddSubCommand(&cli.Command{'
+          'Name:"list-fonts", Group:"Entry points", ShortDescription: "List monospace fonts"})')
+    print('k.AddClone("", plus_list_fonts).Name = "+list-fonts"')
+
+    # kitty +runpy
     print('plus_runpy := plus.AddSubCommand(&cli.Command{'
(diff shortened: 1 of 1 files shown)


kovidgoyal:
+icat is deprecated, in favor of +kitten icat

page-down:
@kovidgoyal

I have noticed occasional abnormal exits when testing completions with the kitty shell (Ctrl+Shift+ESC), and the last frame seems to have the word `EOF` in it. The keys that have been pressed include tab, backspace, arrow keys, page up, page down, Home, and End.

```conf
map f1 launch --allow-remote-control --hold kitty @
```

After completing `launch -`, the second half of the output is all recognized as a hyperlink and does not appear to end correctly.

I also noticed that the `help` command in the kitty shell does not use pager (which keeps the last displayed content on the screen after exit without clearing it), while `launch --help` uses pager.

kovidgoyal:
On Wed, Nov 16, 2022 at 06:22:20AM -0800, page-down wrote:
> @kovidgoyal
> 
> I have noticed occasional abnormal exits when testing completions with the kitty shell (Ctrl+Shift+ESC), and the last frame seems to have the word `EOF` in it. The keys that have been pressed include tab, backspace, arrow keys, page up, page down, Home, and End.

Yes, my custom readline is pretty beta quality right now. It would help
if you can figure out a reproducer, or get a stack trace, which you
should be able to see if you use sh -c "kitty @; read" or just run
kitty @ in a normal kitty window since you dont need to actually execute
any commands.

> 
> ```conf
> map f1 launch --allow-remote-control --hold kitty @
> ```
> 
> After completing `launch -`, the second half of the output is all recognized as a hyperlink and does not appear to end correctly.

This is now fixed.

> 
> I also noticed that the `help` command in the kitty shell does not use pager (which keeps the last displayed content on the screen after exit without clearing it), while `launch --help` uses pager.

help by itself doesnt, since its just a simple list, but help launch
should do it.

page-down:
> ... if you can figure out a reproducer ...

The simplest: hold down a key, for example the left arrow key, long enough and you will get an `Error: EOF`.
Tested under macOS 13.0, either from nightly's standalone kitty-tool, or compiled from the source master branch.

kovidgoyal:
What is long enough? I tried holding it down for 30 secs no errors. Although this is Linux, will try macOS later.

kovidgoyal:
Should be fixed by https://github.com/kovidgoyal/kitty/commit/fc20e8d04b1dc07b5026a8654191ffe8ee521434

PR #5684, opened 2022-11-17
[docs]: don't show config codeblocks that only has a name

ppwwyyxx:
# Before this PR:

I understand that the second line means the "default value is empty", but showing two lines with same content feels a bit ugly. If the default is empty, maybe there is not much reason to show it at all.

# After this PR:

I ran a diff of before vs after on the `docs` directory and verified that nothing else has changed except for a few cases like what's shown in the image.

### DIFF
--- a/kitty/conf/types.py
+++ b/kitty/conf/types.py
@@ -268,11 +268,12 @@ class Option:
             return ans
         mopts = [self] + option_group
         a('.. opt:: ' + ', '.join(f'{conf_name}.{mo.name}' for mo in mopts))
-        a('.. code-block:: conf')
-        a('')
-        sz = max(len(x.name) for x in mopts)
-        for mo in mopts:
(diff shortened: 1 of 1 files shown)


page-down:
There is another problem with the documentation here.

This place is supposed to be the **default** configuration, however, in the rst documentation for repeatable configuration items the **examples** are placed here, leading to confusion for the user. This is not actually the default configuration.

For the commented default `kitty.conf`, I have dealt with this issue here before in one of the PRs.

```conf
# symbol_map
#: E.g. symbol_map U+E0A0-U+E0A3,U+E0C0-U+E0C7 PowerlineSymbols
```

kovidgoyal:
Should be fixed by https://github.com/kovidgoyal/kitty/commit/38547c9e7ab062e86b0dd2dbe33427361461e948

PR #5688, opened 2022-11-18
Note the Unicode version in the generated files instead of the build date

page-down:
Now that we have the Unicode version information, we don't need to use the build date for comments.

Files generated from the same Unicode version will be consistent regardless of the date they were built.


page-down:
This PR has been merged.

PR #5708, opened 2022-11-25
Docs: Fix missing backslashes

page-down:
https://sw.kovidgoyal.net/kitty/conf/#shortcut-kitty.Send-arbitrary-text-on-key-presses
https://sw.kovidgoyal.net/kitty/remote-control/#kitty-send-text

Two backslashes are needed in Python, and two backslashes are needed in the generated rst, for a total of four.

However, `kitty-tool @ send-text --help` ends up with an extra backslash, and needs a little help here.

> ... The text follows Python escaping rules. ...

Also I noticed that `send-text -m id:1 "\x7f"` worked fine in the previous version of the kitty shell, but now it doesn't escape and send properly in the version implemented in Go.

When a command is executed without any output, a blank line is output in the kitty shell, while in previous versions the behavior is the same as in the regular shell, which prints the next prompt if there is no output.

### DIFF
--- a/docs/conf.rst
+++ b/docs/conf.rst
@@ -27,7 +27,7 @@ line.
 
 .. _include:
 
-You can include secondary config files via the :code:`include` directive.  If
+You can include secondary config files via the :code:`include` directive. If
 you use a relative path for :code:`include`, it is resolved with respect to the
 location of the current config file. Note that environment variables are
 expanded, so :code:`${USER}.conf` becomes :file:`name.conf` if
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -3969,7 +3969,7 @@ when pressing specified shortcut keys. For example::
 
 This will send "Special text" when you press the :kbd:`Ctrl+Alt+A` key
 combination. The text to be sent is a python string literal so you can use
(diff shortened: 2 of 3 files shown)


page-down:
@kovidgoyal
I saw the latest changes and noticed that I forgot to mention one issue.
In the kitty shell, type the command `send-text ... '\x7f'` and enter, it will only display `send-text ... ' ` (without the later part starting from `\x7f`.)

Also `send-text \x7f`, `send-text "\x7f"` don't work, only single quotes, is this expected?

kovidgoyal:
On Sat, Nov 26, 2022 at 12:46:13AM -0800, page-down wrote:
> @kovidgoyal
> I saw the latest changes and noticed that I forgot to mention one issue.
> In the kitty shell, type the command `send-text ... '\x7f'` and enter, it will only display `send-text ... ' ` (without the later part starting from `\x7f`.)

Works fine for me though there was a bug where if the sursor was not at
the end of the line pressing enter would truncate the line after the
cursor position.

> 
> Also `send-text \x7f`, `send-text "\x7f"` don't work, only single quotes, is this expected?

You need to escape the backslash, I have added a note about it.

PR #5712, opened 2022-11-26
DRYer: Get hostname

page-down:
The cached hostname cannot be used in `open_url` as it may have been changed.

---

About launching with current cwd, I found an issue.

```shell
kitty --config=NONE -d=/tmp -o 'shell zsh' -o 'map f1 new_window_with_cwd'

ssh ksi-enabled-remote
# Ctrl+D
# F1
```

After ssh disconnects, create a new window with the current cwd, which will then use the remote cwd.
The remote path probably does not exist locally and the shell starts at `/`, which is annoying.

I can think of a few ways to improve this user experience.

Check if the hostname matches the local hostname before `return reported_cwd`.

```python
def hostname_from_osc7_url(url: str) -> str:
    netloc = ''
    if url.startswith('kitty-shell-cwd://'):
        netloc = url.split('/', 3)[2]
    if url.startswith('file://'):
        from urllib.parse import urlparse
        netloc = urlparse(url).netloc
    return netloc.partition(':')[0]
```

Provide more information when reporting cwd in shell integration. For example:
```text
# local
kitty-shell-cwd://host/path#kitty-pid=$KITTY_PID
# ssh kitten
kitty-shell-cwd://host/path#kitty-ssh-kitten=yes
# ssh (with shell integration)
kitty-shell-cwd://host/path
# ssh
file://host/path
```

Store to `last_reported_local_cwd` when the reported kitty-pid is the same as the current one when processing cwd notifications.
When `window.child_is_remote` is False and the last reported cwd explicitly has `kitty-ssh-kitten` or no `kitty-pid`, it falls back to the last recorded `last_reported_local_cwd`.

Outputting the current cwd in prompt would be more accurate, but repeated output of the same cwd is too wasteful of resources and not really worth it for me.

Do you have a better idea?

### DIFF
--- a/kittens/hyperlinked_grep/main.py
+++ b/kittens/hyperlinked_grep/main.py
@@ -4,11 +4,12 @@
 import os
 import re
 import signal
-import socket
 import subprocess
 import sys
 from typing import Callable, cast
 from urllib.parse import quote_from_bytes
+
+from kitty.utils import get_hostname
 
 
 def write_hyperlink(write: Callable[[bytes], None], url: bytes, line: bytes, frag: bytes = b'') -> None:
@@ -76,7 +77,7 @@ def main() -> None:
     num_pat = re.compile(br'^(\d+)([:-])')
 
     in_result: bytes = b''
-    hostname = socket.gethostname().encode('utf-8')
+    hostname = get_hostname().encode('utf-8')
 
     try:
         for line in p.stdout:
--- a/kitty/cli.py
+++ b/kitty/cli.py
@@ -4,7 +4,6 @@
 import os
 import re
 import shlex
-import socket
 import sys
 from collections import deque
 from enum import Enum, auto
@@ -305,6 +304,12 @@ def env(x: str) -> str:
 role_map['envvar'] = role_map['env']
 
 
+@run_once
+def hostname() -> str:
+    from .utils import get_hostname
+    return get_hostname(fallback='localhost')
+
+
 def hyperlink_for_url(url: str, text: str) -> str:
     if sys.stdout.isatty():
         return f'\x1b]8;;{url}\x1b\\\x1b[4:3;58:5:4m{text}\x1b[4:0;59m\x1b]8;;\x1b\\'
@@ -315,7 +320,7 @@ def hyperlink_for_path(path: str, text: str) -> str:
     path = os.path.abspath(path).replace(os.sep, "/")
     if os.path.isdir(path):
         path += path.rstrip("/") + "/"
-    return hyperlink_for_url(f'file://{socket.gethostname()}{path}', text)
+    return hyperlink_for_url(f'file://{hostname()}{path}', text)
 
 
 @role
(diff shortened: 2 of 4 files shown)


kovidgoyal:
On Sat, Nov 26, 2022 at 12:05:59AM -0800, page-down wrote:
> The cached hostname cannot be used in `open_url` as it may have been changed.
> 
> ---
> 
> About launching with current cwd, I found an issue.
> 
> ```shell
> kitty --config=NONE -d=/tmp -o &#39;shell zsh&#39; -o &#39;map f1 new_window_with_cwd&#39;
> 
> ssh ksi-enabled-remote
> # Ctrl+D
> # F1
> ```
> 
> After ssh disconnects, create a new window with the current cwd, which will then use the remote cwd.
> The remote path probably does not exist locally and the shell starts at `/`, which is annoying.

I dont see how that's possible unless you have disabled shell
integration. After ssh exits, the local shell will draw a new prompt and
that prompt will report its cwd. SO the last reported cwd will be
correct.

page-down:
> I dont see how that's possible ... the local shell will draw a new prompt ...

Currently the new cwd is only reported when the cwd changes, at least for zsh and fish (the official fish one and the one that comes with kitty), which I think is good practice.

shell-integration/zsh/kitty-integration
```zsh
chpwd_functions=(${chpwd_functions[@]} "_ksi_report_pwd")
```

shell-integration/bash/kitty.bash
```bash
# unfortunately bash provides no hooks to detect cwd changes
# in particular this means cwd reporting will not happen for a
# command like cd /test && cat.
```

Since the cwd is not changed after exiting ssh, the problem I mentioned will occur.

kovidgoyal:
Then the way to do it is to store last_reported_local_cwd (when
is_remote returns false) and last_reported_remote_cwd. 

Then, use the one that matches the current value of is_remote.

kovidgoyal:
The downside to that is it will only work for ssh it wont work for instance if you another program that reports cwd. The only robust fix is to report on every prompt draw.

kovidgoyal:
I have changed the zsh integration to report cwd at every new prompt draw. So now bash and zsh both do that. Could you do the same for fish please, as it will take me a lot of time to figure out the best way to do it for fish.

page-down:
> Check if the hostname matches ...

To answer my own question: No, for example when connecting from one "localhost" to another "localhost" via IP, this will not work.

> ... when is_remote returns false ...

Need to check this every time CWD is received. This will be expensive.

> The only robust fix is to report on every prompt draw.

Unfortunately, there doesn't seem to be a better way to get acceptable improvements at minimal cost.

> ... Could you do the same for fish ...

I will implement it in another PR.
In my personal configuration I will use `no-cwd` and just update the cwd asynchronously via events after the execution of programs like `ssh`.

This does not apply to normal users, as a list of programs needs to be maintained.
And the shell does not know the OSC7 sent by the executed program, so it also cannot automatically know that cwd needs to be updated again.

Issue #1494, opened 2019-03-24
Setting tab title width?


ddddavidmartin:
Good day,

I was wondering whether there is a way of setting the tab title to a fixed width. The default is to expand the tab when the title is longer and I'd like to avoid that.

See for example this screenshot with different tab widths.

<img width="830" alt="Screen Shot 2019-03-25 at 1 03 19 am" src="https://user-images.githubusercontent.com/710898/54880565-f5a98080-4e99-11e9-8ebb-587dff8414d7.png">

Thanks!

kovidgoyal:
Why would you want to leave unused available space?

ddddavidmartin:
> Why would you want to leave unused available space?

I feel it helps with parsing the open tabs at a glance when they are all equally sized.

Though I have now worked with the tabs as they are for a bit and I may actually mind it a lot less than I initially thought. :)

kovidgoyal:
I see, I dont really feel the need for this myself, but I will accept a
PR adding an option for it.

PR #5718, opened 2022-11-28
Add option to limit length of tab titles

fratajczak:
This PR adds an option to truncate tab titles above a certain length.

I find it annoying sometimes to have very long commands take up lots of screen space for their tab titles. Related to #1494, but it's not exactly what was requested: tab_min_length would need to be added to have a fixed tab size, but I find just setting a max length good enough for my use case.


kovidgoyal:
You dont need ctype in the option definition as this option is only used by python code.

fratajczak:
> You dont need ctype in the option definition as this option is only used by python code.

Okay, fixed

page-down:
@fratajczak

I believe the naming of this configuration item is not accurate, however I did not raise it in time before the merge.
This option should be named `tab_title_max_length`. (We already have `tab_title_template`, prefixed by `tab_title_`)

I also have a question about the placement of this option in `definition.py`, why is it in between a bunch of `tab_bar_`?
Logically, similar configuration items should be placed together as much as possible and ordered according to importance and dependency.

page-down:
I looked at it again and the option names look accurate, however I am not sure this is what the user expects.

The length of the tab includes left and right decorations, abbreviations `...`, and I think it is more useful to limit the length of the title rather than the total length of the tab.

fratajczak:
Indeed, I think it would be better to truncate the title instead.

@kovidgoyal I think your commit 5c50e3869ca07565f79df45cfaa80ff8dd99b5e6 doesn't actually fix the problem: try
```
tab_bar_style fade
tab_title_max_length 5
```
The fading separators get cropped. 

Screenshot from 2022-11-29 16-39-25


The `max_tab_length` parameter of `draw_title` doesn't do anything, it's just used to tell how much title length is available in the title_template. 

Something like:
```python
 def draw_title(draw_data: DrawData, screen: Screen, tab: TabBarData, index: int, max_title_length: int = 0) -> None:
+    tab_title = tab.title
+    if draw_data.max_title_length > 0 and len(tab_title) > draw_data.max_title_length:
+            tab_title = tab.title[:draw_data.max_title_length] + '…'
     ta = TabAccessor(tab.tab_id)
     data = {
         'index': index,
         'layout_name': tab.layout_name,
         'num_windows': tab.num_windows,
         'num_window_groups': tab.num_window_groups,
-        'title': tab.title,
+        'title': tab_title,
         'tab': ta,
```
Would crop only the title

kovidgoyal:
max_tab_length is not used in draw_title its used *after* draw title.

kovidgoyal:
See https://github.com/kovidgoyal/kitty/commit/722a1020fae50a36462cfdc5cb20abbdaa35f9f4

PR #5738, opened 2022-12-05
Fix the base path for envinclude and minor docs improvements

page-down:
Remove extra backslashes from commented config.
Update docs to use `\e` in the example.
Fix the base path for envinclude.

```log
TEST='include target.conf' kitty -o 'envinclude TEST'

Could not find included config file: ~/.config/target.conf, ignoring
```

Do you think the base path in all environment variables should default to the kitty config file path regardless of where the config file uses envinclude?
So in the following example the `target.conf` file in the configuration folder will be loaded instead of `a/b/target.conf`.

```shell
TEST='include target.conf' kitty -o 'include a/b/envinclude-test.conf'

# -> ~/.config/kitty/target.conf
```

The current logic is equivalent to inserting environment variables as configuration lines into the line where `envinclude` is located, which is also understandable and reasonable for the user. I'm not sure which one you prefer.

Also, would you reject a PR that allows kitty to be configured via Python scripts?
For example by adding `pyinclude myconfig.py`. The function in the script will output configuration lines that will be backward compatible.

### DIFF
--- a/kitty/conf/types.py
+++ b/kitty/conf/types.py
@@ -113,6 +113,8 @@ def remove_markup(text: str) -> str:
         if key in ('ac', 'opt'):
             t, q = extract(m)
             return f'{t} {q}' if q and q != t else t
+        if key == 'code':
+            return m.group(2).replace('\\\\', '\\')
 
         return str(m.group(2))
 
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I prefer the current logic and in any case the case for the alternate is not strong enough to justify breaking backwards compat.

As for pyinclude, I am not a big fan of using Turing complete languages for configuration, it tends to make it very hard to statically change/analyze config down the road when you need to migrate things. If it turns out there is a serious need for it maybe I will reconsider, at the moment I cant think of many real world use cases that the existing include mechanism does not address.

PR #5758, opened 2022-12-12
Remote control: Keep the same naming style for ls command output

page-down:
Since we don't have `tab_title`, `is_focused_tab`, `window_title`, `is_focused_window` ... so use the same naming style `is_active`.

Also remove the new `last_focused`, because there is only one active OS window in a kitty instance, and when is_active is true, it means the last focused OS window.

New fields are added only when the current data cannot be expressed. Or add new fields only if this alias brings significant value to the user.

Do you think it is possible that `is_active` is not the last focused OS window?

Please review, thanks.


kovidgoyal:
It's possible for no window to be focused. That was why last_focused was introduced.

page-down:
> It's possible for no window to be focused. ...

Looks like this is what I missed. I have updated the PR, thanks.

Issue #2790, opened 2020-06-23
OpenGL ES 3: Fail to start on aarch64: [glfw error 65543]


tgunnoe:
**Describe the bug**
Fails to start

**To Reproduce**
`$ kitty`

**Expected behavior**
Startup

**Screenshots**
```
[175 14:38:59.616720] [glfw error 65543]: EGL: Failed to create context: Arguments are inconsistent
[175 14:38:59.621982] Traceback (most recent call last):
  File "/nix/store/b9zhn4izh0c2xnhxk94s4rqshzc8jwr8-kitty-0.16.0/bin/../lib/kitty/kitty/main.py", line 351, in main
    _main()
  File "/nix/store/b9zhn4izh0c2xnhxk94s4rqshzc8jwr8-kitty-0.16.0/bin/../lib/kitty/kitty/main.py", line 344, in _main
    run_app(opts, args, bad_lines)
  File "/nix/store/b9zhn4izh0c2xnhxk94s4rqshzc8jwr8-kitty-0.16.0/bin/../lib/kitty/kitty/main.py", line 145, in run_app
    _run_app(opts, args, bad_lines)
  File "/nix/store/b9zhn4izh0c2xnhxk94s4rqshzc8jwr8-kitty-0.16.0/bin/../lib/kitty/kitty/main.py", line 129, in _run_app
    args.cls or appname, load_all_shaders)
ValueError: Failed to create GLFWwindow
```

**Environment details**
OS: NixOS 20.03 aarch64
on Pinebook Pro
RK-3399 and Mali-T860 
Panfrost on Mesa 20.1, experimental OpenGL ES 3.0 support

```
 $ env PAN_MESA_DEBUG=gles3 glxinfo | grep OpenGL                            

OpenGL vendor string: Panfrost
OpenGL renderer string: Mali T860 (Panfrost)
OpenGL version string: 2.1 Mesa 20.1.0
OpenGL shading language version string: 1.20
OpenGL extensions:
OpenGL ES profile version string: OpenGL ES 3.0 Mesa 20.1.0
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.00
OpenGL ES profile extensions:
```

Output of kitty --debug-config
```
kitty 0.16.0 created by Kovid Goyal
Linux gammu 5.6.0 #1-NixOS SMP Thu Jan 1 00:00:01 UTC 1970 aarch64
[...]

PR #5795, opened 2022-12-20
Downgrade OpenGL version requirement to 3.1

marcan:
There are only a few features required from newer versions, and they can be achieved via extensions. This significantly improves compatibility.


kovidgoyal:
Could you tell me a little about what is the increase in compatibility. Is there some specific device/line of devices that support 3.1 but not 3.3 that this will help with?

marcan:
See #2790. In particular, GL 3.2 requires geometry shader support, which a bunch of hardware does not support (like ARM and Apple GPUs). Emulating that somehow is not impossible, but Mesa has no such emulation today and this is unlikely to be fixed any time soon. This should make kitty work on ARM GPUs (Panfrost) today, and Apple GPUs (Asahi) in the near future. See https://mesamatrix.net/.

In general, you should always aim to use the lowest reasonable GL version and request extensions for additional functionality, to improve compatibility.

kovidgoyal:
OK I'll merge it if it works on all the GPUs I have to test with, thanks.

PR #5815, opened 2022-12-24
ssh kitten: Compile terminfo with tic coming from pkgsrc under NetBSD

### LINKED ISSUE #5814: xterm-kitty not setting up correctly in NetBSD
**Describe the bug**
After trying [everything in the FAQ](https://sw.kovidgoyal.net/kitty/faq/#i-get-errors-about-the-terminal-being-unknown-or-opening-the-terminal-failing-when-sshing-into-a-different-computer), NetBSD 9.3 server shows me "Error opening terminal: xterm-kitty." when trying to run certain commands (htop in this case) but the `.terminfo.cdb` file and `~/.terminfo` directory was all created when running `kitty +kitten ssh myserver`.

**To Reproduce**
1. Type `kitty +kitten ssh myserver`, which connects to the NetBSD server.
2. Type `htop`
3. Get the error `Error opening terminal: xterm-kitty`, even though all the files were created on the NetBSD server via the kitty client.

**Screenshots**

**Environment details**
Kitty terminal running on Arch/Artix linux work correctly with all other OSes I've tried when running the commands in the FAQ.

**Additional context**
Unsure if Issue https://github.com/kovidgoyal/kitty/issues/4622 was fixed.

page-down:
Allow programs that depend on terminfo.cdb as well as programs installed via pkgsrc that depend on the new version of ncurses to run.

When the program fails with `~/.terminfo.cdb` as specified by TERMINFO, it will fall back to `~/.terminfo/x/xterm-kitty` compiled with the new version of tic.

Fixes #5814

Please review, thanks.

There is also a related problem, not limited to NetBSD, if the user compiles `~/.terminfo/x/xterm-kitty` with the new version of tic (`/whatever/path/to/tic`) and programs rely on the new version of ncurses, it will not work by default because ssh kitten overwrites terminfo unconditionally.
Should we check that `~/.terminfo/x/xterm-kitty` is older than `kitty.terminfo` before compiling and overwriting?

I don't really think too many checks should be added, as this can lead to added complexity in troubleshooting issues. And not many people will encounter this edge case.

### DIFF
--- a/shell-integration/ssh/bootstrap-utils.sh
+++ b/shell-integration/ssh/bootstrap-utils.sh
@@ -25,7 +25,14 @@ compile_terminfo() {
 
     if [ -e "/usr/share/misc/terminfo.cdb" ]; then
         # NetBSD requires this file, see https://github.com/kovidgoyal/kitty/issues/4622
-        command ln -sf "../../.terminfo.cdb" "$1/$tname/x/xterm-kitty"
+        # Also compile terminfo using tic installed via pkgsrc,
+        # so that programs that depend on the new version of ncurses automatically fall back to this one.
+        if [ -x "/usr/pkg/bin/tic" ]; then
+            /usr/pkg/bin/tic -x -o "$1/$tname" "$1/.terminfo/kitty.terminfo" 2>/dev/null
+        fi
+        if [ ! -e "$1/$tname/x/xterm-kitty" ]; then
+            command ln -sf "../../.terminfo.cdb" "$1/$tname/x/xterm-kitty"
+        fi
         tname=".terminfo.cdb"
     fi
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
On Fri, Dec 23, 2022 at 08:17:42PM -0800, page-down wrote:
> Allow programs that depend on terminfo.cdb as well as programs installed via pkgsrc that depend on the new version of ncurses to run.
> 
> When the program fails with `~/.terminfo.cdb` as specified by TERMINFO, it will fall back to `~/.terminfo/x/xterm-kitty` compiled with the new version of tic.
> 
> Fixes #5814
> 
> Please review, thanks.
> 
> There is also a related problem, not limited to NetBSD, if the user compiles `~/.terminfo/x/xterm-kitty` with the new version of tic (`/whatever/path/to/tic`) and programs rely on the new version of ncurses, it will not work by default because ssh kitten overwrites terminfo unconditionally.
> Should we check that `~/.terminfo/x/xterm-kitty` is older than `kitty.terminfo` before compiling and overwriting?
> 
> I don&#39;t really think too many checks should be added, as this can lead to added complexity in troubleshooting issues. And not many people will encounter this edge case.

Yes, I agree. Let's see if it becomes an actual problem before we try to
address it.

PR #5818, opened 2022-12-25
Fix scrolling images within margins

sergei-grechanik:
This commit fixes an off-by-one error in image scrolling that caused images to not be scrolled or to be cropped at the wrong line.

### DIFF
--- a/kitty/graphics.c
+++ b/kitty/graphics.c
@@ -1421,12 +1421,12 @@ scroll_filter_func(ImageRef *ref, Image UNUSED *img, const void *data, CellPixel
 
 static bool
 ref_within_region(const ImageRef *ref, index_type margin_top, index_type margin_bottom) {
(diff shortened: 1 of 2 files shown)


sergei-grechanik:
Just in case, I used the following script to test it:
```bash
#!/bin/bash

reset

# Print some text to fill the scrolling region
for i in {1..20}; do
  if [[ "$i" == 4 || "$i" == 16 ]]; then
    echo "=================="
  else
    echo "Line $i"
  fi
done

kitty icat --place 22x9@9x5 ./logo/kitty.png

sleep 0.5

# Move lines down
for i in {1..3}; do
  echo -en "\033[5;15r\033[5;1H\033M"
  sleep 0.5
done


# Move lines up
for i in {1..6}; do
  echo -en "\033[5;15r\033[15;1H\033D"
  sleep 0.5
done

echo -en "\033[30;1H"
```

PR #5830, opened 2022-12-28
Show OSC8 link URL at bottom of screen on mouse hover

### LINKED ISSUE #5827: Show URL when hovering on OSC8 hyperlink
Currently, when hovering on an OSC8 hyperlink, the text is only wave-underlined. 
By default (i.e., `allow_hyperlink yes`), users do not know what the link is until the link is opened.

This is a security concern: if the link comes from a malicious app, it could point to malicious web pages, or even worse, directly exploit apps using other schemas like `app_name://xxx`.

`allow_hyperlink ask` addresses the issue by asking the user, but it's less convenient for users: users need an extra click or keystroke to open any page. This feels like too much overhead for the majority of URLs that are safe.

I'd advocate for an experience just like ANY browser: when hovering on a link, show the link to users (e.g., at the bottom left corner). This way, users can easily tell that the link is legit without having to use `allow_hyperlink=ask`.

**Additional context**
Alacritty shows the URL at bottom left corner when hovering:

ppwwyyxx:
Fix #5827 

URL is displayed at bottom by default. If cursor is at the bottom, then URL is displayed at top. Below are screenshots of what it looks like in a terminal with two windows.

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -340,8 +340,13 @@ handle_mouse_movement_in_kitty(Window *w, int button, bool mouse_cell_changed) {
 
 static void
 detect_url(Screen *screen, unsigned int x, unsigned int y) {
-    if (screen_detect_url(screen, x, y)) mouse_cursor_shape = HAND;
-    else set_mouse_cursor_for_screen(screen);
+    if (screen_detect_url(screen, x, y)) {
+      mouse_cursor_shape = HAND;
+      screen->mouse_on_url = true;
+    } else {
+      set_mouse_cursor_for_screen(screen);
+      screen->mouse_on_url = false;
+    }
 }
 
 static bool
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -2913,6 +2913,14 @@ link when clicked.
 '''
     )
 
+opt('preview_hyperlinks', 'yes',
+    option_type='to_bool', ctype='bool',
+    long_text='''
+Preview the URL of :term:`hyperlink <hyperlinks>` escape sequences (OSC 8)
+when mouse hovers on the link.
+'''
+    )
+
 opt('shell_integration', 'enabled',
     option_type='shell_integration',
     long_text='''
--- a/kitty/options/parse.py
+++ b/kitty/options/parse.py
@@ -1136,6 +1136,9 @@ class Parser:
 
     choices_for_pointer_shape_when_grabbed = frozenset(('arrow', 'beam', 'hand'))
 
+    def preview_hyperlinks(self, val: str, ans: typing.Dict[str, typing.Any]) -> None:
+        ans['preview_hyperlinks'] = to_bool(val)
+
     def remember_window_size(self, val: str, ans: typing.Dict[str, typing.Any]) -> None:
         ans['remember_window_size'] = to_bool(val)
 
--- a/kitty/options/to-c-generated.h
+++ b/kitty/options/to-c-generated.h
@@ -903,6 +903,19 @@ convert_from_opts_allow_hyperlinks(PyObject *py_opts, Options *opts) {
 }
 
 static void
+convert_from_python_preview_hyperlinks(PyObject *val, Options *opts) {
+    opts->preview_hyperlinks = PyObject_IsTrue(val);
+}
+
+static void
+convert_from_opts_preview_hyperlinks(PyObject *py_opts, Options *opts) {
+    PyObject *ret = PyObject_GetAttrString(py_opts, "preview_hyperlinks");
+    if (ret == NULL) return;
+    convert_from_python_preview_hyperlinks(ret, opts);
+    Py_DECREF(ret);
+}
+
+static void
 convert_from_python_macos_option_as_alt(PyObject *val, Options *opts) {
     opts->macos_option_as_alt = PyLong_AsUnsignedLong(val);
 }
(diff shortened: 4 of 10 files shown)


page-down:
`screen_draw_overlay_text` is mainly used for drawing IME pre-edited text, you can't just use it like that.

This leads to at least the following problems:
- Pre-edited text disappears.
- Cursor is in the wrong position after canceling input.
- During input, the displayed URL address flies to the cursor position.

When displaying URL addresses, I believe that some common practices to prevent phishing should be used rather than just outputting the original address as is.
For example, [Punycode](https://en.wikipedia.org/wiki/Punycode) should be used when displaying domain names, and the host name part may also be bolded. Also, when displaying the path, it should be decoded instead of displaying Unicode in URI encoded form.

page-down:
When I was trying to test "URLs that exceed the screen width and height", kitty just crashed.
I think when implementing this, the length should be taken into account, e.g. only half the screen height can be used and the longer part should be cropped.

ppwwyyxx:
Thanks for the comment. I am aware that screen_draw_overlay_text was mainly used for drawing IME pre-edited text. I'm giving it more capabilities here. Of course, it's debatable whether to redo the overlay or reuse the existing one.

> Pre-edited text disappears.
Cursor is in the wrong position after canceling input.

Thanks for mentioning this issue. I think the latest commit fixed it.

> During input, the displayed URL address flies to the cursor position.

I'm not sure I understand what it means. If I do the following:
1. hover on a URL
2. start typing with IME

The URL will just disappear and IME overlay will appear. This behavior is reasonable to me.

>  I believe that some common practices to prevent phishing should be used rather than just outputting the original address as is.

1. To prevent phishing, having the URL displayed is much better than not. So at least it's an improvement in security.
2. A user is not expected to be able to copy or click the displayed URL (I'd be surprised if it's copyable or clickable). In that case, it doesn't feel as necessary to me to apply sophisticated rules on how to display the URL.

> When I was trying to test "URLs that exceed the screen width and height", kitty just crashed.

I'm not able to reproduce this: 


The last character that's displayed in the overlay looks weird (this happens for IME as well), but I didn't observe a crash.

kovidgoyal:
I dont think using overlay line is the right way to do this. Instead use
existing facility to draw window titles when doing a visual window select. See
the function render_window_title.

Basically, the algorithm would be:

1) When hovering over a hyperlink (as opposed to a detected URL) set a
property on the Screen object with the URL
2) In draw_cells check if the property exists and if so use a function
similar to render_window_title to draw the URL
3) (2) should come with a setting in kitty.conf to turn it off
4) The location of the rendered URL should be either top or bottom of
window depending on where the cursor is currently.

page-down:
> I'm not sure I understand what it means.

Switch step 2 with step 1. This is for the first commit, and after the second commit I think you have resolved it.

In any case, if you use this, you want to make sure that at all states, you don't enter a state that is not expected. As kovid said, this is not a proper facility to implement this feature.

>> ... some common practices to prevent phishing should be used ...
>
> ... it doesn't feel as necessary to me to apply sophisticated rules on how to display the URL ...

I don't quite understand.
The reason you suggested the need for this feature is precisely to avoid phishing. And what I am proposing is the most basic thing that needs to be considered for this after implementing "Show URL".
To be honest, personally, I don't care much about fancy emoji domains, I just want to make sure that, for example, `U+0430` is distinguishable from `U+0061`.
And for the path part, if you insist on using URI encoding, then for Unicode this is not for human eyes and it makes no sense to display it.

kovidgoyal:
Regarding the display of the URL I suggest you just display it as is. I will take care of sanitizing it to make it useful for phishing detection.

ppwwyyxx:
Sorry, `render_window_title` looks quite difficult to me as I'm not familiar with glfw. 
> 2. In draw_cells check if the property exists and if so use a function similar to render_window_title to draw the URL

Would it make sense for me to do the following instead?
In `handle_move_event` if a hyperlink is found, use a function similar to `screen_draw_overlay_text` to draw the URL.

kovidgoyal:
On Wed, Dec 28, 2022 at 12:04:15AM -0800, Yuxin Wu wrote:
> Sorry, `render_window_title` looks quite difficult to me as I'm not familiar with glfw. 

You dont need to know glfw, just copy paste the existing
render_window_title function changing the use of title_bar_data to a new
url_status_data struct. It will draw at the top of the window with a
direct copy paste, you just need to adjust it to optionally draw at bottom instead depending on cursor position.

> > 2. In draw_cells check if the property exists and if so use a function similar to render_window_title to draw the URL
> 
> Would it make sense for me to do the following instead?
> In `handle_move_event` if a hyperlink is found, use a function similar to `screen_draw_overlay_text` to draw the URL.

The problem with using an overlay line is that, it moves with the text
when things change on screen. So there needs to be a whole
infrastructure to deal with that. That's a lot harder and more fragile
than just drawing the text separately with render_window_title().
Semantically this is basically a tooltip, so it should not be rendered
as part of the screen text, unlike the IME overlay.

ppwwyyxx:
@kovidgoyal thanks for the guidance. I added `render_hyperlink_preview` similar to `render_window_title`. Please let me know if it looks reasonable. Thanks!

kovidgoyal:
OK, I will review when I have a moment.

page-down:
@ppwwyyxx
I have tested the latest commit and found an issue.

When the mouse cursor is on top, the URL is still displayed on top, leading to the area pointed by the mouse being covered.

> The location of the rendered URL should be either top or bottom of
> window depending on where the cursor is currently.

I assume that we are talking about the mouse cursor here.

Steps to reproduce:
- Create a new window
- `ls --hyperlink=yes`
- Point to hyperlink with the mouse and you can see the URL displayed at the bottom.
- `ls -l; ls --hyperlink=yes`
- Use `ls` to output the text that fills the window. Now the URL is displayed at the top.
- Scroll the window to the top and move the mouse cursor to point to the hyperlink output by the first `ls` command located at the top of the window. However, the URL is still displayed at the top and covers the text that the mouse is pointing to.

PR #5858, opened 2023-01-06
Themes kitty: change `current-theme` but not conf

4e554c4c:
I change between themes often (twice a day). However, I also version-control my kitty config.
For this reason, I would like to change themes without having to update my kitty config.

This PR allows one to update only the `current-theme.conf` file when running `kitty +kitten themes`, and still have the ability to reload with the `--update-current` flag.

### DIFF
--- a/kittens/themes/collection.py
+++ b/kittens/themes/collection.py
@@ -530,20 +530,21 @@ class Theme:
     def save_in_dir(self, dirpath: str) -> None:
         atomic_save(self.raw.encode('utf-8'), os.path.join(dirpath, f'{self.name}.conf'))
 
-    def save_in_conf(self, confdir: str, reload_in: str, config_file_name: str = 'kitty.conf') -> None:
+    def save_in_conf(self, confdir: str, reload_in: str, config_file_name: str = 'kitty.conf', modify_conf: bool=True) -> None:
         os.makedirs(confdir, exist_ok=True)
         atomic_save(self.raw.encode('utf-8'), os.path.join(confdir, 'current-theme.conf'))
-        confpath = os.path.realpath(os.path.join(confdir, config_file_name))
-        try:
-            with open(confpath) as f:
-                raw = f.read()
-        except FileNotFoundError:
-            raw = ''
-        nraw = patch_conf(raw, self.name)
-        if raw:
-            with open(f'{confpath}.bak', 'w') as f:
-                f.write(raw)
-        atomic_save(nraw.encode('utf-8'), confpath)
+        if modify_conf:
+            confpath = os.path.realpath(os.path.join(confdir, config_file_name))
+            try:
+                with open(confpath) as f:
+                    raw = f.read()
+            except FileNotFoundError:
+                raw = ''
+            nraw = patch_conf(raw, self.name)
+            if raw:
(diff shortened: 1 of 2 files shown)


kovidgoyal:
You can do that already simply add include current-theme.conf into your version controlled kitty.conf. And pass --config-file-name=current-theme.conf to the kitten.

4e554c4c:
This is not the desired behavior. `--config-file-name` changes the file to add 
```
# BEGIN_KITTY_THEME
# <theme-name>
include current-theme.conf
# END_KITTY_THEME
```
If you run `kitty +kitten themes --config-file-name=current-theme.conf` it causes the error
```
Errors parsing configuration

In file /home/pounce/.config/kitty/current-theme.conf:
48:maximum recursion depth exceeded while calling a Python object in line: include current-theme.conf

48:maximum recursion depth exceeded while calling a Python object in line: include current-theme.conf
```
because it *changes current-theme.conf to include itself*.

I am proposing that there should be a way to modify `current-them.conf` with the desired theme, and reload kitty, but not modify the main config file, or add infinite include loops.

4e554c4c:
Of course this _can_ be solved using three files

you can do
```
# kitty.conf
include intermediary.conf
```
then
```
# intermediary.conf
# BEGIN_KITTY_THEME
# <theme-name>
include current-theme.conf
# END_KITTY_THEME
```
and then run 
```
$  kitty +kitten themes --config-file-name=intermediary.conf
```
but this is a bit messy, no?

kovidgoyal:
I dont follow. --config-file-name changes the file that the kitten
edits.

4e554c4c:
the kitten modifies two files. First, it modifies `current-theme.conf` and adds the desired theme to that file. Then, it modifies the config file specified by `--config-file`.

If these two files are the same, it creates an infinite loop

kovidgoyal:
Ah right I forgot the name of the actual generated theme file. So just
change the name in my advice above to something else.

4e554c4c:
Yes, so how I said above you can do this with three files: `kitty.conf`, `intermediary.conf` and `current-them.conf` `intermediary.conf` is a file that *only* contains an include to `current-them.conf`?

Is this the desired way to solve the problem? I think it would be cleaner with two files (what this PR allows for)

kovidgoyal:
Yes, that's the desired way. In order to operate robustly the kitten
does more than just add the include it also comments out existing color
definitions see the patch_conf() function. Therefore if you really want
it to operate with only two files, the correct way would be to add an
option to have it not include the theme name in kitty.conf. That way it
when it changes the theme there will be no actual changes to kitty.conf
unless, the user has added some colors to since the last time it was 
run kitty.conf.

kovidgoyal:
Oh and what you want to do is already possible with --dump-theme

kitty +kitten themes --dump-theme whatever > ~/.config/kitty/current-theme.conf

4e554c4c:
The problem I have with `--dump-theme` is that it doesn't allow you to reload

kovidgoyal:
You can run

kitty +runpy "from kitty.utils import *; reload_conf_in_all_kitties()"

which is the same code the kitten uses.

PR #5883, opened 2023-01-13
Improve appearance of undercurl

Dragon-Hatcher:
Make the undercurl thicker and have its amplitude scale more closely with font size. Also half the period so there are two peaks per character.

I feel that the way it is right now looks too flat and too thin. Imo this change will look far better for editors using the undercurl for error messages.

Before and After
image

See also VSCode's undercurl. 
image

I think this looks even better but the period isn't a multiple of the character width and so it can't be implemented using the current kitty rendering strategy. (Summary: ~1.5 curls / char > 2 curls / char > 1 curl / char).

### DIFF
--- a/kitty/fonts/render.py
+++ b/kitty/fonts/render.py
@@ -249,16 +249,8 @@ def add_dline(buf: CBufType, cell_width: int, position: int, thickness: int, cel
 
 def add_curl(buf: CBufType, cell_width: int, position: int, thickness: int, cell_height: int) -> None:
     max_x, max_y = cell_width - 1, cell_height - 1
-    xfactor = 2.0 * pi / max_x
-    thickness = max(1, thickness)
-    if thickness < 3:
-        half_height = thickness
-        thickness -= 1
-    elif thickness == 3:
-        half_height = thickness = 2
-    else:
-        half_height = thickness // 2
-        thickness -= 2
+    xfactor = 4.0 * pi / max_x
+    thickness = half_height = max(1, cell_height // 21)
 
     def add_intensity(x: int, y: int, val: int) -> None:
         y += position
--- a/kitty/fonts/render.py
+++ b/kitty/fonts/render.py
@@ -206,7 +206,7 @@ def set_font_family(opts: Optional[Options] = None, override_font_size: Optional
     if debug_font_matching:
         dump_faces(ftypes, indices)
     set_font_data(
-        render_box_drawing, lambda *a: prerender_function(*a, opts), descriptor_for_idx,
+        render_box_drawing, prerender_function, descriptor_for_idx,
         indices['bold'], indices['italic'], indices['bi'], num_symbol_fonts,
         sm, sz, font_features, ns
     )
@@ -247,8 +247,9 @@ def add_dline(buf: CBufType, cell_width: int, position: int, thickness: int, cel
         ctypes.memset(ctypes.addressof(buf) + (cell_width * y), 255, cell_width)
 
 
-def add_curl(buf: CBufType, cell_width: int, position: int, thickness: int, cell_height: int, opts: Options) -> None:
-    (thickness_opt, density_opt) = opts.undercurl_style
+def add_curl(buf: CBufType, cell_width: int, position: int, thickness: int, cell_height: int) -> None:
+    from kitty.fast_data_types import get_options
+    (thickness_opt, density_opt) = get_options().undercurl_style
 
     max_x, max_y = cell_width - 1, cell_height - 1
     xfactor = (2.0 if density_opt == 'sparse' else 4.0) * pi / max_x
@@ -310,8 +311,7 @@ def render_special(
     strikethrough_position: int = 0,
     strikethrough_thickness: int = 0,
     dpi_x: float = 96.,
-    dpi_y: float = 96.,
-    opts: Options = None
+    dpi_y: float = 96.
 ) -> CBufType:
     underline_position = min(underline_position, cell_height - sum(divmod(underline_thickness, 2)))
     CharTexture = ctypes.c_ubyte * (cell_width * cell_height)
@@ -333,7 +333,7 @@ def render_special(
         t = underline_thickness
         if underline > 1:
             t = max(1, min(cell_height - underline_position - 1, t))
-        dl([add_line, add_line, add_dline, lambda *a: add_curl(*a, opts), add_dots, add_dashes][underline], underline_position, t, cell_height)
+        dl([add_line, add_line, add_dline, add_curl, add_dots, add_dashes][underline], underline_position, t, cell_height)
     if strikethrough:
         dl(add_line, strikethrough_position, strikethrough_thickness, cell_height)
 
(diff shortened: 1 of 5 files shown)


kovidgoyal:
This is very much a matter of personal preference. For instance I like my
error indications to be subtle. I am OK with adding an option to control
this, say

undercurl_style thin|thick

Dragon-Hatcher:
That sounds reasonable to me. Can you point to any places in the codebase where config options are implemented so I can learn how to make them? And do you think it should be one option for both thickness and period or two seperate options?

Also how would you feel about keeping the amplitude change either way? Right now the amplitude noticebly changes as you increase the font size. (i.e. the amplitude becomes a smaller percentage of the text size as it gets bigger).

Just amplitude changes. Before, After
image
image
image

kovidgoyal:
You add options by editing options/definition.py then running
./gen-config.py

I definitely dont want two options for this. You can have the option
take values like 

(thin|thick)-(sparse|dense)

The first word controlling line thickness and the second wave length.
Feel free to choose better words if you can think of any.

Leave the amplitude changes in for the moment, I will review them in
detail and make a final decision later.

Dragon-Hatcher:
Ok great. I'll make those changes when soon.

Dragon-Hatcher:
Alright. I've made it an option. The lambda use feels a little messy but it seems like the easiest way to get the options to the relevant functions considering stuff gets passed through various C apis. Let me know if you can think of a better way to do it. I left the amplitude as the changed version for now as discussed.

kovidgoyal:
You don't need to pass opts around to the functions, it can be imported
like this

```py
from kitty.fast_data_types import get_options
opts = get_options()
```

Where does the magic constant 21 come from?

undercurl_style should be simplified to:

```py
def undercurl_style(x: str) -> Tuple[str, str]:
    width, density = x.partition('-')[::2]
    if width not in ('thin', 'thick'):
        raise ValueError(f'Invalid undercurl thickness: {width!r} allowed values are thick and thin')
    if density not in ('dense', 'sparse'):
        raise ValueError('Invalid undercurl density: {density!r} allowed values are sparse and dense')
    return width, density
```

Dragon-Hatcher:
I've made those changes. Thanks for the tips. I don't do much python.

21 doesn't come from anywhere. I just picked a number that loocked good. I'd be open to a more rigorous suggestion.

kovidgoyal:
Well, the current way it is done is it uses the underline thickness defined by the font file as the height. Underlines are supposed to be that thick. For the case of a curly underline, one can argue that underline thickness is not the correct metric. In which case one has to come up with something else. You have available to you the underline thickness, position and cell height. I suggest using something like twice underline thickness capped to the amount of space between the underline position and the bottom of the cell.

Also, thinking about it, since we only have four values you can ditch the special parse function, and just use choices=(...) in definition.py. That gives us type-checking for free. In add curl you can just do, thick = 'thick' in opts.underline_style and dense = 'dense' in underline_style.

Dragon-Hatcher:
I'm having a problem now where gen-config.py is modifiying a whole bunch of files even after changing nothing. It seems to be just minor stuff like different comments. I won't change the config stuff until I can get that figured out. I'm looking at the rendering changes now.

kovidgoyal:
Dont bother with the config I can do it.

Dragon-Hatcher:
I'm really sorry. I'm having a problem now where I can't run kitty using `KITTY_DEVELOP_FROM`. 

I'm getting this.
```
  File "runpy.py", line 197, in _run_module_as_main
  File "runpy.py", line 87, in _run_code
  File "kitty_main.py", line 7, in <module>
  File "/home/xxxxx/Documents/Programming/kitty/kitty/entry_points.py", line 180, in main
    from kitty.main import main as kitty_main
  File "/home/xxxxx/Documents/Programming/kitty/kitty/main.py", line 12, in <module>
    from .boss import Boss
  File "/home/xxxxx/Documents/Programming/kitty/kitty/boss.py", line 57, in <module>
ImportError: cannot import name 'cocoa_hide_app' from 'kitty.fast_data_types' (/home/xxxxx/.local/kitty.app/lib/kitty-extensions/kitty.fast_data_types.so)
```

This didn't happen before and I don't think I've changed anything. Isn't cocoa a mac library? I'm on linux.

kovidgoyal:
yes, but it is present on Linux as a no-op. Importing it shouldnt fail.
What binary of kitty are you using? Maybe run the nightly.

Dragon-Hatcher:
I'm just going to test rendering options on the commit of the latest release.

Dragon-Hatcher:
How do you feel about keeping the amplitude based on the cell height? The problem is if you don't do that it get's really flat.

kovidgoyal:
You really cant have an underline rendered outside the designated area for the underline by the font. That could cause all sorts of issues. The designated area is basically underline_position - underline_thickness  // 2 to bottom of cell. You can potentially use that entire height but going beyond that is problematic, IMO.

Dragon-Hatcher:
How about this:? (ignoring the options for now)
```python
max_height = cell_height - (position - thickness // 2)
desired_amplitude = max(1, cell_height // 21)
thickness = half_height = min(max(1, desired_amplitude), max_height // 2)
```

kovidgoyal:
Its workable though magic constants really make me uncomfortable. Since
it yields the same rendering at smaller font sizes maybe document 21 as
being the ratio of underline_thickness to cell height at some font size
for some font.

Dragon-Hatcher:
I've been testing with Fira Code size 13.0. Where were you thinking of putting that documentation. In the option description?

kovidgoyal:
On Sat, Jan 14, 2023 at 10:39:38PM -0800, Dragon-Hatcher wrote:
> I've been testing with Fira Code size 13.0. Where were you thinking of putting that documentation. In the option description?

No, in the add_curl function. 

The more I think about this the more uncomfortable it makes me. The
relevant metric should be the descent (distance from baseline to bottom
of cell) not the cell height. That is fundamentally the space we have to
render underlines. We dont know how different fonts will scale the
position of the baseline at different sizes. So we should be using that
rather than the cell_height as the numerator.
And then the magic constant will likely be something between 1 and 2.
AKA use the full available height or half of it.

Dragon-Hatcher:
This seems to work pretty well:
```python
max_height = cell_height - (position - thickness // 2)
thickness = half_height = max(1, max_height // 4)
```

kovidgoyal:
looks OK to me, just as a safety for small sizes maybe make the lower bound 

```py
min(max(1, underline_thickness), max_height // 4))
```

and have the existing logic for thickness <= 3

Dragon-Hatcher:
Alright. Alltogether:
```python
opts = get_options()
thick = 'thick' in opts.undercurl_style
dense = 'dense' in opts.undercurl_style

max_x, max_y = cell_width - 1, cell_height - 1
xfactor = (4.0 if dense else 2.0) * pi / max_x

max_height = cell_height - (position - thickness // 2)
half_height = max(1, max_height // 4)
thickness = max(1, thickness) - (1 if thickness < 3 else 2) # thin
if thick: thickness = max(half_height, thickness) # thick
```

Dragon-Hatcher:
How do you feel about `xfactor` -> `period` and `half_height` -> `amplitude`?

kovidgoyal:
You're most welcome!

Issue #5716, opened 2022-11-28
Re: Japanese input bug?（#3330）


ghsable:
- https://github.com/kovidgoyal/kitty/issues/3330
  - https://github.com/kovidgoyal/kitty/issues/3330#issuecomment-1273590741
  - https://github.com/kovidgoyal/kitty/issues/3330#issue-812568792
    - `vim` -> NG, `nano` -> OK

astralhpi:
PR #5886 seems to be able to solve it. Can you check if it reproduces in the code in my repo? https://github.com/astralhpi/kitty

page-down:
The current nightly release now contains IME related fixes and is available to try.

https://github.com/kovidgoyal/kitty/releases/tag/nightly

@ompugao

I see that you reported the problem in original issue 3330,
and it would be helpful if you could try to see if this issue can still be reproduced.
Thanks.

ompugao:
The problem I commented before has been fixed!
Thank you for your contribution!

Issue #5027, opened 2022-04-28
Korean pre-edit completing mechanism needs upgrade


exploitJ:
### **Is your feature request related to a problem? Please describe.**

Current implementation takes Korean input similar to Chinese or Japanese.
However, due to Korean's unique nature of composing characters(Jaso), current method of typing is really frustrating.

To name a few bugs that I'm facing,
1. Letter in pre-edit mode is lost when kitty is out of focus.
2. Every keys other than letter keys exits pre-edit mode and just ignored.
3. Korean inputs in vim are overlapped on each other. (it's not related to pasting a complete korean word) - I'll make a separate issue here and vim's.

All issue above is reproduced with `kitty --config NONE`.
Currently, I using macos but I'll test linux version too.

### **Describe the solution you'd like**
Unlike Chinese, Korean doesn't need "completing" by explicitly pressing \<space\> or \<tab\>.

Every combination of character should make a valid character.
Either "Gulja: a letter" or "Jaso: composing unit" itself or together is a valid input.

To simply put, *every* other keys should work right away either a letter is complete or not; by this point, it *should* be complete.

As a result, instead of completing a letter and doing nothing,
- white space keys(\<space\>, \<tab\>, \<return\>)
- any navigation keys(arrow keys, home, page down, etc.)
- deleting keys(backspace, delete)
- control keys(esc, f1~f12, \<ctrl-c\>, \<ctrl-d\>)
should complete the letter then register by itself.

losing focus should complete it too.

### **Describe alternatives you've considered**
I couldn't think of some hacky workaround to this issue.

### **Additional context**
There are other korean related bugs unrelated to this request.
I understand that debugging and developing feature related to regional stuff is hard.
I'm very much willing to contribute PRs as long as you allow me to.

kovidgoyal:
You are most welcome to send PRs. If you need pointers feel free to ask
in this bug report. Closing but feel free to follow up.

kovidgoyal:
On Thu, Apr 28, 2022 at 09:04:45AM -0700, Serim Son wrote:
> 1. which files should I start looking for?

look in glfw/cocoa_window.m (keyDown function) and kitty/glfw.c, kitty/keys.c

> 2. Do you think this issue is likely solvable only on python side or both c and python?

You will need objective-c and C there is no python involved in this code
path.

> 3. Any set of rules that I must follow?

No.

PR #5886, opened 2023-01-15
Also move the overlay line when DECTCEM mode is turned off and on

astralhpi:
When we type something in vim, it send following commands to the terminal.

1. screen_reset_mode 25 1 (Turn off DECTCEM mode)
2. draw {charater}
3. screen_set_mode 25 1 (Turn on DECTCEM mode)

This flow causes an issue when using some IMEs like Korean in Kitty. Korean IMEs generate character confirmation and overlay update with a single key press. However, due to that flow, the overlay position is not updated at the time of the draw key.

To solve this, I modified it to save the overlay line when DECTCEM mode is turned off, and restore it when the mode is turned back on.

Related issues: #5716  #5027

### before
https://user-images.githubusercontent.com/1402102/212526561-3ea62086-a731-447d-af27-8de30a1e9c20.mov

### after
https://user-images.githubusercontent.com/1402102/212526569-ee08fea1-0749-48c9-bf9a-83f44d4c1896.mov

### DIFF
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -751,7 +751,7 @@ struct SaveOverlayLine {
 
 static void
 save_overlay_line(struct SaveOverlayLine *sol) {
-    if (sol->screen->overlay_line.is_active && screen_is_cursor_visible(sol->screen)) {
+    if (sol->screen->overlay_line.is_active) {
         sol->overlay_text = get_overlay_text(sol->screen);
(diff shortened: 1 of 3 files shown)


kovidgoyal:
merged.

kovidgoyal:
Apparentlyt his causes issues with Chinese IME se #5996

Issue #2053, opened 2019-10-14
How to run ` kitty +kitten ssh ...` for second-level servers in chain ?


collegeimprovements:
Hi Sir,
Thanks a lot for your work on kitty. It's super fast.

Provided we have two servers: 
- `jump_server` 
- `destination_server`

And `destination_server` can only be accessed from `jump_server`.

When I run ` kitty +kitten ssh username@jump_server` kitty works great. 
(e.g. `clear` command works.)

How can we run `kitty +kitty ssh username@destination_server` from our local-machine or from `jump_server` ? 
Currently when we run `clear` command we get `'xterm-kitty': unknown terminal type.` error.
This situation arises because we can't download `kitty` on `jump_server`.

kovidgoyal:
You cannot. Instead copy the terminfo files manually, as described in https://sw.kovidgoyal.net/kitty/faq.html#i-get-errors-about-the-terminal-being-unknown-or-opening-the-terminal-failing-when-sshing-into-a-different-computer

nertpinx:
With new enough ssh client you should be able to do `kitty +kitten ssh -J jump_server destination_server`.

glebsam:
> With new enough ssh client you should be able to do `kitty +kitten ssh -J jump_server destination_server`.

does not work for me unfortunately

nertpinx:
> > With new enough ssh client you should be able to do `kitty +kitten ssh -J jump_server destination_server`.
> 
> does not work for me unfortunately

Well, then I guess you can create an ssh tunnel `ssh -L localhost:12345:machine2:22 machine1` and then in another shell do `kitty +kitten localhost -p 12345`. (written from the top of my head, not tested).

glebsam:
Yep, it seems to me will work. But my intention was to use the kitten as drop-in replacement for usual SSH command. Thanks for you answer, anyway!

glebsam:
I have finally figured out what's wrong with my scenario. Issue #2779 helped me. Indeed, I used command 

```
ssh -At -J user@jumper.internal.com user@target-host.internal.com sudo su
``` 

Command `sudo su` is a key to the riddle: ssh kitten will only copy terminfo for user under which the connection was created, ignoring permissions elevation made by just a command (which seems a sane design decision).

chapmanjacobd:
just in case people aren't aware you can install the terminfo files like so in many distros:

```
sudo apt install kitty-terminfo
```

PR #5888, opened 2023-01-16
Document kitty-terminfo package for getting terminfo to work with sudo when SSH'd

ZimbiX:
I'd followed the instructions for using the SSH kitten ages ago, but I've always had trouble with it when SSH-ing to my server, e.g.:

```
➜ sudo systemctl status crashplan-pro
WARNING: terminal is not fully functional
Press RETURN to continue
```

I only realised today that it was only `sudo` commands that were failing this way. In Googling, I came across [this post](https://bbs.archlinux.org/viewtopic.php?id=181907) by @EvanPurkhiser. In case the thread gets removed, here's the content:

> Hey guys,
>
> I use the termite terminal which uses the termcap file 'xterm-termite'. Since this is non-standard I have to copy over the termcap file to remote machines I commonly work on. I put this file in `$HOME/.local/share/terminfo/x/xterm-termite` and set my $TERMINFO environment variable to `$HOME/.local/share/terminfo`.
>
> When I use sudo -E or sudoedit it looks like the TERMINFO environment variable is getting lost. So vim won't display things properly (no colors) and other silly problems (LESS 'Warning terminal is not fully functional").
>
> Why does the TERMINFO get dropped when my other environment variables stick around?
>
> Edit: Looking at the [TROUBLESHOOTING](http://www.sudo.ws/sudo/troubleshooting.html) document, I see that TERMINFO is removed "to guard against shared library spoofing, shell voodoo, and kerberos server spoofing."
>
> What are my options? I would really rather not copy my termcap file into /usr/share/terminfo since I might not always have root access on my remote machines.
>
> Edit 2: I feel a little silly. I just said I may not have root access. But if I'm using sudo it's almost guaranteed for root access! I'll leave this thread open for other solutions though.

Copying the file into the system path did indeed solve my issue! But I wondered if there was a package for it instead; and [it turns out there is](https://github.com/kovidgoyal/kitty/issues/2053#issuecomment-995349712) - `kitty-terminfo`.

Unfortunately, it doesn't appear to be documented anywhere - the [only mention of the package is in the 'Build from source' page](https://github.com/search?q=repo%3Akovidgoyal%2Fkitty%20kitty-terminfo&type=code). I thought documenting this package would be a valuable addition to the SSH FAQ section; hence this PR.

I considered whether putting it in the section below would be more appropriate since it's talking about `sudo`, but it seems to be addressing a slightly different problem around environment variables; and given the title is not SSH-related, I had not realised it might be applicable to my problem.

The sudo troubleshooting page linked from the above post no longer mentions terminfo, so I have not included reference to that.

I'm not familiar with reStructuredText, so I hope I got the syntax right and to your liking.

Edit: Perhaps this should also be mentioned on the [SSH kitten page](https://sw.kovidgoyal.net/kitty/kittens/ssh/)?

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -135,6 +135,12 @@ command to apply your change (on the server)::
 
     cap_mkdb /usr/share/misc/termcap
 
+For terminfo to work with :program:`sudo` on the server, you'll need to either
+install the Linux distribution's |kitty-terminfo| package on the server, or copy
+:file:`~/.terminfo/x/xterm-kitty` to the server's system path::
+
+    sudo cp ~/.terminfo/x/xterm-kitty /usr/share/terminfo/x/
+
 
 Keys such as arrow keys, backspace, delete, home/end, etc. do not work when using su or sudo?
 -------------------------------------------------------------------------------------------------


kovidgoyal:
kitty-terminfo is distro specific, it's not present in all distros, and
all operating systems, so one cant really use it as general advice. 
There is already a section in the FAQ on dealing with this issue for sudo

https://sw.kovidgoyal.net/kitty/faq/#keys-such-as-arrow-keys-backspace-delete-home-end-etc-do-not-work-when-using-su-or-sudo


That tells you how to workaround the fact that sudo stupidly strips
TERMINFO from the environment. You dont actually need to install kitty's
terminfo system-wide.

ZimbiX:
I see. You are remarkably curt, my friend.

While you prefer that solution, I think it is simpler to install the package. Yes, it may not be available for the distro, or be under a different name; I'd meant "the Linux distribution's |kitty-terminfo| package" to cover that. Would alternative wording be acceptable? (why even have a package if it's not mentioned anywhere)

Regardless, the 'Keys such as arrow keys, backspace, delete, home/end, etc. do not work when using su or sudo?' section title is not obviously applicable when the problem is `WARNING: terminal is not fully functional`. Could this be addressed another way?

kovidgoyal:
On Mon, Jan 16, 2023 at 01:25:40AM -0800, Brendan Weibrecht wrote:
> I see. You are remarkably curt, my friend.

You are expecting me to regurgitate my life story on every piece of the
approx 100 bits of user support I have to do *every single day*, for
free?

> 
> While you prefer that solution, I think it is simpler to install the package. Yes, it may not be available for the distro, or be under a different name; I'd meant "the Linux distribution's |kitty-terminfo| package" to cover that. Would alternative wording be acceptable? (why even have a package if it's not mentioned anywhere)

It might be *your* preferred solution, it's not generally applicable
thus it does not belong in a FAQ. 

> 
> Regardless, the 'Keys such as arrow keys, backspace, delete, home/end, etc. do not work when using su or sudo?' section title is not obviously applicable when the problem is `WARNING: terminal is not fully functional`. Could this be addressed another way?

How? I am certainly not going to list every single symptom that can
occur due to missing terminfo.

kovidgoyal:
No worries, I think people just get put off by the closing of
issues/PRs. It doesn't mean I am not open to more discussion, it's just
the most efficient way for me to manage my workload.
 As for this issue I think the FAQ should be re-organized merging the
 two question into one since they are really the same underlying issue.

ZimbiX:
That's understandable. Maybe adding something like 'I'm still open to other ideas' would help with that. If you'd prefer not saying these same things every time, perhaps you could add a note to the [contributing document](https://github.com/kovidgoyal/kitty/blob/master/CONTRIBUTING.md) on how to interpret your responses.

Would you like me to have a go at merging the two FAQ sections, or would you prefer to do that yourself?

kovidgoyal:
On Mon, Jan 16, 2023 at 03:41:16AM -0800, Brendan Weibrecht wrote:
> That's understandable. Maybe adding something like 'I'm still open to other ideas' would help with that. If you'd prefer not saying these same things every time, perhaps you could add a note to the [contributing document](https://github.com/kovidgoyal/kitty/blob/master/CONTRIBUTING.md) on how to interpret your responses.
> 
> Would you like me to have a go at merging the two FAQ sections, or would you prefer to do that yourself?

Already, done https://github.com/kovidgoyal/kitty/commit/61143557a9aea9b35e2fb4f8150b4bc5cfb80c5d

kovidgoyal:
And https://github.com/kovidgoyal/kitty/commit/59170a36276f00cccddc7bd4cd5f5fa5d385bffd

ZimbiX:
Ah, well done; that looks great! Thanks for incorporating all my suggestions =) I just sent you a tip.

I see the changes haven't rolled out just yet. Out of curiosity, what deploys the website? I see the [GitHub Pages repo is here](https://github.com/kovidgoyal/kovidgoyal.github.io) - is it manual?

kovidgoyal:
Thanks. website is automatically updated when a new version is released.
See publish.py

PR #5922, opened 2023-01-23
Fix SHELL env var in open-actions.conf and launch-actions.conf

page-down:
When `shell` is configured as `bash --login`, `kitty +shebang $FILE_PATH $SHELL` will not work.
The PR fixes this.

- Only the executable path is used when expanding `$SHELL`.
- Also fix the non-working example `{SHELL}` in [the documentation](https://sw.kovidgoyal.net/kitty/open_actions/#scripting-the-opening-of-files-with-kitty-on-macos).

I actually tried to fix the problem of not being able to open image files with kitty under macOS.
However I don't know when the next version will be released, so it's better to let you know.

Steps to reproduce:
- Right click on the image file to open it via kitty.app, or drag and drop the file onto the Dock.
- Or run `kitty @ launch --type=os-window kitty +kitten icat --hold /path/to/image`.
- The new process ends and the new OS window is immediately closed. （`This should be run as kitten icat`)

Issues:
- `icat --hold` is not respected in the python placeholder (`raise SystemExit()`), causing the window to close directly.
- `launch` uses the prewarmed process to run kitten. I think for all `wrapped_kitten_names()` the `prewarm` should be skipped.
- The `launch --hold` uses a python kitten to wrap, so it will run the process independently. `launch --hold kitty +kitten icat` will work fine. I'm not sure if we can use `prewarm` in this case (to make the kitten after `launch --hold` run faster), it seems infeasible and unnecessary. This `hold_till_enter` then looks like it can be implemented in golang.

Lastly, how to add `help` and `exit` auto-completion to the kitty shell?

Thank you.

### DIFF
--- a/docs/open_actions.rst
+++ b/docs/open_actions.rst
@@ -68,6 +68,13 @@ some special variables, documented below:
     The path, query and fragment portions of the URL, without any
     unquoting.
 
+``EDITOR``
+    The terminal based text editor. The configured :opt:`editor` in
+    :file:`kitty.conf` is preferred.
+
+``SHELL``
+    The path to the shell. The configured :opt:`shell` in :file:`kitty.conf` is
+    preferred, without arguments.
 
 .. note::
    You can use the :opt:`action_alias` option just as in :file:`kitty.conf` to
--- a/kitty/open_actions.py
+++ b/kitty/open_actions.py
@@ -74,7 +74,7 @@ def parse(lines: Iterable[str]) -> Iterator[OpenAction]:
     with to_cmdline_implementation.filter_env_vars(
         'URL', 'FILE_PATH', 'FILE', 'FRAGMENT', 'URL_PATH',
         EDITOR=shlex.join(get_editor()),
-        SHELL=shlex.join(resolved_shell(get_options()))
+        SHELL=resolved_shell(get_options())[0]
     ):
         for (mc, action_defns) in entries:
             actions: List[KeyAction] = []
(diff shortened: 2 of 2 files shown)


kovidgoyal:
On Sun, Jan 22, 2023 at 10:41:08PM -0800, page-down wrote:
> - `icat --hold` is not respected in the python placeholder (`raise SystemExit()`), causing the window to close directly.
> - `launch` uses the prewarmed process to run kitten. I think for all `wrapped_kitten_names()` the `prewarm` should be skipped.

These are both fixed.

> - The `launch --hold` uses a python kitten to wrap, so it will run the process independently. `launch --hold kitty +kitten icat` will work fine. I&#39;m not sure if we can use `prewarm` in this case (to make the kitten after `launch --hold` run faster), it seems infeasible and unnecessary. This `hold_till_enter` then looks like it can be implemented in golang.

No, it already uses kitten __hold_till_enter__

> 
> Lastly, how to add `help` and `exit` auto-completion to the kitty shell?

Done.

page-down:
@kovidgoyal
There is also an auto-completion issue with the kitty shell.
`launch kitty +kitten icat ... ` cannot be completed, while it is possible in the integrated 3rd-party shells.

The `launch` command is still quite commonly used, so please evaluate if it is worth fixing, and if not, kindly ignore the report for now, as it will take time on something with low priority.

Of course there are problems such as `subcmd --password-file` not being able to complete the file path, but these are not important at all.

kovidgoyal:
On Mon, Jan 23, 2023 at 03:33:24AM -0800, page-down wrote:
> @kovidgoyal
> There is also an auto-completion issue with the kitty shell.
> `launch kitty +kitten icat ... ` cannot be completed, while it is possible in the integrated 3rd-party shells.
> 
> The `launch` command is still quite commonly used, so please evaluate if it is worth fixing, and if not, kindly ignore the report for now, as it will take time on something with low priority.

I'm afraid not while one could conceivably implement completion for
launch kitty... it would not be possible to complete any arbitrary
command, in the 3rd party shells completion for these works by
delegating to the shell's builtin completion. There is nothing to
delegate to here.

> 
> Of course there are problems such as `subcmd --password-file` not being able to complete the file path, but these are not important at all.

Fixed.

PR #5929, opened 2023-01-25
Update known textual mimes to Python

page-down:
Also add default aliases for JavaScript and XML in Go.

Should `kitten` (in Go) read `~/.config/kitty/mime.types`?

This will make the user's definition available.
```text
text/plain desktop
```

Is it better to synchronize `known_extensions` already defined in Python to Go? (DRY)
For example: `'pyj': 'text/rapydscript-ng'`

---

Could you re-upload the GPG pubkey to GitHub when you have time? Thanks.
https://github.com/kovidgoyal.gpg

```text
-----BEGIN PGP PUBLIC KEY BLOCK-----
Note: The keys with the following IDs couldn't be exported and need to be reuploaded 06BC317B515ACE7C
...
```

If you don't care, please ignore it, I'm just telling you about this problem I found out years ago.


kovidgoyal:
On Wed, Jan 25, 2023 at 12:58:23AM -0800, page-down wrote:
> Also add default aliases for JavaScript and XML in Go.
> 
> Should `kitten` (in Go) read `~/.config/kitty/mime.types`?

Done.

> Is it better to synchronize `known_extensions` already defined in Python to Go? (DRY)
> For example: `&#39;pyj&#39;: &#39;text/rapydscript-ng&#39;`

That's actually a misfeature, pyj is something I use personally, it
really belongs in ~/.config/kitty/mime.types. It's not important enought
to bother changing.

> 
> ---
> 
> Could you re-upload the GPG pubkey to GitHub when you have time? Thanks.
> https://github.com/kovidgoyal.gpg

Done.

Issue #5956, opened 2023-01-31
edit-in-kitty does not work


grimm26:
**Describe the bug**
The edit-in-kitty command fails.

**To Reproduce**
1. use kitty 0.27.0 on linux
2. try to edit a file using edit-in-kitty
3. See error: Command 'kitten' not found

**Screenshots**

![image](https://user-images.githubusercontent.com/127401/215786834-e860c42d-68d9-42ac-9f42-d2db3b719646.png)

**Environment details**
```
 kitty 0.27.0 () created by Kovid Goyal
Linux mnk-hpbook 5.15.0-58-generic #64-Ubuntu SMP Thu Jan 5 11:43:13 UTC 2023 x86_64
Ubuntu 22.04.1 LTS mnk-hpbook /dev/tty

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"
Running under: Wayland
Frozen: True
Paths:
  kitty: /home/mkeisler/.local/kitty.app/bin/kitty
  base dir: /home/mkeisler/.local/kitty.app/lib/kitty
  extensions dir: /home/mkeisler/.local/kitty.app/lib/kitty-extensions
  system shell: /bin/zsh
Loaded config files:
  /home/mkeisler/.config/kitty/kitty.conf

Config options different from defaults:
copy_on_select                     clipboard
cursor_shape                       3
cursor_stop_blinking_after         0
editor                             nvim
enable_audio_bell                  False
enabled_layouts                    ['grid', 'fat:bias=70;full_size=1;mirrored=false', 'splits', 'tall', 'stack']
env:
{'PATH': '/home/mkeisler/.local/bin:/usr/local/bin:/home/mkeisler/.pyenv/shims:/home/mkeisler/.pyenv/bin:/home/mkeisler/.local/bin:/home/mkeisler/.cargo/bin:/usr/local/bin:/usr/local/sbin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/mkeisler/go/bin:/usr/local/go/bin:/usr/local/nodejs/bin'}
focus_follows_mouse                True
font_family                        Hack Nerd Font Mono
font_size                          12.0
initial_window_height              (600, 'px')
initial_window_width               (1000, 'px')
macos_quit_when_last_window_closed True
remember_window_size               False
[...]

PR #5962, opened 2023-02-01
fish integration: Ensure kitten works even if it is not in PATH

page-down:
Also fix duplicate subcommand names in `Usage`.
```text
Usage: kitten edit-in-kitty edit-in-kitty [options] file-to-edit
```

Env var `KITTY_KITTEN_EXE` is not available when shell integration is **manually enabled**, is this intentional?

Why not use `KITTY_BIN_DIR` and add the bin directory to the PATH?
This will make `kitty` also available, are you not wanting to deal with "don't ever touch my PATH"?

I'll leave the bash/zsh completion to you.

`KITTY_KITTEN_EXE` is used only if `kitten` is not defined by the user or is not in the PATH.
If you think this env var should override the user's `kitten` function or path (like prepending KITTY_BIN_DIR to PATH), please let me know.

Related issue:
https://github.com/kovidgoyal/kitty/issues/5956

### DIFF
--- a/kitty/constants.py
+++ b/kitty/constants.py
@@ -79,6 +79,7 @@ def kitty_exe() -> str:
     return os.path.join(rpath, 'kitty')
 
 
+@run_once
 def kitten_exe() -> str:
     return os.path.join(os.path.dirname(kitty_exe()), 'kitten')
 
--- a/shell-integration/fish/vendor_completions.d/clone-in-kitty.fish
+++ b/shell-integration/fish/vendor_completions.d/clone-in-kitty.fish
@@ -1,7 +1,7 @@
 function __ksi_completions
     set --local ct (commandline --current-token)
     set --local tokens (commandline --tokenize --cut-at-cursor --current-process)
-    printf "%s\n" $tokens $ct | command kitten __complete__ fish | source -
+    printf "%s\n" $tokens $ct | kitten __complete__ fish | source -
 end
 
 complete -f -c clone-in-kitty -a "(__ksi_completions)"
--- a/shell-integration/fish/vendor_completions.d/kitten.fish
+++ b/shell-integration/fish/vendor_completions.d/kitten.fish
@@ -1,7 +1,7 @@
 function __ksi_completions
     set --local ct (commandline --current-token)
     set --local tokens (commandline --tokenize --cut-at-cursor --current-process)
-    printf "%s\n" $tokens $ct | command kitten __complete__ fish | source -
+    printf "%s\n" $tokens $ct | kitten __complete__ fish | source -
 end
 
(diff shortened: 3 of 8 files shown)


kovidgoyal:
Thanks, I decided to just add the kitty bin dir to PATH if not present. I merged the rest of your changes.

page-down:
Some of the users put kitty under `/whatever/path` and then created a symbolic link for kitty, `/bin/kitty`, and reset the PATH in the rc file (`export PATH=/bin`). At this point `edit-in-kitty` does not work.
As that requires the PATH to be modified again after the user‘s shell configuration is executed.

So in this case the user should fix their PATH by themselves or add the kitten symbolic link.

To clarify, the recent commit addresses the case when the PATH is not overridden.

kovidgoyal:
yeah if the user resets PATH in their shell rc files, then its up to
them to make sure kitty and kitten are in that PATH

salmankhilji:
I am wondering if there might be a typo in the following commit:

```diff
diff --git a/docs/binary.rst b/docs/binary.rst
index 1559fb0b..c1f09bad 100644
--- a/docs/binary.rst
+++ b/docs/binary.rst
@@ -46,9 +46,9 @@ particular desktop, but it should work for most major desktop environments.
 
 .. code-block:: sh
 
-    # Create a symbolic link to add kitty to PATH (assuming ~/.local/bin is in
+    # Create symbolic links to add kitty and kitten to PATH (assuming ~/.local/bin is in
     # your system-wide PATH)
-    ln -s ~/.local/kitty.app/bin/kitty ~/.local/bin/
+    ln -sf ~/.local/kitty.app/bin/kitty ~/.local/kitty.app/bin/kitty ~/.local/bin/
     # Place the kitty.desktop file somewhere it can be found by the OS
     cp ~/.local/kitty.app/share/applications/kitty.desktop ~/.local/share/applications/
     # If you want to open text files and images in kitty via your file manager also add the kitty-open.desktop file
```

I don't understand what this line does:

```
ln -sf ~/.local/kitty.app/bin/kitty ~/.local/kitty.app/bin/kitty ~/.local/bin/
```

I _think_ we might have instead meant:

```
ln -sf ~/.local/kitty.app/bin/kitty ~/.local/kitty.app/bin/kitten ~/.local/bin/
```

page-down:
> I think we might have instead meant ...

Yes, the second `kitty` should be `kitten`. This one also slipped under my radar.

@kovidgoyal

Also, I believe `alias` is a `builtin`.
So it should be called using `builtin` to be consistent.

See the recently changed lines in the bash, zsh shell integration scripts.

kovidgoyal:
Thanks, fixed.

Issue #5996, opened 2023-02-08
Output's preview with Ibus Chinese input in neovim messed up, but the content is correct


YaQia:
**Describe the bug**
In the latest kitty 0.27 version, ibus starts to move the cursor to some strange places, however, the typing content is not messed up. It is the display of the cursor has been messed up. (Strangely, it only happens to some files. In \*.c\/\*.cpp files this will never happen.)

I compared it with alacritty's result, and the latter works well.
**To Reproduce**
1. use nvim to open a random txt file (e.g. 1.txt)
2. use ibus to type some characters into it
3. the format will be occasionally messed up

**Screenshots**
[录屏 2023-02-09 00-33-34.webm](https://user-images.githubusercontent.com/61108050/217592652-4d7a8884-3436-43aa-be80-61e80e074364.webm)
> `cat 1.txt` shows the content is correct, but in neovim its output preview is messed up

**Environment details**
```
kitty 0.27.0 () created by Kovid Goyal
Linux legion 6.1.9-1-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Feb  1 20:38:52 UTC 2023 x86_64
Manjaro Linux 6.1.9-1-MANJARO  (legion) (/dev/tty)


DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=22.0.2
DISTRIB_CODENAME=Sikaris
DISTRIB_DESCRIPTION="Manjaro Linux"
Running under: X11
Frozen: False
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /bin/zsh
Loaded config files:
  /home/ethan/.config/kitty/kitty.conf

Config options different from defaults:
background_opacity    0.9
editor                nvim
font_family           Fira Code Nerd Font
font_size             15.0
initial_window_height (460, 'px')
initial_window_width  (800, 'px')
remember_window_size  False
window_border_width   (1.0, 'px')
Changed shortcuts:
	kitty_mod+h →  neighboring_window left
	kitty_mod+j →  neighboring_window down
	kitty_mod+k →  neighboring_window up
	kitty_mod+l →  neighboring_window right
	kitty_mod+n →  next_layout
Colors:
	background            #131413   
	color1                #cc0000   
	color10               #8ae234   
	color11               #fce94e   
	color12               #719ecf   
	color13               #ad7ea7   
	color14               #34e2e2   
[...]

PR #6002, opened 2023-02-10
IME: Fix IME commit text and update pre-edit at the same time

page-down:
Fix the bug when IME uses single key to confirm and update pre-edit text.

- Correctly update the overlay position when cursor visibility changes.
- Restore the overlay line only when the cursor is visible.
- Clear the saved overlay when drawing new pre-edit text.
- Also update the cursor position when screen size changes.
- Use four spaces to indent instead of two.

When IME confirming text with a single key and updating the pre-edit text, the overlay is drawn at the initial x coordinate since the cursor position is not updated. Saving the overlay at this point will deactivate it and update cursor->x.
When deactivated, the line is marked as dirty and immediately redrawn, and it will appear that the confirmed text is drawn after the current pre-edit text.

So when the confirmed text is `kitty` and the next pre-edit text is `t`, the cursor->x will be updated to follow the current overlay, i.e. `cursor->x = 1`. `kitty` will be drawn from the second cell, and `tkitt|y` will appear (the cursor is in the sixth cell).

Before this PR, although the overlay text was saved when entering the cursor hiding mode, the overlay was restored immediately due to SGR and cursor movement that followed. The cursor x was incorrectly updated to xstart, the initial position, when it was saved (deactivate was called), so the second confirmed text position drawn next is also wrong.

When the second confirmed text is `terminal`, even if the program has asked to move the cursor to `1,6`, `terminal` still draws from the first cell and overwrites `kitty`, and finally moves the cursor to `1,14`.

See: https://github.com/kovidgoyal/kitty/issues/5996#issuecomment-1423084528

Related:
https://github.com/kovidgoyal/kitty/pull/5886
https://github.com/kovidgoyal/kitty/issues/5716

Slightly tested:
- the built-in CJK input methods on macOS Ventura 13.2
- RIME on macOS and Linux X11 IBus, Wayland Fcitx5
- VIM v9.0.981
- Neovim v0.8.3
- Kakoune v2022.10.31

@astralhpi @YaQia
I don't want to break anyone else's IME, and it would be nice if you could test to see whether it works for you.
Especially on Linux platforms, enabling status line for vim/neovim, e.g. `vim --clean`, `nvim --clean`, let the cursor change position and draw text when hidden.

Thank you!

### DIFF
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -64,6 +64,11 @@ init_overlay_line(Screen *self, index_type columns) {
     return true;
 }
 
+static void save_overlay_line(Screen *self, const char* func_name);
+static void restore_overlay_line(Screen *self);
+static void deactivate_overlay_line(Screen *self);
+static void clear_saved_overlay_line(Screen *self);
+
 #define RESET_CHARSETS \
         self->g0_charset = translation_table(0); \
         self->g1_charset = self->g0_charset; \
@@ -140,6 +145,7 @@ new(PyTypeObject *type, PyObject *args, PyObject UNUSED *kwds) {
         init_tabstops(self->alt_tabstops, self->columns);
         self->key_encoding_flags = self->main_key_encoding_flags;
         if (!init_overlay_line(self, self->columns)) { Py_CLEAR(self); return NULL; }
+        clear_saved_overlay_line(self);
         self->hyperlink_pool = alloc_hyperlink_pool();
         if (!self->hyperlink_pool) { Py_CLEAR(self); return PyErr_NoMemory(); }
         self->as_ansi_buf.hyperlink_pool = self->hyperlink_pool;
@@ -147,15 +153,14 @@ new(PyTypeObject *type, PyObject *args, PyObject UNUSED *kwds) {
     return (PyObject*) self;
 }
 
-static void deactivate_overlay_line(Screen *self);
 static Line* range_line_(Screen *self, int y);
 
 void
 screen_reset(Screen *self) {
     if (self->linebuf == self->alt_linebuf) screen_toggle_screen_buffer(self, true, true);
     if (self->overlay_line.is_active) deactivate_overlay_line(self);
+    clear_saved_overlay_line(self);
     Py_CLEAR(self->last_reported_cwd);
-    Py_CLEAR(self->overlay_line.save.overlay_text);
     self->render_unfocused_cursor = false;
     memset(self->main_key_encoding_flags, 0, sizeof(self->main_key_encoding_flags));
     memset(self->alt_key_encoding_flags, 0, sizeof(self->alt_key_encoding_flags));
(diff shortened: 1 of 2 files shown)


YaQia:
Unfortunately, my problem is not about single key confirmation, it seems that it just happens in neovim. I guess my problem is about the changes in #5886 . Strangely, my problem only happens in neovim but not in vim, and only in kitty but not in alacritty.

I used `nvim --clean` to type in a txt file, the problem still happens. (Tips:  in neovim with lsp configuration \*\.c files' behavior is normal.)
I tried to let vim has the same configuration with neovim (e.g. set num, let INSERT mode cursor be I-beam), but vim will never have the same problem.

page-down:
@YaQia
> I used nvim --clean to type in a txt file, the problem still happens.

Can you give reproduction steps?
What input method was used (version and configuration)?
It would be helpful if you could run `kitty --config=NONE --dump-commands --debug-input nvim --clean` and provide some logs when problems occur.

page-down:
I believe the root cause of similar issues are misplaced cursor positions.
In `deactivate_overlay_line`, the cursor->x is restored to an outdated position.
```c
deactivate_overlay_line(...) {
    ...
    ... self->cursor->x = self->overlay_line.xstart;
```

Also when overlay is displayed, cursor->x would not be the expected value.

It is better to separate it from the main cursor, just like a tooltip, hide it when finished, and draw the screen strictly according to the program demands.

@kovidgoyal
Do you have a better way to synchronize the cursor coordinates?

It is OK that the overlay does not move to the specified position in time,
but it would be unacceptable to have the text drawn to `1,1` after receiving the command `Move cursor to 1,6 and Draw`.

kovidgoyal:
On Fri, Feb 10, 2023 at 02:36:29AM -0800, page-down wrote:
> I believe the root cause of similar issues are misplaced cursor positions.
> In `deactivate_overlay_line`, the cursor->x is restored to an outdated position.
> ```c
> deactivate_overlay_line(...) {
>     ...
>     ... self->cursor->x = self->overlay_line.xstart;
> ```
> 
> Also when overlay is displayed, cursor->x would not be the expected value.
> 
> It is better to separate it from the main cursor, just like a tooltip, hide it when finished, and draw the screen strictly according to the program demands.
> 
> @kovidgoyal
> Do you have a better way to synchronize the cursor coordinates?

I am fine with showing the overlay line at whatever position the
cursor was when the line was activated and not messing with the cursor
at all afterwards. In fact the cursor can simply be overriden to be
displayed at whatever position is needed for the overlay line. Any
changes to the cursor coming from the application can apply to the
actual cursor, but when an override is set kitty will render the
overriden cursor instead instead of the actual cursor.

YaQia:
> @YaQia
> 
> > I used nvim --clean to type in a txt file, the problem still happens.
> 
> Can you give reproduction steps? What input method was used (version and configuration)? It would be helpful if you Sorry, my fault, probably I didn't remove the build cache and got the wrong result from the cached `make` result. It performs perfectly now. Appreciate the great contribution you made!

page-down:
@YaQia
> Sorry, my fault, ... It performs perfectly now.

Glad to hear that your case was confirmed as resolved.

I believe the Japanese IME issue has also improved at some level, but at least one confirmation is needed to close that issue so it doesn't get repeatedly reopened.

@kovidgoyal
> ... kitty will render the overriden cursor.

Yes, the overlay should have had a cursor on it.
The input method supports moving the overlay cursor with the `left` and `right` keys, and undoing partially confirmed text with `backspace` and choosing candidates again word by word without exiting the IME state.

This can currently be done in some other terminals under macOS.
However the hardest part is testing and making sure **ALL** the input methods work properly rather than implementing this, so I will just leave it as is for now.

PR #6008, opened 2023-02-11
Get go version from go.mod

page-down:
No need to update multiple places when bumping the go version.
Also bump versions of the deprecated github actions to the latest.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ env:
     ASAN_OPTIONS: leak_check_at_exit=0
     LC_ALL: en_US.UTF-8
     LANG: en_US.UTF-8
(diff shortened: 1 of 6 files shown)


page-down:
Error E704, which was falsely reported by ruff, will be removed in v0.0.246, so no code changes are needed.
https://github.com/charliermarsh/ruff/commit/ffb4e89a980e638d4d596c02f8cb29515d2d740f

The newly added `go` code scan does not seem to work, CodeQL exits after exhausting all disk space, so it needs to be disabled.

```yaml
- name: Initialize CodeQL
  uses: github/codeql-action/init@v2
  with:
    languages: python, c
    # languages: python, c, go
```

I'm not committing new changes for now, waiting for ruff v0.0.246.

PR #6049, opened 2023-02-22
IME: Render overlay at the last visible cursor position and improvements under macOS

page-down:
Render overlay at the last visible cursor position with a separate cursor.

Fix the problem caused by wrong cursor coordinates. No more messing with the main cursor, instead the cursor is saved when receiving a pre-edit text update and used for drawing later.

Update the overlay to the last visible cursor position before rendering to ensure it always moves with the cursor. Finally, draw the overlay after line rendering is complete, and restore the line buffer after updating the rendered data to ensure that the line text being read is correct at all times.

This also improves performance by only rendering once when changes are made, eliminating the need to repeatedly disable and draw after various commands and not even comprehensively.

Allow IME to actively get the cursor position in real time on macOS.

IME will automatically get the display position when needed, which keeps it consistent with the overlay as much as possible.
Fix the issue that when IME is activated after mouse click, it is displayed at the wrong position.

After the window is created, the initial activation of the IME will be in whatever position the cursor is located, not in the upper left corner.

Tested on:
- macOS, builtin CJK IME
- Linux X11/Wayland, IBUS/Fcitx5, RIME

Supersedes:
https://github.com/kovidgoyal/kitty/pull/6002

Please review, thank you.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -46,7 +46,7 @@ Detailed list of changes
 
 - When changing the cursor color via escape codes or remote control to a fixed color, do not reset cursor_text_color (:iss:`5994`)
 
-- Input Method Extensions: Fix incorrect rendering of IME in-progress text in some situations (:pull:`6002`)
+- Input Method Extensions: Fix incorrect rendering of IME in-progress and commited text in some situations (:pull:`6049`)
 
 - Linux: Reduce minimum required OpenGL version from 3.3 to 3.1 + extensions (:iss:`2790`)
 
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1454,15 +1454,11 @@ is_ascii_control_char(char x) {
 }
 
 void _glfwPlatformUpdateIMEState(_GLFWwindow *w, const GLFWIMEUpdateEvent *ev) {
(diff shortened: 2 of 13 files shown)


page-down:
Thanks for the cleanup.

Do you know how to get `readSelectionFromPasteboard` (`glfw/cocoa_window.m`) to work?
https://github.com/kovidgoyal/kitty/commit/9b5034f904d781cc1cf43d525668e19d91256bda
https://github.com/kovidgoyal/kitty/commit/18fc8dad37c944e9f3cfe708df8648ac234ac3c8

Are there any tips from the contributor of that patch?

I want to test this piece of code to see if it affects the input and if there is another way.

kovidgoyal:
I dont recall I think I tested the writetopasteboard not the readfrompasteboard. But I think there are some chinese translation services or something like that that export text. Also terminal.app support these services so selecting text there might be a way to test it.

page-down:
> ... I tested the writetopasteboard not the readfrompasteboard ...

I did the same.

> ... there are some ... export text.
> ... so selecting text there might be a way to test it.

I believe this is for `writeSelectionToPasteboard`.
And `readSelectionFromPasteboard` is used to handle the text sent to kitty by other services.
Have tried various ways, however I just can't get this function to be executed.

```log
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   fast_data_types.so            	       0x108b2f3e9 get_ime_cursor_position + 57
```

After updating to the latest commit, I'm experiencing a crash just by pressing some keys after startup.

kovidgoyal:
https://github.com/kovidgoyal/kitty/commit/c5149dec24ee9cd7ddff1c19f5c5153636aa815a

PR #6076, opened 2023-03-01
Replace the deprecated imghdr module

page-down:
When running `. /tests.py mypy`, it shows the following message and fails (exit code 1).

```log
DeprecationWarning: 'imghdr' is deprecated and slated for removal in Python 3.13
```

I noticed that the relevant remote control commands cannot be executed via shortcuts.

```shell
kitty -o 'map f1 remote_control set-window-logo /path/to/png'
```

```log
kitty.rc.base.StreamError: No stream_id in rc payload
```

Since this is related to the remote control protocol, could you update the documentation if you have some time? (And maybe the async part, such as `async_id`)
https://sw.kovidgoyal.net/kitty/rc_protocol/

The detailed traceback is printed before the `Press e to see detailed traceback ...` line. This minor issue is not important.

I see the use of `NamedTemporaryFile` in setting the background image and logo, is it possible to use SHM to avoid higher file system IO (writes) when setting dynamically generated images in high frequency?

Is it intentional not to reload the background image after reloading the configuration?

### DIFF
--- a/kittens/show_error/main.py
+++ b/kittens/show_error/main.py
@@ -55,7 +55,7 @@ def real_main(args: List[str]) -> None:
         tb = data['tb']
         for ln in tb.splitlines():
             print('\r\n', ln, sep='', end='')
-        print(flush=True)
+        print(end='\r\n', flush=True)
     hold_till_enter()
 
 
--- a/kitty/rc/set_background_image.py
+++ b/kitty/rc/set_background_image.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
 # License: GPLv3 Copyright: 2020, Kovid Goyal <kovid at kovidgoyal.net>
 
-import imghdr
 import os
 from base64 import standard_b64decode, standard_b64encode
 from typing import TYPE_CHECKING, Optional
 
 from kitty.types import AsyncResponse
+from kitty.utils import is_png
 
 from .base import (
     MATCH_WINDOW_OPTION,
@@ -89,7 +89,7 @@ failed, the command will exit with a success code.
         if path.lower() == 'none':
             ret['data'] = '-'
             return ret
-        if imghdr.what(path) != 'png':
+        if not is_png(path):
             self.fatal(f'{path} is not a PNG image')
 
         def file_pipe(path: str) -> CmdGenerator:
--- a/kitty/rc/set_window_logo.py
+++ b/kitty/rc/set_window_logo.py
@@ -1,13 +1,12 @@
 #!/usr/bin/env python
 # License: GPLv3 Copyright: 2020, Kovid Goyal <kovid at kovidgoyal.net>
 
-
-import imghdr
 import os
 from base64 import standard_b64decode, standard_b64encode
 from typing import TYPE_CHECKING, Optional
 
 from kitty.types import AsyncResponse
+from kitty.utils import is_png
 
 from .base import (
     MATCH_WINDOW_OPTION,
@@ -85,7 +84,7 @@ failed, the command will exit with a success code.
         if path.lower() == 'none':
             ret['data'] = '-'
             return ret
-        if imghdr.what(path) != 'png':
+        if not is_png(path):
             self.fatal(f'{path} is not a PNG image')
 
         def file_pipe(path: str) -> CmdGenerator:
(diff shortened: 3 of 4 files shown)


kovidgoyal:
On Tue, Feb 28, 2023 at 08:33:06PM -0800, page-down wrote:
> When running `. /tests.py mypy`, it shows the following message and fails (exit code 1).
> 
> ```log
> DeprecationWarning: &#39;imghdr&#39; is deprecated and slated for removal in Python 3.13
> ```
> 
> I noticed that the relevant remote control commands cannot be executed via shortcuts.
> 
> ```shell
> kitty -o &#39;map f1 remote_control set-window-logo /path/to/png&#39;
> ```

This is on my TODO list I never got around to porting it after changing
RC to Go.

> 
> ```log
> kitty.rc.base.StreamError: No stream_id in rc payload
> ```
> 
> Since this is related to the remote control protocol, could you update the documentation if you have some time? (And maybe the async part, such as `async_id`)
> https://sw.kovidgoyal.net/kitty/rc_protocol/
> 
> The detailed traceback is printed before the `Press e to see detailed traceback ...` line. This minor issue is not important.
> 
> I see the use of `NamedTemporaryFile` in setting the background image and logo, is it possible to use SHM to avoid higher file system IO (writes) when setting dynamically generated images in high frequency?

It doesn't actually need a file at all, its just there for code re-use.

> 
> Is it intentional not to reload the background image after reloading the configuration?

I dont think so would have to look at the code to be sure.

kovidgoyal:
On Tue, Feb 28, 2023 at 08:33:06PM -0800, page-down wrote:
> When running `. /tests.py mypy`, it shows the following message and fails (exit code 1).
> 
> ```log
> DeprecationWarning: &#39;imghdr&#39; is deprecated and slated for removal in Python 3.13
> ```
> 
> I noticed that the relevant remote control commands cannot be executed via shortcuts.
> 
> ```shell
> kitty -o &#39;map f1 remote_control set-window-logo /path/to/png&#39;
> ```
> 
> ```log
> kitty.rc.base.StreamError: No stream_id in rc payload
> ```
> 
> Since this is related to the remote control protocol, could you update the documentation if you have some time? (And maybe the async part, such as `async_id`)
> https://sw.kovidgoyal.net/kitty/rc_protocol/
> 
> The detailed traceback is printed before the `Press e to see detailed traceback ...` line. This minor issue is not important.
> 
> I see the use of `NamedTemporaryFile` in setting the background image and logo, is it possible to use SHM to avoid higher file system IO (writes) when setting dynamically generated images in high frequency?

Actually NamedTemporaryFile here is a dummy class, it doesnt touch the
filesystem.

page-down:
>> ...  is it possible to use SHM ...
>
> It doesn't actually need a file at all, its just there for code re-use.

In fact, in `handle_streamed_data` -> `StreamInFlight`, `handle_data` is still writing to the temporary file.
The following will not work.
And high frequency updates will result in high file system IO.

```shell
cat image.png | kitten @ set-window-logo /dev/stdin
```

```log
Failed to decode PNG image at: /var/folders/.../.../T/tmpjisxwfhh.png
```

I noticed the relevant codes.
```text
state.c, set_window_logo
window_logo.c, find_or_create_window_logo
graphics.c, png_path_to_bitmap, `FILE* fp = fopen(path, "r"); ...; log_error("Failed to decode PNG image at: %s", path);`

kovidgoyal:
On Tue, Feb 28, 2023 at 09:21:34PM -0800, page-down wrote:
> >> ...  is it possible to use SHM ...
> >
> > It doesn't actually need a file at all, its just there for code re-use.
> 
> In fact, in `handle_streamed_data` -> `StreamInFlight`, `handle_data` is still writing to the temporary file.

Yes of course, actually changing the image requires a file, since
getting libpng to load data from memory is a huge pain. I meant the code
path for clearing the image. We cant really use SHM as there are lots of
systems where SHM is not turned on or broken, so one would have to write
code to very carefully check and fallback.

kovidgoyal:
Although rather than SHM one can use fmemopen assuming it works on macOS

page-down:
> Yes of course, ...

Thanks for the explanation, it's not easy to create a memory file system under macOS.
Unfortunately, even if there are GBs of free memory space, the memory is sitting there and cannot be used.

One use case is to dynamically generate a progress bar or other background task information for the current window.
Rather than generating a temporary file, I would rather send the encoded image directly to the UNIX socket server and render it.

PR #6078, opened 2023-03-01
hints kitten: Allow copying matches to named buffers

page-down:
Use a new prefix `#` to specify the name of the buffer to copy to.

Please review, thanks.


page-down:
> Packages is now experiencing degraded availability. ...

https://www.githubstatus.com/incidents/bypf7bqdh4g0

I noticed that `kitten clipboard` can't be used to copy text to buffer, maybe we need to add something like `:kitty_buffer=NAME` (also in `clipboard.py`).
I'll let you decide on this.

kovidgoyal:
I dont think so, clipboard kitten can copy arbitrary data types. Refactoring the named buffers to store multiple MIME types is not worth the effort for something no one has even requested.

PR #6097, opened 2023-03-08
Redraw old unicode placeholders when a virtual placement is added

sergei-grechanik:


kovidgoyal:
I am not sure about this. Doesn't this mean that placeholders will be refreshed only on screen but not in the scrollback. Which is even more confusing. And scanning through the entire scrollback on every new unicode placeholder graphics command is prohibitive. Maybe better to just specify in the protocol that unicode placeholders must come after the image is created.

sergei-grechanik:
Yes, you are right, the scrollback is not refreshed until something marks it as dirty, like font scaling. One way to work around this without scanning the entire scrollback is to always rerender unicode placeholders when we are displaying lines from the scrollback (we pay for this only if there are placeholders in the scrollback). Not sure if it's actually less expensive though, depends on how much the user scrolls.

kovidgoyal:
Hmm, it pains my stingy soul to have to rescan lines in the scrollback for
this, but I guess there isn't really a good way around it. At least, the
cost is paid only for lines that have previously been marked as having
placeholders.

PR #6109, opened 2023-03-12
Added new icons to the FAQ page

eccentric-j:
- Added the 2d winky terminal icon
- Added the 3d reinterpretation of the original

Have not worked with rst yet. How does one preview the build?

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -338,6 +338,14 @@ homepage:
    :target: https://github.com/samholmes/whiskers
    :width: 256
 
+.. image:: https://github.com/eccentric-j/eccentric-icons/raw/main/icons/kitty-terminal/2d/kitty.png
+   :target: https://github.com/eccentric-j/eccentric-icons
+   :width: 256
+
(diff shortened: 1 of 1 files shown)


page-down:
This will cause to load 1.7MB of new icon image data when opening the FAQ page.
The current maximum size of the other icons in the document page is only 350KB.

I personally recommend providing 256x256 compressed preview icons that are hosted in GitHub.

> ... How does one preview the build?

https://sw.kovidgoyal.net/kitty/build/

After building the documentation, you will be able to see the HTML files in `docs/_build/`.

kovidgoyal:
You can also just preview in github itself, it has a basic rst renderer.
And yes please reduce the icon preview image sizes.

eccentric-j:
I'll make an additional web-friendly, compressed version and make another PR to update those links.

Would there be any benefit to making it 512x512 so it won't look blurry on retina displays?

page-down:
> ... making it 512x512 ...

Yes. That would be acceptable.

PR #6110, opened 2023-03-12
Updated faq with optimized alt icon preview urls

eccentric-j:
Follow up to #6109 to use optimized icon previews

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -338,11 +338,11 @@ homepage:
    :target: https://github.com/samholmes/whiskers
    :width: 256
 
-.. image:: https://github.com/eccentric-j/eccentric-icons/raw/main/icons/kitty-terminal/2d/kitty.png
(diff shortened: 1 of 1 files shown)


page-down:
@eccentric-j

I noticed that in your commit message:

```text
Added ...
Updated ...
```

https://git-scm.com/docs/SubmittingPatches

> The problem statement that describes the status quo is written in the present tense. ...
> ...
>  Describe your changes in imperative mood, e.g. "make xyzzy do frotz" instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy to do frotz", as if you are giving orders to the codebase to change its behavior. ...

https://github.com/git-guides/git-commit

> Commit messages should be present tense and directive, like the following examples:
> - git commit -m "create file structure for Git guides"
> ...

As far as I know, in this project (`kitty`) there are no unnecessary constraints on committing patches as long as your changes to the code are reasonable and meaningful, but other projects may not be.

This is just a friendly reminder (that is with a mainstream style bias opinion).

PR #6129, opened 2023-03-23
feat: support moving macos windows with hidden decorations

isaksamsten:
Hi!

I'm using kitty on Mac and to maximise available vertical space I hide the window decorations. It has always nagged me that the window cannot be dragged with mouse. This WIP pull request "hacks" the view and overrides the `mouseDownCanMoveWindow` so it returns `YES`.

I don't expect this pull-request to be merged as-is but instead of raising an issue I thought it can serve as a starting point for discussing this feature (move the window by drag even if the decorations are hidden).

**Short video showcase**
kitty-move-recording

**Ideas**
- [ ] Right-pad tabs to allow space for buttons if always visible

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -34,7 +34,13 @@
 #import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
 #include <float.h>
 #include <string.h>
+#import <objc/runtime.h>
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
dup of #3744

isaksamsten:
Missed that. Sorry for the spam.

PR #6135, opened 2023-03-26
add background blur on macos

BIKA-C:
This is an attempt to add background blur to macos. I am not familiar to the code base and a little confused about how to properly read config. Any help would be appreciated. Here is a fixed blur radius preview:
image


kovidgoyal:
This is a non-public API correct? And dont you have to set a color for the blur?  window.backgroundColor IIRC

Not to mention there is no generally applicable solution for this on other platforms that I know of. I am not very keen to add platform specific features that anyway use a private API and so may stop working at any time.

As for your actual implementation, move it into kitty/cocoa_window.m then you can access the config value as simply OPT(whatever).

page-down:
@BIKA-C

You should use `NSVisualEffectView` and set the blending mode instead of using the private API.

hungvx-dev:
> 

You can apply `NSVisualEffectView` to this PR.

hungvx-dev:
@kovidgoyal You can update this PR and merge it for the next release. I very need this feature. Please help me.

FormalSnake:
I need it too!

hungvx-dev:
@FormalSnake Yes. Because, when using background opacity is below 0.8 then it will be difficult to see, eyestrain. So we need a degree of blur

Screenshot 2023-06-08 at 10 36 44

kovidgoyal:
Kindly stop spamming my inbox. I already stated my concerns with this PR no one has addressed them. Come up with a proper PR and we can discuss further.

Issue #6108, opened 2023-03-11
Powerline10k Shading doesn't render nicely (Unicode block elements e.g U+2591)


Dadams2:
**Describe the bug**
I just did my powerline10k config for Zsh and noticed kitty was not rendering the shading characters nicely like it does for block characters.

I have a strong suspicion this might be a known issue/decision something along the lines of "This is how Unicode presents the character so we are going to display it as such" but I was unable to find a similar issue that had already been marked as a duplicate. At the very least I will be happy if this is marked as dup as it may help others answer this question. 

Specifically the characters: 
U+2591 | ░ | Light shade
U+2592 | ▒ | Medium shade
U+2593 | ▓ | Dark shade

Are not smooth. 

I also apologies if this is something to do with sub-pixel rendering which I know is an issue of contention and in the works.

However if there is a nice workaround with `symbol_map` or something I would be very happy to hear it 

**To Reproduce**
1. `p10k configure` 
2. choose unicode prompt
3. select blurred prompt heads/tails

**Screenshots**
Example in regular kitty (without tmux) 
![image](https://user-images.githubusercontent.com/37862994/224483210-0fc09e4e-04f5-4755-bca4-44dca4deaead.png)

Example in default alacritty:
![image](https://user-images.githubusercontent.com/37862994/224483229-b2747898-c8ac-4ddb-8fa7-1eb63b483329.png)

**Environment details**
```
kitty 0.27.1 () created by Kovid Goyal
Linux effervescence 6.2.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 03 Mar 2023 15:58:31 +0000 x86_64
Arch Linux 6.2.2-arch1-1 (/dev/tty)

DISTRIB_ID="Arch"
DISTRIB_RELEASE="rolling"
DISTRIB_DESCRIPTION="Arch Linux"
Running under: X11
Frozen: False
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /bin/zsh
Loaded config files:
  /home/dadams/.config/kitty/kitty.conf

Config options different from defaults:
background_opacity      0.9
clipboard_control       ('write-clipboard', 'write-primary', 'no-append')
font_family             Fira Code
hide_window_decorations 1
shell                   /bin/zsh -l -c "tmux attach || tmux"
Colors:
	background              #2e3440   
	color0                  #3b4252   
[...]

PR #6146, opened 2023-03-29
Add shade_transparency option

MithicSpirit:
This makes the shade characters (U+2591 through U+2593) to be drawn as semitransparent blocks rather than discrete, full opacity dots in a dither effect. It is disabled by default to retain compatibility with previous versions of kitty and with the standard, though note that many other terminals (e.g., alacritty and most/all VTE-based terminals) do this by default.

I have seen that this was already addressed in kovidgoyal/kitty#6108, and was just making this as a patch for personal use. Since there is at least some desire in the community for this feature, I thought that it would be a good idea to contribute it to upstream.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -2375,7 +2375,7 @@ class Boss:
         patch_global_colors(spec, configured)
 
     def apply_new_options(self, opts: Options) -> None:
-        from .fonts.box_drawing import set_scale
+        from .fonts.box_drawing import set_scale, set_shade_transparency
         # Update options storage
         set_options(opts, is_wayland(), self.args.debug_rendering, self.args.debug_font_fallback)
         apply_options_update()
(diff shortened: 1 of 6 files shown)


kovidgoyal:
I am afraid I am not going to accept such a PR. One of my core goals is to keep terminals a platform that developers of terminal applications can rely on. Part of that means adhering to standards. So for example, kitty will not support an option to interpret bold as bright, as it makes bold un-useable in applications. Similarly, having the appearance of characters whose appearance is fixed by the unicode standard appear differently based on a config setting makes those characters unuseable in applications.

MithicSpirit:
I've taken a look at [the Unicode standard for these blocks](https://www.unicode.org/charts/PDF/U2580.pdf), and you might be interested in the fact that it states that the light, medium, and dark shades should have 25%, 50%, and 75% covering, respectively. Currently, in kitty, they have approximately 12.5%, 25%, and 75% covering, resp. I could submit a PR to fix this if desired.

kovidgoyal:
On Wed, Mar 29, 2023 at 10:51:45AM -0700, MithicSpirit wrote:
> I've taken a look at [the Unicode standard for these blocks](https://www.unicode.org/charts/PDF/U2580.pdf), and you might be interested in the fact that it states that the light, medium, and dark shades should have 25%, 50%, and 75% covering, respectively. Currently, in kitty, they have approximately 12.5%, 25%, and 75% covering, resp. I could submit a PR to fix this if desired.

Sure, that would be welcome.

PR #6218, opened 2023-04-27
Fix color fringing caused by gamma-incorrect blending in compositor when background_opacity is used

### LINKED ISSUE #6209: `text_composition_strategy legacy` is noticeably worse than pre-0.28 rendering
**Describe the bug**

Relating to Kitty 0.28 `text_composition_strategy legacy`:
 
1. it's always subtly different from actual pre-0.28 rendering, and
2. it looks much worse with transparent windows in front of light backgrounds.

**To Reproduce**
1. Use config as shown below
3. Update to kitty 0.28
4. Observe text (as shown in screenshots)

**Screenshots**

Note: Screenshots were taken in a kitty instance with full transparency — the background color is exclusively from blur (via my compositor) over my desktop background.

**Post-0.28**

Screenshot of the funky rendering

Screenshot of rendering on a darker background looking a-ok

**Pre-0.28**

Screenshot of text looking a-ok

Screenshot of rendering on a darker background looking a-ok

**Note:** With the darker background, there is _also_ a slight change that post-0.28 rendering is slightly thicker - should hopefully be easy enough to see if you compare the two.

**Environment details**

Couldn't get Ctrl+Shift+F6 working, for some reason. Pasting the relevant pieces of my config file. `kitty -v` outputs `kitty 0.28.1 created by Kovid Goyal`. Happy to provide more details upon request.

```
### Fonts ###
font_family Ubuntu Mono
font_size   20.0

disable_ligatures always
text_composition_strategy legacy

### Color scheme ###
foreground #eaeaea
background #263238

# Terminal colors
#
# These are given first with the normal variety, and then with the
# bold variety.
#
# Black
color0 #000000
color8 #666666
# Red
color1 #d54e53
color9 #ff3334
# Green
color2  #b9ca4a
color10 #9ec400
# Yellow
color3  #e6c547
color11 #e6c547
# Blue
color4  #7aa6da
color12 #7aa6da
# Magenta
color5  #c397d8
color13 #b77ee0
# Cyan
color6  #70c0ba
color14 #54ced6
# White
color7 #eaeaea
color15 #ffffff

### Advanced options ###
background_opacity 0.0

# note: lines unrelated to rendering were omitted.
```

I _think_ the font color in the screenshots is `color4` (normal blue), but not 100% sure.

m4rw3r:
This PR fixes #6209 by adjusting the alpha channel to account for gamma-incorrect blending performed by the compositor when background_opacity is < 1.0. At least wlroots, picom, GNOME, and MacOS perform gamma-incorrect blending by default with transparent windows.

It also includes a more precise sRGB-linear conversion to pretty much eliminate fringing completely.

This PR will however cause the following issues when using a transparent background:

* Some text will appear bolder, depending on the foreground color, the background, and the color of the pixels behind the semi-transparent window.
* Some of the transparent pixels will have incorrect hues due to incorrect blending.
* `background_opacity` will no longer be linear

See the following examples, the third variant of each is the current version with issues (100% transparency, old).

image
image (1)

### DIFF
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -37,6 +37,22 @@ in float colored_sprite;
 out vec4 final_color;
 
 // Util functions {{{
+float linear2srgb(float x) {
+    // Linear to sRGB conversion.
+    float lower = 12.92 * x;
+    float upper = 1.055 * pow(x, 1.0f / 2.4f) - 0.055f;
+
+    return mix(lower, upper, step(0.0031308f, x));
+}
+
+float srgb2linear(float x) {
+    // sRGB to linear conversion
+    float lower = x / 12.92;
+    float upper = pow((x + 0.055f) / 1.055f, 2.4f);
+
+    return mix(lower, upper, step(0.04045f, x));
+}
+
 vec4 alpha_blend(vec4 over, vec4 under) {
     // Alpha blend two colors returning the resulting color pre-multiplied by its alpha
     // and its alpha.
@@ -103,19 +119,8 @@ vec4 vec4_premul(vec4 rgba) {
 #ifdef NEEDS_FOREGROUND
 // sRGB luminance values
 const vec3 Y = vec3(0.2126, 0.7152, 0.0722);
-const float gamma_factor = 2.2;
 // Scaling factor for the extra text-alpha adjustment for luminance-difference.
 const float text_gamma_scaling = 0.5;
-
-float linear2srgb(float x) {
-    // Approximation of linear-to-sRGB conversion
-    return pow(x, 1.0 / gamma_factor);
-}
-
-float srgb2linear(float x) {
-    // Approximation of sRGB-to-linear conversion
-    return pow(x, gamma_factor);
-}
 
 float clamp_to_unit_float(float x) {
     // Clamp value to suitable output range
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -184,6 +184,12 @@ void main() {
     vec4 fg = calculate_foreground(background);
 #ifdef TRANSPARENT
     final_color = alpha_blend_premul(fg, vec4_premul(background, bg_alpha));
+
+    // Adjust the transparent alpha-channel to account for incorrect
+    // gamma-blending performed by the compositor (true for at least wlroots,
+    // picom, GNOME, MacOS).
+    // This is the last pass:
+    final_color.a = linear2srgb(final_color.a);
 #else
     final_color = alpha_blend_premul(fg, background);
 #endif
@@ -207,6 +213,10 @@ void main() {
 
 #ifdef FOREGROUND
     final_color = calculate_foreground();  // pre-multiplied foreground
+
+    // This is the last pass, adjust alpha to compensate for gamma-incorrect
+    // blending in compositor:
+    final_color.a = linear2srgb(final_color.a);
 #endif
 
 }
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -198,7 +198,7 @@ void main() {
 #endif
 
 #ifdef BACKGROUND
-#if defined(TRANSPARENT)
+#ifdef TRANSPARENT
     final_color = vec4_premul(background, bg_alpha);
 #else
     final_color = vec4(background, draw_bg);


kovidgoyal:
LGTM merged. I wish there was a way to tell compositors to blend correctly. I hate these kinds of kludges.

Thesola10:
@m4rw3r there is some weird translucent color distortion going on in latest release, wondering if this is what you described as "Some of the transparent pixels will have incorrect hues due to incorrect blending."?

see #6250

m4rw3r:
@Thesola10 The change in this PR makes the whole `background_opacity`-setting non-linear for the transparent background, to fix the color fringing caused by incorrect blending. You can get a similar look by applying the sRGB curve to the opacity value, ie. `pow(old)` (for the case of 0.7 that would be 0.425). Hopefully that improves things?

I have no clue what is happening with the red background in the window-only screenshots however. Did this happen before this PR landed?

@kovidgoyal Maybe I should add a reverse correction for the `background_opacity` value itself, to have it behave the linearly as it did before?

kovidgoyal:
On Sat, May 06, 2023 at 11:44:28AM -0700, Martin Wernstål wrote:
> @Thesola10 The change in this PR makes the whole `background_opacity`-setting non-linear for the transparent background, to fix the color fringing caused by incorrect blending. You can get a similar look by applying the sRGB curve to the opacity value, ie. `pow(old)` (for the case of 0.7 that would be 0.425). Hopefully that improves things?
> 
> I have no clue what is happening with the red background in the window-only screenshots however. Did this happen before this PR landed?
> 
> @kovidgoyal Maybe I should add a reverse correction for the `background_opacity` value itself, to have it behave the linearly as it did before?

Hmm, maybe worth it. It will avoid a whole bunch of duplicate bug
reports. And having linear-ish behavior is more intuitive for most
people. I know that from days as a physicist :)

m4rw3r:
Seems like the cause for the brighter-appearing transparent background in the screenshots from @Thesola10 in #6250 is caused by premultiplication occurring in linear space. Looking at options for "fixing" this. So this PR is not related to the brighter transparent backgrounds.

Thesola10:
Hence why #6250 needs to be a separate issue -- we can't know for sure what is causing it, but it exists and it's on release.

Please reopen @kovidgoyal

neddey:
Would there an easy way to calculate a value that would be equivalent to using a liner opacity?

kovidgoyal:
Note that Wayland recently released support for specifying color properties for surfaces (windows) this might allows us to remove this hack on Wayland, see
https://wayland.app/protocols/color-management-v1

In particular it has a gamma22 sRGB transfer function specification.

fenhl:
I came here after finding #8355 as I have the same issue, and after reading through this and some linked issues, and trying a few things, I am thoroughly confused. I have sway configured to have window border color `#ffffffa6` (approximately 65% opacity) so if I'm understanding https://github.com/kovidgoyal/kitty/pull/6218#issuecomment-2907967009 correctly, setting kitty's `background_opacity` to 0.3813260 should result in the appearance I want. However, the result I'm observing is that the window contents are still much brighter than the border. Am I missing something? Alternatively, would it be possible to add a config option to revert to the linear behavior?

zuloo:
So why are all issues concerning this closed? This is not fixed, right? 0.42.2 should actually have this commit, but it still has the same issue. It worked perfectly with 0.26.2, so why not just take the blending code from 0.26.2 and port it to 0.42.2?

m4rw3r:
> So why are all issues concerning this closed? This is not fixed, right? 0.42.2 should actually have this commit, but it still has the same issue. It worked perfectly with 0.26.2, so why not just take the blending code from 0.26.2 and port it to 0.42.2?

0.26.2 was using the old gamma-incorrect blending. The gamma-correct blending was introduced in 0.28.0. This pull request is adding a correction pass to ensure all blending is srgb-based (ie. correct according to gamma) which also includes any transparent background.

Attempting to port anything from 0.26.2 would cause these issues again, or force the use of the old gamma-incorrect blending for everything which causes incorrect colors, perceived brightness, and font-weight.

To get it to use gamma-incorrect blending just for the background we would have to come up with a new way of blending the edges of the text with the background to avoid any fringing or changes in perceived font-weight.

Using the new Wayland support for color properties might solve this as we would be able to remove the correction-pass, but that is something which needs to be tested.

I'm not sure if that is best done as a separate issue or where that effort should be directed.

kovidgoyal:
On Wed, Jul 30, 2025 at 12:36:51PM -0700, Martin Wernstål wrote:
> Using the new Wayland support for color properties might solve this as we would be able to remove the correction-pass, but that is something which needs to be tested.
>
> I'm not sure if that is best done as a separate issue or where that effort should be directed.

If you want to work on it let me know and you can open a PR for it.
Otherwise, it's on my relatively short term TODO list to at least
add support for that Wayland protocol to kitty's glfw fork. At which
point it should be relatively easy to test if it works as advertised on
the major compositors. I think currently, KDE, GNOME and Hyprland
support it.

kovidgoyal:
I found that there is still color fringing, opened a new issue to track it: #8869

PR #6256, opened 2023-05-08
Fix shade characters to follow unicode standard

MithicSpirit:
Addresses discussion here: https://github.com/kovidgoyal/kitty/pull/6146#issuecomment-1489047598. (Sorry for taking so long on submitting this PR; I was busy with classes).

The characters were implemented as a checkerboard of the correct fill density (like in the example characters in the Standard), but this might look a little dated. Let me know if you want me change their appearance.
- Light shade: 25% fill
- Medium shade: 50% fill
- Dark shade: 75% fill (implemented as inverse of light shade)

Additionally, it is extremely easy to add support for the inverse medium shade character (`🮐`, U+1FB90) now, as it should just be `[p(shade, invert=True)]`. Let me know if you want me to include that in this PR (I have also enabled "Allow edits by maintainers" if you want to add it yourself).

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -543,6 +543,7 @@ START_ALLOW_CASE_RANGE
         case 0xe0b0 ... 0xe0bf:  // powerline box drawing
         case 0x1fb00 ... 0x1fb8b:  // symbols for legacy computing
         case 0x1fba0 ... 0x1fbae:
+        case 0x1fb90:  // inverse medium shade
             return BOX_FONT;
         default:
             *is_emoji_presentation = has_emoji_presentation(cpu_cell, gpu_cell);
@@ -585,6 +586,9 @@ START_ALLOW_CASE_RANGE
             return 0x151 + ch - 0x1fba0;
         case 0x2800 ... 0x28ff:
             return 0x160 + ch - 0x2800;
+        case 0x1fb90:
+            //  Allocated to allow for 0x1fb8c ... 0x1fb94 eventually
+            return 0x25f + ch - 0x1fb8c;
         default:
             return 0xffff;
     }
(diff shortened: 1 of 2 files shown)


MithicSpirit:
Updated the appearance because I wasn't happy with how old-timey the checkerboard looked. Now it uses antialiased diagonal lines that tile perfectly.

Feel free to revert (or tell me to revert) that last commit (c883a02) if you want the previous look.

MithicSpirit:
### Comparison of appearances
- Original look
original
- Previous look
previous
- New look
new

All tests were done with the command `./kitty/launcher/kitty --config NONE --hold cat /tmp/shade` without changing the font size (original and previous do look nicer at larger font sizes or higher resolution, but since that's not the default I thought it'd be better to show it as it is), where `/tmp/shade` is a file containing
```
░░░░░░░░░░░░░░░░░░░░░░░░
░░░░░░░░░░░░░░░░░░░░░░░░
▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
```

kovidgoyal:
I think I prefer the checkerboard, I want to stick to the defined appearance in the standard. If we are going to change that we might as well go the whole way and just implement them as transparency. Also please add the inverse medium to this PR.

MithicSpirit:
> I think I prefer the checkerboard, I want to stick to the defined appearance in the standard.

Understandable. It has been reverted.

> Also please add the inverse medium to this PR.

Added. I've also looked into the neighborhood of that character and there seem to be other, similar blocks that wouldn't be hard to implement in a future PR, so I've left space for them in fonts.c's `box_glyph_id` to prevent too much fragmentation in the future.

PR #6283, opened 2023-05-18
Add foreground override contrast threshold

lmcnulty:
This adds a parameter to text_composition_strategy that specifies a percentage difference of luminance below which the foreground color will be overridden. The foreground color is set to white if the background is dark or black if the background is light.

Many programs output colors that look good with the author's terminal's color scheme but which are completely illegible with other color schemes. This allows the user ensure that there is always sufficient contrast to read the text on the screen.

Since we want existing configs to continue working, this also makes it so that rather than taking exactly two parameters, text_composition_strategy takes one--three parameters, using the default values for those not specified.

### DIFF
--- a/kitty/cell_fragment.glsl
+++ b/kitty/cell_fragment.glsl
@@ -1,6 +1,7 @@
 #version GLSL_VERSION
 #define {WHICH_PROGRAM}
 #define NOT_TRANSPARENT
+#define NO_FG_OVERRIDE
 
 #if defined(SIMPLE) || defined(BACKGROUND) || defined(SPECIAL)
 #define NEEDS_BACKROUND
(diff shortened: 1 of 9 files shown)


kovidgoyal:
1) Using conditionals (ifs) in shaders is bad performance wise
instead use mix(), step() and friends.

2) I am not sure this belongs with text_composition_strategy it doesnt
have much to do with composition. 

3) Ideally I would like to avoid any performance overhead for what is a
rarely used option to workaround bugs in other programs. This means
using defines in the shader. So you would check if the option is set to
non-zero value and recompile the shader accordingly.

lmcnulty:
@kovidgoyal Thanks for your feedback! Changes per your comments:

1. Now uses step().
2. Changed to separate option: `text_fg_override_threshold`
3. Now included only when `FG_OVERRIDE` is defined.

lmcnulty:
This affects colored emoji badly. Investigating.

kovidgoyal:
Merged, thanks. It might be nice if you added a sentence to the docs of the option suggesting a good value.

Issue #6382, opened 2023-06-21
window size (resize) to absolute value and choice of axis


ershov:
Hi,
I have two questions:

1. Is there a way to specify the window size in absolute units? `resize-window` only accepts positive or negative increments.
2. Is there a way to ask Kitty to resize the window regardless the axis? If I have a split, the axis is known and I should't be required to specify it. But it's a required parameter and using a wrong axis fails the command.

kovidgoyal:
1. no
2. no

ershov:
Is there any specific reason for that or it would require significant effort to implement?

kovidgoyal:
Layouts work by distributing available space between windows based on
biases. There is no concept of fixed size windows in kitty.
https://sw.kovidgoyal.net/kitty/layouts/

ershov:
That's right. I was thinking of 'splits' layout. I thought it was fully manual. Does it support the 'bias' parameter?

kovidgoyal:
Yes, every split has a bias associated with it.

ershov:
`layout_action bias` unfortunately doesn't work for `splits` payout and it can't be changed dynamically. Maybe, because the Splits one doesn't override the `apply_bias` function? https://github.com/search?q=repo%3Akovidgoyal%2Fkitty%20apply_bias&type=code

I also tried a trick like `enabled_layouts splits,splits:bias=90` but it looks like the `bias` option is ignored by Splits layout.

I however noticed a useful feature: it created two separate 'splits' layouts between which I can cycle. (but it doesn't solve the problem of setting a pre defined split size via the `bias` option)

ershov:
Hmm ok it does implement `modify_size_of_window` though.

I think in layout_action it needs to implement the action name `bias` or some special action for this layout.

ershov:
Can you please have a loot at https://github.com/kovidgoyal/kitty/pull/6386

ershov:
I also have a neat solution for no.2 (increase/decrease/set size of window regardless the slit axis). Would you be interested in a pull request for that?

kovidgoyal:
On Thu, Jun 22, 2023 at 02:22:13AM -0700, Yury Ershov wrote:
> I also have a neat solution for no.2 (increase/decrease/set size of window regardless the slit axis). Would you be interested in a pull request for that?

Let's focus on one thing at a time.

PR #6386, opened 2023-06-22
Add window resize to absolute values

ershov:
This is a proof-of-concept for enabling the 'split' layout to use the 'bias' option and allowing to specify the absolute values for 'resize_window' command.

The 'bias' option was previously ignored by 'splits' layout, now it's possible to do:

```
enabled_layouts splits:bias=90
```

Only increments or decrements were previously accepted. 'width' and 'height' are for absolute percentages:

```
map <KEY>          resize_window width 20
map <KEY>          resize_window height 70
```

The issue is mentioned here: https://github.com/kovidgoyal/kitty/issues/6382

Please let me know if this kind of change can potentially be accepted.

### DIFF
--- a/kitty/layout/base.py
+++ b/kitty/layout/base.py
@@ -237,6 +237,9 @@ class Layout:
     def apply_bias(self, window_id: int, increment: float, all_windows: WindowList, is_horizontal: bool = True) -> bool:
         return False
 
+    def apply_bias_abs(self, window_id: int, bias: float, all_windows: WindowList, is_horizontal: bool = True) -> bool:
+        return False
+
     def remove_all_biases(self) -> bool:
         return False
 
@@ -245,6 +248,12 @@ class Layout:
         if idx is None:
             return False
         return self.apply_bias(idx, increment, all_windows, is_horizontal)
+
+    def modify_size_of_window_abs(self, all_windows: WindowList, window_id: int, bias: float, is_horizontal: bool = True) -> bool:
+        idx = all_windows.group_idx_for_window(window_id)
+        if idx is None:
+            return False
+        return self.apply_bias_abs(idx, bias, all_windows, is_horizontal)
 
     def parse_layout_opts(self, layout_opts: Optional[str] = None) -> LayoutOpts:
         data: Dict[str, str] = {}
--- a/kitty/layout/splits.py
+++ b/kitty/layout/splits.py
@@ -18,11 +18,11 @@ class Extent(NamedTuple):
 
 class Pair:
 
-    def __init__(self, horizontal: bool = True):
+    def __init__(self, horizontal: bool = True, bias: float = 0.5):
         self.horizontal = horizontal
         self.one: Optional[Union[Pair, int]] = None
         self.two: Optional[Union[Pair, int]] = None
-        self.bias = 0.5
+        self.bias = bias
         self.top = self.left = self.width = self.height = 0
         self.between_borders: List[Edges] = []
         self.first_extent = self.second_extent = Extent()
@@ -119,17 +119,17 @@ class Pair:
             q = self.one if one_count < two_count else self.two
             return q.balanced_add(window_id)
         if not isinstance(self.one, Pair) and not isinstance(self.two, Pair):
-            pair = Pair(horizontal=self.horizontal)
+            pair = Pair(horizontal=self.horizontal, bias=self.bias)
             pair.balanced_add(self.one)
             pair.balanced_add(self.two)
             self.one, self.two = pair, window_id
             return self
         if isinstance(self.one, Pair):
             window_to_be_split = self.two
-            self.two = pair = Pair(horizontal=self.horizontal)
+            self.two = pair = Pair(horizontal=self.horizontal, bias=self.bias)
         else:
             window_to_be_split = self.one
-            self.one = pair = Pair(horizontal=self.horizontal)
+            self.one = pair = Pair(horizontal=self.horizontal, bias=self.bias)
         assert isinstance(window_to_be_split, int)
         pair.balanced_add(window_to_be_split)
         pair.balanced_add(window_id)
@@ -142,7 +142,7 @@ class Pair:
             pair.horizontal = horizontal
             self.one, self.two = q
         else:
-            pair = Pair(horizontal=horizontal)
+            pair = Pair(horizontal=horizontal, bias=self.bias)
             if self.one == existing_window_id:
                 self.one = pair
             else:
@@ -256,6 +256,21 @@ class Pair:
         if parent is not None:
             which = 1 if parent.one is self else 2
             return parent.modify_size_of_child(which, increment, is_horizontal, layout_object)
+        return False
+
+    def set_size_of_child(self, which: int, bias: float, is_horizontal: bool, layout_object: 'Splits') -> bool:
+        if is_horizontal == self.horizontal and not self.is_redundant:
+            if which == 2:
+                bias = 1 - bias
+            new_bias = max(0.1, min(bias, 0.9))
+            if new_bias != self.bias:
+                self.bias = new_bias
+                return True
+            return False
+        parent = self.parent(layout_object.pairs_root)
+        if parent is not None:
+            which = 1 if parent.one is self else 2
+            return parent.set_size_of_child(which, bias, is_horizontal, layout_object)
         return False
 
     def borders_for_window(self, layout_object: 'Splits', window_id: int) -> Generator[Edges, None, None]:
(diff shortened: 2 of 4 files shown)


kovidgoyal:
By absolute sizes do you mean size in cells or size in bias units?  

If you mean size in cells, there are various problems. If you set the
size in cells for multiple windows and they conflict, what happens? What
happens on an OS window resize, when an absolute size cannot be
respected because there isnt enough space? 

You could possibly mean "absolute" in the sense that specifying a size
in cells will adjust the corresponding bias to achieve that size, and
that's all. If that's the case I am OK with it.

But note that to be acceptable such a PR has to work for *all* layouts
not just splits.

ershov:
That's by far about an absolute bias value. In the current implementation, there's no way to specify that other than changing the global layout setting which will forcedly reshape all existing windows.

I was looking for something similar to 'resize_window' for one window but in the sense of absolute value (not relative increments/decrements).

The current solution allows to set absolute 'bias' for a window in a split. But I can see a relatively simple solution like adding units suffix (like 'px', 'pt', 'cells', '%') to the number argument. This feature might come later as an incremental improvement after an initial solution.

Regarding all other layouts, I agree that consistency is important. I can see how it can be implemented as a more general case in the base class for all layouts.

Does it also imply adding support for 'bias' option to all layouts or it's a separate thing?

What do you think about this feature?

kovidgoyal:
On Thu, Jun 22, 2023 at 03:43:48AM -0700, Yury Ershov wrote:
> That's by far about an absolute bias value. In the current implementation, there's no way to specify that other than changing the global layout setting which will forcedly reshape all existing windows.

I am OK with that.

> 
> I was looking for something similar to 'resize_window' for one window but in the sense of absolute value (not relative increments/decrements).
> 
> The current solution allows to set absolute 'bias' for a window in a split. But I can see a relatively simple solution like adding units suffix (like 'px', 'pt', 'cells', '%') to the number argument. This feature might come later as an incremental improvement after an initial solution.

Just cells absolute and relative are fine as far as I am concerned.
Indeed since windows only have sizes in cells, you cant really use other
units they will get rounded to nearest cells anyway.

> 
> Regarding all other layouts, I agree that consistency is important. I can see how it can be implemented as a more general case in the base class for all layouts.

Yes in order for me to consider a PR it has to be implemented for all
layouts.

> 
> Does it also imply adding support for 'bias' option to all layouts or it's a separate thing?

All layouts internally use bias already otherwise they would not be
resizeable at all. Except for the stack layout, obviously.

> 
> What do you think about this feature?

I'm OK with reviewing a PR for it.

ershov:
> Just cells absolute and relative are fine as far as I am concerned. Indeed since windows only have sizes in cells, you cant really use other units they will get rounded to nearest cells anyway.

Is it ok to start with just bias value (which is equal to percent) and adding cells later as a separate PR?

> Yes in order for me to consider a PR it has to be implemented for all layouts.

Sounds good.

>> Does it also imply adding support for 'bias' option to all layouts or it's a separate thing?

> All layouts internally use bias already otherwise they would not be resizeable at all. Except for the stack layout, obviously.

Ah yes, I mean, bias is used internally but not every layout supports the 'bias' option as a parameter to 'enabled_layouts'. Are you also looking to enable it for all layouts too? This also might go as a separate PR as it's an independent feature which would be good for consistency.

kovidgoyal:
On Thu, Jun 22, 2023 at 02:20:23PM -0700, Yury Ershov wrote:
> > Just cells absolute and relative are fine as far as I am concerned. Indeed since windows only have sizes in cells, you cant really use other units they will get rounded to nearest cells anyway.
> 
> Is it ok to start with just bias value (which is equal to percent) and adding cells later as a separate PR?

It's fine to start with bias but not that bias is only a percent in the
case of two windows involved. If there are more than two then it is no
longer a percent.

> 
> > Yes in order for me to consider a PR it has to be implemented for all layouts.
> 
> Sounds good.
> 
> >> Does it also imply adding support for 'bias' option to all layouts or it's a separate thing?
> 
> > All layouts internally use bias already otherwise they would not be resizeable at all. Except for the stack layout, obviously.
> 
> Ah yes, I mean, bias is used internally but not every layout supports the 'bias' option as a parameter to 'enabled_layouts'. Are you also looking to enable it for all layouts too? This also might go as a separate PR as it's an independent feature which would be good for consistency.

One cannot have bias as a parameter for all layouts, it is meaningless
for some layouts such as stack/horiontal/vertical, it only really works
for splits/vertical/horizontal. And I dont much see the point of it for
splits, why would you want every split you make to start out already
biased?

ershov:
> One cannot have bias as a parameter for all layouts, it is meaningless for some layouts such as stack/horiontal/vertical, it only really works for splits/vertical/horizontal. And I dont much see the point of it for splits, why would you want every split you make to start out already biased?

Since I use 'splits' a lot, I can tell that I expected it to be more consistent. If it has the 'bias' parameter, I don't see why it ignores it in options to 'enabled_layouts  splits:bias=X'.

One of possible uses is: when I open a secondary window, I might prefer it to be smaller than the main one. One of the ways to achieve it would be to specify the 'bias' parameter.

By the way, the 'Pair' currently doesn't clear its 'bias' when one of windows is closed. It leads to an odd behavior of recalling the last 'bias' on the next split.

To reproduce:
1. Enable 'splits' layout.
2. Split the window in two.
3. Resize one of the windows.
4. Close one of splits.
5. Open a split again.

Result: the split ratio is not 1:1 but the same as it was adjusted.

kovidgoyal:
On Fri, Jun 23, 2023 at 01:09:20AM -0700, Yury Ershov wrote:
> > One cannot have bias as a parameter for all layouts, it is meaningless for some layouts such as stack/horiontal/vertical, it only really works for splits/vertical/horizontal. And I dont much see the point of it for splits, why would you want every split you make to start out already biased?
> 
> Since I use 'splits' a lot, I can tell that I expected it to be more consistent. If it has the 'bias' parameter, I don't see why it ignores it in options to 'enabled_layouts  splits:bias=X'.
> 
> One of possible uses is: when I open a secondary window, I might prefer it to be smaller than the main one. One of the ways to achieve it would be to specify the 'bias' parameter.

But setting an overall bias means that *every* split you open in that
layout will be biased.

> 
> By the way, the 'Pair' currently doesn't clear its 'bias' when one of windows is closed. It leads to an odd behavior of recalling the last 'bias' on the next split.

Yes, that's by design. If you want a pair of windows biased, I dont see
why closing one of them neccessarily means you no longer want the bias.

ershov:
>> By the way, the 'Pair' currently doesn't clear its 'bias' when one of windows is closed. It leads to an odd behavior of recalling the last 'bias' on the next split.
> Yes, that's by design. If you want a pair of windows biased, I dont see why closing one of them neccessarily means you no longer want the bias.

Ok but the bias is not saved for the next split level.

cybardev:
@ershov Hi, are you still working on this? Even just this change would be really useful to have, since Fat/Tall already have it.

`kitty/layout/splits.py`:
```diff
-   def __init__(self, horizontal: bool = True):
+   def __init__(self, horizontal: bool = True, bias: float = 0.5):
        self.horizontal = horizontal
        self.one: Optional[Union[Pair, int]] = None
        self.two: Optional[Union[Pair, int]] = None
-       self.bias = 0.5
+       self.bias = bias
```

kovidgoyal:
As of kitty 0.46.0 you can resize windows in all layouts by dragging their borders, making this rather unneccessary

cybardev:
Dragging with the mouse imperatively everytime is no substitute for declarative configuration.

kovidgoyal:
declarative configuration already exists.

cybardev:
No. For splits layout the bias is a hardcoded 50% and currently cannot be changed in configuration. See the file I referenced in my comment above.

kovidgoyal:
No, you can launch individual windows with whatever bias you like
overriding the default bias.
https://sw.kovidgoyal.net/kitty/launch/#cmdoption-launch-bias

cybardev:
That is no substitute for a one-for-all bias like `enabled_layouts splits:bias=N`.

If Fat/Tall can have it, Splits having it as well would make sense for the sake of consistency and convenience. As you mentioned, it would make it simple to achieve this:
> setting an overall bias means that *every* split you open in that layout will be biased.

kovidgoyal:
Sigh, why dont you read the PR description? That's not what it was
about. And I can conceive of no possible use case where anyone would
want a splits layout with a default bias of something other than 50%.
What would be the point? In any case if you do have such a use case feel
free to describe it in an actual PR for that purpose, without the rest
of the changes in this PR.

PR #6388, opened 2023-06-23
Add terminfo entries for cursor shape/color and underline color

mrak:
Kitty already supports the following, but it is nice to include them in the terminfo description so that more applications are aware of its capabilities:

* `Cs` - set the cursor color
* `Cr` - reset the cursor color
* `Ss` - set the cursor shape/style (DECSCUSR)
* `Se` - return cursor shape to original (DECSCUSR)
* `Setulc` - set the underline/undercurl/etc color
* ~~`ol` - reset the underline/undercurl/etc color~~

I am aware that `Setulc` is redundant with `Su` (similar to `Tc` and `setrgbf`?), but there doesn't seem to be consensus yet across terminals/applications/ncurses/tmux/etc about how to specify the capability in terminfo. Adding `Setulc` in particular makes colored underlines work in `nvim` through `tmux`, which is a remote pairing setup that is used at my workplace. I know your feelings toward terminal multiplexers, but this simple change helps to unlock these kitty features in the current complicated terminfo landscape.

Thank you for producing kitty and pushing the envelope when it comes to terminal emulators. It is simply the best cross-platform solution that has the ligatures, performance, and terminal capabilities that I enjoy :)

EDIT: Remove `ol` as per request.


kovidgoyal:
I am not in principle opposed to adding these but they are not defined
in official terminfo, as far as I know. Is there some
documentation/references for who made these and what applications use
them and why?

mrak:
Thank you for the prompt reply!

* tmux source for assigning terminal capabilities related to underline and others: https://github.com/tmux/tmux/blob/29a5dfc7c0574fcb7ff742a2a6566a8b0800adaa/tty-features.c#L161
* NeoVim source for detecting terminal capabilities: https://github.com/neovim/neovim/blob/master/src/nvim/tui/tui.c#L2165
* tmux man page for terminal extensions: https://man7.org/linux/man-pages/man1/tmux.1.html#TERMINFO_EXTENSIONS

kovidgoyal:
As far as I can see these are just random strings invented by tmux
because it insists on using terminfo as a configuration language for
unfathomable reasons.

Why not use terminal-feature to configure tmux yourself?
All modern terminals support these so set them once and forget about them.

Do any other terminals support these terminfo entries? I am not at all
keen to pollute the terminfo namespace, which is a globally shared one,
with tmux specific crud. tmux should at least prefix its custom property
names with tmux-. Especial the shorter two-letter ones are quite likely
to conflict with something Dickey decides to add to official terminfo at
a later date.

mrak:
I believe wezterm has these in it's terminfo at the moment. One convenience of kitty is using its kitten to easily copy its terminfo to a remote server where one may not have control over a shared tmux configuration, but a user's terminfo can be accepted upon attaching as a client.

I agree the current state of "official" or de facto terminfo can feel gated and glacial at times, but at least you can specify extensions such as these that ease interop with applications.

The ones in this PR seem to have some greater community acceptance than, say, emacs' `setf24` for RGB.

I've been using these additions as a personal terminfo fix and thought I'd contribute upstream.

kovidgoyal:
On Thu, Jun 22, 2023 at 11:29:15PM -0700, Eric Mrak wrote:
> I believe wezterm has these in it's terminfo at the moment. One convenience of kitty is using its kitten to easily copy its terminfo to a remote server where one may not have control over a shared tmux configuration, but a user's terminfo can be accepted upon attaching as a client.

Note that the ssh kitten can copy arbitrary files and env vars, so you
can have it auto-install your tmux configuration in addition to
terminfo.

> 
> I agree the current state of "official" or de facto terminfo can feel gated and glacial at times, but at least you can specify extensions such as these that ease interop with applications.

My problem with these is that there is basically one, let's say not very
professional, person in charge of the terminfo official database. He
could at any time decide to introduce new capabilities and they could
conflict with these ones, especially the short two letter ones.  That
said, Ss and Se, at least, are in xterm's terminfo since 2011, so they
should be safe, and Setulc is too long to likely conflict. 

I am a bit iffy on Cs/Cr/ol ol especially given there is already a ul
in official terminfo. I will note that ol is not present in wezterm's
terminfo, as per 

infocmp -1x wezterm | grep '\bol'

Also wezterm has a *lot* less usage than kitty so its likely to trigger
far fewer edge cases. That said, I am willing to merge a subset of
these.

> 
> The ones in this PR seem to have some greater community acceptance than, say, emacs' `setf24` for RGB.
> 
> I've been using these additions as a personal terminfo fix and thought I'd contribute upstream.

I suggest you drop ol and squash the commits, and I will merge.

mrak:
Dropped ol and squashed as requested. Thank you!

PR #6389, opened 2023-06-24
Add option to enable/disable "fake" scroll

tep:
This change adds a new boolean option "wheel_scroll_use_fake" which, when set to "no", will disable the "fake" scroll that is performed when Kitty is in "alternate screen" mode.

NOTE: Since this option's default value is "yes", this change will
      not alter Kitty's behavior unless "wheel_scroll_use_fake" is
      explicitly set to "no".

### DIFF
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -936,7 +936,7 @@ scroll_event(double xoffset, double yoffset, int flags, int modifiers) {
                 }
             } else {
                 if (screen->linebuf == screen->main_linebuf) screen_history_scroll(screen, abs(s), upwards);
-                else fake_scroll(w, abs(s), upwards);
+                else if (OPT(wheel_scroll_use_fake)) fake_scroll(w, abs(s), upwards);
             }
         }
     }
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -420,6 +420,11 @@ negative number, the minimum number of lines will always be added.
 '''
     )
 
+opt('wheel_scroll_use_fake', 'yes',
+    option_type='to_bool', ctype='bool',
+    long_text='Perform a "fake" scroll if kitty is in alternate screen mode.'
+    )
+
 opt('touch_scroll_multiplier', '1.0',
     option_type='float', ctype='double',
     long_text='''
--- a/kitty/options/parse.py
+++ b/kitty/options/parse.py
@@ -1332,6 +1332,9 @@ class Parser:
     def wheel_scroll_multiplier(self, val: str, ans: typing.Dict[str, typing.Any]) -> None:
         ans['wheel_scroll_multiplier'] = float(val)
 
+    def wheel_scroll_use_fake(self, val: str, ans: typing.Dict[str, typing.Any]) -> None:
+        ans['wheel_scroll_use_fake'] = to_bool(val)
+
     def window_alert_on_bell(self, val: str, ans: typing.Dict[str, typing.Any]) -> None:
         ans['window_alert_on_bell'] = to_bool(val)
 
--- a/kitty/options/to-c-generated.h
+++ b/kitty/options/to-c-generated.h
@@ -188,6 +188,19 @@ convert_from_opts_wheel_scroll_min_lines(PyObject *py_opts, Options *opts) {
 }
 
 static void
+convert_from_python_wheel_scroll_use_fake(PyObject *val, Options *opts) {
+    opts->wheel_scroll_use_fake = PyObject_IsTrue(val);
+}
+
+static void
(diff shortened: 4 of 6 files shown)


kovidgoyal:
What's the motivation? Why would producing nothing on wheel input be
desirable? And does it apply to wheel sonly or touchpad based scrolling
as well?

tep:
I have to start by saying, "Thank you!" for kitty. It's a very nice piece of work. I've been using `urxvt` for 6 or 7 years but became frustrated with its limitations of only allowing 256 colors and its lack of support for non-BMP unicode. Kitty covers both of those and a whole lot more.

My motivation for this change is that I have way too many years of muscle memory to **paste** text by pressing the middle button of my mouse -- and, since the middle button is now almost always combined with the mouse wheel, the "fake scroll" is sometimes problematic if I accidentally _nudge_ the cursor in one direction or the other (right before engaging the middle button) and end up pasting text on the wrong line.

This is not a problem with a _normal_ scroll since that simply moves the screen; the cursor remains in the same location and pasted text doesn't end up in the wrong place. But, if the cursor moves due to an accidental fake scroll, the pasted text ends up on the wrong line.

Finally, I don't recall ever suffering this problem with a trackpad but I suppose that's because they don't have the same troublesome, ergonomic design flaw that the mouse-wheel+middle-button combo has given us.

kovidgoyal:
On Sat, Jun 24, 2023 at 07:15:25AM -0700, Tim Peoples wrote:
> I have to start by saying, "Thank you!" for kitty. It's a very nice piece of work. I've been using `urxvt` for 6 or 7 years but became frustrated with its limitations of only allowing 256 colors and its lack of support for non-BMP unicode. Kitty covers both of those and a whole lot more.

You are welcome, I created kitty because I was dissatisfied with all
existing terminals personally.

> 
> My motivation for this change is that I have way too many years of muscle memory to **paste** text by pressing the middle button of my mouse -- and, since the middle button is now almost always combined with the mouse wheel, the "fake scroll" is sometimes problematic if I accidentally _nudge_ the cursor in one direction or the other (right before engaging the middle button) and end up pasting text on the wrong line.
> 
> This is not a problem with a _normal_ scroll since that simply moves the screen; the cursor remains in the same location and pasted text doesn't end up in the wrong place. But, if the cursor moves due to an accidental fake scroll, the pasted text ends up on the wrong line.
> 
> Finally, I don't recall ever suffering this problem with a trackpad but I suppose that's because they don't have the same troublesome, ergonomic design flaw that the mouse-wheel+middle-button combo has given us.

OK, but does setting wheel_scroll_multiplier to zero not work? and if
not, a patch making it work would be more appropriate than a separate
setting.

tep:
> does setting wheel_scroll_multiplier to zero not work? 

That's the first thing I tried -- but, that seemed to also disable the arrow keys in `vim`.  However, I just tried again and that's not the case.

Granted, that was with an older version of kitty and I likely had other config changes in play as well (or maybe I'm just confused).  Either way, I believe setting `wheel_scroll_multiplier` to zero is my solution; I'll retract this PR.

tep:
**Aha!**

The issue I mentioned above (about setting `wheel_scroll_multiplier` to zero) wasn't that the arrow keys stopped working in `vim` -- it was that the mouse wheel no longer scrolls the terminal session when the screen is in _normal_ mode.

I suspect that behavior is working as intended.  If not, I'll see what I can do to fix it. Otherwise, this PR may still be warranted.  To reiterate, my goal is prevent wheel scrolling in alternate mode but not in normal mode.

@kovidgoyal -- What say you?

kovidgoyal:
It is working as intended.

Do you not paste in normal mode? I dont follow your use case in that
case. I thought the problem was you dont like accidental scrolls when
pasting using the middle mouse button?

tep:
Scrolling in normal mode moves the whole screen, not the cursor -- therefore, scroll or not, the text always gets pasted in the desired location.

However, the _fake_ scroll in alternate mode moves the cursor, not the screen -- so an accidental scroll will put the pasted text on the wrong line.

kovidgoyal:
Hmm, well sorry, this use case isn't compelling enough to justify another
option. You are, of course, welcome to run with a local patch. From back
when I used to use a wheel mouse myself I suggest picking up one with a
"click wheel" they tend to be much less easy to accidentally scroll
while clicking.

kovidgoyal:
On Tue, Jun 27, 2023 at 10:11:31AM -0700, Tim Peoples wrote:
> That's unfortunate.
> 
> The other features of Kitty aren't compelling enough to justify the sacrifices you demand so it looks like I'll be switching to another terminal program.

Don't let the door hit you on the way out.

PR #6416, opened 2023-07-03
Enable shadow for transparent window on macOS (fix #2827)

### LINKED ISSUE #2827: Some window decorations missing on macOS when window background_opacity < 1
It looks like some window decorations are missing on macOS when the background opacity is set to < 1: the title bar is drawn, but outside borders and shadows between the window and the rest of the OS are not displayed. (When the window is opaque, window decorations appear normally.)

This is an example of the expected window decorations when a window is positioned over Kitty:
Screen Shot 2020-07-02 at 12 11 08 PM

However, when Kitty is positioned over another window, you can see that there are no borders or shadows, and the Kitty window just blends in with the other window. (Window borders between terminals within Kitty itself are fine.):
Screen Shot 2020-07-02 at 12 11 38 PM

This is in v0.18.1 on macOS Catalina 10.15.5

henryhchchc:
This PR enables the window shadow on macOS when `background_opacity` is set to a value smaller than 1.0.
Related issue was discussed in #2827.

A comparison is as follows.
image

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1825,7 +1825,7 @@ static bool createNativeWindow(_GLFWwindow* window,
     {
         [window->ns.object setOpaque:NO];
         [window->ns.object setHasShadow:YES];
-        [window->ns.object setBackgroundColor:[NSColor clearColor]];
+        [window->ns.object setBackgroundColor:[NSColor colorWithWhite: 0 alpha: 0.001f]];
     }
 
     [window->ns.object setContentView:window->ns.view];
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1824,7 +1824,7 @@ static bool createNativeWindow(_GLFWwindow* window,
     if (fbconfig->transparent)
     {
         [window->ns.object setOpaque:NO];
-        [window->ns.object setHasShadow:NO];
+        [window->ns.object setHasShadow:YES];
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This doesnt interact well with the new background_blur option, try it
with for example

kitty -o background_blur=64 -o background_opacity=0.5

It looks like the shadows in the corners are getting blurred as well.

henryhchchc:
Do you mean the tiny areas in the window corners with a different color?

I spent some time but failed to figure out why this happens. I need some help from who knows macOS and OpenGL well.

Would you accept this PR given that the area is nearly unnoticeable? It appears when `background_blur` is greater than 0 but does not change with the value of `background_blur`.

kovidgoyal:
It doesn't have anything to do with OpenGL, shadows and blur are done by
the OS after opengl is finished rendering. As I said, it will be because
the shadows are blurred too. 

I could accept this PR if turning on the shadows is made conditional on
blur being 0. You will need to add a private API to glfw to toggle the
shadows. And then use it in kitty/glfw.c in create_os_window() and
update_background_blur()

henryhchchc:
Thanks for the tips. I set the background color with black with a very small opacity (i.e. `0.001f`, which is nearly invisible) and the blurred shadow is gone.

As shown in the screenshot above, the colors in the corners come from the wallpaper. As long as the background color is not `[NSColor clearColor]`, the corner will not blend with the wallpaper.

PR #6441, opened 2023-07-11
use 64-bit off_t by default and stop using lfs64 interface

nekopsykose:
fixes the build against musl 1.2.4  

```
kitty/fast-file-copy.c:86:9: error: unknown type name 'off64_t'; did you mean 'off_t'?
   86 |         off64_t r = in_pos;
      |         ^~~~~~~
      |         off_t
```

this is because the \*64 names are no longer exposed by default https://github.com/bminor/musl/commit/25e6fee27f4a293728dd15b659170e7b9c7db9bc

### DIFF
--- a/kitty/fast-file-copy.c
+++ b/kitty/fast-file-copy.c
@@ -83,7 +83,7 @@ copy_with_file_range(int infd, int outfd, off_t in_pos, size_t len, FastFileCopy
 #ifdef HAS_COPY_FILE_RANGE
     unsigned num_of_consecutive_zero_returns = 128;
     while (len) {
-        off64_t r = in_pos;
+        off_t r = in_pos;
         ssize_t n = copy_file_range(infd, &r, outfd, NULL, len, 0);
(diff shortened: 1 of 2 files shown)


kovidgoyal:
You cant just change the size of off_t. It is used in other places in
the code and it would require careful review to make sure doing so is
safe. Not to mention linking against libraries that are compiled without
it and that use off_t will break.

nekopsykose:
it is generally expected these days that everything does actually use file_offset_bits=64 on glibc because it is the prerequisite for time_offset_bits=64 which is required for time_t still working in 2038. pretty much every project passes exactly that.

in this case, the code already handles every case of off_t being 64-bit fine (on 32-bit architectures where it's not the default)

kovidgoyal:
I'd rather not take the chance that everything will just work fine. As I said to be actually sure of that, one has to be sure that all libraries kitty links against are also compiled with off_t=int64_t which is impractical to do.

Issue #6400, opened 2023-06-27
Allow stacking images when using unicode placeholders


Nogesma:
**Is your feature request related to a problem? Please describe.**
The app I am using this for, [twitch-tui](https://github.com/Xithrius/twitch-tui), is displaying images in between text, using them as emojis.
Before unicode placeholders were available I had to keep track of the position of the text and move the images accordingly. 
Now that they are available it simplifies my code a lot, but it's missing a feature I was using, stacking images by specifying a z-index.

**Describe the solution you'd like**
I was thinking of two possible solutions to this problem: 

- Specify another image to stay at the same position of when displaying an image.
ex: `<ESC>_Ga=p,i=<image id>,p=<placement id>,z=<z-index>,b=<image id>,c=<placement id>;`
Where b is the id, and c is the placement id of the other image to "stick" to.
This would allow to display the first image with unicode placeholders, and the rest of images with a display command.
b and c are examples of the keys that could be used, I haven't checked if they're available.

- Join unicode placeholders with a zero width space. 
ex: `\e[38;5;42m\U10EEEE\U0305\U10EEEE\U200B\e[38;5;43m\U10EEEE\U0305\U10EEEE`
This would display the image with id 42 at z-index 0 and the image with id 43 at z-index 1.
I don't know if this one is possible.

I am not familiar with the codebase at all so I might have missed some better way to do this, please let me know.

I'm open to try to make a PR if no one is interested in doing so.

kovidgoyal:
Your approach (2) is not feasible. The way unicode placeholders work is
that the image data is encoded in the unicode data which is stored in
cells. cells cannot store arbitrary amounts of data. Besides I think
this approach is much less flexible than (1), see below.

Some random thoughts on approach 1:

Having image placements that track existing ones is potentially do-able,
but it has some subtleties. One has to consider what happens when the
tracked image is deleted, and avoid endlessly long lookup chains and
cycles in the lookup chains. Then there will be issues with
implementation, I dont know how efficient the current way image
placeholders are stored in kitty is for having one placement track
another. 

From a protocol design perspective, I don't see why this should be
limited to tracking unicode placeholder placements. It would be
generally useful to be able to say create a placement that tracks
this other placement with an offset of X,Y cells. So a complex UI
could be made of 10 different images all positioned relative to each
other, the application can move just the base one and all the rest
will maintain relative position.

So if you are interested in designing such a protocol extension I will
be happy to provide any needed guidance/review. Relevant code is in
[...]

PR #6617, opened 2023-09-12
Graphics protocol: add image chains

Nogesma:
This PR adds the ability to display images relative to other images, discussed a bit already in #6400.

- Images can be positioned relative to a parent image by passing `F=1` in a put command.
- You need reference the parent with its id (`e`) and placement id (`g`).
- You can add a cell offset relative to parent with the `j` and `k` keys, corresponding to x and y offset.
- If you move an image, all of its child will move with it.
- When an image is deleted, all of its child will also be deleted.

The code should work, but I only tested it with simple scenarios (mostly chains of lengths 1 starting with a unicode placeholder).

I'm making this a draft PR as there is still some small issues to be addressed: 
- There is a few recursive functions, they should probably be converted to iterative functions, or I can set a recursion limit.
- The control data keys used should probably be renamed to something else, but most meaningful letters are already used.
- I need to write some tests.
- I need to write the documentation for the website.

Please tell me if you have any suggestions or any more changes are needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ __pycache__/
 .cache
 bypy/b
 bypy/virtual-machines.conf
+.idea
--- a/gen-apc-parsers.py
+++ b/gen-apc-parsers.py
@@ -276,6 +276,10 @@ def graphics_parser() -> None:
         'z': ('z_index', 'int'),
         'C': ('cursor_movement', 'uint'),
         'U': ('unicode_placement', 'uint'),
+        'P': ('parent_id', 'uint'),
+        'g': ('parent_placement_id', 'uint'),
+        'j': ('parent_offset_x', 'int'),
+        'k': ('parent_offset_y', 'int')
     }
     text = generate('parse_graphics_code', 'screen_handle_graphics_command', 'graphics_command', keymap, 'GraphicsCommand')
     write_header(text, 'kitty/parse-graphics-command.h')
--- a/kitty/graphics.c
+++ b/kitty/graphics.c
@@ -682,6 +682,119 @@ update_dest_rect(ImageRef *ref, uint32_t num_cols, uint32_t num_rows, CellPixelS
     ref->effective_num_cols = num_cols;
 }
 
+ImageChainData get_img_ref_data(GraphicsManager *self, ImageRef *parent,
+                                uint32_t img_id, uint32_t ref_id) {
+    Image *child_img = NULL;
+    ImageChainData img_data = {NULL, NULL, NULL};
+
+    child_img = img_by_client_id(self, img_id);
+    if (child_img != NULL) {
+        for (size_t i = 0; i < child_img->refcnt; i++) {
+            if (child_img->refs[i].client_id == ref_id) {
+                img_data.img = child_img;
+                img_data.ref = child_img->refs + i;
+                img_data.parent = parent;
+                return img_data;
+            }
+        }
+    }
+
+    return img_data;
+}
+
+/// This function returns the next element in the image chain.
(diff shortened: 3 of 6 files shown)


kovidgoyal:
I'll review it in detail when I have a moment, probably in a few days.
But I will require tests for this functionality. See
kitty_tests/graphics.py for examples of writing imaging related tests.

kovidgoyal:
Apologies for the delay in reviewing. Some comments:

Escape code design:

1) There is no need for a separate F chain key. Chaining is implied by setting the parent id to non-zero.
2) Use P for parent id and d for placement id
3) d=0 should imply using one of the placements of the specified image, which one is undefined, this simplifies the common case when there is only one placement of the parent image.
4) Why does F=1 mean that the cursor is not moved? (line 1097 in your patch)
5) We should possibly allow refering to parents by image numbers in addition to ids

Code naming:

1) Use anonymous structs for parent data in ImageRef. Something line

struct { uint32 id, placement_id; int32 row_offset, col_offset; } parent, root;
struct { size_t capacity, count; uint32 *child; } children;

2) Use "children" rather than "childs" in function/variable names

Docs and tests:

1) You will need to document your changes in docs/grpahics-protocol.rst

2) There need to be some tests for this functionality in kitty_tests/graphics.py
I am particularly concerned about the interaction of unicode placements and chains

Finally, I have yet to review the actual function implementations in detail, I will update with further comments if I do.

kovidgoyal:
On Tue, Sep 26, 2023 at 08:31:30AM -0700, Mano Ségransan wrote:
> Regarding point 4, when displaying an image in a chain, its position does not depend on the cursor position. I am not sure if it would be expected behavior for the cursor to move.

Ah, makes sense.

Nogesma:
I have pushed an update with all your suggestions. I also rewrote the recursive functions so that there is no recursion anymore.

I don't think I can use the d key for the parent_placement_id, as it is used already for the delete_action, I have left is as g for now.

Nogesma:
I noticed that unicode placeholders were centered by using a source rectangle with values outside of the [0,1] range. I haven't found a way to center a normal image, as the source rectangle is restricted to [0,1] in `handle_put_command`, and the image can only be padded to the right using the `X` key.

This causes issues when trying to overlay images centered on top of Unicode placeholders. 

Do you have any solution on how this could be done?

kovidgoyal:
Not sure I follow, there is no centering in the graphics protocols. If a
client wants an image centered it needs to position it appropriately
based on its knowledge of the screen dimensions.

Nogesma:
What I meant is centering an image inside a terminal cell. 

If for example you want to display an image in a 2x1 cell rectangle, and scale it to fit the rectangle, whilst keeping its aspect ratio, the only way to do that, if I am not mistaken, would be to pad it to the right/bottom by providing a start cell offset with the `X`/`Y` key.
The image would then be offset to either the right or bottom of the rectangle.

What I am looking for is a way to center it inside this cell rectangle.

Unicode placeholders do this centering by using a source rectangle with values outside of the [0,1] range.
Providing a way to pad the image to the left/top by giving an end cell offset would also achieve the same result.

kovidgoyal:
If you want to center an image inside some cell block you size the image so that its dimensions are smaller than the cell block, then in each dimension you subtract the size of the image from the size of the cell block, divide by 2, and set X and Y to those values. X and Y are offsets added to the origin at the top left corner of the cell.

kovidgoyal:
I ended up implementing this differently using hash tables. Feel free to comment.

Nogesma:
Nice thanks a lot

Issue #6363, opened 2023-06-14
Selecting word and until end of line


warpfork:
**Is your feature request related to a problem? Please describe.**

A feature I love from other terminals is tripleclick (or your choice of action) to select the current word (much like doubleclick does by default in kitty) and then extend the selection also to the end of the current line.

"Current word and to end of current line" is very often the ideal "do what I mean" for selecting part of a script, etc.  By contrast, selecting the whole line is often _not_ exactly what I mean, because it will also grab part of the terminal prompt text (e.g. the `$` or `#`, typically).

**Describe the solution you'd like**

I think what I'm asking for is the ability to say something like `mouse_map left triplepress ungrabbed mouse_selection line_from_word`.

Or, some way to combine `mouse_selection word` and `mouse_selection line_from_point`.

**Describe alternatives you've considered**

I've been trying to find some incantation combining existing features like `word` and `line_from_point` and possibly `extend` or `move-end`.  Or really any combination of keywords I see in https://sw.kovidgoyal.net/kitty/conf/#conf-kitty-mouse-mousemap .  I haven't yet found the correct incantation, though, if one exists :)

If there's a way to chain selection adjustments, it seems like that would work, but I don't think I've found examples of that.  (I see comma-separators in that mousemap config docs page, but those don't appear to be for the composing of actions.)

**Additional context**

Think that should do it.  Kitty is pretty great, thanks :)

kovidgoyal:
Double click to select word then right click to extend selection.

warpfork:
Is there a way I can map that to a single action, though?

kovidgoyal:
I dont understand why you would want to select a word and then extend to end of line. Instead select to end of line from the start of the word. I suppose it requires a bit more precision to move your mouse to the start of the word rather than at any point in the word, but that's not sufficient motivation for me to implement such an action. A PR is most welcome.

warpfork:
Right, it's a precision-required vs ease thing :)  I find it's a very very common action for me, so making it easy and require lower precision of motion is a large life-improvement.

I may try to make a PR, but I'm not very familiar with either Python or C and I might have some questions about the build process. :)

PR #6663, opened 2023-10-02
Add word-and-line-from-point as a mouse selection action

warpfork:
Hello, this patch attempts to introduce a new selection mode called "word_and_line_from_point".

The behavior of this is to select the word under the cursor and then also out to the end of the line.  It's functionally equivalent to using `word` and `line_from_point` in sequence and taking the combined result.  (The desire for this was previously described in https://github.com/kovidgoyal/kitty/issues/6363 .)

It can be used in `kitty.conf` with a config line such as:

```
mouse_map left triplepress ungrabbed mouse_selection word_and_line_from_point
```

I attempted to keep the diff as minimal as possible, and here's what is included:

- `MOUSE_SELECTION_WORD_AND_LINE_FROM_POINT` is added to mouse selection types in C and python bindings, so it's communicable from the config options layer.
- the string "line_from_word" is mapped to `MOUSE_SELECTION_WORD_AND_LINE_FROM_POINT` in `kitty/options/utils.py` function, so that string can be recognized in the config file.
- in `mouse.c`, `mouse_selection` function gains a new case so it can forward the `MOUSE_SELECTION_WORD_AND_LINE_FROM_POINT` action to the `screen_selection_*` functions.
- `EXTEND_WORD_AND_LINE_FROM_POINT` is added to the `SelectionExtendModes` enum in C, so our mode can be communicated to the `screen_*_selection` functions.
- `screen_update_selection` gains a very small branch for `EXTEND_WORD_AND_LINE_FROM_POINT` in the middle of the overall `EXTEND_LINE` logic, which performs the search for word.  If there's a word, the selection start moves to the start of the word.  (If there's no word, the selection is made as it would be otherwise by regular `EXTEND_LINE`, meaning it starts at the cursor.)
- documentation added in `kitty/options/definition.py`.

I'm testing this manually and it seems to have the desired behavior.  It works on normal lines; it works on long lines that wrapped; it works on the non-first rendered line of wrapped lines; it combines fine with extending selections.  I've been driving it around for a few days now and encountered no surprising behavior.

I'm happy to accept any feedback about the feature name, or the code.  I think it works, and I tried to keep the change minimal.  But this is my first attempt to submit a patch to Kitty, and also the first time I've touched C code in many years, so I cannot guarantee it is perfect on the first pass. :)  Thank you very much for your review!


kovidgoyal:
At a quick glance it looks fine. Some comments:

1) Add some docs to options/definition.py -- wherever line_from_point is used to its docs add a sentence saying that one can instead use word_and_line_from_point to have it select the word under the cursor and extend to end of line

2) rename line_from_word to something like word_and_line_from_point

I'll review in detail after you make these changes.

warpfork:
That's a much better name, thanks.  Rebased to that.

Brief docs also added to `definition.py`.

warpfork:
:tada: :star_struck: :woman_dancing: Thank you!!  This made my day.

PR #6671, opened 2023-10-03
kitten: Use /dev/fd/2 for stderr

glensc:
This is a fix for termux under android:

> $ kitten --version
/data/data/com.termux/files/home/.local/share/kitty-ssh-kitten/kitty/bin/kitten: 9: cannot create /dev/stderr: Permission denied


glensc:
low priority,. as seems it's not going to work anyway:

```
$ kitten --version
Unknown CPU architecture armv7l
```

PR #6672, opened 2023-10-03
kitten: Add armv7l alias to arm

glensc:
Fix `Unknown CPU architecture armv7l` on raspberry pi 400


kovidgoyal:
kitty is not available for arm only arm64 so you can only modify the kitten for this case.

PR #6817, opened 2023-11-14
command done notification

jinliu:
Just a Proof of Concept at the moment.

This implements "send desktop notification when a command is done in a background window" as plugins in various shells do (e.g. https://github.com/franciscolourenco/done), but should work in any shells with OSC 133 shell integration.

TODO:
1. configuration (notify_cmd_done=off/unfocused/invisible, notify_cmd_done_min_duration)

Open questions:
1. Whether to hard-code it as in the PoC, or to extend the Watcher API and implement it as a built-in Watcher.
2. To detect the end of command, should we generate and parse OSC 133;D (end of command output), instead of using OSC 133;A (start of prompt) as in the PoC? 
3. Should the notification include the cmdline? By generating and parsing OSC 133;B? Or by injecting code in each shell and sending it via OSC 1337?
4. Should we support changing the title (or somehow make it "flash") of an inactive tab in a focused window, instead of sending a notification?
5. How to use a C enum (e.g. PromptKind) in Python?

### DIFF
--- a/docs/launch.rst
+++ b/docs/launch.rst
@@ -145,6 +145,10 @@ functions for the events you are interested in, for example:
         # when a title change was requested via escape code from the program
         # running in the terminal
 
+    def on_cmd_startstop(boss: Boss, window: Window, data: Dict[str, Any]) -> None:
+        # called when the shell starts/stops executing a command. Here
+        # data will contain is_start and time.
+
 Every callback is passed a reference to the global ``Boss`` object as well as
 the ``Window`` object the action is occurring on. The ``data`` object is a dict
 that contains event dependent data. Some useful methods and attributes for the
--- a/kitty/launch.py
+++ b/kitty/launch.py
@@ -398,6 +398,15 @@ def load_watch_modules(watchers: Iterable[str]) -> Optional[Watchers]:
         w = m.get('on_focus_change')
         if callable(w):
             ans.on_focus_change.append(w)
+        w = m.get('on_set_user_var')
+        if callable(w):
+            ans.on_set_user_var.append(w)
+        w = m.get('on_title_change')
+        if callable(w):
+            ans.on_title_change.append(w)
+        w = m.get('on_cmd_startstop')
+        if callable(w):
+            ans.on_cmd_startstop.append(w)
     return ans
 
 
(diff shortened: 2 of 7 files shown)


kovidgoyal:
On Mon, Nov 13, 2023 at 08:08:44PM -0800, Jin Liu wrote:
> Just a Proof of Concept at the moment.
> 
> This implements &quot;send desktop notification when a command is done in a background window&quot; as plugins in various shells do (e.g. https://github.com/franciscolourenco/done), but should work in any shells with OSC 133 shell integration.
> 
> TODO:
> 1. configuration (notify_cmd_done=off/unfocused/invisible, notify_cmd_done_min_duration)
> 
> Open questions:
> 1. Whether to hard-code it as in the PoC, or to extend the Watcher API and implement it as a built-in Watcher.

I suggest doing both. That way the config option can be relatively
simple, and people that want uber custom behavior can write a simple
watcher for it.

> 2. To detect the end of command, should we generate and parse OSC 133;D (end of command output), instead of using OSC 133;A (start of prompt) as in the PoC? 

I dont see the need it would be pretty rare that you get end of command
not followed by start of prompt.

> 3. Should the notification include the cmdline? By generating and parsing OSC 133;B? Or by injecting code in each shell and sending it via OSC 1337?

Not sure, I dont see much value in the notification having the command
line. command lines are usually quite long not really suitable for 
a notification popup.
If you do want it then I would suggest using 1337 as this information
can potentially have other uses as well.

> 4. Should we support changing the title of an inactive tab of a focused window, instead of sending a notification?

Why not, its fairly easy to do.

leiserfg:
I was looking into this as well but was trying to add the logic into the fish/bash... code. This approach requires less work so may I suggest adding a setting for minimum time and one for blacklisting commands?

jinliu:
> I was looking into this as well but was trying to add the logic into the fish/bash... code. This approach requires less work so may I suggest adding a setting for minimum time and one for blacklisting commands?

It already has a minimum time option:
https://github.com/kovidgoyal/kitty/pull/6817/files#diff-9f4d52d09d15bdd4c9acdad3c8d90c648be21236a00f79d0b68b4b88070448c9R3175

Blacklisting commands would require knowing the command first, i.e. the open question 3 above. Probably better done in the injected code for each shell (send back via OSC 1337).

PR #6819, opened 2023-11-14
Add a `notify_on_cmd_finish_use_bell` option

jinliu:
When enabled, ring a terminal bell instead of a desktop notification when a command finishes.

Since I exported screen_bell() in the Screen Python type, do I need to manually update fast_data_types.pyi, or is it somehow auto-generated? Seems it doesn't contain every Screen methods?


kovidgoyal:
It's manually maintained. Also combine this into the existing notify
option, something like

notify_on_cmd_finish when duration action

where action can be one of
notify, bell, command

If it is command, all subsequent arguments are the cmdline to run.

PR #6875, opened 2023-11-30
fix: Correct Nix shell for building correctly

gabyx:
Some small fixes to build on NixOS.


kovidgoyal:
See https://sw.kovidgoyal.net/kitty/build/ for how to build and run from
source. If you want to build kitty.app you do make app

Issue #6930, opened 2023-12-20
New os_window with `--singe-instance` and `--instance-group` spawns on non-active display on macOS


FelixKratz:
**Describe the bug**
I am using kitty on macOS 14.2 and use
```bash
open -na /Applications/kitty.app --args --single-instance --instance-group kk
```
to create a new kitty os_window. 
* The first window is spawned on the active display.
* All further os_windows are spawned on the display where the first window was spawned, even if it is no longer the active one. 
* Even `cmd+n` will spawn the new os_window on the "wrong" display, if invoked on a window located on a different display than the first created window.

**To Reproduce**
1. 
```bash
open -na /Applications/kitty.app --args --single-instance --instance-group kk
```
2. Switch focus to another display
3.
```bash
open -na /Applications/kitty.app --args --single-instance --instance-group kk
```
4. Notice that the new window is not spawned on the currently active display but on the one where the first os_window was created.

**Environment details**
```
kitty 0.31.0 (b247fda672) created by Kovid Goyal
Darwin Felixs-MBP 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:53:18 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6000 arm64
ProductName:		macOS ProductVersion:		14.2 BuildVersion:		23C64
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /bin/zsh
Loaded config files:
  /Users/felix/.config/kitty/kitty.conf

Config options different from defaults:
active_tab_font_style   (True, False)
bold_font               Liga SFMono Nerd Font Bold
bold_italic_font        Liga SFMono Nerd Font Bold Italic
confirm_os_window_close -2
cursor_blink_interval   0.0
cursor_text_color       None
font_family             Liga SFMono Nerd Font
font_size               13.0
hide_window_decorations 2
input_delay             1
italic_font             Liga SFMono Nerd Font Regular Italic
macos_colorspace        default
remember_window_size    False
repaint_delay           8
resize_debounce_time    (0.001, 0.5)
symbol_map:
	U+f8ff - U+f8ff → SF Pro
	U+f0001 - U+f1af0 → Hack Nerd Font
	U+100000 - U+1018c7 → SF Pro
[...]

PR #6932, opened 2023-12-20
spawn new os_window on active screen on macOS

FelixKratz:
This fixes the bug discussed in #6930.


kovidgoyal:
Rather than disabling cascading which would break expected behavior on this platform why not check if the cascade point is on the same monitor and only use it then?

FelixKratz:
The newest commit checks if the cascade point is on the same screen, if not, it reset it to the window origin, such that cascading of further windows will work on that monitor.

PR #7009, opened 2024-01-17
switch to recent os window: nth_os_window -1

### LINKED ISSUE #7008: feature: Switch to recent os window
### Discussed in https://github.com/kovidgoyal/kitty/discussions/7007

<div type='discussions-op-text'>

<sup>Originally posted by **jackielii** January 17, 2024</sup>
I'm trying to setup my kitty to fully replace Tmux, one thing I use very often is tmux session switching. Specifically switch to last session.

In kitty, there is no session, but I use a separate OS window for a session which works great. But there doesn't seem to be a way to switch to the "recent focused" OS window.

There is the `nth_os_window` mappable action, but it doesn't accept `-1` to switch to last-focused window like `nth_window -1` does.

In `boss.py` there also doesn't seem to be a way. Is this possible?</div>

jackielii:
fix #7008

1. expose `last_focused_counter(os_window_id) -> int`
2. allow mappable action `nth_os_window -1` to switch to last focused OS window

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -92,6 +92,7 @@ from .fast_data_types import (
     get_os_window_size,
     global_font_size,
     is_modifier_key,
+    last_focused_counter,
     last_focused_os_window_id,
     mark_os_window_for_close,
     os_window_font_size,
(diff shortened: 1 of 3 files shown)


kovidgoyal:
As I mentioned on #7007 that function is for the case when no OS window has focus. So it is correct. I suggest you instead create another function that returns the value of last_focused_counter given and os_window_id. Then any history algorithm can be implemented in python code using that function.

jackielii:
added per your suggestion. Thanks

jackielii:
Thanks

PR #7012, opened 2024-01-18
Add Python APIs to get and set os-window position

ad-chaos:
Useful when using floating window managers to get a floating terminal placed appropriately with either a shell, compiler output or a man page or other reference material to be temporarily available at a glance.

### DIFF
--- a/kitty/fast_data_types.pyi
+++ b/kitty/fast_data_types.pyi
@@ -1396,6 +1396,12 @@ def get_os_window_size(os_window_id: int) -> Optional[OSWindowSize]:
     pass
 
 
+def get_os_window_pos(os_window_id: int) -> Tuple[int, int]:
+    pass
+
+def set_os_window_pos(os_window_id: int, x: int, y: int) -> None:
+    pass
+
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I'm not a big fan of this. On Wayland both getting and setting OS window position is not allowed for the good lord alone knows what reasons. On X11, setting wont work with non-floating windows/window managers. That is why I have not implemented a --geometry flag for kitty to set OS window geometry. I dont like to implement features that work only some of the time, it leads to lot of support noise for me.

ad-chaos:
Understandable, this could be a little ugly, but how do you feel about only allowing this under `if is_macos:` and a noop otherwise?

kovidgoyal:
Well, I suppose I am ok with exposing the glfw functions to get/set os window pos via state.c but not as part of the Boss object. That way its clearly internal API, use at your own risk. As for add_os_window() there is no real need for it there, as you can just call setwindowpos after adding. Practically it wont make any difference unless the position is on a different monitor or similar.

ad-chaos:
Done

ad-chaos:
Thank you!

PR #7017, opened 2024-01-19
Explicit GO111MODULE=on when getting required Go version

Maytha8:
Rather than assuming that GO111MODULE hasn't been modified, this PR makes it so GO111MODULE is explicitly set when running `go list` to extract the required Go version from go.mod, ensuring the version check works even outside of module-aware mode.

Taken from [Debian](https://salsa.debian.org/debian/kitty/-/blob/debian/sid/debian/patches/0011-fix-go-version-check.patch).


Maytha8:
Fixed

PR #7120, opened 2024-02-11
Add support for using main font for box drawing

erikolofsson:
I have a font that is designed for pixel perfect rendering. Box drawing characters from the font works without the need for rendering them separately, and I would prefer to use the font characters.

This implements an option to prefer the main font for box drawing:

image


kovidgoyal:
I am afraid not. As I have said before, box drawing characters are
supposed to have a defined appearance. Allowing users to override that
appearance means TUI programs cant rely on the appearance.

erikolofsson:
Both iTerm2 and Alacritty has options for this, but if you can't be persuaded I can always keep a fork around. Thanks.

PR #7121, opened 2024-02-11
Drawing method for U+1FB98 and U+1FB99

ReubenM:
I don't know if perhaps there are any corner cases where the implementation in this new shading method might have undesired results. I've only tested with the cell size configured on my system.

Produces some interesting moiré patterns.

Screenshot_20240211_140956


kovidgoyal:
Renders as solid blocks on my system

screenshot

ReubenM:
What is your cell size? I was testing using 120x40, but that's taller aspect ratio than most people probably use.

To reduce line density, change
```slice_count = width // 10 if width // 10 >= 2 else 2```
increase the value "10" to something higher like "20" and see if that separates the lines enough.

PR #7134, opened 2024-02-17
Add `new-tab-neighbor` option to `detach_window`

ces42:
This should be pretty self-explanatory.

Rationale: I usually have plenty of different tabs open in which I'm doing different things (work/system troubleshooting/configuring nvim/htop/...) and use `detach_window` when my current tab gets too crowded. I would like related tabs to stay together, so an option to detach a window to a new tab, _which gets created next to the current tab_ would be useful.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -2675,10 +2675,9 @@ class Boss:
                         tm = q
                     else:
                         tm = self.os_window_map[target_os_window_id]
-                    if target_tab_id == 'new':
-                        target_tab = tm.new_tab(empty_tab=True)
-                    elif target_tab_id == 'new_as_neighbor':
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Not sure I follow wont tab-right or tab-left do what you want already?

ces42:
Uh, it seems to me tab-right and tab-left don't create new tabs. I just want new-tab except the new tab is created next to the active tab.

kovidgoyal:
ah ok you want a new empty tab. Maybe have new-tab-left and
new-tab-right then. Am sure someone in the future will want to detach to
a tab to the left instead of neighbor which IIRC is the tab to the
right.

ces42:
Yes probably new-tab-left is also useful. Do you want me to implement it?

kovidgoyal:
yes.

PR #7291, opened 2024-04-01
Add fullscreen on startup option

IAmMoltony:
This adds a new feature: fullscreen on startup, which is controlled by a new configuration option: `fullscreen_on_startup`, which is set to `no` by default.


kovidgoyal:
There is already a command line flag to do this. It makes no sense to have this as a config option, config options apply to all kitty instances. If you really want all kitty instances to be full screen, then create an alias or a edit the desktop file to use the kitty facility for specifying command line flags to macOS apps.

If there is some use case I havent covered feel free to comment.

PR #7293, opened 2024-04-01
Add link to kitty-toggle-terminal kitten

shardulbee:


kovidgoyal:
Good you wrote a kitten, but for this particular task you dont need one. You can use a conditional mapping, see https://sw.kovidgoyal.net/kitty/mapping/#conditional-mappings-depending-on-the-state-of-the-focused-window

create a conditional mapping that is active only when in editor it can use the combine action to switch to the fat layout and open a new window.

shardulbee:
Thanks for the pointer and the reminder to read the docs more thoroughly 😄

Issue #7200, opened 2024-03-06
[fish shell integration] repaint removes text printed by fish key binding


iacore:
**Describe the bug**
text printed by keybind in fish disappear when screen is repaint (such as resizing the window)

**To Reproduce**
1. open new window with `kitty --config NONE fish`
2. run `bind \eg 'echo; git status; git remote -v; echo; commandline -f repaint'`
3. press Alt+g in a git repo
4. resize the window to be wider or narrower (in width)

**Screenshots**

The screenshots below show the whole window width.

After `git init`, I pressed Alt+g, and git information is shown.

![image](https://github.com/kovidgoyal/kitty/assets/74560659/9909270a-0e78-44be-8219-a8f819f6d9cb)

Then, I resize the kitty window to be wider, and the git info text is gone, leaving an empty space.

![image](https://github.com/kovidgoyal/kitty/assets/74560659/843c77a6-2f7b-455f-98d5-4f62c5580e6e)

**Environment details**
```
❯ kitty --version
kitty 0.31.0 created by Kovid Goyal

❯ fish --version
fish, version 3.6.4
```

**Additional context**

Ctrl+Shift+G (show last command output in pager) ignores the output printed by fish keybind. I think this is the appropriate behavior.

Can reproduce with `kitty --config NONE`. Latest fish version (3.7.0) probably works too.

The issue is first discussed in https://github.com/fish-shell/fish-shell/discussions/10289.

iacore:
I have "fixed" the problem by calling `__ksi_mark_prompt_start` at the end.

e.g. `bind \eg 'echo; git status; git remote -v; echo; commandline -f repaint; __ksi_mark_prompt_start'`

kovidgoyal:
I dont know of anyway to fix this. If commandline -f repaint doesnt run
the prompt marking hooks, there is nothing the shell integration script
can do about it. Really, the correct fix for this is for fish to
implement prompt marking itself instead of every terminal emulator
having to write a script for it.

Issue #7168, opened 2024-02-28
Fish shell integration automatically completes currently shown completion


martonperei:
**Describe the bug**
When the `prompt-mark` shell integration is enabled with fish, occasionally the currently shown completion is auto-completed.
The problem doesn't happen with `shell_integration no-prompt-mark`.

Sometimes it happens right after opening a new tab and typing a command, ending up in something like `git checkout developit clone` when I want to type `git clone`

**To Reproduce**
1. Launch kitty with the following config: `shell /opt/homebrew/bin/fish -N --login --interactive`
2. Copy paste the manual shell integration for fish from here: https://sw.kovidgoyal.net/kitty/shell-integration/
3. Start writing a command, after some time (mouse clicks sometimes trigger it), the command will auto-complete

**Screenshots**
Completion is shown:
<img width="238" alt="Screenshot 2024-02-28 at 21 04 09" src="https://github.com/kovidgoyal/kitty/assets/5266196/e3970db3-cd24-49b6-a09b-78a7e1f005f6">
Completion is auto-completed automagically.
<img width="232" alt="Screenshot 2024-02-28 at 21 04 15" src="https://github.com/kovidgoyal/kitty/assets/5266196/a79c8383-b904-4057-ba50-e0a83359fcc4">

**Environment details**
```
kitty 0.32.2 (925043d645) created by Kovid Goyal
Darwin martons-mbp.lan.****.*** 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64
ProductName:		macOS ProductVersion:		14.3.1 BuildVersion:		23D60
Frozen: True
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /opt/homebrew/bin/fish
Loaded config files:
  /Users/mperei/.config/kitty/kitty.conf

Config options different from defaults:
shell /opt/homebrew/bin/fish -N --login --interactive
Removed shortcuts:
	cmd+, →  edit_config_file
	cmd+enter →  new_window
	cmd+h →  hide_macos_app
	cmd+k →  clear_terminal to_cursor active
	cmd+m →  minimize_macos_window
	cmd+n →  new_os_window
[...]

PR #7316, opened 2024-04-06
fish integration: drop redundant OSC 133 markers in upcoming fish 3.8

krobelus:
The upcoming fish 3.8 release will output OSC 133 sequences unconditionally [1].

I tested ctrl-shift-{g,x,z} bindings both without and with kitty's shell integration on top; everything seems to work.

Let's simplify kitty integration by removing the markers for the upcoming fish >= 3.8.

I have hopes that the native OSC 133 implementation address #7200 though I'm not sure if I could reproduce this bug
(I only saw a similar bug when `fish_handle_reflow` was not enabled, which fish also does now (same commit)).
cc @iacore, let me know if you can reproduce #7200 with latest fish master.

[1]: https://github.com/fish-shell/fish-shell/commit/3b9e3e251bf9d4c7d0b31275cac55df68fe0127a

### DIFF
--- a/docs/keyboard-protocol.rst
+++ b/docs/keyboard-protocol.rst
@@ -52,6 +52,7 @@ In addition to kitty, this protocol is also implemented in:
 * The `yazi file manager <https://github.com/sxyazi/yazi>`__
 * The `awrit web browser <https://github.com/chase/awrit>`__
 * The `nushell shell <https://github.com/nushell/nushell/pull/10540>`__
+* The `fish shell <https://github.com/fish-shell/fish-shell/commit/8bf8b10f685d964101f491b9cc3da04117a308b4>`__
 
 .. versionadded:: 0.20.0
 
--- a/docs/shell-integration.rst
+++ b/docs/shell-integration.rst
@@ -85,6 +85,8 @@ no-cwd
 no-prompt-mark
     Turn off marking of prompts. This disables jumping to prompt, browsing
     output of last command and click to move cursor functionality.
+    Note that for the fish shell this does not take effect, since fish always
+    marks prompts.
 
 no-complete
     Turn off completion for the kitty command.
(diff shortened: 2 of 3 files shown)


kovidgoyal:
Cool, that's great news. One question, the bug we recently collaborated on, mouse clicks to move the cursor #7168, this requires prompt marking to include the special_key=1 field, so that kitty knows to use a special key instead of the arrow keys for clicks. Is that included in the fish builtin prompt mark?

krobelus:
My bad, I'll push a fix. I also need to fix some other things with fish's CSI u parsing for this to work.
My preliminary testing shows that the `forward-char-passive` approach has noticeable delay, at least with fish in debug mode.
If that's not just a bug in fish we should probably use `commandline --cursor 123` to set the cursor position directly (the argument is a 1-based code point index from the start of the command line).
I'll look into it because it also needs a new feature to work with the search field..

krobelus:
I've pushed the functional fixes to fish (`866585c~3..866585c`).

I haven't found a source of slowness in fish yet, still looking.
I'm trying to run `kitty/launcher/kitty --debug-input` but that doesn't include a timestamp for the mouse click.
A relative timestamp would be enough, I think I'll add relative timestamps to fish logs.

---

Also whenever I type `kitty <TAB>` in fish I get an error

`Error: Invalid shell state specification: fish2`

I don't know where `fish2` is coming from. I'm using 0.33, probably same on master. cc @pagedown

krobelus:
Ok I figured out the source of "slowness" - it's the double click interval (wayland here).
Seems to be somewhere below 500ms here. So this is fine.

krobelus:
found the completion issue as well, I had a stale `kitty.fish` from a old `make install` of fish.. I wish I didn't need to install, it'd be much better to provide a relocatable binary :(

kovidgoyal:
merged, changing the code to only run bind --function-names once

iacore:
> cc @iacore, let me know if you can reproduce #7200 with latest fish master.

The problem is solved. There is another weird thing though.

When I start fish shell and not run anything, when I press Ctrl+Shift+G, the following text is shown in pager:

```
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
```

This is not the output of a command, but the welcome prompt.

krobelus:
Yeah I think that's a feature, not a bug. I expect that it's the same for motd (which can be somewhat long)

PR #7443, opened 2024-05-17
Try to workaround race condition for @ run

amosbird:
Address https://github.com/kovidgoyal/kitty/commit/38fed8b391ebbc1c2ac8c72102e6cd18418f6bab#commitcomment-142137229

### DIFF
--- a/tools/cmd/at/socket_io.go
+++ b/tools/cmd/at/socket_io.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"time"
 
+	"golang.org/x/sys/unix"
 	"kitty/tools/tui/loop"
 	"kitty/tools/utils"
 	"kitty/tools/wcswidth"
@@ -59,11 +60,16 @@ func read_response_from_conn(conn *net.Conn, timeout time.Duration) (serialized_
 		var n int
 		(*conn).SetDeadline(time.Now().Add(timeout))
 		n, err = (*conn).Read(buf)
+		if err == unix.EAGAIN || err == unix.EINTR {
+			err = nil
+		}
 		if err != nil {
 			keep_going = false
 			break
 		}
-		p.Parse(buf[:n])
+		if n > 0 {
+			p.Parse(buf[:n])
+		}
 	}
 	return
 }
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Rather than mess with the loop in socket-io I chose to just run the
background process after an event loop tick, which *should* take care of
it, cant be sure as I cant reproduce.

amosbird:
> I chose to just run the background process after an event loop tick

This definitely looks better. However, after some tests, it seems that `add_timer(callback, 0, False)` doesn't queue the callback properly to make sure `AsyncResponse()` has been sent.

I've tested `add_timer(callback, 0.01, False)`, 18/20 work fine. With `add_timer(callback, 0.1, False)`, 100% good.

amosbird:
> cant be sure as I cant reproduce.

I'm on a remote ssh connection with kitty domain socket forwarding, which makes it easy to reproduce. I guess you can simply add time.sleep(...) before returning `AsyncResponse()` to reproduce it :)

kovidgoyal:
That does not reproduce it. I dont see why it would either. add_timer
schedules the callback it wont actually run till after
response_from_kitty returns as response_from_kitty is running on the
main thread and blocking the event loop. You can trace this sequence of
events. In child-monitor.c around line 500 you can see the return value
being sent. You can add

log_error("%s", msg)

at the top of send_response_to_peer() that will show you everything that
is sent to the peer in order.

amosbird:
Interesting. Both good and bad cases print the same event log which is indeed in correct order:

```
[35.176] ␛P@kitty-cmd{"ok": true, "stream": true}␛\
[35.182] ␛P@kitty-cmd{"ok": true, "data": {"stdout": "MQo=", "stderr": "", "exit_code": 0, "exit_status": 0}}␛\
[35.191] (null)
```

I guess something is wrong at the kitten side.

kovidgoyal:
You can add

forward_stdio yes

to kitty.conf

then in socket_io.go

you can do

import 	"kitty/tools/tty"

and then use

tty.DebugPrintln(...)

in simple_socket_io() to see whats going on. The output of the prints
will be in the stderr of the kitty inside which the kitten runs.

amosbird:
https://github.com/amosbird/kitty/blob/master/tools/cmd/at/socket_io.go#L61

Here bad cases read two responses in one `Read` call and parse them together, which will swallow the first `AsyncResponse()`.

kovidgoyal:
Should be fixed by: https://github.com/kovidgoyal/kitty/commit/703068f884479bbfbfcb641ae75ed7b9e56e1ecd

kovidgoyal:
Or actually no maybe not quite, the parser can call handledcs multiple times in one read

kovidgoyal:
Next commit fixes it, I hope.

amosbird:
Unfortunately, `kitten @ run` gets stuck when packet coalescing happens with a661f0065  Fix the previous fix.

kovidgoyal:
https://github.com/kovidgoyal/kitty/commit/a9924d2ab76ae3399ced776e8df9931320f1e560

amosbird:
a9924d2 works! Thank you!

PR #7526, opened 2024-06-08
Add `--help` option to `kitty list-fonts`

ilyagr:
I haven't tested this, but it should work. Feel free to take this as an FR instead of a PR.

I'm assuming that `kitty list-fonts` and `kitty +list-fonts` are equivalent.


kovidgoyal:
list-fonts is becoming a full fledged font selection UI in the next
kitty release, see the var branch, with support for font features and
variable fonts.

Issue #7532, opened 2024-06-11
Feature Request: Option to clamp background outside center or scale window logo


amuDev:
**Is your feature request related to a problem? Please describe.**
The `window_logo` parameter seems to have the same goals that I am trying to achieve but with one major shortfall, there is no scaling applied when the window size changes. Why might this be a problem? Sometimes I just want to have a nice logo on that scales in my terminal. Using the current window_logo creates either a logo far too small when using a larger terminal window, or a logo that blocks far too much content on smaller terminal windows.

**Describe the solution you'd like**
Ideally an additional parameter that is tied either to background or window logo. In the case of the background having a choice to set the anchor point to cscale (instead of forced centre). In the case of the logo, a simple flag to enable scaling, though this seems a bit more complex since there would also likely need to be a ratio set to determine what to scale the logo to.

**Describe alternatives you've considered**
Not many relevant alternatives that arent listed. I have created a [fork branch](https://github.com/kovidgoyal/kitty/compare/master...amuDev:kitty:bg-br-clamping) that simply forces this behaviour when using cscaled, though this is purely for my use and a further demonstration to what the goal is. Although the actual implementation is just changing the scaling behaivour on the `draw_background_image` method with slightly different calculations for left and top to allow clamping on bottom right.

**Additional context**
![image](https://github.com/kovidgoyal/kitty/assets/51936412/f941addc-cb64-48bf-ba2b-5813aa54adfc)

kovidgoyal:
I dont follow your use case. If you want a logo that scales with the
terminal window, simply use a background image that is completely
transparent other than the logo part in the bottom right (or wherever)
and make sure this image is larger than the max terminal size you use.
It will be scaled automatically and your logo will be always displayed
at the bottom right scaled alongwith it.

As for your feature request, I am in principle fine with adding a
window_logo_scale type option but I dont really see the need given what
I said above. In any case I am not going to implement such a feature
myself, but a PR is welcome. If you do choose to submit a PR we can
discuss the semantics of such an option.

amuDev:
The problem with using cscaled in conjunction with an oversized image is that when the aspect ratio of the image and window are different there is some portion of the image that gets cut. This can't be mitigated much since (at least on a tiling window manager) the aspect ratio of a given terminal session can vary from the whole screen to a small box, in cases where there is a greater discrepancy between the usable aspect ratios and the images aspect ratio, it becomes very difficult to have a properly positioned "logo" while still scaling.

[...]

PR #7534, opened 2024-06-12
add option to scale window_logo via window_logo_scale

amuDev:
Implements what was discussed in #7532.

New option `window_logo_scale` takes int 0-100, scales the window logo image to x% of the window dimensions (smallest dimension sets scale).

Problem, using the current window_logo, you are able to position and clamp, but unable to enable scaling such that the logo does not take up too much space on a small terminal or be too small on a large terminal. Alternative solution would be to use background_image with cscale, yet if you want to have a logo anywhere other than the centre of the terminal, this quickly falls apart once you begin resizing the terminal outside the aspect ratio of the original image.
image

Adding the option to scale the window logo allows for the original goal of #7532 while being quite generic and overall a very minor code addition.
Here is an example of the proposed implementation to demonstrate the new option.
image

### DIFF
--- a/kitty/options/definition.py
+++ b/kitty/options/definition.py
@@ -1160,11 +1160,12 @@ faded and one being fully opaque.
 '''
     )
 
-opt('window_logo_scale', '0',
-    option_type='int', ctype='int',
+opt('window_logo_scale', '0.0 -1.0',
+    option_type='window_logo_scale', ctype='!window_logo_scale',
     long_text='''
-The percentage [1-100] of the window which the logo should scale to.
-0 to disable.
+The percentage [1-100] of the window which the logo should scale to. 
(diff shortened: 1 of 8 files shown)


kovidgoyal:
Looks OK but perhaps make it

window_logo width% [height%]

so that each can be specified independently. Having one number means %
of the smallest dimension.

A value of zero means no scaling in that dimension. A single zero (the
default) means no scaling at all.

And make it a float rather than an int so we can get higher precision.
Dont really gain anything by having it be an integer.

amuDev:
Implemented comments, not the cleanest way to implement the cases but should be quite clear the intent of each case at a glance. Was especially unsure on implementing the optional parameter for the height, decided -1 as a default would be sufficient for initial implementation but if there is a more clear way to handle this with the framework that I didn't catch point me to it!

PR #7553, opened 2024-06-17
Treat Fn as modifier key

pcc:
On some machines such as M series Macs the Fn key generates a keyboard event on its own. Add it to the list of modifiers so that, among other things, pressing it does not cause the scrollback buffer to scroll to the bottom.

### DIFF
--- a/kitty/keys.c
+++ b/kitty/keys.c
@@ -11,6 +11,10 @@
 #include "glfw-wrapper.h"
 #include <structmember.h>
 
+#ifndef __APPLE__
+#include <xkbcommon/xkbcommon.h>
+#endif
+
 // python KeyEvent object {{{
 typedef struct {
     PyObject_HEAD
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I am travelling now without access to my mac so cant really review this. Under Linux at least on ThinkPads it is mapped to XF86WakeUp not XF86Fn. And eyeballing your patch it is not mapping Fn to anything under cocoa at all. So it would have no effect. 

Are you running linux on a mac? And if so and your actual issue is pressing fn should not cause the scrollback to return then use the native_key value and check against both XF86Fn and XF86WakeUp constants. Create a function like is_no_action_key() which checks for both is_modifier or is one of the above and replace calls to is_modifier_key() in on_key_input() with calls to this function.

pcc:
Yes, I'm running Linux on a Mac where Fn is mapped to XF86Fn because it doesn't act as a wakeup key. Your suggestion makes sense, I had realized after sending my patch that we probably don't want to expose Fn to the extended keyboard protocol etc. I'll implement that.

pcc:
Done, PTAL

Issue #7425, opened 2024-05-12
GNOME 3 Wayland title bar is completely missing


sesse:
I'm running GNOME 3 with Wayland. After a recent change, kitty no longer draws a title bar; it is there, and you can see the outlines of it, but is completely transparent with no text or anything. This is what it looks like:

![Skjermbilde fra 2024-05-12 15-35-26](https://github.com/kovidgoyal/kitty/assets/238700/a31570ef-7aca-49d6-8dee-0b61d03194c1)

This started happening after this commit:
```
commit 235b8dc2e44e2c0f66f3bbedaff752704f2138ef
Author: Kovid Goyal <kovid@kovidgoyal.net>
Date:   Sat Apr 6 10:59:28 2024 +0530

    Assume all capabilities on compositors that dont support reporting of capabilities

 glfw/wl_init.c     | 5 +++--
 glfw/wl_platform.h | 1 +
 glfw/wl_window.c   | 8 +++++---
 3 files changed, 9 insertions(+), 5 deletions(-)
```

**Environment details**
```
kitty 0.34.1 created by Kovid Goyal
Linux kos 6.6.9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.6.9-1 (2024-01-10) x86_64
Debian GNU/Linux trixie/sid kos /dev/tty

Running under: Wayland (GNOME Shell 44.9) missing: preferred_scale
Frozen: False
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /bin/zsh
Loaded config files:
  /home/sesse/.config/kitty/kitty.conf

Config options different from defaults:
bold_font               Noto Sans Mono Bold
cursor_blink_interval   0.0
enable_audio_bell       False
font_family             Noto Mono
font_size               12.0
initial_window_height   (24, 'cells')
initial_window_width    (82, 'cells')
input_delay             0
placement_strategy      top-left
remember_window_size    False
repaint_delay           2
shell_integration       frozenset({'no-cursor'})
sync_to_monitor         False
touch_scroll_multiplier 5.0
wayland_enable_ime      False
Added shortcuts:
        shift+page_down →  scroll_page_down
        shift+page_up →  scroll_page_up
Colors:
        active_border_color     #c0bfbc   
        active_tab_background   #b0afac   
        active_tab_foreground   #504e55   
        background              #fcfcfc   
[...]

PR #7554, opened 2024-06-17
Fix Gnome 44 invisible title bar on Wayland

adamschmalhofer:
a.k.a. title bar is completely missing

Fixes: https://github.com/kovidgoyal/kitty/issues/7425


kovidgoyal:
I have merged this, but the patch makes no sense to me. Turning off all capabilities on a compositor that does support capability reporting is wrong since in such cases one should assume all capabilities exist. However, given that practically speaking this only matters for out of date GNOME, and you say it works for you and I am not bothering to install out of date gnome to test with, I have merged.

Does fullscreen via ctrl+shift+f11 still work after your change?

PR #7564, opened 2024-06-19
Fix config patching to only modify values provided by the theme

mattsacks:
Checks to see if a theme supplies a color before commenting out it's value in the patched config. This fixes an issue where a setting like `macos_titlebar_color background` may not be provided by a theme but switching themes would disable it.


kovidgoyal:
This is by design. There is no way for kitty to know if a particular
color setting clashes with a particular theme. If you want some color
settings to survive regardless of theme put them in a separate conf file
and include it, then the kitten wont overwrite them.

PR #7604, opened 2024-07-07
Update graphics-protocol.rst

Canop:
Add broot to the list of applications using Kitty's graphics protocol

IMO the list would be easier to read with an alphabetical sorting and by removing the gratuitous "blazing fast" assertions but that's out of scope of this PR.


kovidgoyal:
feel free to send a PR re-arranging the list into alphabetical order,
maybe separate libraries and applications into two lists. As for the one
liners for each project, I generally copy that from the project
description on github or leave it to the person sending the PR. You are
most welcome to use whatever adjectives you like to describe broot.

PR #7652, opened 2024-07-21
Add Window.mouse_position

novastrider:
This is to allow custom kittens to access the mouse position.
The use case for this for when a kitten is mapped to a mouse button, so that something useful can be done with the clicked mouse position.


kovidgoyal:
The ac decorator is for actions the user can map to key presses.

novastrider:
Thanks. Please note however your current Py_BuildValue call crashes on the `O`, because it will interpret the bool as a ptr.

kovidgoyal:
Right, sorry, fixed.

PR #7705, opened 2024-08-08
fix: add `hollow` option to `cursor_shape` config option

Axlefublr:
I prefer a hollow cursor, and would like for it to be allowed.


kovidgoyal:
It isnt allowed because hollow cursors are used to indicate lack of
input focus.

Issue #7201, opened 2024-03-06
Mouse pointer abnormally large under KDE Plasma 6


liamhays:
**Describe the bug**
I think this bug is similar if not identical to #4860. I upgraded my Arch Linux laptop to Plasma 6 this morning and the I-beam-shaped cursor that appears when I hover over the kitty window is abnormally large compared to other windows. I have Plasma global scaling set to 100%, though I had set it to 125% to test and then restored it to 100%. This issue persists after ending a Plasma session and starting it again.

**To Reproduce**
1. Open kitty
2. Hover over the window

**Screenshots**
Top image is kitty with mysterious large cursor, bottom image is Kate with normal sized cursor. Both windows are the exact same size.
![image](https://github.com/kovidgoyal/kitty/assets/5673976/437902de-f1fd-4264-b5c1-1e399242e3f4)
![image](https://github.com/kovidgoyal/kitty/assets/5673976/7884b644-f01d-49bd-b174-8eae6ee416ac)

**Environment details**
```
kitty 0.31.0 created by Kovid Goyal
Linux [hostname] 6.7.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 03 Mar 2024 00:30:36 +0000 x86_64
Arch Linux 6.7.8-arch1-1 (/dev/tty)

Running under: Wayland
Frozen: False
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /bin/bash
Loaded config files:
  /home/liam/.config/kitty/kitty.conf

Config options different from defaults:
enable_audio_bell       False
font_family             JetBrains Mono
touch_scroll_multiplier 4.0

Important environment variables seen by the kitty process:
	PATH                                /home/liam/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/var/lib/flatpak/exports/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/liam/.local/bin
	LANG                                en_US.UTF-8
	EDITOR                              nano
	SHELL                               /bin/bash
	DISPLAY                             :0
	WAYLAND_DISPLAY                     wayland-0
	USER                                liam
	XDG_CONFIG_DIRS                     /home/liam/.config/kdedefaults:/etc/xdg
	XDG_CURRENT_DESKTOP                 KDE
[...]

Issue #6914, opened 2023-12-13
Cursor size is inconsistent with the rest of the DE in Plasma 6/Wayland


MadCatX:
According to KDE developers, the cause is kitty's lack of support for `input-shape-v1` Wayland protocol. See https://bugs.kde.org/show_bug.cgi?id=478107 for the original report at KDE bugzilla.

kovidgoyal:
Yeah I might someday get around to supporting that protocol which IIRC
was released a few *months* ago. If KDE expects applications to support
a months old protocol they are insane. They need to set XCURSOR_SIZE
or use the desktop settings portal to make that size available to
applications.

kovidgoyal:
Not to mention that that protocol is both unstable and provides no way to know the cursor size for custom cursors and so is not a solution to the DE not providing cursor size information in backward compatible ways.

MadCatX:
Thanks for explaining the situation. Any hints where to look in case I'd like to give it a go myself? I assume that the actual protocol name is [cursor_shape_v1](https://hoyon.github.io/wayland-protocol-docs/protocols/cursor_shape_v1.html)?

kovidgoyal:
yes you would need to implement that in glfw/wl_*

kovidgoyal:
I had a few minutes so I implemented it, but it is currently disabled by an if (false) since no actual released wayland compositors support it so I cant test it. See line 726 in wl_init.c to enable. https://github.com/kovidgoyal/kitty/commit/cff490f881a8eba6636a32387a00d1c91c6ee4d4

MadCatX:
Just tested it with Plasma 6 and it seems to work fine. Thanks for doing this.

kovidgoyal:
Problem is it causes hyprland to crash so I can enable it.

MadCatX:
I understand. Any chance in can be hidden behind an env var or a setting? I'm okay with enabling it manually because it does improve my Kitty experience under Plasma 6 a lot.

kovidgoyal:
I suggest opening an issue with hyprland, if they fix it, I can just
enable it in kitty. This is rather too small a thing for a dedicated
setting/env var.

MadCatX:
Just FTR, I just tried the patched kitty with the latest Hyprland from git and it seems to work correctly for me.

Issue #4860, opened 2022-03-22
Mouse cursor large when on the kitty window under GNOME


hellobbn:
**Describe the bug**
I am using a system running GNOME 42.0 (Wayland). The mouse cursor becomes large when it is
moved to the kitty window.

The scaling factor of the screen is 1, while I do enable [Fractional Scaling](https://wiki.gnome.org/Initiatives/FracionalScaling) on the other screen with a scaling factor 1.25.

The kitty does print some error messages, see the logs below

**To Reproduce**
1. Run under GNOME 42.0
2. Open kitty
3. Observe the mouse cursor size

**Screenshots**
Mouse cursor on kitty:
![image](https://user-images.githubusercontent.com/10683168/159412383-3e8148bb-9016-427c-87fa-9b27b5e7ce74.png)

Mouse cursor on other applications:
![image](https://user-images.githubusercontent.com/10683168/159412431-b9aac877-8792-4d98-9ad7-3cffe38f4743.png)

**Environment details**
The logs from kitty `kitty --debug-gl`
```
[081 13:17:49.447111] Ignoring unknown config key: open_url_modifiers
[081 13:17:49.447208] Ignoring unknown config key: rectangle_select_modifiers
[081 13:17:49.447224] Ignoring unknown config key: terminal_select_modifiers
Created decoration buffers at scale: 1 vertical_height: 836 horizontal_width: 824
top-level configure event: size: 0x0 states: 
final window content size: 800x800
Setting window geometry in configure event: x=0 y=-24 800x824
GL version string: '4.6 (Core Profile) Mesa 21.3.7' Detected version: 4.6
top-level configure event: size: 800x824 states: TOPLEVEL_STATE_ACTIVATED 
final window content size: 800x800
Setting window geometry in configure event: x=0 y=-24 800x824
Resizing framebuffer to: 800x800 at scale: 1
prompt_marking: x=0 y=0 op='A'
prompt_marking: x=4 y=0 op='B'
[081 13:17:49.659322] [glfw error 65544]: on_gnome_cursor_theme_read: failed with error: [org.freedesktop.DBus.Error.ServiceUnknown] The name org.freedesktop.portal.Desktop was not provided by any .service files
[081 13:17:49.659337] [glfw error 65544]: on_gnome_cursor_size_read: failed with error: [org.freedesktop.DBus.Error.ServiceUnknown] The name org.freedesktop.portal.Desktop was not provided by any .service files
[...]

PR #7712, opened 2024-08-11
A small doc patch.

teika-kazura:
Hi. I created a small documentation patch. I can't build the doc, so it's not tested. If you like it, please use it. Feel free to modify.

Changes are:

1 Added links to the website (https://sw.kovidgoyal.net/kitty/) and the github repo in index.rst.
2 Mouse pointer config note.
3 Waring about "ssh" kitten; do not use it to forward X.
4 Glossary -> kitten: Added a link to kitten_intro.

Most points must be ok, but the rationale for the point 1 is: I always install the doc locally in my PC. In that case the kitty.png points the local home, not the upstream home.

Thanks a lot for developping kitty. Best wishes.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -540,3 +540,8 @@ The correct command will depend on your situation and hardware.
 system, which will wake the NVIDIA card during device enumeration.
 ``MESA_LOADER_DRIVER_OVERRIDE`` also assures that Mesa won't offer any NVIDIA
 card during enumeration, and will instead just use :file:`radeonsi_dri.so`.
+
+Mouse pointer configuration in X11
+----------------------------------
+To change the mouse pointer, run kitty with environment variables, like this:
+    XCURSOR_SIZE=<size> XCURSOR_THEME=<cursor theme> kitty <options>...
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -50,6 +50,7 @@ Glossary
       inside kitty windows and provide it with lots of powerful and flexible
       features such as viewing images, connecting conveniently to remote
       computers, transferring files, inputting unicode characters, etc.
+      See :doc:`the index of kittens</kittens_intro>` for more.
 
    easing function
       A function that controls how an animation progresses over time. kitty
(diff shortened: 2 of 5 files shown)


kovidgoyal:
1. This should be included only when building locally, its just clutter otherwise.
2. I'm afraid kitty docs are not the place to learn about configuring X11 The XCURSOR env vars are generic X11 config mechanisms supported by many programs
3. X11 forwarding works fine with the ssh kitten
4. This is ok

teika-kazura:
Thanks.

For the mouse cursor, my proposal is legit, since there are many ways to change the cursor for X apps - at least .Xdefaults - but none of them apply to kitty. It took me quite a lot to find it.

Furthermore the mouse pointer size is a faq: https://github.com/kovidgoyal/kitty/issues/4860 https://github.com/kovidgoyal/kitty/issues/6914 https://github.com/kovidgoyal/kitty/issues/7201 . If you are generous enough and don't mind further repeated questions, it's ok, but I doubt if it really will benefit you.

Anyway your decision has to be respected. Let me repeat my gratitude again. Best regards.

kovidgoyal:
Eh, on X11, there are myriad ways to configure mouse cursor size, kitty
currently supports the env var and gnome settings read via DBUS.
See https://wiki.archlinux.org/title/Cursor_themes

I am not particularly interested in documenting this mess. What would be
useful is a patch to read cursor size from xrdb as well and use that
when defined in preference to the env vars on X11.

kovidgoyal:
Actually, I looked at the code, and kitty on X11 uses XCreateFontCursor
which should automatically respect Xcursor.theme and Xcursor.size so
nothing to do. kitty follows standard X11 APIs on X11. On Wayland it
first gets the compositor to draw standard cursors for it, if the
compositor doesnt support that it fallsback to DBUS + env vars.

teika-kazura:
Sorry for making noise, and thank you again for taking extra effort of checking it.
Yes, today Xcursor.theme|size works for me, too

# Last time I tried these two didn't work. It is sure that it was the case for me; at that time all other apps' mouse pointer was changed except kitty, and I was frantic seeking a solution. Strange.

PR #7777, opened 2024-08-26
Security: Check UID of remote socket

fluffysatoshi:
This prevents remote access to Unix domain sockets when using an insecure umask or abstract sockets.

Socket permissions have no meaning for abstract sockets: the process umask(2) has no effect when binding an abstract socket.


fluffysatoshi:
Prevents access to the RC socket from different UIDs:

```
# sudo -u nobody kitten @ --to unix:@kitten send-text "Hello"

```

PR #7800, opened 2024-08-29
Set IN_NIX_SHELL=1 for tests

booxter:
This helps isolating system-wide zsh instance running in tests from the broader system shell initialization for limited nix environments.

This fixes failures as reported here:
https://github.com/NixOS/nixpkgs/issues/312692

### DIFF
--- a/kitty_tests/shell_integration.py
+++ b/kitty_tests/shell_integration.py
@@ -42,6 +42,7 @@ def basic_shell_env(home_dir):
         'BASH_SILENCE_DEPRECATION_WARNING': '1',
         'PYTHONDONTWRITEBYTECODE': '1',
         'WEZTERM_SHELL_SKIP_ALL': '1',  # dont fail if WezTerm's system wide, default on (why?) shell integration is installed
+        'IN_NIX_SHELL': '1',
     }
     for x in ('USER', 'LANG'):
         if os.environ.get(x):


kovidgoyal:
What exactly does IN_NIX_SHELL do and can you please add a comment explaining what it does in basic_shell_env()

kovidgoyal:
As far as I can tell from a quick google valid values for it are pure/impure and its effects are undefined, depending on individual pkgs. https://nix.dev/manual/nix/2.18/command-ref/nix-shell#env-IN_NIX_SHELL

booxter:
It very well may be that this is not a correct fix. Let me describe what I'm trying to achieve and why this helps.

The reason why test suite fails is that while `zsh` in the nix build environment comes from an injected package, it still reads from `/etc/zshenv`. From the man page for `zsh`:

```
STARTUP/SHUTDOWN FILES
       Commands are first read from /etc/zshenv; **this cannot  be
       overridden**.   Subsequent behaviour is modified by the RCS
       and GLOBAL_RCS options; the former  affects  all  startup
       files, while the second only affects global startup files
       (those  shown  here  with an path starting with a /).  If
       one of the options is unset at any point, any  subsequent
       startup  file(s)  of  the  corresponding type will not be
       read.  It is also possible for  a  file  in  $ZDOTDIR  to
       re-enable  GLOBAL_RCS. Both RCS and GLOBAL_RCS are set by
       default.
```

So tests adding `--noglobalrcs` is not enough to isolate in-test `zsh` from the system.

And here is what is in `/etc/zshenv` (managed by `nix-darwin`):

```
$ cat /etc/zshenv
# /etc/zshenv: DO NOT EDIT -- this file has been generated automatically.
# This file is read for all shells.

# Only execute this file once per shell.
# But don't clobber the environment of interactive non-login children!
if [ -n "$__ETC_ZSHENV_SOURCED" ]; then return; fi
export __ETC_ZSHENV_SOURCED=1

# Don't execute this file when running in a pure nix-shell.
if test -n "$IN_NIX_SHELL"; then return; fi

if [ -z "$__NIX_DARWIN_SET_ENVIRONMENT_DONE" ]; then
  . /nix/store/5kgqn44swwbivdxx1gsvpkz1qcvfhq16-set-environment
fi



# Read system-wide modifications.
if test -f /etc/zshenv.local; then
  source /etc/zshenv.local
fi
```

If the variable is **not** set, my full zsh environment is initialized, which triggers the following command (comes from `/nix/store/5kgqn44swwbivdxx1gsvpkz1qcvfhq16-set-environment`):

```
/nix/store/0mnp04vh94qi0cir0qhy29dn675hmpj5-gnupg-2.4.5/bin/gpg-connect-agent --quiet updatestartuptty /bye > /dev/null
```

...which produces some errors on the screen and make the tests fail.

Setting `IN_NIX_SHELL` allows me to work around the issue since then my env is not initialized.

booxter:
I'm looking at this a bit closer, and I wonder if perhaps the root of the issue is not in the test suite, or not even in zsh, but in the fact that `nix-darwin` configures its `zshenv` such that it may dump errors from `gpg-connect-agent` in some scenarios.

Here is where the command is injected into the env: https://github.com/LnL7/nix-darwin/blob/ac5694a0b855a981e81b4d9f14052e3ff46ca39e/modules/programs/gnupg.nix#L46

I wonder if it should be `gpg-connect-agent --quiet updatestartuptty /bye > /dev/null 2>&1` (note the additional `2>&1`) to suppress all errors.

Do you think I should do this instead?

Issue #7871, opened 2024-09-18
Allow `tab_bar_margin_color` to set vertical and horizontal margins separately


wincent:
**Is your feature request related to a problem? Please describe.**

Take a config like the following, which has the intent of drawing a border between the tab bar and the main content:

```
# This is the default:
tab_bar_margin_width 0.0

# This overrides the default (0.0 0.0):
tab_bar_margin_height 0.0 2.0

# This overrides the default (none):
tab_bar_margin_color #e0e0e0

# This overrides the default (center):
placement_strategy top
```

You will see the desired margin between the tab bar and the main content.

But, because kitty pads windows when the dimensions don't match integer multiples of the text cell size, you will either see the color on the left and right of the tab bar:

![Image](https://github.com/user-attachments/assets/f75bbfaa-dfe1-4539-9605-93e76f8b2d38)

Or no color on the left and right:

![Image](https://github.com/user-attachments/assets/c12bbb19-120b-4e80-8864-88e50ce1241d)

**Describe the solution you'd like**

The left and right margins popping in and out depending on font size is distracting, so it would be nice to be able to set the colors of vertical margins independently from the horizontal ones.

ie. something like this (using the semantics of [CSS shorthand properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) to allow passing 1, 2, 3, or 4 values):

```
# 1. Sets the color for all margins, unconditionally:
tab_bar_margin_color #color1

# 2. Sets first color for top/bottom margins and second color for left/right margins:
tab_bar_margin_color #color1 #color2

# 3. Special case of the above — using the keyword "none" to indicate
# "use color that matches neighboring tab background":
tab_bar_margin_color #color1 none

# 4. Separate colors for top, left/right, and bottom margins:
tab_bar_margin_color #color1 #color2 #color3

# 5. Sets separate colors for top, right, bottom, and left margins:
tab_bar_margin_color #color1 #color2 #color3 #color4
```

Now, I don't know if you want to incur the complexity of implementing all those variants. The only one I actually care about here is "3", but I listing them all out anyway in case you wanted to get ahead of potential future feature requests.

**Describe alternatives you've considered**

I looked at:

- https://github.com/kovidgoyal/kitty/issues/5719

which was asking for the same thing:

> Add a second argument to the `tab_bar_margin_color` option, allowing users to set a separate left/right margin color

But it was noted there:

[...]

PR #7876, opened 2024-09-19
feat: don't draw visible tab bar margins if width is 0

wincent:
On thinking more about the related issue (#7871) I realized that the desired result can be obtained much more simply by suppressing the display of left/right margins whenever the user has set `tab_bar_margin_width` to `0.0` (which is the default).

### Before: Tab bar with three different font sizes

Left and right margins pop in and out depending on blank space:

before large

before medium

before small

### After: Tab bar with three different font sizes

Left and right margins are not shown:

after large

after medium

after small

Closes: https://github.com/kovidgoyal/kitty/issues/7871

### DIFF
--- a/kitty/tab_bar.py
+++ b/kitty/tab_bar.py
@@ -599,7 +599,7 @@ class TabBar:
                     blank_rects.append(Border(0, tab_bar.bottom + 1, vw, central.top, bg))
         g = self.window_geometry
         left_bg = right_bg = bg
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Oh an incidentally, I was a big user of your command-t years ago, in fact, I added a command-t quick file opener to the calibre ebook editor inspired by it. So thanks :)

PR #7938, opened 2024-10-04
nix: fix shell config for darwin

booxter:
With the change, I was able to build the binary with `make` and successfully execute the tests with `./test.py` on darwin aarch64.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -13,6 +13,9 @@ in
           ncurses
           lcms2
           xxhash
+          xxHash
+          simde
+          go_1_23
         ]
         ++ optionals stdenv.isDarwin [
           Cocoa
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Why is there both xxhash and xxHash in buildInputs?

booxter:
@kovidgoyal One is python package; another is a C library.

kovidgoyal:
But kitty doesn't use a python xxhash package as far as I know, so what's that for?

booxter:
@kovidgoyal If that is the case, then we should drop `xxhash` from the list. (I've just checked and indeed it doesn't seem to affect the build or tests.)

This is not strictly speaking related to this PR, but I will include it in the next push.

kovidgoyal:
Yeah kitty uses only a C and a Go xxhash, no Python.

kovidgoyal:
Merged, however it would be nice if there was some way to read the required go version from go.mod otherwise the next time that is updated shell.nix will break again.

Issue #2460, opened 2020-03-22
Feature suggestion: Animated cursor


Asheq:
https://github.com/Kethku/neovide/blob/master/assets/AnimatedCursor.gif

The above gif is from a neovim front-end and demonstrates an animated cursor. Using it is actually quite a joy, and useful because it's easier for you and onlookers to track the cursor position.

@kovidgoyal Just wondering if you would be willing to consider something like this for kitty (as an option that can be disabled or enabled, of course). Thanks!

ctrlcctrlv:
Wow, cool. I support this idea. I'm already thinking what kind of clever shader can do this.

Luflosi:
Looks cool but I have no idea how this might be implemented, this is probably not that simple.

kovidgoyal:
Seems somewhat headache inducing to me, but am fine with an option for it that defaults to off. To implement it you would basically need to use the timers facility to implement the animation, similar to how it is done for blinking cursors. 

If someone else wants to work on it feel free to post if you need help and I will provide assistance with getting started.

egmontkob:
Let's note that in terminals, due to the very nature of the architecture, this can easily result in faulty jumps, i.e. jumps to intermediate positions where the contents are updated but the cursor is not meant to rest.

This can be mitigated for cooperating apps that either hide the cursor while repainting, or use the protocol extension for atomic updates (I can't recall if Kitty already supports it, or we just discussed it, sorry).

Still, the terminal needs to figure out what to do if a cursor animation is in progress and in the mean time the cursor moves to a new location, e.g. should it complete the pending animation first, or should it change direction immediately, or maybe should it consider the pending animation an unintended one, cancel it, and restart a new animation from the previous starting point (maybe if this new animation is not more than x% complete). It's heuristics, and no matter what do you, there will be cases when the user sees undesired glitches.

Also there are many more kinds of apps to consider, not just text editors; e.g. what would the cursor do while editing a multi-line commandline, cat'ing a large file, or using mc, less, top, alsamixer, tmux with split panes etc...

Would be interesting to see a prototype, see all these in live action ;) Without that, I can't really foresee if this feature would be more on the "super useful, with very few glitches" side or on the "quite useless because of way too many glitches" side. If the latter, you might consider introducing an escape sequence, so that it can be enabled for specific apps only.

kovidgoyal:
On Mon, Mar 23, 2020 at 02:29:08AM -0700, egmontkob wrote:
[...]

PR #7970, opened 2024-10-13
Add Cursor Trail Feature to Enhance Cursor Visibility

jinhwanlazy:
This pull request introduces a new feature that adds a trailing effect behind the cursor as it moves within the terminal window. The cursor trail enhances cursor visibility, making it easier to track cursor movement, especially during rapid navigation or in complex outputs. This feature is particularly beneficial when demonstrating tasks in the terminal, as it allows viewers to easily follow the cursor's movements.

related issue: https://github.com/kovidgoyal/kitty/issues/2460

video - cursor trail moves smoothly between kitty windows, tmux panes and vim splits.

https://github.com/user-attachments/assets/65a07571-fe7c-491b-9b29-d4640696f60b

### Key Changes

- **New Configuration Option: `enable_cursor_trail`**
  - **Type:** Boolean (`yes` or `no`)
  - **Default:** `no`
  - **Description:** Enables or disables the cursor trail effect. When set to `yes`, a trailing effect is rendered behind the cursor as it moves, creating a motion trail.

- **New Configuration Option: `cursor_trail_decay`**
  - **Type:** Tuple of two positive floats
  - **Default:** `0.1 0.3`
  - **Description:** Controls the decay times for the cursor trail effect when `enable_cursor_trail` is set to `yes`. This option accepts two positive float values specifying the fastest and slowest decay times in seconds. The first value corresponds to the fastest decay time (minimum), and the second value corresponds to the slowest decay time (maximum). The second value must be equal to or greater than the first value. Smaller values result in a faster decay of the cursor trail. Adjust these values to control how quickly the cursor trail fades away.

### Implementation Details

- **Cursor Trail Concept**
  - Instead of animating the cursor itself, this feature introduces the concept of a cursor trail. The trail constantly follows the current active cursor position, creating a visual path that enhances the user's ability to track cursor movement.
  - The existing cursor rendering logic remains mostly untouched, ensuring minimal impact on the current behavior.

- **CursorTrail Data Structure**
  - The `CursorTrail` instance resides within the `Tab` structure, while the cursor itself is part of the `Screen`. This design choice allows the trail to represent cursor movements across different windows within a tab.

- **Rendering Logic**
  - Added `cursor_trail.c` to handle the updating and rendering of the cursor trail.
  - Modified `child-monitor.c` and `shader.c` to integrate the cursor trail update and rendering into the existing rendering pipeline.

- **Trail Updating Rule**
  - Implemented exponential ease-out animation for the trail movement towards the cursor, creating a smooth trailing effect without altering the cursor's own animation settings.
  - The trail only follows cursors that have stayed in their position longer than `input_delay`, which means it ignores fast-moving cursors. This prevents trails from appearing for cursors that rapidly change positions, such as those used for drawing UI elements in certain applications.
  - In most applications, this behavior works as intended, enhancing visibility without unnecessary trails.
  - **Note:** In Neovim, during sequential `n` commands, the trail may jump between the search bar and the search result. This appears to be an issue with Neovim's cursor handling, as running Neovim with `nvim -u NONE` does not exhibit this behavior.
  
video - bug in nvim(?)

https://github.com/user-attachments/assets/a39818f4-d285-4c01-85b9-c5d9cdc143fc

- **Visual Artifacts**
  - **Cursor and Trail Visibility:** Both the cursor and the trail are visible simultaneously. This is a key difference from Neovide's implementation, where the cursor itself is morphing.
  - **Underlying Text Visibility:** Fonts under the moving trail are not visible while the trail is present.


kovidgoyal:
Fix the CI failures please. WIll review in detail when I have a moment. You can send some trivial PR that I will merge so that this PR does not need me to approve running CI for every change.

kovidgoyal:
Regarding the nvim issue, IIRC nvim uses synchronized updates. I havent looked at your code in detail to se eif it already does this, but it may be possible to work around it by only triggerring the trail based on cursor movements after the synchronized update is over (this is exposed in kitty as screen->paused_rendering

jinhwanlazy:
Thanks for the insight. I wasn't aware of the details surrounding synchronized updates and terminal modes, so this has been a great learning opportunity for me!

I tried monitoring the pause state and cursor visibility inside the update_cursor_trail function, like this:

```    
    log_error("DECTCEM: %d", (int)WD.screen->modes.mDECTCEM);
    log_error("Pause_expires_at: %u", (unsigned int)WD.screen->paused_rendering.expires_at);
```

Here’s what I observed:
1. During a regular search with `/`, for every new character input:
```
[1186.136] DECTCEM: 0
[1186.136] Pause_expires_at: 0
```

2. During sequential search with `n` (the condition causing the issue):
```
[1400.140] DECTCEM: 1
[1400.140] Pause_expires_at: 0
```

It seems that Neovim doesn't trigger paused_rendering during searches, instead, it's controlling the visibility of the cursor. I considered using the DECTCEM state to decide whether to follow the cursor, as it seemed logically sensible, though it doesn't fully address the issue mentioned above. However, I encountered another problem: DECTCEM remains off while in Neovim's terminal mode. Since terminal mode is used not only for regular terminal usage but also by many plugins that implement floating windows inside Neovim, I can't rely on this state, as I want the trail to work in terminal mode as well.

3. neovim terminal mode
```
[1518.859] DECTCEM: 0
[1518.859] Pause_expires_at: 0
```

jinhwanlazy:
FYI, the issue does not occur in VIM 9.1

kovidgoyal:
On Tue, Oct 15, 2024 at 08:15:01AM -0700, Jinhwan Choi wrote:
> I still believe this is a Neovim issue. I found that preventing the `filetype.lua` plugin (one of the default plugins) from loading resolves the problem. It is slowing down searching ui to update somehow. It's strange because this is a filetype plugin, and I wouldn't expect filetypes to have anything to do with searching.

Report it to neovim then, you can get a trace of the commands neovim is
sending kitty with kitty --dump-commands that should be enough to
demonstrate the issue without your PR on released kitty. Look for
move_cursor, hide_cursor etc in the log. neovim is presumably redrawing
the search bar/status bar when filetype.lua is loaded and doing a
search, causing the cursor jump. Maybe they will be willing to fix that.

In general though, I think you should take synchronized updates into
account. That is cursor movement command received while paused_rendering
is enabled should not trigger trails. This is consistent with the
semantics of synchronized updates which is that nothing on screen should
change while rendering is paused for an update.

jinhwanlazy:
> In general though, I think you should take synchronized updates into
account. That is cursor movement command received while paused_rendering
is enabled should not trigger trails. This is consistent with the
semantics of synchronized updates which is that nothing on screen should
change while rendering is paused for an update.

The commit I just pushed will do that. Can you suggest an app or a situation that triggers paused_rendering for an example? So that I can see it actually affecting the behavior or the trail.

> Report it to neovim then,

I'll try that when I have a moment.

kovidgoyal:
On Tue, Oct 15, 2024 at 04:08:48PM -0700, Jinhwan Choi wrote:
> > In general though, I think you should take synchronized updates into
> account. That is cursor movement command received while paused_rendering
> is enabled should not trigger trails. This is consistent with the
> semantics of synchronized updates which is that nothing on screen should
> change while rendering is paused for an update.
>
> The commit I just pushed will do that. Can you suggest an app or a situation that triggers paused_rendering for an example? So that I can see it actually affecting the behavior or the trail.


The various kittens like diff, themes, choose_fonts all use it. I think
some of the more modern terminal text editors also use it but I am not
sure.

kovidgoyal:
I have merged, so that your work can form the basis of further changes.

I have changed the cursor_trail option to allow specifying the wait time instead of using input_delay.

Some notes, please fix these in a new PR

When cursor=none (i.e. the cursor uses a reverse video effect the
trail color seems to be black). It should be the current cursor color or
perhaps the foreground color instead.

Move the calculation of the trail color into shader.c and pass a
reference to the currently active screen, if any to draw_trail. Use that
to calculate the color, see the pick_cursor_color() function or
alternately store the last rendered cursor color and re-use it.

As far as I can tell, the trail is drawn even when the cursor is hidden
by the application. Is this intentional?

Regarding paused_rendering. Currently as far as I can see, when
rendering is paused the cursor trail is not updated. This is correct,
but also the last cursor move time that you stored on the cursor object
should possibly be updated when rendering is unpaused. Essentially the
semantic should be as if all commands that happen during a paused
rendering all happen at the same instant, when the rendering is
unpaused. You can probably implement this by setting a flag on the
paused rendering object indicating cursor position was updated and if
that flag is set at unpause update the time.

jinhwanlazy:
Thanks for merging and the great feedback! I'll dive into these points and work on addressing them in a new PR soon.

```
As far as I can tell, the trail is drawn even when the cursor is hidden
by the application. Is this intentional?
```
When cursor is hidden. The intended behavior of the trail is it still be visible until it reaches the position where the cursor  was visible last, then vanishes. I think it feels more natural than it suddenly disappear in the middle of its transition. 

```
Regarding paused_rendering. Currently as far as I can see, when
rendering is paused the cursor trail is not updated. This is correct,
but also the last cursor move time that you stored on the cursor object
should possibly be updated when rendering is unpaused. Essentially the
semantic should be as if all commands that happen during a paused
rendering all happen at the same instant, when the rendering is
unpaused. You can probably implement this by setting a flag on the
paused rendering object indicating cursor position was updated and if
that flag is set at unpause update the time.
```
Same as above, the idea is that when entering paused_rendering, the trail moves until it reaches the known position of the cursor, stay there, and resume following once unpaused. So it shouldn't need to track cursor moves during the pause.

Let me know if you want me to rethink these behaviors!

kovidgoyal:
On Fri, Oct 18, 2024 at 11:38:28PM -0700, Jinhwan Choi wrote:
> Thanks for merging and the great feedback! I'll dive into these points and work on addressing them in a new PR soon.
>
> ```
> As far as I can tell, the trail is drawn even when the cursor is hidden
> by the application. Is this intentional?
> ```
> When cursor is hidden. The intended behavior of the trail is it still be visible until it reaches the position where the cursor  was visible last, then vanishes. I think it feels more natural than it suddenly disappear in the middle of its transition.

My concern is that trails should not start due to movements that happen
while the cursor is hidden. This is because many applications hide the
cursor then update their UI then reshow the cursor. This might fix your
nvim search bounce issue as well. It should be possible to implement
this by either not changing Cursor->position_changed_by_client_at when
the cursor is hidden or more correctly not starting a trail when the
cursor is hidden.

>
> ```
> Regarding paused_rendering. Currently as far as I can see, when
> rendering is paused the cursor trail is not updated. This is correct,
> but also the last cursor move time that you stored on the cursor object
> should possibly be updated when rendering is unpaused. Essentially the
> semantic should be as if all commands that happen during a paused
> rendering all happen at the same instant, when the rendering is
> unpaused. You can probably implement this by setting a flag on the
> paused rendering object indicating cursor position was updated and if
> that flag is set at unpause update the time.
> ```
> Same as above, the idea is that when entering paused_rendering, the trail moves until it reaches the known position of the cursor, stay there, and resume following once unpaused. So it shouldn't need to track cursor moves during the pause.

Similarly, I want to basically ensure no trail is started while in
rendering paused mode and that for cursor movements that happen in that
mode the position_changed_by_client_at field is the time at which
rendering was resumed. This is the "correct" value for this field and
ensures that any code conditional on the value of this field behaves as
expected. If you want a pre-existing trail to continue being rendered
upto the position of the cursor when rendering was paused, then that
should be handled separately.

mikesmithgh:
https://github.com/user-attachments/assets/7290ab00-e988-4cd5-8567-5e7302c241aa

awesome work! just a video of me playing around 😂

```
cursor_trail 3000
cursor_trail_decay 1 2
```

BingCoke:
Thank you for creating the animation! It's really stunning, but in my view, there's a minor shortcoming that's quite bothersome. When the beam cursor moves over a very short distance, it can cause a trailing shadow, and even feel like it's vibrating, which is a bit dizzying to the eyes. For example, when I'm typing in insert mode in nvim and deleting text (this issue is particularly noticeable at such times).

https://github.com/user-attachments/assets/42a92b26-698e-4007-9969-764d734e6bc4

kovidgoyal:
Yeah there should probably be a minimum distance threshold say if the cursor has moved less than 2-3 cells in the same window don't start the trail. It should not affect normal typing then.

benfrain:
Only just started using it but so far also really appreciate this feature 👍. Agree that the minimum 2-3 cells sounds like a sensible caveat.

jinhwanlazy:
> Thank you for creating the animation! It's really stunning, but in my view, there's a minor shortcoming that's quite bothersome. When the beam cursor moves over a very short distance, it can cause a trailing shadow, and even feel like it's vibrating, which is a bit dizzying to the eyes. For example, when I'm typing in insert mode in nvim and deleting text (this issue is particularly noticeable at such times).

> This feature is so cool. However, when input is entered, the cursor flashes consistently. I don't know if I can cancel this effect when input is entered in the future.

> Yeah there should probably be a minimum distance threshold say if the cursor has moved less than 2-3 cells in the same window don't start the trail. It should not affect normal typing then.

Thank you for sharing your experiences and suggestions. I see the issues you're pointing out. 

Unfortunately, implementing a minimum distance threshold isn't trivial in the current implementation, as the trail is simply a dumb follower of the cursor. One way to implement this would be to make the trail jump to the cursor when the distance between them is exactly an integer multiple of the cell grid size. However, this could introduce a rare (roughly one in a million) glitch where the trail suddenly disappears while in normal jump.

A simpler workaround could be adding an option to disable the cursor trail for beam-shaped cursors. This should be easy to implement and might address the cases you've shared.

Another thing I've noticed is that the trail is significantly shorter when moving horizontally compared to vertical or diagonal movement. This effect is most significant with beam shaped cursor. This makes it look less smooth when typing text. It's known since the beginning of the implementation. But Initially, I thought it was a good idea to keep the cursor trail beam-shaped while inserting, but perhaps I should reconsider this approach.

I'll try implementing these ideas over the coming weekend. Stay tuned for updates.

kovidgoyal:
On Tue, Oct 22, 2024 at 05:31:09AM -0700, Jinhwan Choi wrote:
> > Thank you for creating the animation! It's really stunning, but in my view, there's a minor shortcoming that's quite bothersome. When the beam cursor moves over a very short distance, it can cause a trailing shadow, and even feel like it's vibrating, which is a bit dizzying to the eyes. For example, when I'm typing in insert mode in nvim and deleting text (this issue is particularly noticeable at such times).
>
> > This feature is so cool. However, when input is entered, the cursor flashes consistently. I don't know if I can cancel this effect when input is entered in the future.
>
> > Yeah there should probably be a minimum distance threshold say if the cursor has moved less than 2-3 cells in the same window don't start the trail. It should not affect normal typing then.
>
> Thank you for sharing your experiences and suggestions. I see the issues you're pointing out.
>
> Unfortunately, implementing a minimum distance threshold isn't trivial in the current implementation, as the trail is simply a dumb follower of the cursor.

When starting a trail if one doesn't already exist cant one check if the
distance to cursor is small? Basically it needs one single additional
boolean flag on the cursor trail object to track if a trail is already
started or not. Or maybe I am missing something.

--

Dr. Kovid Goyal
https://www.kovidgoyal.net
https://calibre-ebook.com

jinhwanlazy:
> When starting a trail if one doesn't already exist cant one check if the
distance to cursor is small? Basically it needs one single additional
boolean flag on the cursor trail object to track if a trail is already
started or not. Or maybe I am missing something.

Yes, you are right. Using a flag would make it easier. We don't even need to add a new one. We can use `needs_render` 
 flag which is already there in `CursorTrail`. Maybe I was too focused on my initial idea that keeping trail updating rules stateless would be better.

kovidgoyal:
On Tue, Oct 22, 2024 at 05:48:46PM -0700, Jinhwan Choi wrote:
> > When starting a trail if one doesn't already exist cant one check if the
> distance to cursor is small? Basically it needs one single additional
> boolean flag on the cursor trail object to track if a trail is already
> started or not. Or maybe I am missing something.
>
> Yes, you are right. Using a flag would make it easier. We don't even need to add a new one. We can use `needs_render`
>  flag which is already there in `CursorTrail`. Maybe I was too focused on my initial idea that keeping trail updating rules stateless would be better.

It needs to track some state, namely its position and last update time
already so one more bit of state will hopefully not be too bad. Or as
you say, needs_render is enough I havent checked it semantics
thoroughly.

--

Dr. Kovid Goyal
https://www.kovidgoyal.net
https://calibre-ebook.com

krbylit:
Loving this feature. My one problem is that I’m using yazi for file management and the trail is not ideal there.

I’ll usually open yazi with a key bind I made, and have half way solved my problem with this (typically use zsh but fish opens a lot faster even compared to when I tried zsh sourced with a stripped down rc file specific for yazi):

map kitty_mod+y launch —allow-remote-control —type=tab —cwd=current fish -ic “source ~/.config/fish/config.fish; kitten @ load-config —override cursor_trail=0; yazi”

However, overriding the kitty config persists after quitting yazi, so I don’t have cursor trail anymore.

Anyone know if there’s a solution to reset that option or load my normal kitty.conf upon quitting yazi?

kovidgoyal:
map kitty_mod+y launch —allow-remote-control —type=tab —cwd=current fish -ic “source ~/.config/fish/config.fish; kitten @ load-config —-override cursor_trail=0; yazi; kitten @ load-config —-override cursor_trail=3”

krbylit:
> map kitty_mod+y launch —allow-remote-control —type=tab —cwd=current fish -ic “source ~/.config/fish/config.fish; kitten @ load-config —-override cursor_trail=0; yazi; kitten @ load-config —-override cursor_trail=3”
> 

Ohh wow okay cool. Thank you!

jinhwanlazy:
Wow, yazi is amazing. I didn't know it exist.

https://github.com/kovidgoyal/kitty/pull/7985#issuecomment-2424716409
> One potential compromise could be to add some opacity to the trail, so it slowly fades when moving to a window where the cursor is hidden. But I'm not sure if this extra complexity is worth implementing, as I'm pretty happy with how it works now.

This will fix the problem with yazi. I'll make a PR soon. maybe tomorrow.

jinhwanlazy:
@BingCoke @chaozwn @benfrain @krbylit 
New changes are merged. Please try it. 🙏
1. Added a new option `cursor_trail_distance_threshold`. The cursor trail will not be initiated when the cursor jump distance is below this threshold.
2. Trail animation will not be visible within a window where the cursor is hidden (e.g. htop, **yazi**)

chaozwn:
> @BingCoke @chaozwn @benfrain @krbylit New changes are merged. Please try it. 🙏
> 
> 1. Added a new option `cursor_trail_distance_threshold`. The cursor trail will not be initiated when the cursor jump distance is below this threshold.
> 2. Trail animation will not be visible within a window where the cursor is hidden (e.g. htop, **yazi**)

thank you so much. i will try it.

benfrain:
> @BingCoke @chaozwn @benfrain @krbylit New changes are merged. Please try it. 🙏
> 
> 1. Added a new option `cursor_trail_distance_threshold`. The cursor trail will not be initiated when the cursor jump distance is below this threshold.
> 2. Trail animation will not be visible within a window where the cursor is hidden (e.g. htop, **yazi**)

Above it says that the new config option is `enable_cursor_trail yes` but that doesn't work on the nightly I just pulled down, I still need to use `cursor_trail 3`

jinhwanlazy:
Sorry for confusion. The option actually changed after we merged this PR. 

> I have changed the cursor_trail option to allow specifying the wait time instead of using input_delay.

I fixed the original PR description.

BingCoke:
> @BingCoke @chaozwn @benfrain @krbylit New changes are merged. Please try it. 🙏
> 
> 1. Added a new option `cursor_trail_distance_threshold`. The cursor trail will not be initiated when the cursor jump distance is below this threshold.
> 2. Trail animation will not be visible within a window where the cursor is hidden (e.g. htop, **yazi**)

aaaaawesome!!! Thank you for your contribution to the animation, the current is incredibly smooth!

leiserfg:
@jinhwanlazy did you file the issue for neovim?

WilliamAnimate:
resizing the terminal seems to trigger the trail

https://github.com/user-attachments/assets/3ca7610c-2036-46e7-a4b6-6d87e6bb30d2

kovidgoyal:
That is because the shell redraws the prompt on resize which involves cursor movement.

jinhwanlazy:
> @jinhwanlazy did you file the issue for neovim?

@leiserfg Not yet. I tried `--dump-commands` as kovidgoyal suggested, but I got identical output with and without loading filetype.lua, even though the trail's behavior is different. I'm planning to profile commands with exact timestamps and read some Neovim code over the coming weekend.

kuro337:
For neovim I see an issue where if the cursor isn't on the first column and I navigate up/down - for some reason the updates are significantly slower




https://github.com/user-attachments/assets/f96ed78f-915e-4f09-977a-d0886116413c

jinhwanlazy:
> For neovim I see an issue where if the cursor isn't on the first column and I navigate up/down - for some reason the updates are significantly slower
> 
>  kitty_cursor_s.mov

Looks like the trail animation is blocking key input from what I see in the video, but I have no idea why that's happening. 

Can you give me more details about your setup so I can try to reproduce this? Machine, OS, desktop environment, and your full kitty config would be helpful.

kuro337:
Seeing the same thing with zero config nvim and normal nvim, don't think it's related to any config/plugins - 

```bash
nvim --version
NVIM v0.11.0-dev-996+g395f420fc
Build type: Release
LuaJIT 2.1.1727870382

# nvim build command  
git clone https://github.com/neovim/neovim.git && cd neovim
cmake -S cmake.deps -B .deps -G Ninja -D CMAKE_BUILD_TYPE=Release -D ENABLE_WASMTIME=ON 
cmake --build .deps
cmake -B build -G Ninja -D CMAKE_BUILD_TYPE=Release -D ENABLE_WASMTIME=ON
cmake --build build
sudo cmake --install build



kitty --version
kitty 0.37.0 created by Kovid Goyal
```

kitty.conf

```bash
# vim:fileencoding=utf-8:foldmethod=marker
include kitty-theme.conf
listen_on unix:/tmp/mykitty
allow_remote_control yes
editor nvim

#: Performance tuning {{{
repaint_delay 0
input_delay 0
sync_to_monitor no
macos_colorspace displayp3
text_composition_strategy 1.2 20
#: }}}

# shell_integration enabled
shell_integration no-cursor
macos_quit_when_last_window_closed no
macos_show_window_title_in none
macos_option_as_alt yes
cursor_blink_interval 0
initial_window_width 170c
initial_window_height 64c
enable_audio_bell no
strip_trailing_spaces smart
open_url_with default
update_check_interval 24
env LANG=en_US.UTF-8
env XDG_DATA_DIRS=/usr/local/share:/usr/share:/Users/kuro/.nix-profile/share:/nix/var/nix/profiles/default/share
env EDITOR=nvim
env VISUAL=nvim
env LC_CTYPE=UTF-8
confirm_os_window_close 1
map alt+r load_config_file
map alt+h toggle_layout stack
map opt+cmd+, debug_config
mouse_map left click ungrabbed mouse_handle_click prompt
mouse_map ctrl+left click ungrabbed mouse_handle_click link
map cmd+shift+t launch --cwd=current 
map cmd+shift+enter launch --cwd=current
map cmd+shift+w close_window
map cmd+shift+k launch --cwd=current --type=os-window
map cmd+] next_window
map cmd+shift+. move_tab_forward
map cmd+shift+, move_tab_backward
map ctrl+f   toggle_fullscreen
map cmd+shift+equal change_font_size current +1
map cmd+shift+minus change_font_size current -1
map cmd+shift+0 change_font_size current 0
map ctrl+k scroll_line_up
map shift+ctrl+k scroll_page_up
map ctrl+j scroll_line_down
map shift+ctrl+j scroll_page_down
map cmd+s send_text all \x1b:write\ \x0d
map cmd+shift+c copy_to_clipboard
map cmd+c send_text all \x1b[57376u
map cmd+1 next_window
map cmd+p show_first_command_output_on_screen
mouse_hide_wait -1.0 
font_size 18.0
draw_minimal_borders yes
# inactive_text_alpha 0.8
tab_bar_edge bottom
window_border_width 1px
window_margin_width 0
window_padding_width 7
active_tab_background #0f4b6e
inactive_tab_background #1e1f2e
active_tab_foreground   #e5e7eb
inactive_tab_foreground #7f8694
tab_bar_edge bottom
tab_bar_align left
tab_bar_style custom
tab_bar_min_tabs 1
tab_activity_symbol none
bell_on_tab no
tab_separator ""
tab_bar_margin_width 0.0
tab_bar_margin_height 0.0 0.0
tab_title_template "{f'{title[:30]}…' if title.rindex(title[-1]) + 1 > 30 else (title.center(6) if (title.rindex(title[-1]) + 1) % 2 == 0 else title.center(5))}"
active_tab_font_style italic
active_border_color none
inactive_border_color #717171
color0 #1E1E2E 
color15 #FFFFFF 
color16 #FFFF00  

# https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points
# Seti-UI + Custom
symbol_map U+e5fa-U+e6b5                                Symbols Nerd Font
# Devicons
symbol_map U+e700-U+e7c5                                Symbols Nerd Font
# Font Awesome
symbol_map U+ed00-U+f2ff                                Symbols Nerd Font
# Font Awesome Extension
symbol_map U+e200-U+e2a9                                Symbols Nerd Font
# Material Design Icons
symbol_map U+f0001-U+f1af0                              Symbols Nerd Font
# Weather
symbol_map U+e300-U+e3e3                                Symbols Nerd Font
# Octicons
symbol_map U+f400-U+f533,U+2665,U+26A1                  Symbols Nerd Font
# [Powerline Symbols]
symbol_map U+e0a0-U+e0a2,U+e0b0-U+e0b3                  Symbols Nerd Font
# Powerline Extra Symbols
symbol_map U+e0a3,U+e0b4-U+e0c8,U+e0ca,U+e0cc-U+e0d7    Symbols Nerd Font
# IEC Power Symbols
symbol_map U+23fb-U+23fe,U+2b58                         Symbols Nerd Font
# Font Logos (Formerly Font Linux)
symbol_map U+f300-U+f375                                Symbols Nerd Font
# Pomicons
symbol_map U+e000-U+e00a                                Symbols Nerd Font
# Codicons
symbol_map U+ea60-U+ec1e                                Symbols Nerd Font
# Additional sets
symbol_map U+276c-U+2771,U+2500-U+259f                 Symbols Nerd Font

#: }}}
#
# symbol_map U+f300-U+f313,U+e000-U+e00a,U+ea60-U+ebeb,U+e0a0-U+e0c8,U+e0ca,U+e0cc-U+e0d4,U+e200-U+e2a9,U+e300-U+e3e3,U+e5fa-U+e6b1,U+e700-U+e7c5,U+f000-U+f2e0,U+f300-U+f372,U+f400-U+f532,U+f0001-U+f1af0 Symbols Nerd Font Mono
#
#font_family family='SF Pro' 
font_family family='JetBrains Mono' postscript_name=JetBrainsMono-Regular
bold_font auto
italic_font auto
bold_italic_font auto
font_features JetBrainsMono-Regular +calt +zero +cv07
font_features JetBrainsMono-Italic +calt +zero 

cursor_trail 3
cursor_trail_decay 0.2 0.3
cursor_trail_start_threshold 3
```
arm64 m3 max

```bash
      System Version: macOS 15.2 (24C5073e)
      Kernel Version: Darwin 24.2.0
      Boot Volume: Macintosh HD
      Boot Mode: Normal
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 41 minutes, 33 seconds
```

jinhwanlazy:
@kuro337 
I was able to reproduce the issue on my M2 MacBook. Here's the minimal config file that causes the problem:

```
repaint_delay 0
sync_to_monitor no
cursor_trail 3
```

Simply setting `repaint_delay 1` will fix it for the memoment. 
Make sure you completely close the Kitty app and reopen it, rather than applying the config with a shortcut.

TRPB:
thanks so much for this! 

Is there any chance of adding an option to hide the cursor during the animation?

ie.

- Cursor is hidden
- Animation starts
- Animation ends
- Cursor is shown

That way the cursor would animate into position rather than instantly move but have a trail. Since kitty supports hiding the cursor for the blink animation I assume this is possible and fairly trivial to implement?

jinhwanlazy:
@TRPB Thanks for the suggestion! I understand that you're looking for a seamless experience similar to what Neovide provides, where the cursor itself appears to morph into position. That’s the essence of Neovide’s "animated cursor," which differs from the "cursor trail" implementation in Kitty.

Unfortunately, it’s not as straightforward as it might seem. Here’s what happens when we simply toggle cursor visibility:

https://github.com/user-attachments/assets/037fd5f6-6f32-4520-9f72-2ff224b9a808

(Ignore the blinking artifact.)

The real issue is that the trail hides text underneath it, regardless of cursor visibility. This happens because the cursor trail is essentially rendered as a plain opaque quad. I had to use a trick like this to ensure the glyph remains visible on the cursor while the trail is moving:

https://github.com/kovidgoyal/kitty/blob/a5f70fbee52919f36910a0df5ac4473ac613b659/kitty/trail_fragment.glsl#L10-L12

It skips rendering on the cursor’s position. It would look even more awkward with this if the cursor is turned off.

I’ve spent some time thinking about how to achieve Neovide’s animated cursor look and feel. @kovidgoyal and @mxple had a great discussion about it even before I started working on this feature. See: https://github.com/kovidgoyal/kitty/issues/2460#issuecomment-2362853770. Here are the main options I’ve considered:

1. Extend cell shaders to support arbitrary shapes. This would require minimal new code and likely have the least performance impact. However, it’s too intrusive for a feature not everyone wants or is willing to maintain.

2. Redraw foreground text or images on top of the cursor trail. This involves minimal changes to the existing codebase but is fairly complex, and I’m concerned the added complexity could become a maintenance burden.

3. Use a two-pass rendering approach. This would pass through foreground pixels on top of the cursor trail. While it might be simpler than the other two options, it would significantly hurt performance, making it impractical.

I’ll revisit this if I come up with a better way to implement it.

TRPB:
Thanks for looking into that so quickly! That's definitely a problem and not ideal, you're right, it's not as simple as I'd hoped it might be. Good effort so far!

Looking at neovide (just using it, not the code) it appears that the cursor trail is always below the text. Could you draw the cursor trail on a layer between the background and the text? I don't know how the terminal differentiates the parts but kitty knows to apply transparency to the background. Whether it's even feasible to put another drawn layer in-between the background and the text I have no idea.

Sorry for multiple requests but this is such an awesome feature. One other thing that would be nice is antialiasing on the trail. The lack of it is particularly apparent in your slowed down video. I would hope that it would be just enabling it as part of the shader but what do I know, I suggested hiding/showing the cursor should be "fairly trivial"

jinhwanlazy:
OK, that could be option 4. There are 3 main rendering methods implemented in kitty: one that draws every element (background, cursor, text, image, etc.) in a single draw call, which is supposed to be concise and fast, and the other 2 that draws them one by one, each is triggered when transparency and a background image is involved. I might be able to make kitty to use the latter methods and insert a draw call for cursor_trail between the background and text, specifically here: https://github.com/kovidgoyal/kitty/blob/8e388acab7512ef23994c388bd0b390d81558298/kitty/shaders.c#L898-L899

The downside is this would mess with existing rendering logic, which I'm trying to avoid (Same reason as I'm unhappy with option 1).  Resulting code would look messy unless some careful refactoring is done. e.g. the function should be renamed from `draw_cells_interleaved` to something like `draw_cells_interleaved_and_trail_if_exist`.

kovidgoyal:
draw_cells_interleaved just means the cells are drawn in multiple passes
there is no need to rename it for cursor trail in particular, you just
need to update the logic in draw_cells to choose the correct draw method
and then insert an extra draw call to draw the cursor trail after
drawing the background and special cells.

TRPB:
Looking again I don't think it will actually be quite that simple either, unfortunately. Because the character under the cursor is a different colour to the character before the cursor is over it. You original shader overlay might be better

For example, a white cursor with a white character under it like your video, the semicolon is a different colour before and after the cursor is on it

jinhwanlazy:
@kovidgoyal regarding your recent commit https://github.com/kovidgoyal/kitty/commit/4b12bffcb19375f74d6fba3e59adf5cdf8062dca
> Branches have an outsize penalty in shaders as does discard. We don't
need the check that the fragment is in the region anyway ans the
rendering pipeline takes care of that for us, I think. At least I can
see no visual difference with it removed.
> 
> @jinhwanlazy please review

These videos below are with and without your change. 
For me, the difference is clearly visible. The text at the cursor is masked by the trail while it is moving in the version with the commit. I think it would be better to revert this change.

https://github.com/user-attachments/assets/128b9981-7ada-4257-b486-b8c9fffc7d37

https://github.com/user-attachments/assets/af27d5c3-c707-4212-8029-6ad3e5fdc62a

kovidgoyal:
If you want to restore it then implement it without using if and discard. Instead use step() to set a multiplier that will be zero outside the region and 1 inside and multiple the opacity by that.

kovidgoyal:
And I will note that I cant see any difference in your videos. In both cases during the large horizontal movement text is obscured to the same extent.

I tested it with
kitty -o cursor_trail=3 -o "cursor_trail_decay=1 4"

so that it can be easily followed by eye. And the result are identical with and without the patch.

kovidgoyal:
Ah never mind I think I got it, you mean you dont want the cursor trail drawn over the actual cursor position.

kovidgoyal:
This should take care of it:
https://github.com/kovidgoyal/kitty/commit/4839233676ccb49521266795481baa639a03317f

linkarzu:
@jinhwanlazy mentioned by Kovid, and on Live TV
https://youtu.be/8PYLPC3dzWQ?si=PuWVbVshAVrVUEH2&t=4319

jinhwanlazy:
guess I’m famous now lol. thanks for the shoutout :tada: @kovidgoyal your pronounciation is perfect!

kovidgoyal:
Glad to hear it and you totally deserved the shoutout, like I said cursor trails was not even on my radar until you popped up.

PR #7980, opened 2024-10-18
Fix segfault when font descriptor point to invalid font file

matubu:
For some reason, `d`, the descriptor used to load the font face, points to a file that does not exist in my system, so `face_from_descriptor` returns `NULL`, and then `has_cell_text` segfaults. I don't know if it's a normal behavior. I also had to add a seemingly missing `return` (all the other lines had it, and it was segfaulting at `init_ft_face` otherwise). Maybe there is another root cause issue that should be fixed, but I don't know the code base enough to know.

Either way, have a great day!

### DIFF
--- a/kitty/fontconfig.c
+++ b/kitty/fontconfig.c
@@ -520,7 +520,7 @@ face_from_descriptor:
             }
         }
         ans = face_from_descriptor(d, fg);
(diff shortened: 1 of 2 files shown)


kovidgoyal:
You should firgure out why fontconfig is reporting non-existent paths on your system. While kitty should definitely not be crashing for that, it still should be fixed there.

PR #7985, opened 2024-10-19
Minor fixes about cursor trail 

jinhwanlazy:
two main changes
* Reuse cursor color computed in `cell_update_uniform_block` to render cursor trail
* Refactored cursor trail update logic. So its behavior can be more easily grasped looking at the code.

> Similarly, I want to basically ensure no trail is started while in
rendering paused mode and that for cursor movements that happen in that
mode the position_changed_by_client_at field is the time at which
rendering was resumed. 

I believe the current implementation already addresses your concern. Here's why:

1. The trail only displays when the current trail position and target cursor position are not equal:

```c
ct->needs_render = false;

if (update_cursor_trail_corners(ct, w, now, os_window)) {
    ct->needs_render = true;
}
```

2. The target cursor position does not change while in the paused_rendering state:

```c
if (!WD.screen->paused_rendering.expires_at && OPT(cursor_trail) < now - WD.screen->cursor->position_changed_by_client_at) {
    update_cursor_trail_target(ct, w);
}
```

3. Given points 1 and 2, while in paused rendering, the cursor movement doesn't change target position of the trail(`cursor_edge_x` and `cursor_edge_y`). Instead, the trail position updated toward the target position last set before entering paused_rendering. Once trail reaches its target position, it keeps position and rendering not be triggered until pausing end.

### DIFF
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -45,6 +45,7 @@ update_cursor_trail_target(CursorTrail *ct, Window *w) {
         EDGE(x, 1) = right;
         EDGE(y, 0) = top;
         EDGE(y, 1) = bottom;
+        ct->screen = WD.screen;
     }
 }
 
@@ -119,13 +120,9 @@ update_cursor_trail(CursorTrail *ct, Window *w, monotonic_t now, OSWindow *os_wi
         ct->needs_render = true;
     }
 
-    if (ct->needs_render) {
-        ColorProfile *cp = WD.screen->color_profile;
-        ct->color = colorprofile_to_color(cp, cp->overridden.cursor_color, cp->configured.cursor_color).rgb;
-    }
-
-#undef WD
-#undef EDGE
     // returning true here will cause the cells to be drawn
     return ct->needs_render || needs_render_prev;
 }
+
+#undef WD
+#undef EDGE
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -1,4 +1,5 @@
 #include "state.h"
+#include "float.h"
 
 inline static float
 norm(float x, float y) {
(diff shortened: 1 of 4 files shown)


jinhwanlazy:
I tried kittens(diff, themes and choose_fonts). All of them seems to work smoothly at least not inside tmux.

kovidgoyal:
Merged, changing the code to not store a pointer to Screen which is
dangerous as its lifetime is not guaranteed to be the same as that of
CursorTrail.

Your condition 1 above applies only to paused rendering, not when the
cursor is hidden. We need the trail to not be started when the cursor
position is changed while it is hidden as well.

jinhwanlazy:
Thanks for the fix! 

Regarding the "hidden" cursor, you're referring to when DECTCEM is 0, right? I have some thoughts on why the cursor trail shouldn't rely on this state. Here's my reasoning:

Imagine we have three windows open in a Kitty tab. Two are running bash, and the third is running htop, where the cursor is constantly hidden. If we don't update the trail's target position when the cursor is hidden, we might see some unexpected behavior:

1. When moving focus from window 1 to window 3 (htop), the trail wouldn't start.
2. Then, moving from window 3 to window 2, the trail would suddenly jump from window 1 to window 2, because it never "moved" to window 3.

This doesn't seem like the ideal user experience. I think it's better for the trail to move regardless of the destination window's cursor visibility. This way, users can always see where their focus is moving from and to.

The paused_rendering state works differently because it's usually short-lived(as far as I can see). In contrast, some applications (like nvim's terminal mode or htop) keep the cursor hidden constantly.

One potential compromise could be to add some opacity to the trail, so it slowly fades when moving to a window where the cursor is hidden. But I'm not sure if this extra complexity is worth implementing, as I'm pretty happy with how it works now.

I'm definitely open to your thoughts on this. If you notice any glitches or issues with the cursor trail, please let me know, and I'll do my best to address them!

kovidgoyal:
OK, this is not a feature I use myself, so I defer to your judgement on
how it should work. If users report bugs about it in the future, I will
assign them to you and you can figure out whatever you think is best.

PR #7998, opened 2024-10-27
Enhance Cursor Trail: Improved Animation, Fades, and New Distance Threshold Option

jinhwanlazy:
continued from https://github.com/kovidgoyal/kitty/pull/7970. 3 main changes

1. Improved cursor trail animation. It now looks uniform regardless of its moving direction. The movement appears smoother, especially when the cursor moves horizontally.

https://github.com/user-attachments/assets/9c089fb9-666a-4057-80ed-ab1036ce5231

https://github.com/user-attachments/assets/618fbf65-487a-423a-9ad3-afda22c4de2d

https://github.com/user-attachments/assets/4ec4c3c7-3aeb-4285-bcc5-c539f041b352

https://github.com/user-attachments/assets/a06cc8c8-2bcf-4f8e-b975-c918e4da6441

2. The trail now fades in and out when moving between windows with different cursor visibility. One caveat: the trail will not be visible when moving from a window where the cursor is hidden to another window where the cursor is also hidden. However, this scenario is likely rare.

https://github.com/user-attachments/assets/e2372378-b243-4e9d-9657-7e0464999a6e

https://github.com/user-attachments/assets/fc767669-f9c5-414c-8db5-6fc85b78ed10

3. Added a new option `cursor_trail_distance_threshold`. The cursor trail will not be initiated when the cursor jump distance is below this threshold.

### DIFF
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -52,11 +52,10 @@ should_skip_cursor_trail_update(CursorTrail *ct, Window *w, OSWindow *os_window)
         return true;
     }
 
-    int distance_threshold = 2;  // make it option
-    if (distance_threshold > 0 && !ct->needs_render) {
+    if (OPT(cursor_trail_distance_threshold) > 0 && !ct->needs_render) {
         int dx = (int)round((ct->corner_x[0] - EDGE(x, 1)) / WD.dx);
         int dy = (int)round((ct->corner_y[0] - EDGE(y, 0)) / WD.dy);
-        if (abs(dx) + abs(dy) <= distance_threshold) {
+        if (abs(dx) + abs(dy) <= OPT(cursor_trail_distance_threshold)) {
             return true;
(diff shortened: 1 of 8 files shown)


kovidgoyal:
I have merged, but you should check how the trail rendering works ina
semi-transparent window (i.e. with background_opacity < 1) I dont think
BLEND_ONTO_OPAQUE is correct in that case.

kovidgoyal:
On Sun, Oct 27, 2024 at 06:02:06AM -0700, Jinhwan Choi wrote:
> Sorry, but I don't see what's wrong with `BLEND_ONTO_OPAQUE`.

Nothing wrong with it, in graphics programming there is no right and
wrong. Just whatever looks good. I just wanted to make sure you had
considered the case where the surface being blended onto was not opaque.
As long as you checked and you feel it looks ok, I am good with it.

PR #8058, opened 2024-11-18
tui: explicitly turn off in-band resize mode

rockorager:
Explicitly turn off in-band resize notifications if they were requested as part of the loop. Kitty implicitly resets this mode when `XTRESTORE` (`CSI ? r`) is seen, however some terminals implementing mode 2048 (eg foot and Ghostty) follow the xterm behavior of only restoring the state of explicitly passed modes. Some of the kittens have become very useful outside of just Kitty (`show-key`, for example) - but when mode 2048 is left enabled in this way, escape codes are dumped to the screen. 

Add an explicit reset of mode 2048 to support the behavior of other terminal emulators.

### DIFF
--- a/tools/tui/loop/terminal-state.go
+++ b/tools/tui/loop/terminal-state.go
@@ -169,6 +169,9 @@ func (self *TerminalStateOptions) ResetStateEscapeCodes() string {
 	} else {
 		sb.WriteString(SAVE_CURSOR)
 	}
+	if self.in_band_resize_notification {
+		reset_modes(&sb, INBAND_RESIZE_NOTIFICATION)
+	}
 	sb.WriteString(RESTORE_PRIVATE_MODE_VALUES)
 	if self.restore_colors {
 		sb.WriteString(RESTORE_COLORS)


kovidgoyal:
Yeah it's a kitty extension to support restoring all modes to saved values without having to list them explicitly. Seems a no brainer to me, you might suggest to the developers of foot and ghostty to follow kitty in that. That said, this is sufficiently harmless to implement, its just a few extra bytes at kitten shutdown. But I do want to make it clear that in general, kittens working with other terminal emulators is a non-goal for me and not something I will sped time/effort on.

mitchellh:
> But I do want to make it clear that in general, kittens working with other terminal emulators is a non-goal for me and not something I will sped time/effort on.

Author of Ghostty here. This is a reasonable take and I respect it. 

> Yeah it's a kitty extension to support restoring all modes to saved values without having to list them explicitly.

I'll consider it.

✌️

rockorager:
On Mon Nov 18, 2024 at 8:36 PM CST, Kovid Goyal wrote:
> That said, this is sufficiently harmless to implement, its just a few extra
> bytes at kitten shutdown. But I do want to make it clear that in general,
> kittens working with other terminal emulators is a non-goal for me and not
> something I will sped time/effort on.

I assumed this was a case, hence why I went straight for the PR rather than
making an issue :).

Thanks for understanding!

-- 
Tim

kovidgoyal:
> I'll consider it.

I have added docs for it, see docs/misc-protocol.rst

Issue #8007, opened 2024-10-31
Feature request: drawing panels in top or overlays layers


robin-carlier:
I really like the new `panel` kitten, but as far as I can tell, it only allows rendering the panels in layers 0 (via `GLFW_LAYER_SHELL_BACKGROUND` or 1 (via `GLFW_LAYER_SHELL_PANEL`). It looks like the various functions in `wl_window.c` don’t allow changing the layer via anything exposed to the user. Would it be possible to add an option/flag to render kitty in layers 2 or 3 (so, top/overlay)? It would allow using a TUI program as a screensaver for instance.

kovidgoyal:
Sure you are welcome to send a PR for it. It's on my long term TODO list but I dont personally have a use case for it at the moment so its not high on that list. Although, how would it work for a screensaver, doesnt a screensaver also need to inform the compositor somehow and lock the screen etc? You can add edge=overlay/whatever for it. 
Ideally, there would be some way to specify geometry so it could be used to draw popups like notifications or dialogs as well. Though I dont remember what the layer shell protocol supports in the way of geometry specification, wayland is queerly allergic to allow windows to position themselves.

Closing the issue feel free to ask more questions if needed.

robin-carlier:
> Although, how would it work for a screensaver, doesnt a screensaver also need to inform the compositor somehow and lock the screen etc?

I was thinking about letting swaylock/hyprlock/whatever handle that.

>Ideally, there would be some way to specify geometry so it could be used to draw popups like notifications or dialogs as well.

Agreed, this would be nice to be able to use kitty as a rofi/wofi/dmenu kind of thing.

> Sure you are welcome to send a PR for it.

I’ll see if I can cook up something in a relatively near future, but that’d be my first time doing something like this and working with this protocol, so this will probably need to be very carefully reviewed once ready.

kovidgoyal:
On Fri, Nov 01, 2024 at 04:22:02AM -0700, Robin Carlier wrote:
>
> &gt; Although, how would it work for a screensaver, doesnt a screensaver also need to inform the compositor somehow and lock the screen etc?
>
> I was thinking about letting swaylock/hyprlock/whatever handle that.
>
> &gt;Ideally, there would be some way to specify geometry so it could be used to draw popups like notifications or dialogs as well.
>
> Agreed, this would be nice to be able to use kitty as a rofi/wofi/dmenu kind of thing.

I doubt you will find the layer_ext protocol capable of doing this. But I will be
[...]

PR #8068, opened 2024-11-23
More panel options

robin-carlier:
I finally found some time to implement what we discussed (and a bit more) in #8007

The main additions/modifications are

- A new `--layer` option for the kitten panel to specify on which layer the panel should be drawn.
- A new `--edge=none` option for the kitten, when set, the width, height and margins of the kitten can be freely set to get some flexibility in the geometry. When set, the panel leaves its exclusive_zone to `-1`, so it might get drawn on top of other things. Maybe we could also add a flag to conthol this as well?
- `--lines` and `--colums` are now split as two distinct parameters to control the size (in cells) when using `--edge=none`.
- New `--margin-...` options to set some margin if desired.
- New `focus-policy` flag.

With this, I think the kitten should be reasonably flexible to be used in fun ways.

I could test that the options behave as expected using Hyprland, but I don’t know if they do for other compositors.


kittens/panel/main.py
```diff
@@ -39,6 +41,14 @@
 kitten.
 
 
+--layer
+choices=background, bottom, top, overlay
+default=bottom
+On a compositor that supports the wlr layer shell protocol, specifies the layer
```
kovidgoyal, line 82: On a *Wayland* compositor to make it clear this option is Wayland specific

glfw/wl_window.c
```diff
@@ -998,6 +998,10 @@ layer_set_properties(_GLFWwindow *window) {
                     panel_width = window->wl.width;
                     exclusive_zone = window->wl.width;
                     break;
+                case GLFW_EDGE_NONE:
+                    panel_width = window->wl.width;
+                    panel_width = window->wl.height;
```
kovidgoyal, line 1004: Shouldnt this be panel height?

kittens/panel/main.py
```diff
@@ -24,14 +25,20 @@
...
+
+--columns
+type=int
+default=1
+The number of columns shown in the panel if vertical. Ignored for background panels.
 
 
 --edge
-choices=top,bottom,left,right,background
+choices=top,bottom,left,right,background,none
```
kovidgoyal, line 68: Add an explanation for the meaning of none to the help text

kittens/panel/main.py
```diff
@@ -28,13 +28,37 @@
...
 type=int
 default=1
-The number of columns shown in the panel if vertical. Ignored for background panels.
+The number of columns shown in the panel. Ignored for background and horizontal panels.
+
+
+--margin-top
+type=int
+default=0
+Request a given top margin to the compositor.
```
kovidgoyal, line 46: Add that the margin only works on Wayland compositor support layer shell. Add to help text for all margin options.

kovidgoyal:
I am fine with adding a flag to control exclusive_zone.

glfw/wl_window.c
robin-carlier: Yes, good catch!

robin-carlier:
> I am fine with adding a flag to control exclusive_zone.

I split it into two flags: `override-exclusive-zone` to allow override for panels anchored through e.g, `edge=top`, and a `exclusive-zone` flag to actually pass the value in case it is overriden.

kovidgoyal:
Merged. It might be good to add some "recipes" or cookbook style
examples for doing common things to the docs for the panel kitten in
docs/kittens/panel.rst

PR #8194, opened 2025-01-05
ssh kitten: Fix zsh shell integration expanding variables in title

### LINKED ISSUE #8193: Bug: SSH with ZSH shell integration expands variables quoted in ' when setting title
## Describtion
When executing a command on a remote machine that contains some variables inside single quotes, these variables are expanded by `_ksi_preexec` if the shell integration is configured to set the tile.

## Steps to reproduce
1. Make sure the [`shell_integration`](https://sw.kovidgoyal.net/kitty/kittens/ssh/#opt-kitten-ssh.shell_integration) for `kitten ssh` is turned on and `no-title` is **not** set.
2. SSH into a remote machine using `kitten ssh ...`
3. Execute a command that contains a variable in single quotes. For example:
    ```shell
    # Delete file "somefile" if it starts with "foo"
    find . -type f -name somefile -exec bash -c '[[ $(head -c 3 "{}") = "foo" ]] && rm -- "{}"' ';'
    ```

This has been tested with `kitty --config NONE`, shell is `zsh`.
## Details

There are two main issues with this:
1. The title will be wrong, because the variable is expanded. In this case, the title will look like this: `... -exec bash -c '[[ = "foo" ]] && ...`
2. When expanding a command like in the example above using `$(...)`, `_ksi_preexec` will **execute** the command in the variable. I think there is no need to say that this is potentially dangerous.

In the example above, `head` will print an error to `stderr` in the current shell:
```text
head: cannot open '{}' for reading: No such file or directory
```
This was how I originally stumbled upon this bug.

### Origin of the bug
When executing the command, `_ksi_preexec` will execute the following command:
```shell
builtin print -Prnu 13 $'\e]2;'"%m: ${(V)1}"$'\a'
```
thereby expanding (because of the `-P` option) the variable.

Another very minimal example of this would be the following:
```shell
echo '$(echo "hi" 1>&2)'
# Prints the following: $(echo "hi" 1>&2)
```
While this example is not a real world example of code someone would actually write, it is far simpler than the `find` example. With this, I tested how `_ksi_preexec` expands the variable:
```shell
command_text='echo '\''$(echo "hi" 1>&2)'\'
builtin print -Prn "%m: $command_text"
# Output:
# hi
# machine-name: echo ''
```

### Potential fix
The origin of the problem lies here:
https://github.com/kovidgoyal/kitty/blob/98c1e0f7aa93078ae8afea875a9831aced602f6b/shell-integration/zsh/kitty-integration#L273-L274

The option `-P` is used to expand `%m` to the machine name. Instead, we could define a variable:
```shell
machine_name=$(builtin print -Prn "%m")
functions[_ksi_preexec]+="
    builtin print -Prnu $_ksi_fd \$'\\e]2;'\"$machine_name: \${(V)1}\"\$'\\a'"
```

> [!IMPORTANT]
> I have not yet checked if the same issue occurs in Bash, I have only tested Zsh so far.

I'd be happy to fix this in a pull request. Let me know what you think :)

3j14:
Remove usage of the `-P` option from the print inside the `_ksi_precmd` and `_ksi_preexec` functions. Instead, the result of `builtin print -Pn "%m"` is stored in the local variable `hostname`. This mitigates accidental expansion of variables in the command to be executed.

Fixes #8193.

### DIFF
--- a/shell-integration/zsh/kitty-integration
+++ b/shell-integration/zsh/kitty-integration
@@ -268,10 +268,15 @@ _ksi_deferred_init() {
 
         if [[ "$is_ssh_session" == "y" ]]; then
             # show the hostname via %m for SSH sessions
+            builtin local hostname="$(builtin print -Pn "%m")"
+            # The variable `hostname` is used such that we don't have to use
+            # the -P option in the two print commands below. This mitigating
+            # accidental variable expansion, e.g. of the variables in the
+            # command string (V).
             functions[_ksi_precmd]+="
-                builtin print -Prnu $_ksi_fd \$'\\e]2;'\"%m: \${(%):-%(4~|…/%3~|%~)}\"\$'\\a'"
+                builtin print -rnu $_ksi_fd \$'\\e]2;'\"$hostname: \${(%):-%(4~|…/%3~|%~)}\"\$'\\a'"
             functions[_ksi_preexec]+="
-                builtin print -Prnu $_ksi_fd \$'\\e]2;'\"%m: \${(V)1}\"\$'\\a'"
+                builtin print -rnu $_ksi_fd \$'\\e]2;'\"$hostname: \${(V)1}\"\$'\\a'"
         else
             functions[_ksi_precmd]+="
                 builtin print -rnu $_ksi_fd \$'\\e]2;'\"\${(%):-%(4~|…/%3~|%~)}\"\$'\\a'"


3j14:
Note that the variable `$hostname` is expanded when adding the print command to the `_ksi_precmd` and `_ksi_preexec` functions, so the functions look as follows:
```shell
_ksi_preexec () {
        # ...
        builtin print -Prnu 13 $'\e]2;'"SOMEHOSTNAME: ${(V)1}"$'\a'
        # ...
}
```
where `SOMEHOSTNAME` is the name of the remote at the time of initializing the shell integration. If the hostname changes during the session, this will not update.

I think this is fine, I don't even know if `print -P "%m"` updates immediately upon changing the hostname of a machine or only after reboot.

Alternatively, we would have to move the variable inside the function, but I'm not a fan of that.

3j14:
Oh, you were a bit quicker than me! Thanks for merging!

kovidgoyal:
Yeah I dont care about hostname being changed dynamically, that's too much of an edge case. Thanks for the fix!

romkatv:
A better fix would be to replace `%m` with `\${(V)HOST-}` and remove `-P` from `print`.

3j14:
@romkatv I was already wondering if there is a builtin parameter referring to the hostname. Thanks for pointing me to `HOST`!

I totally agree that using `$HOST` is probably way more elegant and eliminates the extra `print -P` and the variable `$hostname`

According to the [documentation](https://zsh.sourceforge.io/Doc/Release/Parameters.html#Parameters-Set-By-The-Shell), the `HOST` variable refers to the current hostname. [`%m`](https://zsh.sourceforge.io/Doc/Release/Prompt-Expansion.html#Login-information) on the other hand refers to the hostname up to the first '.'.

**Depending on whether you want to print the hostname up to the first '.' or in its entirety, I propose the following change**

```shell
${(V)${HOST-}/%.*/}
```
which works as follows:

1. Expand `${HOST-}`. The `-` substitutes the expression with an empty string is `HOST` is not set
2. `${HOST-}/%.*/` replaces the longest possible match of `.*` with an empty string, where `%` indicates that the patter must match at the end of the string.
3. Finally, the [`(V)` flag](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Parameter-Expansion-Flags) makes any special characters in the resulting words visible.

Let me know which you prefer:
```shell
${(V)HOST-}
# or
${(V)${HOST-}/%.*/}
```

I can open another pull request if you think we should use the `$HOST` variable.

3j14:
Nevermind, just saw that you @kovidgoyal already updated it, thanks for the fix!

romkatv:
@3j14 That's a very good point. When I made my comment, I didn't remember that `%m` truncates the hostname. I knew this at some point -- as I've raid the manual once through, and then many times patchily -- but then forgot. I apologize, as I only now realize that my comments here may be treated with authority, which would actually be entirely logical given my zsh projects. I'll keep this in mind the next time I post here, and will be more careful in verifying the correctness of my words, abiding by the same standards I followed when implementing zsh integration.

In case my main message was lost in meta comments: I agree that `${(V)${HOST-}/%.*/}` would be better.

romkatv:
> Also, I am not even sure if the distinction between the abbreviated or full hostname is even important here :)

Oh, I'm not saying it's important but the truncated hostname is clearly correct there. That's why `%m` -- a prompt escape sequence -- truncates. Terminal titles are just like prompts in terms of brevity-exactness tradeoff.

kovidgoyal:
Interesting do people really use fully qualified names as hotsname? I have made the change.

3j14:
I'm not 100% certain but I think in da3bb0a4cbd02a3c794a4012db42c9a694e040ca you are missing an escape (`\`) for that second `$`. You don't want to expand `${HOST-}` immediately but only later (the outer expansion includes a `\`):

```shell
\${(V)\${HOST-}/%.*/}
# instead of
\${(V)${HOST-}/%.*/}
```

kovidgoyal:
Right, doesn't really matter since I doubt zsh re-reads hostnames after
startup but, no harm I guess.

romkatv:
tl;dr: The fix suggested by @3j14 is correct and important.

`HOST` is a read-write parameter, so it can indeed change after startup. `%m` stays tied to `HOST` even if the latter is modified.

```zsh
% print -P '%m'
adam
% HOST=foo
% print -P '%m'
foo
```

`\${(V)${HOST-}/%.*/}` would attempt to resolve a parameter whose name is equal to the hostname. It would likely be empty but could also be anything.

```zsh
% typeset -p HOST adam
typeset: no such variable: adam
typeset HOST=adam
% eval "print -r -- \${(V)${HOST-}/%.*/}"

% adam=foo
% eval "print -r -- \${(V)${HOST-}/%.*/}"
foo
```

Lastly, this portion of the code is generating a function. It's best not to open the can of worms that would result from injecting the value of a parameter in it.

PR #8195, opened 2025-01-05
Enable macOS universal autofill

memco:
Implements #8173

This is the bare minimum of menu items I could add to coax macOS into providing the requested AutoFill > Passwords... menu. Unfortunately you cannot just create an Edit menu, it has to have at least one child for the AutoFill (and also the Start Dictation... menu item) to appear: hence the lone `separatorItem`

After changes:

Screenshot 2025-01-04 at 10 19 37 PM

Before changes:

Screenshot 2025-01-04 at 10 33 12 PM

### DIFF
--- a/kitty/cocoa_window.m
+++ b/kitty/cocoa_window.m
@@ -755,6 +755,13 @@ cocoa_create_global_menu(void) {
     MENU_ITEM(appMenu, ([NSString stringWithFormat:@"Quit %@", app_name]), quit);
     [appMenu release];
 
+    NSMenuItem* editMenuItem = [bar addItemWithTitle:@"Edit"
+                                              action:NULL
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Well if we are going to be adding an edit menu might as well add copy, paste, clear scroll back and cear screen actions to mimic Terminal.app

zharmany:
I noticed that in macOS, this change also enabled the use of the native macOS emoji picker. For better or worse, I had been using `ctrl-cmd-space` as a sort of leader or prefix key for Kitty since it explicitly did not support the emoji picker. 

I was able to work around this by reassigning the "Emoji & Symbols" menu item to a different keyboard shortcut, which allowed me to continue to use `ctrl-cmd-space` in Kitty.

PR #8334, opened 2025-02-17
Fix issues with prompt scrolling and cmd output paging

rivenirvana:
Currently, the cmd output paging logic does not properly page for:
- `mouse_show_command_output` on the current prompt with no output
- `mouse_show_command_output` on wrapped prompt lines (long commands)
- `show_first_command_output_on_screen` when the prompt of the first cmd output is just above the first line of the screen

Upon investigating, I came across a tangential issue, an inconsistency between how `scroll_to_prompt` finds prompt lines that are in the linebuf vs in the historybuf. After performing a resize that wraps prompt lines already in the historybuf, using `scroll_to_prompt` scrolls the terminal to individual lines of the wrapped prompt line, instead of only where `PROMPT_START` is assumed to be, at the beginning of the prompt.

With this change, `scroll_to_prompt` works properly for prompt lines in historybuf even after resize events, and all cases mentioned above are addressed. `make test` passes all tests.


kovidgoyal:
Please add some tests that exercise these cases to test_prompt_marking()
in kitty_tests/screen.py

Let me know if you have questions about how the testing environment
works.

You can run a single test with:

make debug && ./test.py prompt_marking

kovidgoyal:
I fixed the prompt attrs issue slightly differently and added a test for it.

rivenirvana:
Had a couple of issues especially with `lvco()`, but I hope the tests are sufficient.

Some things I noticed that I'd like to take note of, for future reference.
- The only way to set the last visited y-value in tests is to use a successful `scroll_to_prompt`, meaning it scrolled. And that only sets it to the first line on the screen, making the tests a bit longer than they have to be.
- If I am in scrollback and there are 2 prompts below, 1 at `y = 0` and another at `y > 0` (can't be scrolled to), using `s.scroll_to_prompt(2)` will scroll and return True (it counts inside even when it can't scroll past `y = 0`), but using `s.scroll_to_prompt(1)` twice will be False for the second call since it does not scroll any further. Using `s.scroll_to_prompt(3)` returns False and does not scroll. Doesn't really affect the user interface since we're only allowed to jump one prompt at a time, but something I tripped on for a moment.
- On resize, a visible prompt with supposedly no ouput will have its wrap eaten and a `<` character replace it. Wondering if this is an issue cause this is from `preserve_blank_output_start_line()` which seems to be cleared anyway. Was having issues with it in the beginning of making tests cause I was losing track of lines as it was eating the prompt.
```py
def pls():
    print()
    print('<---')
    for i in range(s.lines):
        print(f'"{s.visual_line(i)}"')

s = self.create_screen()
draw_prompt('abc')
pls()
s.resize(s.lines, s.columns - 1)
pls()
```

Output:
```
<---
"$ abc"
""
""
""
""

<---
"$ ab"
"<"
""
""
""
```
In any case, none of it are in need of urgent attention, and tests pass nonetheless.

kovidgoyal:
On Tue, Feb 18, 2025 at 09:17:00AM -0800, Arvin Verain wrote:
> rivenirvana left a comment (kovidgoyal/kitty#8334)
>
> Had a couple of issues especially with `lvco()`, but I hope the tests are sufficient.
>
> Some things I noticed that I'd like to take note of, for future reference.
> - The only way to set the last visited y-value in tests is to use a successful `scroll_to_prompt`, meaning it scrolled. And that only sets it to the first line on the screen, making the tests a bit longer than they have to be.

You can add some optional/keyword args to scroll_to_prompt() to
manipulate last_visited_prompt for tests.

> - If I am in scrollback and there are 2 prompts below, 1 at `y = 0` and another at `y > 0` (can't be scrolled to), using `s.scroll_to_prompt(2)` will scroll and return True (it counts inside even when it can't scroll past `y = 0`), but using `s.scroll_to_prompt(1)` twice will be False for the second call since it does not scroll any further. Using `s.scroll_to_prompt(3)` returns False and does not scroll. Doesn't really affect the user interface since we're only allowed to jump one prompt at a time, but something I tripped on for a moment.

I dont think this is an issue in practice because in the UI we cant
scroll past y = 0 anyway.

> - On resize, a visible prompt with supposedly no ouput will have its wrap eaten and a `<` character replace it. Wondering if this is an issue cause this is from `preserve_blank_output_start_line()` which seems to be cleared anyway. Was having issues with it in the beginning of making tests cause I was losing track of lines as it was eating the prompt.

This definitely sounds like a bug.

kovidgoyal:
And I looked into it, it's not a bug. It happens in the prompt case
because of prevent_current_prompt_from_rewrapping(). Basically, we are
expecting the shell to redraw everything from the start of the prompt
downwards so the contents of the buffer are undefined. I dont think this
is a case we need to worry about unless it causes real world problems.

PR #8373, opened 2025-02-24
Adjust the size of the alternative icon on the FAQ page

igrmk:
The FAQ page currently uses the SVG version of the whiskers icon. This SVG serves as the 'source' for PNG icons, with additional padding added during PNG creation, varying by OS (on macOS, the icon has slightly more padding). Since the source lacks any padding, the icon appears too large


igrmk:
Actually, I'd better fix it on my side by providing an SVG icon with the correct padding instead of the 'source' SVG

igrmk:
Done

PR #8415, opened 2025-03-08
Fix regression in setup.py

leiserfg:


leiserfg:
There is a regression in `setup.py --shell-integration=enabled\ no-rc` because of this.

PR #8508, opened 2025-04-04
Prevent mouse cursor from showing on short, unintentional flicks

irisjae:
I use kitty daily, and hiding the mouse in my terminal helps my focus a lot. Currently, kitty always shows the mouse cursor immediately after I move the mouse, even if it happens due to unintentional flicks on my trackpad, breaking my focus.

In this patch, I've added three new options `mouse_show_wait`, `mouse_show_threshold`, and `mouse_scroll_show`.  `mouse_show_wait` controls how long to wait before mouse activity is treated an intentional, `mouse_show_threshold` controls the threshold to treat mouse activity as intentional, and `mouse_scroll_show` controls whether mouse scrolls trigger unhiding the mouse cursor. The current behaviour can be recovered by setting `mouse_show_wait` to 0, and is configured to be the default.


irisjae:
Thanks for your quick response! I've made the changes you mentioned, see if that's right?

kovidgoyal:
Looks generally OK after a quick once over will review in detail when I have some time.

irisjae:
Hi, I see that in your cleanup commit 3d440cf0f68cec80858ef2e18be5c99d272c8e07 you replaced the `/ 1e9` calculation with `monotonic_t_to_ms`, I think that should be `monotonic_t_to_s_double`? If we make unhide event threshold is set as ms * threshold, that's 3 orders of magnitude bigger than what I intended in the documentation, and the threshold option is too coarse to allow unhiding with reasonable mouse activity.

kovidgoyal:
yup, sorry, typo. fixed.

Issue #540, opened 2018-05-15
Support blinking text


RedBeard0531:
I noticed this because the lightning bolts from `curl wttr.in` weren't blinking.

Example that blinks in gnome-terminal but not kitty: `echo  -e "normal|\e[5mblink\e[0m|normal"`

kovidgoyal:
I'm afraid I am not interested in supporting blinking text. It is a feature that has a large overhead for a very limited use case.

BornstellarEternal:
Any updates on whether you plan to support blinking text?

jctratt:
The limited use case for blinking text is at times a critical feature; I happen to need the feature.

ppwwyyxx:
Some users prefer rich features over speed - although kitty are already very strong on both.

Would be great if owners can reconsider this feature, even if it's disabled by default with a warning on performance.

residualmind:
I really love kitty. But these opinionated choices make it really hard to finally adopt the kitty. No ability to disable bold (historically, the ANSI bold tag has been used to render *bright* text - that is why many (most) applications use the tag to render bright text, not bold - which is a very valid reason to disable bold rendering in a terminal. I mean, it's still a terminal, not a graphical WYSIWYG word processor.) And now I find out blink is not supported.. sad puppy.

yermulnik:
Author is [so disgustingly pompous](https://github.com/kovidgoyal/kitty/issues/8030#issuecomment-2561952762) speaking to people in general 😒 Sadly making `kitty` less attractive 🙁

kovidgoyal:
Excellent, another issue to lock down. Blinking text is an awesome filter.

kovidgoyal:
Implemented in #8551

Issue #8030, opened 2024-11-07
Shell text blink


yermulnik:
**Describe the bug**
A shell escape codes to make text blink doesn't work in Kitty.

**To Reproduce**
1. `echo '\033[33;5mTitle of the Program\033[0m'`
2. The text `Title of the Program` should be yellow on black and blinking
3. Instead in Kitty it is not blinking

**Environment details**
`kitty 0.37.0 (2531c31d02) created by Kovid Goyal`

**Additional context**
Same behavior with `kitty --config NONE`.
Is it maybe I'm doing something wrong? 🤔

kovidgoyal:
kitty does not and will not support blinking text.

kovidgoyal:
already discussed search the closed issues.

yermulnik:
![image](https://github.com/user-attachments/assets/245e61df-3a89-46c3-9e49-85795272dceb)
334 issues, 14 pages — it's probably easier to update README and provide brief outline of why Kitty doesn't support standard shell feature.

kovidgoyal:
github search is shit google kitty blink issues github and you will find #540 as your first match

yermulnik:
Thanks for the pointer. Pity you've declined this feature. Kitty is great, though those a few missing features from common/traditional terminal apps (e.g. blink and bold) are the rough decision to consider switching to Kitty finally 🙁

glad2rest:
"I won't do a blink, viu support is a rare case, i won't do that, i won't do this blah-blah-blah overhead blah-blah-blah".
You're just lazy. Half measures, half product, half kitty.

kovidgoyal:
ROFL what a moron. Issue closed since it has started attracting fools.

PR #8551, opened 2025-04-20
[POC] Blinking text (SGR 5/25) in kitty

ctrlcctrlv:
Closes kovidgoyal/kitty#540.

This PR is not yet ready to be merged, and may never be depending on @kovidgoyal's feedback given his skepticism in the original issue (and my own, to be fair); critically, I have not made it optional yet.

I wrote this for my own personal use. Because most TE's ironically ignore SGR 5/25 due to its lack of support, I knew I could rely on it to really draw my attention in scripts, without giving reallt any other application the ability to blink as there's no check for it.

I wanted to just leave this patch in kovidgoyal/kitty#540. But, I couldn't comment, despite having contributed PR's in past. Kovid said over email to post this PR as an alternative.

https://github.com/user-attachments/assets/8d1bb7b9-c2bf-4f48-ab16-9b6a75860653


kovidgoyal:
See the discussion in https://github.com/kovidgoyal/kitty/discussions/8315

With regards to your implementation, I dont allow branches in kitty shaders. The way to implement this would be basically to use cursor_opacity to control the foreground opacity of text using mix() and step() in the shader. This suffers from the issue that blinking text will stop blinking when the cursor stops blinking, but it is the only reasonable way to implement this. 

Drawing attention to things in scripts is best done with reverse video not blink. Or even using the marker feature in kitty. 

I still dont see any actual use case for blink and given it cannot be implemented in a non-hacky way, I am not inclined to add support for it.

kovidgoyal:
On Mon, Apr 21, 2025 at 08:12:36AM -0700, Fredrick Brennan wrote:
> ctrlcctrlv left a comment (kovidgoyal/kitty#8551)
>
> Wouldn't this mean if the cursor blinking is disabled, text would not blink?

You would not actually use cursor_opacity, but instead do the same
calculation on CPU for blink_opacity. Disabling both blinks after
cursor_stop_blinking_after.

ctrlcctrlv:
Oh, I see your meaning now. I'll amend my patch.

zzhaolei:
Just a thought — would it be helpful to note in the blink-related issue (e.g. #540, #8030) that kitty already has support for this feature?

kovidgoyal:
Sure, feel free, its locked to contributors only so you should be able to post, if not let me know and I will do it.

zzhaolei:
> This conversation has been locked and limited to collaborators.

I’m unable to post a comment.

kovidgoyal:
Ok updated #540 myself

PR #8580, opened 2025-04-27
Add k-nine plotting tool to the integrations documentation

talwrii:
I made a small tool called k-nine (kitty-plotnine) to do quick-and-dirty plotting within kitty-compatible terminals with bash one-liners. 

See this [reddit post](https://www.reddit.com/r/KittyTerminal/comments/1k7u34z/knine_a_tool_for_plotting_in_the_shell_with_bash/). It was suggested that I add this tool to the kitty documentation so I am creating a pull request to do so. 

Thanks,
readwithai

### DIFF
--- a/docs/integrations.rst
+++ b/docs/integrations.rst
@@ -209,6 +209,20 @@ Add this to bashrc and then to plot a function, simply do:
 
     iplot 'sin(x*3)*exp(x*.2)'
 
+.. _tool_k-nine:
+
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Thanks, it needs to be a single line describing the tool and linking to
its homepage, that's the style for all other tools (except gnuplot and
that is because gnuplot doesnt support kitty itself out of the box and
has no documentation about how to do it).

talwrii:
Here is a shorter version.

Issue #1197, opened 2018-12-02
Mac OS: kitty doesn't restore


toonn:
Mac OS offers the option to resume all your open apps after a shutdown, Apple's version of hibernate, I guess. Terminal.app actually restores the ouput/history in the windows that were open when you shut the system down.

While that's nice, just restoring windows to the positions and sizes they had when shutting down would be plenty for me. Previous versions of kitty restored the positions, not the size. The current version (0.12.3) doesn't for me.

Note that I don't like the `remember_window_size` setting, I want new terminals to open at 80x24, it's just some long running terminals I'd like to restore to whatever size I made them.

On a similar note, I've been unable to change kitty windows's sizes using applescript, this was a nice QoL improvement with Terminal.app because the Mac OS window manager is so impotent.

kovidgoyal:
kitty has no code to restore window positions, so I am surprised it would ever have happened. As I recall, cocoa has some special APIs for saving/restoring application state. These are not wrapped by glfw (the toolkit kitty uses) so they would need to be implement in kitty/cocoa_window.m instead. This is too much effort for me personally, but patches are welcome.

toonn:
I assume this holds for applescript control as well?
How about simply opening at startup if it was open at shutdown? Does that also require implementation?

How hard/easy would it be to build kitty? Maybe I can git bisect to find where the behavior was lost?

kovidgoyal:
No idea about aplescript control. Isn't applescript going to be retired by apple? opening at startup is the same thing as far as i know. see https://sw.kovidgoyal.net/kitty/build.htmlfor how to build from source

pencilcheck:
It would be very nice to have this feature. :)

zoidbergwill:
After looking into this a little, I think [this](https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/CoreAppDesign/CoreAppDesign.html#//apple_ref/doc/uid/TP40010543-CH3-SW27) is the necessary related documentation.

[setFrameAutosaveName](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/WinPanel/Tasks/SavingWindowPosition.html) might also work to solve it too, apparently.

[Here](https://stackoverflow.com/questions/12779681/nswindow-restorable-not-always-working) is the related StackOverflow where I found them. It took a while to figure out the right thing to search for.

JunfengJia:
[...]

PR #8600, opened 2025-05-04
Retains window size when quit

dgshanee:
Solves some of #1197 

The issue was that the cache was being saved only when the window was marked for closing, but on macOS, when restarting or using cmd + q, it skips this process and jumps straight to terminating the application.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -1853,6 +1853,8 @@ class Boss:
             for qt in q:
                 windows += list(qt)
         active_window = self.active_window
+        active_window_size = get_os_window_size(active_window.id)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Not sure I follow. Boss.quit  doesn't quit straightaway, it sets the IMPERATIVE_CLOSE_REQUEST flag on the os window, which is checked in child-monitor.c. Any OSWindows that have that flag set are destroyed and the on_os_window_closed callback in boss.py is called, where the window size is saved. Furthermore starting kitty, resizing the window and then quitting with cmd+q and then restarting kitty restores the window size, as epected.

dgshanee:
@kovidgoyal you're right. Sorry, I just realized I wasn't solving anything

Issue #1119, opened 2018-11-03
Kitty doesn't remember position on macOS screen


iandol:
Hi, using Kitty 0.12.3 on macOS 10.14.1 and `remember_window_size` enabled, when Kitty first starts it does not remember its position on screen. As I also have `draw_minimal_border` enabled I have to drag edges to get the window where I want it. It would be nice if kitty could remember its position as well as its size on a restart.

Thanks for a great minimalist terminal!

kovidgoyal:
IMO window placement should be the responsibility of th eOS, not individual applications. For macOS I believe there are third party tools that allow you to save and restore window positions for individual applications, such as: https://www.macworld.com/article/2036191/mac-gems-stay-keeps-your-windows-exactly-where-you-want-them.html

iandol:
@kovidgoyal — thanks for the reply, all other apps on macOS that I'm aware of do have their position remembered, I assume because they all use the OS API. I'm not a mac developer so cannot tell you how other apps do this, this was all I found, and it is part of the Cocoa frameworks:

https://developer.apple.com/documentation/appkit/nswindow/1526255-restorable 
https://www.raywenderlich.com/613-windows-and-windowcontroller-tutorial-for-macos 

Kitty's behaviour is currently different to most other apps, and yes this is a minor issue that can be worked around with 3rd party window managers.

kovidgoyal:
The cocoa APIs are not really suitable, since kitty, being a terminal has special sizing requirements, its window size can either be remembered or depend on the font size, dpi and number of cells specified in its config.  It's probably possible to make ethis work with cocoa api, but I'm afraid its too much work, at least for me. Patches are welcome.

orthodoX:
Started experimenting with kitty recently and I'm amazed how light and fast it is. I know it's been a while since this issue was closed, I must agree with @iandol, it's a bit irritating to readjust the position and size of the window all the time or have a third party app only for managing kitty window 😞

rix0rrr:
+1, I would consider dropping kitty over something as trivial as this. I don't need lots of tiny needle pricks added to my daily routine, all day every day.

bogdancss:
any updates on this?

chrisgrieser:
I actually wanted to switch to kitty, and set up configs and a theme etc – but having to resize the window *every time* really is a deal breaker.

Luflosi:
[...]

PR #8601, opened 2025-05-04
Remembers window position after quit

dgshanee:
Adds a cache partition to remember the window position after quitting. Solves #1119

### DIFF
--- a/kitty/.dmypy.json
+++ b/kitty/.dmypy.json
@@ -1,1 +1,1 @@
-{"pid": 8237, "connection_name": "/var/folders/5f/y1sf62j518ndbk27w_bvy8b40000gn/T/tmpl469rc06/dmypy.sock"}
+{"pid": 1521, "connection_name": "/var/folders/5f/y1sf62j518ndbk27w_bvy8b40000gn/T/tmp3aw21ec7/dmypy.sock"}
--- a/kitty/.dmypy.json
+++ b/kitty/.dmypy.json
@@ -1,0 +1,1 @@
+{"pid": 8237, "connection_name": "/var/folders/5f/y1sf62j518ndbk27w_bvy8b40000gn/T/tmpl469rc06/dmypy.sock"}
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -86,6 +86,7 @@ from .fast_data_types import (
     get_options,
     get_os_window_size,
     get_os_window_pos,
+    glfw_primary_monitor_size,
     global_font_size,
     is_layer_shell_supported,
     last_focused_os_window_id,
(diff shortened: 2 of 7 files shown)


kovidgoyal:
It's not that simple, this can for example lead to the window being
positioned off screen if the screen resolution has changed. And it wont
work on tiling window managers, and Wayland and so on.

It definitely needs to be behind a default off option and if we are
implementing this then implement a --position command line flag for
kitty as well.

dgshanee:
@kovidgoyal it works for Rectangle for macos but I don't know about Wayland. Perhaps I could make the x and y values offsets of the screen resolution?

dgshanee:
@kovidgoyal I just tested it with different resolutions on macos and nothing appears off-screen. I think glfw accounts for off-screen positioning when making windows. Let me know.

kovidgoyal:
You need to store the monitor geometry and the window position. On Cocoa
and X11 the window co-ordinates are in the same co-ordinate space as the
monitor co-ordinates. Then the restoring position code needs to handle
the case when the current monitor layout has changed, glfw absolutely
does not do that, see the _glfwplatformcreatewindow function in
cocoa_window.m

This behavior needs to be made optional, defaulting to off, the default
may be changed in the future if it works well.

There needs to be a --position CLI flag added to kitty and all this
needs to be documented, noting that on Wayland it doesn't work at all
since applications are not allowed to set window positions.

dgshanee:
@kovidgoyal I just pushed a working version of what you described. Thanks!

kovidgoyal:
OK I'll review when I have a moment

kovidgoyal:
You need to store all available monitor geometries not just the primary one (the os window being closed could be on any monitor not necessarily the primary one), and then in the restore code check that the geometries are unchanged and only then restore, otherwise let the window manager decide where to place the window. Use glfwGetMonitors, glfwGetMonitorPos and glfwGetMonitorWorkarea and glfwgetmonitorphysicalsize. Store all that data in the cache and when restoring use it to decide whether to restore or abort or move and restore.

dgshanee:
Got it, thanks.

dgshanee:
@kovidgoyal It now only restores if the monitor workarea is the exact same. Also, <pyerrors.h> was throwing undefined function errors, so I fixed it by importing <Python.h> in glfw.c. Let me know!

dgshanee:
@kovidgoyal It seems my pull request changed some of the formatting of the code. I'll fix that right now

dgshanee:
@kovidgoyal Working version is here. Sorry about the confusion.

kovidgoyal:
Please fix the CI failures

dgshanee:
I think it was because of including Python.h in kitty/glfw.c

dgshanee:
@kovidgoyal should be working now

Issue #7700, opened 2024-08-05
Missing package in shell.nix (Go)


nomisreual:
**Describe the bug**
I attempted to build kitty from source following the documentation provided. *Go* seems to be missing as a build input in shell.nix.

**To Reproduce**
1. Dowloaded and extracted the latest release tarball.
2. Dropped myself in a nix shell with `nix-shell --pure`.
3. Attempted to build kitty with `make linux-package`.
4. Build failed, as *Go* was not available.

kovidgoyal:
I am afraid I dont use nix so the nix build script is best effort. Feel
free to send a PR to update it as needed.

PR #8621, opened 2025-05-10
fix: add missing pkgs and manually add font for nix shell

nomisreual:
I added three missing packages to the shell environment provided by the *Nix* shell (two of them are only needed on Linux). Also, I manually copy the needed font to *./fonts* using *shellHook*.

The provided *Nix Shell* now allows for successful builds (tested on x86_64-linux and x86_64-darwin).

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -15,6 +15,7 @@ in
           xxHash
           simde
           go_1_23
+          matplotlib
         ]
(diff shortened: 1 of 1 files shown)


nomisreual:
I once opened an issue regarding the nix shell: #7700
now I finally found time to work on it

PR #8639, opened 2025-05-15
panel kitten: on Wayland, use app-id for the LayerShellV1 `namespace` parameter

jinliu:
The Wayland LayerShellV1 protocol doesn't allow setting window class/name/role/type/appId. Instead, it has a `namespace` parameter that we currently hardcode to `kitty`. It seems suitable to use app-id for this parameter.

Example use case:

kwin use the `namespace` parameter and an undocumented mapping to set window type for Layer Shell windows. E.g., "dock" maps to NET:Dock. So if we set `app_id dock` in quick_access_terminal.conf, the panel would be considered a dock, and won't have the normal scale in/out animation on shown/hidden.


jinliu:
It might worth mentioning the kwin use case, but I'm unsure where to put it.

PR #8647, opened 2025-05-18
Fixed typos and file permissions

polluks:


kovidgoyal:
Python files should not be executable, so drop that commit.

kovidgoyal:
merged dropping the mods commit.

PR #8650, opened 2025-05-18
docs: panel: update niri compatibility information

alex-huff:
As of the latest niri release, more specifically https://github.com/YaLTeR/niri/commit/3e31c134a602dd876f2b446071e5802fa4a0c3ec, focus no longer gets stuck on an unmapped layer surface.

### DIFF
--- a/docs/kittens/panel.rst
+++ b/docs/kittens/panel.rst
@@ -160,8 +160,8 @@ Compatibility with various platforms
         🟠 **niri**
            Breaks when hiding (unmapping) layer shell windows. This means the quick
            access terminal is non-functional, but background and dock panels work.
-           More technically, keyboard focus gets stuck in the hidden window and when trying
-           to remap the hidden window niri never sends configure events for the remapped surface.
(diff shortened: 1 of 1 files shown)


arthsmn:
Also [this PR](https://github.com/YaLTeR/niri/pull/1592) might fix the problems.

YaLTeR:
> As of the latest niri release, more specifically https://github.com/YaLTeR/niri/commit/3e31c134a602dd876f2b446071e5802fa4a0c3ec, focus no longer gets stuck on an unmapped layer surface.

Good thing I stumbled upon this PR because I wasn't thinking about this edge case when doing that change; I'll add a comment in the niri code there

alex-huff:
> Good thing I stumbled upon this PR because I wasn't thinking about this edge case when doing that change; I'll add a comment in the niri code there

Yes, when I was reading the diff it did seem quite accidental 😆. Also, at first I was quite confused why it even worked because I didn't understand that the `?` operator in rust would magically return out of the function in the case that the optional value wasn't present.

PR #8750, opened 2025-06-23
Escape inputs in zsh integration's _ksi_preexec

jirassimok:
I noticed that the zsh integration sometimes emits errors or extraneous text when you start a command line with `-`, because `_ksi_preexec` treats those input lines as arguments.

```console
% -f(not-a-command)
(not-a-command)zsh: command not found: -fabc
% - sh -c 'echo $0'
_ksi_preexec:print:4: bad option: - 
-sh
```

I fixed it by adding the option-terminating `--`, added a test, and fixed a bug in the test function I used.


kovidgoyal:
thanks!

PR #8755, opened 2025-06-26
Add scroll_offset parameter to `scroll_to_prompt`

jcgl17:
This PR allows you to specify an offset when calling `scroll_to_prompt`, such that `n` previous lines from before the prompt are shown. This is somewhat like `less`'s `--jump-target` option or vim's `scrolloff` setting.

[Screencast_20250626_224818.webm](https://github.com/user-attachments/assets/75a0eed5-e2fa-4dd4-93f1-c319c8e91a79)

I find this helpful when repeatedly running the same command, especially when the beginning output of the command is the same but the end might be different.

Ansible playbook runs are prime example for me--pages and pages of same-ish output, with the interesting part (i.e. failures) at the bottom. Of late, I've been using [this little interactive task runner tool I've been building](https://codeberg.org/jcgl/wnl) for such repeated commands. It emits the prompt escape codes (which I learned of from Kitty!) to delimit consecutive executions. So with this PR, I'm able to quickly jump around and identify if I'm on the invocation I'm interested in.

### DIFF
--- a/kitty/fast_data_types.pyi
+++ b/kitty/fast_data_types.pyi
@@ -1305,7 +1305,7 @@ class Screen:
     def scroll_to_next_mark(self, mark: int = 0, backwards: bool = True) -> bool:
         pass
 
-    def scroll_to_prompt(self, num_of_prompts: int = -1) -> bool:
+    def scroll_to_prompt(self, num_of_prompts: int = -1, scroll_offset = 0) -> bool:
         pass
 
     def set_last_visited_prompt(self, visual_y: int = 0) -> bool:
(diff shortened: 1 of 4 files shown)


jcgl17:
Thank you for the quick merge and cleaning things up a little too!

Should I have added a line to the changelog as well? It hadn't occurred to me that that might be customary. I see now from the changelog's git history that others have added to it in the past.

kovidgoyal:
No need, I would have added it in my cleanup if needed. This is a
sufficiently small feature that it doesn't really warrant a changelog
entry.

PR #8765, opened 2025-06-29
feat: Improve Windows compatibility in setup.py

Mirza-Samad-Ahmed-Baig:
This PR introduces several changes to setup.py to improve its compatibility and behavior on Windows operating systems.

The primary goal is to allow the build script to run more gracefully on Windows, provide more informative error messages when Unix-specific tools are missing, and lay some groundwork for potential future Windows development.

Key changes include:

Windows platform detection: Added an is_windows flag for clearer platform-specific logic.

Robust run_tool function: Modified the run_tool function to correctly handle command execution on Windows by using shell=True and passing commands as single strings, which is often necessary for Windows-native commands and shell built-ins.

Improved pkg-config error handling: Enhanced the pkg_config function to provide a more specific and helpful error message on Windows if the pkg-config tool is not found, guiding users towards potential solutions like MSYS2 or WSL.

C compiler detection on Windows: Updated cc_version to prioritize cl.exe (MSVC) detection on Windows, providing a dummy version to allow the script to proceed and offering better guidance if no suitable compiler is found.

Conditional execution of macOS/Linux-specific tools: Wrapped calls to macOS-specific tools (iconutil, lipo) and Linux-specific bundle creation (create_linux_bundle_gunk) within if is_macos: or if not is_macos and not is_windows: blocks, preventing errors on unsupported platforms.

Conditional font detection: Modified add_builtin_fonts to skip fc-match on Windows, acknowledging that Windows-specific font detection would be required for full functionality.

Robust Git VCS revision retrieval: Improved get_vcs_rev to use shutil.which('git') for better Git executable discovery and provide a warning if Git is not found.

Refined clean function: Adjusted the clean function to conditionally remove .dylib files only on macOS and to run go clean only if the Go executable is available.

These changes aim to make the setup.py script more resilient and user-friendly for those attempting to build or interact with the project on Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ is_freebsd = 'freebsd' in _plat
 is_netbsd = 'netbsd' in _plat
 is_dragonflybsd = 'dragonfly' in _plat
 is_bsd = is_freebsd or is_netbsd or is_dragonflybsd or is_openbsd
+is_windows = sys.platform == 'win32'
 is_arm = platform.processor() == 'arm' or platform.machine() in ('arm64', 'aarch64')
 c_std = '' if is_openbsd else '-std=c11'
 Env = glfw.Env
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm somewhat unclear on the purpose of this. kitty doesnt work under windows at all, so why would one want to build it under Windows? Is this preliminary work to porting kitty to Windows?

Mirza-Samad-Ahmed-Baig:
Thank you for reviewing, sure let me share my thinking and point of view, it's for contributors. many open-source contributors use Windows, even just to inspect or audit code. This update helps avoid unnecessary roadblocks and reduces confusion when someone tries to build or test the project from that environment—even if only partially.

Issue #8806, opened 2025-07-11
make kitten ssh from new_X_with_cmd disconnection return to a shell


barr-israel:
**Is your feature request related to a problem? Please describe.**
Using `new_X_with_cmd`(any of the variations) is convenient in all cases except when using `kitten ssh`.  
When using it with `kitten ssh`, it creates another ssh shell, as expected, but closing the connection results in the entire window being closed, instead of just the ssh shell. My expectation was that I would go back to my local shell.

**Describe the solution you'd like**
I would like disconnecting from a `kitten ssh` created using `new_X_with_cmd`(or launch --cwd=current) to put me back on a system shell.

**Describe alternatives you've considered**
The two alternatives I found were:

- Have different sets of keybinds for these 2 scenarios(outside and inside `kitten ssh`)
- Disconnecting from ssh, creating the new window/tab and reconnecting to ssh
Neither are very convenient.

Thank you

kovidgoyal:
new_window_with_cwd is behaving as expected. If it is to open an new
window at the current dir, it *must* be via ssh otherwise opening at the
current dir is impossible. And when you open a shell in a new window,
quitting that window closes the shell. There is no expectation that the
window will not close because the shell is a remote one not a local one.
Indeed, the behavior you describe would be extremely counter-intuitive, IMO.

That said I am happy to accept a PR adding an option to have launch
--cwd=current (the underlying action of new_window_with_cwd) run a
local shell after the ssh kitten exits. It is a fairly simple PR, about
10 lines of code in total, feel free to ask for guidance if you need it.
Relevant code is line 165 onwards in window.py you simply need to
prepend argv with [kitten_exe(), "run-shell]. The rest is simply adding
an option to control that behavior, defaulting to off.

PR #8807, opened 2025-07-11
allow spawning a shell after kitten ssh exit first step

barr-israel:
I have successfully gotten the behavior I wanted in #8806 , but I am having trouble adding a flag for it, **this fork currently does not build** due to 
```
Cant map fields: ssh_spawn_shell for cmd: launch
Generating go code failed with exit code: 1
The following build command failed: /home/barr/Documents/code/kitty/dependencies/linux-amd64/bin/python setup.py develop
exit status 1
```
After adding `ssh_spawn_shell` to the launch.py `protocol_spec`.  
I have also added it to `clone_safe_opts` expecting it to be necessary.
Additionally, there is a little parameter drilling here from `launch` to `modify_argv_for_launch_with_cwd` that im not sure if I can avoid.


kovidgoyal:
You need to add it to options_spec() in kitty/launch.py and implement it in the _launch() function

kovidgoyal:
Once you are happy with your implementation please squash it down to one commit and force push, then I will review.

barr-israel:
it was already implemented in _launch when you made that comment, but i have now added it to options_spec and LaunchKwds and the build still fails, Ill keep investigating

kovidgoyal:
You need two blank lines before --ssh-spawn-shell

barr-israel:
Thanks, it seems to be working now.  
I dont see any options available on the `new_X_with_cwd` commands, so I guess I should just use `launch` and set the relevant type.  
Ill play around with it more to make sure nothing breaks.  
Are any new tests needed for this?

kovidgoyal:
No, this functionality is pretty simple, it shouldnt need tests. 

0) argv[:] = ... should be argv = to avoid an extra copy

1) I dont like the name ssh-spawn-shell as it is not clear what that means? It could mean spawn a shell on the remote machine. Instead something like, --hold-after-ssh this matches the usage in kitty of --hold

2) In the docs for the option instead of cwd=current use something like :option:`--cwd`=current so that it becomes a link and also add a similar link to this option in the docs for the --cwd option.

3) Rebase into a single commit

barr-israel:
0. at least one argv[:] is required because the caller does not get it back as a response but instead expects the value inside it to be changed(I've tried with only argv= and it does not work). Ive minimized it to one instead of two
1. done
2. done
3. done

barr-israel:
fixed a ruff lint error

kovidgoyal:
merged.

barr-israel:
Great, thanks for all the help

PR #8873, opened 2025-08-04
Kitten desktop UI portal xdg compat

azertyfun:
Hi! Trying to get the `desktop-ui` kitten to work I had a couple issues running `kitten desktop-ui enable-portal`:

1. On my machine `/etc/profile.d/flatpak.sh` sets `XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share` -> The `.local/share/flatpak/exports/share` directory was being used to create the portal configuration, which is incorrect. The [XDG Spec](https://specifications.freedesktop.org/basedir-spec/0.8/#referencing) mentions that `XDG_DATA_DIRS` should be applied _on top of_ `XDG_DATA_HOME`.  
   As a workaround I could run `XDG_DATA_DIRS=$HOME/.local/share:$XDG_DATA_DIRS kitten desktop-ui enable-portal` but that is obviously not spec-compliant.
2. After running into the first issue, I now had a `.local/share/flatpak/exports/share/xdg-desktop-portal/portals` directory. `enable_portal`'s current behavior is to first find a writable directory and write the portal definition to it, then fall back to creating a directory in the first available configuration directory. So even though `.local/share` had precedence it did not yet have a `xdg-desktop-portal` directory which meant that the flatpak one was used to create the portal definition. This MR changes the behavior to find-or-create.
3. Since I did not already have a `portals.conf`  file, an empty one was created. The second commit now makes sure that in the case where the portals file doesn't exist it is populated.


kovidgoyal:
On Mon, Aug 04, 2025 at 06:57:31AM -0700, Nathan Monfils wrote:
> Hi! Trying to get the `desktop-ui` kitten to work I had a couple issues running `kitten desktop-ui enable-portal`:
>
> 1. On my machine `/etc/profile.d/flatpak.sh` sets `XDG_DATA_DIRS=$HOME/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share` -&gt; The `.local/share/flatpak/exports/share` directory was being used to create the portal configuration, which is incorrect. The [XDG Spec](https://specifications.freedesktop.org/basedir-spec/0.8/#referencing) mentions that `XDG_DATA_DIRS` should be applied _on top of_ `XDG_DATA_HOME`.

Does it though? Quoting the spec:

```
$XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory. The directories in $XDG_DATA_DIRS should be seperated with a colon ':'.

The order of base directories denotes their importance; the first directory listed is the most important. When the same information is defined in multiple places the information defined relative to the more important base directory takes precedent. The base directory defined by $XDG_DATA_HOME is considered more important than any of the base directories defined by $XDG_DATA_DIRS.
```

Seems quite clear XDG_DATA_HOME takes precedence.
https://specifications.freedesktop.org/basedir-spec/0.8/#variables

The kitten will put its files into the most important directory, which
is XDG_DATA_HOME when it is defined. Otherwise, they would not take
effect.

azertyfun:
That's what I meant yes. Sorry, I got confused in my own troubleshooting. The original code took precedence into account properly **IF** XDG_DATA_HOME was defined, but in my case it's undefined. Now it coalesces to `$HOME/.local/share` properly. I'll amend the MR to clarify.

PR #8894, opened 2025-08-12
desktop-ui kitten: allow portal config for non-fullscreen panel

azertyfun:
Hi!

Me again with a UX PR this time.

I've got a 3440x1440 ultrawide monitor where `kitten panel --layer=overlay --edge=center` is not very usable on such a huge screen. Not only is it too much information over an area too large, but when scrolling in big folders it causes significant amounts of lag.

This allows me to keep the spawned kitten at a usable resolution:

```
# ~/.config/kitty/desktop-ui-portal.conf
panel_lines 50
panel_columns 160
```

There is one downside to this implementation though, if only one of `panel_lines` and `panel_columns` is set then the panel crashes. But IMO that's a separate issue to be fixed in the panel. I didn't want to enforce both options being set at the same time because it doesn't look to me like this is done elsewhere in kitty?


kovidgoyal:
But the panel is supposed to cover all other windows while the file
chooser is active, if we restrict its size, it will no longer do that.

Might be better to change the actual choose files kitten to optionally
have a max width/height and draw itself centered if those constraints
are provided. Will be more work to implement however.

azertyfun:
The behavior on my Hyprland at least is that the panel window is drawn smaller, but I can't "click out" out of it. That makes intuitive sense to me, especially as that's basically how the GTK and QT file portals already behave.

Is there a specific design philosophy I'm missing for why the panel should be drawn over other windows on the focused monitor?

kovidgoyal:
On Tue, Aug 12, 2025 at 05:44:15AM -0700, Nathan Monfils wrote:
> azertyfun left a comment (kovidgoyal/kitty#8894)
>
> The behavior on my Hyprland at least is that the panel window is drawn smaller, but I can't "click out" out of it. That makes intuitive sense to me, especially as that's basically how the GTK and QT file portals already behave.

That would be fine if that behavior is guaranteed under the layer shell
spec, I am not sure it is, however, see
https://wayland.app/protocols/wlr-layer-shell-unstable-v1

azertyfun:
What is the reason to require exclusive input? I just checked and neither the GTK nor QT portal request any kind of exclusive input. Here in firefox the file save window opens as just another non-floating window and even allows me to keep interacting with firefox while the file save window is open.

Is it to avoid a panel kitten being "forgotten", forcing the user to close it before continuing? But in that case why provide the `--edge=center-sized` option at all for panel? Or is input exclusivity a requirement specific to the desktop-ui kitten for some reason?

kovidgoyal:
On Tue, Aug 12, 2025 at 07:00:01AM -0700, Nathan Monfils wrote:
> azertyfun left a comment (kovidgoyal/kitty#8894)
>
> What is the reason to require exclusive input? I just checked and neither the GTK nor QT portal request any kind of exclusive input. Here in firefox the file save window opens as just another non-floating window and even allows me to keep interacting with firefox while the file save window is open.

Its not the exclusive input so much as modality. Either the file
selection dialog needs to be application modal or window modal, but that is not possible
here, so desktop modal is next best. And if firefox's file
save files window is not modal, that would be a bug in firefox.

PR #8916, opened 2025-08-20
always focus new os windows on macos

j-c-m:
I have been having an issue where sometimes a new OS windows start behind the foreground application on MacOS. I cannot reliably reproduce the issue. I am testing this change, which is hopefully in the correct spot, to see if it addresses the issue.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -460,6 +460,8 @@ class Boss:
             wclass = self.args.cls or appname
         tm = TabManager(os_window_id, self.args, wclass, wname, startup_session)
         self.os_window_map[os_window_id] = tm
+        if is_macos:
+            self.focus_os_window(os_window_id, False)
         return os_window_id
 
     def add_os_panel(self, cfg: LayerShellConfig, wclass: str | None = appname, wname: str | None = appname) -> int:


kovidgoyal:
That's very heavyweight, for example on Wayland it would cause a full roundtrip to the compositor to get an activation token. You should instead call just

```
if is_macos:
    focus_os_window(os_window_id, True)
```

However, more generally speaking I dont think this is the correct thing to do. What calling focus_os_window() does on macOS is end up calling _glfwPlatformFocusWindow() which in turn calls [NSApp activateIgnoringOtherApps] which brings the kitty application to the front "activating" it. I dont think the application should force itself to the front if the window manager has decided for whatever reason to not give the newly opened OS window focus. 

If you can instead isolate in what exact circumstances this happens, then that particular case could be addresses more narrowly, if needed.

j-c-m:
Admittedly heavy-handed.

A few reasons I went this way first, I tried working around the issue with a startup_session containing "focus_os_window", which if I traced correctly eventually calls focus_os_window(id, True). I didn't realize this only applied to the first launch in a single instance situation, this is initially why I made the change False (if_needed_only) to always force it.
I couldn't find measurable difference in startup time on macos with this either True or False, then when researching how I could reproduce it I found [39ea727](https://github.com/alacritty/alacritty/commit/39ea7271e32ad88280191d8040d6f8feafe4307a) , but no reasoning which made me comfortable in thinking some behavior may have changed in macos.

--

Going to try and isolate it to a reproducible scenario, I also will try running normally with:
```
if is_macos:
    focus_os_window(os_window_id, True)
```

I typically start, or open a new kitty window via hotkey with skhd.
```
ctrl + cmd - t : /opt/homebrew/bin/kitty -1 -d ${HOME}
```

The issue has happened both with a browser (chrome & edge) and vscode in the foreground where at some point all new kitty os windows just continue to open behind them, until I exit the kitty instance entirely.

kovidgoyal:
It probably is a macOS bug (Lord knows there are a ton of those), but I still feel we need to identify the exact scope of the bug and limit the workaround to that scope rather than doing it unconditionally, which could have unintended side effects.

j-c-m:
I can somewhat reproduce it now.

* I open kitty connect to a remote server I monitor and minimze it.
* I start terminals (single instance, kitty -1) throughout the day, at some point they just stop taking focus.
* If I restore the minimized kitty os window, and then minimize it again, new os windows are focused again.

I am going to run it patched now with
```
if is_macos:
    focus_os_window(os_window_id, True)
```

and see if any change.

j-c-m:
Still persists with
```
if is_macos:
    focus_os_window(os_window_id, True)
```

trying
```
if is_macos:
    focus_os_window(os_window_id, False)
```

j-c-m:
It doesn't seem to make much of a difference, even without any changes _glfwPlatformFocusWindow is still being called when a new os window is created. For some reason macos isn't allowing it to take position.

glfw has updated focus window, wrapping it in autoreleasepool, I am doubtful this will make any difference either.

```
 @autoreleasepool {
    // Make us the active application
    // HACK: This is here to prevent applications using only hidden windows from
    //       being activated, but should probably not be done every time any
    //       window is shown
    [NSApp activateIgnoringOtherApps:YES];
    [window->ns.object makeKeyAndOrderFront:nil];
    } // autoreleasepool
   ```

**Comment by kovidgoyal** [comment]
Must be part of some random cleanup, there is nothing to release in the
function, so no point in using an autoreleasepool. In any case in
kitty's glfw fork IIRC each event loop tick is within an
autoreleasepool.

Try changing canBecomeKeyWindow() in glfw/cocoa_window.m to always
return YES.

**Comment by j-c-m** [comment]
canBecomeKeyWindow is YES when it happens... 

```
[GLFW Debug] Final state: isKeyWindow: NO, isActive: NO, isOnActiveSpace: YES, canBecomeKeyWindow: YES
```

**Comment by j-c-m** [comment]
turns out I was never actually forcing the call because in state.c:902
```python
PYWRAP1(focus_os_window) {
    id_type os_window_id;
    int also_raise = 1;
    const char *activation_token = NULL;
    PA("K|pz", &os_window_id, &also_raise, &activation_token);
    WITH_OS_WINDOW(os_window_id)
        if (!os_window->is_focused || (activation_token && activation_token[0])) focus_os_window(os_window, also_raise, activation_token);
        Py_RETURN_TRUE;
    END_WITH_OS_WINDOW
    Py_RETURN_FALSE;
}
```

there is another os_window->is_focused check.

j-c-m:
I am now 99% sure this was a "design choice" (bug) introduced in Sonoma. I will close this.

https://github.com/koekeishiya/yabai/issues/1858

kovidgoyal:
Hmm, I guess apple decided non active apps can no longer become active
on their own. Similar to "focus stealing" prevention in Linux land.
Paternalistic protection of users from their software.

This issue made me realize I had broken switching focus to minimized
windows while fixing another bug, I dont think this will affect your
issue, but anyway: https://github.com/kovidgoyal/kitty/commit/ac34c2a2aa8fa8c95ea86831e1b027ce8f64a74b

PR #8920, opened 2025-08-22
Fix the background color rendering

zzhaolei:
Currently, when background_opacity < 1.0 and macos_traditional_fullscreen = no, switching to fullscreen with Cmd+Ctrl+F causes the window’s background color to differ from the specified background.
After the fix, the behavior with macos_traditional_fullscreen = yes remains unaffected.


zzhaolei:
@kovidgoyal This is ready to be reviewed.

PR #8943, opened 2025-09-02
Add branch drawing symbols for compact graphs

rbong:
## Summary

This includes 10 new branch drawing symbols for [drawing compact graphs](https://github.com/rbong/vim-flog/issues/163).
Commit drawing code has been slightly updated to allow for slightly more complex lines.

Here is a screenshot of how commit symbols appear with the new branch drawing code and new symbols:

new_flog_commit_symbols

## Consistency

The order is consistent with other branch drawing and box drawing symbols.
I've tried to make the new code consistent with other box drawing functions.

These symbols are a logical extension of branch drawing symbols and are exactly where we would want to put them in the symbol table.
They're just commits with 3/4 way connections instead of just 1/2 way connections, which is all we had before.

## Future-proofing

The only reason these weren't included before was that I believe no one asked for them even after I reached out for feedback, and I couldn't think of a use case myself.
In retrospect, it would have been better to include them before for future-proofing, but I think it's fairly low-impact to include them now as no one else has extended branch drawing symbols yet.
The [use case for adding them makes sense to me](https://github.com/rbong/vim-flog/issues/163).

There's no other combinations of symbols we could reasonably add: fade-out symbols don't make sense combined with anything else, curved lines would just be obscured by commit symbols if we combined them, every other combination exists.

I say this so I hope it's obvious I won't just keep adding symbols.
This should be the last updates unless if someone comes up with a good case for something essentially new, and we can weigh the number of new symbols against the usefulness.

## Next steps

I'm submitting it here first because this was the first terminal to accept a PR for these symbols.

I plan to reach out to other terminals that have already implemented these symbols to let them know about this development.
If you'd like me to do that first, let me know.

In the interest of more standardization, I think it's also time to submit these symbols to Nerd Fonts.
I'm not confident in my ability to create these as font symbols, but since no one else has done it yet I will give it a shot.
Again, if you'd like to see this happen first, let me know.

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -608,7 +608,7 @@ START_ALLOW_CASE_RANGE
         case 0xe0b0 ... 0xe0bf: case 0xe0d6 ... 0xe0d7:    // powerline box drawing
         case 0xee00 ... 0xee0b:    // fira code progress bar/spinner
         case 0x1fb00 ... 0x1fbae:  // symbols for legacy computing
-        case 0xf5d0 ... 0xf5fb:    // branch drawing characters
+        case 0xf5d0 ... 0xf605:    // branch drawing characters
             return BOX_FONT;
         default:
             *is_emoji_presentation = has_emoji_presentation(cpu_cell, gpu_cell);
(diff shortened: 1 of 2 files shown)


rbong:
My mistake, the symbols were already added in the last PR. I didn't fetch from upstream before checking. Apologies.

PR #8945, opened 2025-09-03
Add interactive scrollbar with full mouse support

lxe:
Replace the minimal scroll position indicator with a full-featured interactive scrollbar that provides mouse-based navigation for scrollback history.

https://github.com/user-attachments/assets/d5412e47-07cf-41a1-9e33-a6f301847675

Features:
- Full-height scrollbar positioned flush with the right window edge (independent of the row/cols coordinates)
- Configurable autohide to match current scrollback indicator behavior
- Two click behaviors: Click track to jump to position OR scroll by page (configurable)
- Configurable everything: color, track color, border radius, width, autohide, and even revert to non-interactive mode!
- Good HCI patterns: expanded (and configurable) hitbox around thumb for easier grabbing, 

Configuration options:
  - scrollbar_color: Controls the color. Default is `foreground`
  - scrollbar_opacity: Controls thumb opacity (default: 0.5)
  - scrollbar_track_opacity: Controls track background opacity (default: 0)
  - scrollbar_interactive: Enable/disable mouse interaction (default: yes)
  - scrollbar_gap: Spacing from edges in pixels (default: 5)
  - scrollbar_width: Width in pixels (default: 10)
  - scrollbar_min_thumb_height: Minimum thumb height in pixels (default: 50)
  - scrollbar_hitbox_expansion: Extra pixels for easier grabbing (default: 5)
  - scrollbar_autohide: Hide scrollbar when at bottom unless hovering (default: yes)
  - scrollbar_track_behavior: 'jump' to position or 'page' scroll (default: jump)

### DIFF
--- a/kitty/fast_data_types.pyi
+++ b/kitty/fast_data_types.pyi
@@ -1387,7 +1387,8 @@ def set_tab_bar_render_data(
 
 def set_window_render_data(
     os_window_id: int, tab_id: int, window_id: int, screen: Screen,
-    left: int, top: int, right: int, bottom: int
+    left: int, top: int, right: int, bottom: int,
+    spaces_left: int, spaces_top: int, spaces_right: int, spaces_bottom: int
 ) -> None:
     pass
 
--- a/kitty/shaders.c
+++ b/kitty/shaders.c
@@ -10,9 +10,11 @@
 #include "cleanup.h"
 #include "colors.h"
 #include <stddef.h>
+#include <string.h>
 #include "window_logo.h"
 #include "srgb_gamma.h"
 #include "uniforms_generated.h"
+#include "state.h"
 
 enum {
     CELL_PROGRAM, CELL_FG_PROGRAM, CELL_BG_PROGRAM, CELL_PROGRAM_SENTINEL,
@@ -739,22 +741,9 @@ draw_visual_bell(const UIRenderData *ui) {
 
 static bool
 has_scrollbar(Screen *screen) {
-    return OPT(scrollback_indicator_opacity) > 0 && screen->linebuf == screen->main_linebuf && screen->scrolled_by;
+    return OPT(scrollbar_opacity) > 0 && screen->linebuf == screen->main_linebuf && screen->historybuf->count > 0;
 }
 
-static bool
-draw_scroll_indicator(color_type bar_color, GLfloat alpha, float frac, const UIRenderData *ui) {
-    bind_program(TINT_PROGRAM);
-#define C(shift) srgb_color((bar_color >> shift) & 0xFF) * alpha
-    glUniform4f(tint_program_layout.uniforms.tint_color, C(16), C(8), C(0), alpha);
-#undef C
-    float bar_width = 0.5f * gl_size(ui->cell_width, ui->screen_width);
-    float bar_height = gl_size(ui->cell_height, ui->screen_height);
-    float bottom = -1.f + MAX(0, 2.f - bar_height) * frac;
-    glUniform4f(tint_program_layout.uniforms.edges, 1.f - bar_width, bottom + bar_height, 1.f, bottom);
-    draw_quad(true, 0);
-    return true;
-}
 
 static unsigned
 render_a_bar(const UIRenderData *ui, WindowBarData *bar, PyObject *title, bool along_bottom) {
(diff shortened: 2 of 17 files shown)


kitty/fast_data_types.pyi
```diff
@@ -1387,7 +1387,8 @@ def set_tab_bar_render_data(
 
 def set_window_render_data(
     os_window_id: int, tab_id: int, window_id: int, screen: Screen,
-    left: int, top: int, right: int, bottom: int
+    left: int, top: int, right: int, bottom: int,
+    spaces_left: int, spaces_top: int, spaces_right: int, spaces_bottom: int
```
lxe, line 1391: I needed this, otherwise the viewport geometry will be based on quantized rows/cols

kitty/shaders.c
```diff
@@ -880,9 +869,67 @@ static void
 draw_scrollbar(const UIRenderData *ui) {
     if (!has_scrollbar(ui->screen)) return;
     Screen *screen = ui->screen;
+    Window *window = ui->window;
+    if (!window) return;
+    
     color_type bar_color = colorprofile_to_color(screen->color_profile, screen->color_profile->overridden.highlight_bg, screen->color_profile->configured.highlight_bg).rgb;
     float bar_frac = (float)screen->scrolled_by / (float)screen->historybuf->count;
-    draw_scroll_indicator(bar_color, OPT(scrollback_indicator_opacity), bar_frac, ui);
```
lxe, line 885: I guess I can retain this name

lxe:
Addresses #7599

kovidgoyal:
IMO it should be an overlay rather than taking up width. Like macOS
scrollbars. And it should work in the following way:

1) Be invisible normally
2) When the user starts scrolling back the scroll indicator should
become visible, as it does now.
3) When the mouse hovers near the right edge of the window the full
scrollbar should become visible and respond to mouse clicks etc.
4) Clicking in the track should not move the indicator there but rather
scroll by one page (one screenful) (this can be an option)
5) The full scrollbar should render wider than the scroll indicator,
maybe one cell wide.

That said I am not opposed to having an option for the behavior in your
PR, but the default needs to be as described above.

Good work! Thanks.

lxe:
Thanks @kovidgoyal 

All great ideas. I specifically didn't implement the fade behavior to keep it simple (performant?), but since you're not opposed to it, I'll give it a shot. 

I'll adjust the click to jump behavior and make it configurable as well. Let me see what I can do.

lxe:
I've implemented most of  the @kovidgoyal 's suggestions:

>     1. Be invisible normally
>     2. When the user starts scrolling back the scroll indicator should
>        become visible, as it does now.
>     3. When the mouse hovers near the right edge of the window the full
>        scrollbar should become visible and respond to mouse clicks etc.

Added `scrollbar_autohide` config option

>     4. Clicking in the track should not move the indicator there but rather
>        scroll by one page (one screenful) (this can be an option)

Added `scrollbar_track_behavior` config option

>     5. The full scrollbar should render wider than the scroll indicator,
>        maybe one cell wide.

This is actually handled via scrollbar_width, scrollbar_gap and the hitbox parameters -- independently from the quantized row/cols coordinates. This way the bar is hugging the pane/window instead of aligning with a row/column

I also added a very rudimentary test.

kovidgoyal:
OK I'll review the actual code when I have some time, unfortunately am
rather swamped at the moment. Hopefully will get done in the next few
days.

lxe:
Thanks, I'll be cleaning this thing up and running benchmarks too

lxe:
Demo of the config and behavior:


https://github.com/user-attachments/assets/23ae69b8-b479-449d-aecc-c1f6425a7518

kitty/shaders.c
```diff
@@ -880,9 +869,69 @@ static void
 draw_scrollbar(const UIRenderData *ui) {
     if (!has_scrollbar(ui->screen)) return;
     Screen *screen = ui->screen;
+    Window *window = ui->window;
+    if (!window) return;
+
     color_type bar_color = colorprofile_to_color(screen->color_profile, screen->color_profile->overridden.highlight_bg, screen->color_profile->configured.highlight_bg).rgb;
     float bar_frac = (float)screen->scrolled_by / (float)screen->historybuf->count;
-    draw_scroll_indicator(bar_color, OPT(scrollback_indicator_opacity), bar_frac, ui);
```
lxe, line 885: TODO: backwards compat

lxe:
Mac OS example:

https://github.com/user-attachments/assets/40e07d32-f3b9-4634-b6a4-31fde54f34a2

lxe:
I think I'm going to need some help profiling memory usage and rendering performance of this

kovidgoyal:
I will look into those when I am reviewing. In principle there should be
no significant impact as extra rendering is just a couple of draw calls
over a small area of the framebuffer. The fade animation (if you
implemented it) might cause some minor loading but its transient and on
par with the cursor blink animation in principle.

I don't see why it should have any memory impact at all, but I will need
to look at your implementation to check. Unfortunately I am travelling
for a few days, so...

lxe:
>  The fade animation (if you
implemented it) might cause some minor loading but its transient and on
par with the cursor blink animation in principle.

I actually removed the fade animation to keep the shader code relatively simple

kovidgoyal:
On Mon, Sep 08, 2025 at 11:24:51AM -0700, Aleksey Smolenchuk wrote:
> lxe left a comment (kovidgoyal/kitty#8945)
>
> >  The fade animation (if you
> implemented it) might cause some minor loading but its transient and on
> par with the cursor blink animation in principle.
>
> I actually removed the fade animation to keep the shader code relatively simple

Doesnt fade just require a single opacity uniform in the shader to implement?

lxe:
To have a autohide, yes. In fact I don't even draw it if it's hidden or auto-hidden due to 0 scrollback: https://github.com/kovidgoyal/kitty/pull/8945/files#diff-0dad3090fe3ad3d31d01ba687ffea0cc81029a25bd08dd186d1349c2537eba6fR897

I have an opacity uniform for drawing it. For smooth transition I needed to track a timer and I didn't like how it turned out. I can revisit smooth fading later when I have the time.

kovidgoyal:
I have looked over the code briefly and it looks pretty good. I have
merged into a "scrollbar" branch. I will work on making some minor
changes/fixes and possibly redoing the option settings into a single
option to control the scrollbar that takes key/value pairs
rather than having so many individual options.

One minor issue I noticed is that the code to specify colors only has
special colors of background and foreground, but the docs say any
configured color variable is supported. I dont think using any
configured color variable is needed. Maybe have only a small set
background/foreground/highlight_bg/highlight_fg is good enough. Also the
track color should probably be configurable as well.

Another thing I think should change is the dimensions of the scrollbar
(width/min handle height etc) should be specified in units of cells
rather than pixels. The cell is a natural dimension associated with the
terminal and using it means the scrollbar size will scale with the font
size, which is nice. If you have some strong reason to prefer pixels,
let me know and I will make the dimensions support pts, pixels and cells
as units.

I will work on these in the scrollbar branch and merge into master when
I feel its ready.

Thanks for this PR!

lxe:
> should be specified in units of cells
rather than pixels. The cell is a natural dimension associated with the
terminal and using it means the scrollbar size will scale with the font
size, which is nice. 

I like this as long as we allow for certain fractional sizing. Wonder if this makes calcs harder or MORE natural.

kovidgoyal:
On Mon, Sep 15, 2025 at 03:00:46PM -0700, Aleksey Smolenchuk wrote:
> lxe left a comment (kovidgoyal/kitty#8945)
>
> > should be specified in units of cells
> rather than pixels. The cell is a natural dimension associated with the
> terminal and using it means the scrollbar size will scale with the font
> size, which is nice.
>
> I like this as long as we allow for certain fractional sizing. Wonder if this makes calcs harder or MORE natural.

Yes, of course fractional sizes are allowed. The settings are all float.

PR #8959, opened 2025-09-06
fix: when parsing `python_string`s options, don't fail if `'` is last character

poliorcetics:
Previously, if the last character was `'` parsing would fail: from `abc'` it would produce the literal `'''abc''''`, which has one too many unescaped single quote at the end.

This also fixes the issue for solo `'''`, where before it would produce `''''\\'''''`, again with one too many single quote at the end.

I added tests for both cases.


poliorcetics:
@kovidgoyal should I have added a changelog entry ? I didn't see it in the contributing guide but I can make a follow up MR if necessary

kovidgoyal:
No need, this is an extreme edge case. This bug has existed for about a
decade and no one has run into it in all that time.

poliorcetics:
Yeah, it's certainly extremely small, just wanted to be sure, thanks!

Issue #8533, opened 2025-04-12
[RFC] Specifying how terminals process Unicode text


kovidgoyal:
Hi all,

Following on from the text-sizing work in #8226 I have decided to specify the exact algorithm terminals should use to split Unicode text into cells and implement it in kitty. It is based on the Unicode specification's Grapheme segmentation rules but in addition also specifies things particular to terminals not covered in the Unicode spec. It fixes various long standing issues such as #3810 (emoji with zwj) and #8433 (Korean text).

The specification is [here](https://sw.kovidgoyal.net/kitty/text-sizing-protocol/#the-algorithm-for-splitting-text-into-cells). Feel free to read it and comment. There might well be differing opinions on the parts not covered by the Unicode spec. I am open to suggestions for modification.

From kitty users, I would appreciate if some of you can run [nightly](https://sw.kovidgoyal.net/kitty/binary/#customizing-the-installation) and report if there are any issues. It's possible you will have issues if you use ZWJ based emoji in your workflows, as the width kitty assigns to these has changed, and terminal programs may use a different width than the correct one.

In master, there is also a kitten that can be run easily to test a terminal's compliance with the spec. It uses grapheme test data from the Unicode consortium. Run it as:

```
kitten __width_test__
```

Here are results of running it on various terminals.

| Terminal name        | Number of tests failed |
| -------------------- | ---------------------- |
| kitty (master)       | 0                      |
| kitty 0.41.1         | 45                     |
| wezterm 5046fc22     | 179                    |
| foot 1.21.0          | 186                    |
| konsole 24.12.3      | 280                    |
| iTerm2 3.5.13        | 289                    |
| gnome-term 3.56.0    | 317                    |
| kitty-master+tmux-3.5| 347                    |
| xterm 397            | 371                    |
| Apple terminal 2.14  | 479                    |

And finally, we have ghostty 1.1.3, on which the test kitten failed to run because ghostty returned way more cursor position reports than were expected, something badly broken there. I did happen to look at its code as it claims to do grapheme segmentation, and it [doesn't implement the segmentation algorithm correctly anyway](https://github.com/ghostty-org/ghostty/discussions/6931).

kovidgoyal:
Closing since there were zero comments in a month and kitty 0.42.0 has been released with the new code anyway. If anyone points out serious concerns with the spec I am still open to revisions, but the bar is now higher.

Wukuyon:
I think the proposal looks great. I have one general comment.

[...]

PR #8971, opened 2025-09-14
Text sizing protocol docs: specify UTF-8 decoding, use unambiguous Unicode terminology

Wukuyon:
This commit makes the following changes to the text sizing protocol's docs, to describe Kitty's current Unicode behavior more precisely in the text sizing protocol docs (#8533).

These changes describe **behavior that Kitty *already* has**. It does not require any changes to any actual Kitty source code. (I don't know if Foot currently matches this behavior, but it probably does.)

1. The algorithm for splitting text into cells now uses the phrases "code point" or "Unicode scalar value" instead of "Unicode character", which is an ambiguous phrase.

2. The algorithm now requires that the terminal uses UTF-8 to decode the bytes it receives into Unicode scalar values.

3. The algorithm now requires that the terminal replace each [maximal subpart of any ill-formed subsequence](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G66453) with U+FFFD REPLACEMENT CHARACTER (�).

<details>
<summary>How to verify that Kitty decodes UTF-8 by replacing maximal subparts of ill-formed subsequences</summary>

Kitty's use of the "maximal subpart" approach can be verified by running the following shell commands (requires `python3` and `node`):

```fish
node --eval "['C0 AF E0 80 BF F0 81 82 41', 'ED A0 80 ED BF BF ED AF 41', 'F4 91 92 93 FF 41 80 BF 42', 'E1 80 E2 F0 91 92 F1 BF 41'].map(h => new TextDecoder('utf-8').decode(Uint8Array.from(h.split(' ').map(b => parseInt(b, 16))))).forEach(s => console.log(s));"

python3 -c "import sys; [sys.stdout.buffer.write(bytes.fromhex(h) + b'\n') for h in ['C0 AF E0 80 BF F0 81 82 41', 'ED A0 80 ED BF BF ED AF 41', 'F4 91 92 93 FF 41 80 BF 42', 'E1 80 E2 F0 91 92 F1 BF 41']]"
```

Kitty displays the Python command's STDOUT bytes with the same reference results as the Node command (which uses the WHATWG Encoding Standard). If Kitty did not use the "maximal subpart" approach, it would have replaced each *entire* ill-formed subsequence with a *single* �. For example, `C0 AF` would have been replaced by only one �, rather than ��. But Kitty does use the "maximal subpart" approach, and it decodes `C0 AF` as ��. (Matching the WHATWG Encoding Standard is a good thing. Almost all modern Unicode implementations match it too.)

</details>

### DIFF
--- a/docs/text-sizing-protocol.rst
+++ b/docs/text-sizing-protocol.rst
@@ -126,7 +126,7 @@ and trailers)::
 
 Note, in particular, how the last character, the cat emoji, ``🐈`` has ``w=2``.
 In practice client applications can assume that terminal emulators get the
-width of all ASCII characters correct and use the ``w=0`` form for efficient
+width of all ASCII code points correct and use the ``w=0`` form for efficient
 transmission, so that the above becomes::
 
    cool- w=2:🐈
@@ -369,16 +369,24 @@ a width one unit in the character grid the terminal displays.
 The basis for the algorithm is the
 `Grapheme segmentation algorithm <https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries>`__
 from the Unicode standard. However, that algorithm alone is insufficient to
-fully specify text handling for terminals. The full algorithm is specified
-below. When a terminal receives a Unicode character:
+fully specify text handling for terminals. The full algorithm is specified below.
 
-#. First check if the character is an ASCII control code, and handle it
-   appropriately. ASCII control codes are the characters less than 32 and the
-   character 127 (DEL). The NUL character (0) must be discarded.
+A terminal using this algorithm must decode the bytes they receive
+into Unicode scalar values (i.e., code points except surrogates) using UTF-8.
+When it encounters any UTF-8 ill-formed subsequences,
+it must be replace each
+`maximal subpart of the ill-formed subsequence <https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G66453>`__
+with a U+FFFD REPLACEMENT CHARACTER (�).
 
-#. Next, check if the character is *invalid*, and if it is, discard it
-   and finish processing. Invalid characters are characters with Unicode category :code:`Cc or Cs`
-   and 66 additional characters: :code:`[0xfdd0, 0xfdef]`, :code:`[0xfffe, 0x10ffff-1, 0x10000]`
+For each decoded code point:
+
+#. First check if the code point is an ASCII control code, and handle it
+   appropriately. ASCII control codes are the code points less than U+0032 and the
+   code point U+0127 (DEL). The NUL code point (0) must be discarded.
+
+#. Next, check if the code point is *invalid*, and if it is, discard it
+   and finish processing. Invalid code points are code points with Unicode category :code:`Cc or Cs`
+   and 66 additional code points: :code:`[0xfdd0, 0xfdef]`, :code:`[0xfffe, 0x10ffff-1, 0x10000]`
    and :code:`[0xffff, 0x10ffff, 0x10000]`.
 
 #. Next, check if there is a previous cell before the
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Add some tests for maximal subpart decoding in the test_utf8_parsing function in kitty_tests/parser.py

Dont know if the existing tests already cover it, they probably do, but no harm in being explicit.

Wukuyon:
I have added more tests to this PR for UTF-8 replacement of maximal subparts.

I discovered one case that Kitty (master branch) currently does not pass:
```python
# Not added to this PR yet because it fails
pb(b'"\xff\x80\x80\x80"', '"\ufffd\ufffd\ufffd\ufffd"')
```

* Consider a byte sequence that would have been a [valid 4-byte UTF-8 sequence](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G27506), except its first byte exceeds `0xF4`.
* This invalid byte sequence is basically trying to encode a code point beyond the maximum U+10FFFF.
* Kitty should replace this byte sequence with `����`.
* Instead, Kitty replaces this byte sequence with an *empty string*.
* This is inconsistent with how Kitty treats 1-, 2-, 3-, and 5-byte sequences that start with bytes that exceed `0xF4`. Kitty correctly replaces those with replacement characters. It treats only 4-byte sequences differently.

```shell
# Kitty prints "�\n"
printf "\xff\n"

# Kitty prints "��\n"
printf "\xff\x80\n"

# Kitty prints "���\n"
printf "\xff\x80\x80\n"

# Kitty should print "����\n", incorrectly prints only "\n"
printf "\xff\x80\x80\x80\n"

# Kitty prints "�����\n"
printf "\xff\x80\x80\x80\x80\n"
```

I believe this should be considered a Kitty bug. If that sounds correct to you, should I try to fix the bug in this PR, or should I try to fix it in a separate PR?

kovidgoyal:
Separate PR would be good. Relevant code is in simd-string-impl.h
utf8_decode_to_esc()

Note that UTF-8 decoding is highly performance sensitive. You can run
benchmarks using

`kitten __benchmark__`

and also ./benchmark.py

kovidgoyal:
There are a few more of your tests that fail, I have commented them out pending fixes in the parser. Assuming a fix is feasible without performance regressions.

Wukuyon:
**Edit**: Sorry, I didn't see your second reply until after I sent this one.

Thanks. I'll work next on that bug, when I can.

It looks like the [CI testing started to fail after you merged this PR](https://github.com/kovidgoyal/kitty/actions/runs/17720631454/job/50352114463). The following assertion is failing:
```python
# Overlong 2-byte sequence for U+0000 (should be `0x00`)
pb(b'"\xc0\x80"', '"\ufffd\ufffd"')
```

I must have missed this before I pushed the tests. Sorry about that.

I can confirm that running `printf "\xc0\x80\x80\n"` in Kitty, master branch, incorrectly prints only a newline.

This is almost certainly the same bug that affects `"\xff\x80\x80\x80"`. It has the same pattern:

```bash
# Kitty prints "�\n"
printf "\xc0\n"

# Kitty prints "��\n"
printf "\xc0\x80\n"

# Kitty should print "���\n", incorrectly prints only "\n"
printf "\xc0\x80\x80\n"

# Kitty prints "����\n"
printf "\xff\x80\x80\x80\n"
```

Sorry about making CI fail on the master branch. How would you like to handle the failing assertion? I can make a new PR to remove it temporarily, if you would like.

kovidgoyal:
I've already commented out the failing tests, pending a fix, see above. I have also merged the utf8_parsing test with simd_decode which is more robust as it tests against all three SIMD modes when the hardware supports it.

Issue #8985, opened 2025-09-18
Hint Label Collision


edralston:
**Describe the bug**

When the number of hint labels exceeds the size of the hint alphabet, some of the labels must be expanded to two or more characters.  However, as currently implemented, it is possible for some longer hint labels to
start with the same text as a shorter label.  This makes it such that it becomes impossible to select some of the shorter hint labels.

For example, if the hint label choices are "j" "k" "ja" "jb" and "jc", then it is not possible for the user to select the mark associated with "j" because the hint engine will still be trying to match against "ja" or "jb" etc.

**To Reproduce**
1. Map a key in `kitty.conf` to a hint that has a small alphabet but potentially a large number of hint matches:
```
map ctrl+shift+w   kitten hints --type=regex --regex '[^-=+:\s<>\[\]()/│]{4,}' --alphabet 'jkl' --program @
```
2. Launch kitty, then make sure there is a lot of text on the screen, then invoke the above hint 
3. Observe that some of the longer hint labels higher on the screen (say, `ljjlk` and `ljjkk`) start with the same text as a shorter hint label near the bottom of the screen (say, `ljj`).
4. Attempting to select the shorter label `ljj` is not possible because the kitten is still attempting to determine whether the user wants to select one of the still-matching longer labels.

Note that the short alphabet and the eager regex above are just to help reproduce this issue.  I still often run into this issue even with a longer alphabet when there is a lot of text on the screen and hence many more opportunities for hint pattern matches.

**Screenshots**

<img width="851" height="564" alt="Image" src="https://github.com/user-attachments/assets/c46cc582-a4ca-4d29-a184-48ade1290fb9" />

**Environment details**
```
kitty 0.42.2 (c30f8cec46) created by Kovid Goyal
Linux spike 6.12.43+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.43-1 (2025-08-27) x86_64
Debian GNU/Linux 13 spike /dev/tty
Running under: Wayland (sway version 1.10.1) missing: blur icon bell window-tag window-state-suspended color-manager
OpenGL: '4.6 (Core Profile) Mesa 25.0.7-2' Detected version: 4.6
Frozen: False
Fonts:
  medium: FiraCodeNF-Reg: /home/eralston/.local/share/fonts/FiraCodeNerdFont-Regular.ttf:0
          Features: ()
[...]

PR #8986, opened 2025-09-18
[bugfix] Fix hint label collision

edralston:
When the number of hint labels exceeds the size of the hint alphabet, some of the labels must be expanded to two or more characters.  However, as currently implemented, it is possible for some longer hint labels to start with the same text as a shorter label.

For example, if the hint label choices are "j" "k" "ja" "jb" and "jc", then it is not possible for the user to select the mark associated with "j" because the hint engine will still be trying to match against "ja" or "jb" etc.

This commit ensures that longer hint labels never start with the same text as any shorter hint labels.

Addresses #8985

### DIFF
--- a/kittens/hints/main.go
+++ b/kittens/hints/main.go
@@ -89,26 +89,55 @@ type Result struct {
 	Cwd                  string           `json:"cwd"`
 }
 
-func encode_hint(num int, alphabet string) (res string) {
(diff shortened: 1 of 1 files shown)


kovidgoyal:
As I noted on #8985 you simply press j then enter or space to select j when ja and jb exist as well.

PR #9000, opened 2025-09-21
SDF implementation for rounded corner

bea4dev:
This is a complete SDF-based implementation, rather than the previous provisional workaround.

# Screenshot
- Before
image

- After
image

### DIFF
--- a/kitty/decorations.c
+++ b/kitty/decorations.c
@@ -313,29 +313,6 @@ add_vholes(Canvas *self, uint level, uint num) {
     }
 }
 
-
-static void
-draw_hline(Canvas *self, uint x1, uint x2, uint y, uint level) {
-    // Draw a horizontal line between [x1, x2) centered at y with the thickness given by level and self->supersample_factor
-    uint sz = thickness(self, level, false);
-    uint start = minus(y, sz / 2);
-    for (uint y = start; y < min(start + sz, self->height); y++) {
-        uint8_t *py = self->mask + y * self->width;
-        memset(py + x1, 255, minus(min(x2, self->width), x1));
-    }
-}
-
-static void
-draw_vline(Canvas *self, uint y1, uint y2, uint x, uint level) {
-    // Draw a vertical line between [y1, y2) centered at x with the thickness given by level and self->supersample_factor
-    uint sz = thickness(self, level, true);
-    uint start = minus(x, sz / 2), end = min(start + sz, self->width), xsz = minus(end, start);
-    for (uint y = y1; y < min(y2, self->height); y++) {
-        uint8_t *py = self->mask + y * self->width;
-        memset(py + start, 255, xsz);
-    }
-}
-
 static uint
 half_width(Canvas *self) {  // align with non-supersampled co-ords
     return self->supersample_factor * (self->width / 2 / self->supersample_factor);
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Looks generally good, some comments:

1) Please remove/squash the first commit, I think that's from your previous PR?

2) Why are you drawing hline() and vline() with anti-aliasing?
Horizontal and vertical lines dont require anti aliasing and that just
makes them look blurry. Is it to match the antialiasing on the rounded
corners? If so I dont think that's worth the cost, hline and vline are
used in many other contexts apart from as part of rounded rectangles and
making them blurry everywhere isnt the right tradeoff IMO.

3) If we are doing SDF based drawing do we really need another round of
anti aliasing on top? Currently rounded_corner() is called with the S()
macro which means a round of anti aliasing is applied after it is run.
That is both inefficient and might actually make things look worse, not
sure. Worth checking if it is needed. If it turns out to be better
without extra antialiasing then remember to change the rest of the call
sites of rounded corner.

4) Please translate the Chinese comments to English.

bea4dev:
image

I’m not sure whether it’s an issue with the existing anti-aliasing or with my code, but I’m having a hard time tuning it at low DPI. Honestly, I think it would be smoother to implement if straight edges also had anti-aliasing pre-applied, the same way as the rounded corners.(In fact, with the approach that pre-applies anti-aliasing, it works well even at low DPI. And, the straight lines also look fine.)

kovidgoyal:
Anti aliasing the straight lines just makes them thicker and a bit fuzzy on the edges, that's not the right way to fix the issue as it makes most things worse to improve one small thing. Since nowadays most people have high DPI displays I am ok with optimising for that case rather than worrying too much about the 1-2 px thickness case.

bea4dev:
I understand your point, but the existing anti-aliasing is applied after mapping to pixels, which makes fine-tuning at low~middle DPI very difficult. Could you build it and take a look at how the straight lines render? (I hope you’ll like this approach.)

bea4dev:
image
Here are comparison screenshots.
The top one uses my proposed method; the bottom one shows the in-progress tuning approach.
To my eye, the top looks better (I like it a lot, and it works well on both low DPI and HiDPI).

kovidgoyal:
On Sun, Sep 21, 2025 at 06:03:44AM -0700, bea4dev wrote:
> bea4dev left a comment (kovidgoyal/kitty#9000)
>
> I understand your point, but the existing anti-aliasing is applied after mapping to pixels,

that's why I suggested in point 4 above to remove that layer of anti
aliasing. Drawing with an SDF should mean we dont require it at all.

> Could you build it and take a look at how the straight lines render?

On my machine it add two pixels in half opacity on each side of the
line, thereby doubling its width, which thanks to the half opacity
makes it look like the line is slightly thicker with fuzzy edge.

kovidgoyal:
On Sun, Sep 21, 2025 at 06:12:48AM -0700, bea4dev wrote:
> bea4dev left a comment (kovidgoyal/kitty#9000)
>
> The top one uses my proposed method; the bottom one shows the in-progress tuning approach.
> To my eye, the top looks better (I like it a lot, and it works well on both low DPI and HiDPI).

Definitely *rounded rectangles* look better if one anti aliases the
straight segments as well, then the whole thing matches, corners and
straights look the same thickness. That's how they are drawn on the GPU
using the rounder rectangle shader as well.

But what if you have normal, not rounded rectangles? Then the lines just
look fuzzy for no gain. I am not sure that is a worthwhile
tradeoff. What happens if you remove the supersample anti aliasing from
the rounded corners?

I am not adamantly opposed to anti aliasing the straight segments, but I
would like to explore other alternatives first.

bea4dev:
Understood, thank you. I’ll split the work into separate branches on my side so they can be compared.
I’ll share the code logic for the non–anti-aliased approach later; when I do, please try it out as well and compare the results.

bea4dev:
I’ve pushed the tuned code.
The sdf branch contains the tuned version, and double-anti-aliasing is the one that applies pre-anti-aliasing.
I’d appreciate your review.

bea4dev:
image
There are some parts with rather hacky tweaks.

kovidgoyal:
OK I'll take a look at it in the next day or two.

kovidgoyal:
And just as motivation I checked out the rounded corners on this page itself. Under magnification we see that the straight segments are not anti aliased but the thickness of the rounded corners is correct, so maybe we should look at how chromium or more correctly skia renders rounded corners. Possibly there is some technique used for this I am not aware of.

kovidgoyal:
Relevant code in skia: https://github.com/google/skia/blob/main/src/core/SkScan_AAAPath.cpp#L1604

kovidgoyal:
@bea4dev do you want to try adapting the skia code? Otherwise I will get to it when I have some time.

kovidgoyal:
Basically all that needs to be done is replace draw_parametrized_curve_with_derivative with an implementation that is based on the skia code. Anti aliasing will be done in that implementation rather than the generic supersampling anti aliasing that is done currently.

That should fix this issue.

bea4dev:
Honestly, this is outside my area of expertise, so I’m not very confident I can implement it (and even if I could, it would likely take some time).
In addition, I believe Skia’s AAA (analytic anti-aliasing) operates at rasterization time—i.e., when mapping to pixels—so its logic would differ from our current post-rasterization anti-aliasing, and it might require non-trivial code changes.
While I’d like to consider your proposed approach positively, to be frank, it would probably be better in the long run if you implemented it yourself.
(That said, I am curious how other terminal emulators like Ghostty address this kind of issue.)

kovidgoyal:
OK, then I will do it. Thanks for your efforts.

bea4dev:
I just realized the blurriness on straight lines may not be due to double-applying anti-aliasing, but rather an issue with my implementation approach.
So I’ll think a bit more about how we might further improve the anti-aliasing. It might be solvable with a simple approach.

kovidgoyal:
OK

kovidgoyal:
I have implemented it, take a look. No antialiasing of straight line segments.

bea4dev:
This looks great—almost perfect.
It works perfectly on HiDPI.
Thank you for implementing it.
That said, on low DPI (cases where features are ≤3 px wide), as you mentioned, some parameter tuning may be necessary.
At 1.5× scale I can see a slight misalignment.
image

kovidgoyal:
See if this helps: https://github.com/kovidgoyal/kitty/commit/89b1c35ac687fa77fb49523e16fb71496101712a

bea4dev:
image

I think it’s working perfectly.
Thank you very much.

kovidgoyal:
SDF based rendering was finally merged in #9091

PR #9004, opened 2025-09-22
resize-os-window: fix '--incremental' option of 'os-panel' action

alex-huff:
This commit addresses a few issues with the implementation of '--incremental':
    - Unspecified settings are reset to their default value, which defeats the purpose of the option.
    - It is assumed that the names of options in 'LayerCLIOptions' map one to one with the names of fields in 'LayerShellConfig' but this isn't true. For example: The 'margin_top' cli option sets the 'requested_top_margin' layer shell config.
    - When some options are set to a certain value, they force other options to be some value. The current implementation doesn't account for this.
    - The documentation is contradictory.

### DIFF
--- a/kittens/panel/main.py
+++ b/kittens/panel/main.py
@@ -5,7 +5,7 @@ import os
 import sys
 from contextlib import suppress
 from functools import partial
-from typing import Iterable, Mapping, Sequence
+from typing import Any, Iterable, Mapping, Sequence
 
 from kitty.cli import parse_args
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Note I moved the incremental update code into the panel kitten module so more chances of it staying in sync with future CLI modifications.

alex-huff:
> Note I moved the incremental update code into the panel kitten module so more chances of it staying in sync with future CLI modifications.

This makes sense.

PR #9007, opened 2025-09-22
resize-os-window: fix crash with '--incremental'

alex-huff:
Before this commit when using '--incremental' with a layer cli option that didn't map to any layer shell configurations would crash with a 'KeyError'.


alex-huff:
For example: `--incremental override=foo`

PR #9008, opened 2025-09-22
Revert "Workaround for latest Apple regression in Tahoe"

ethanwu10:
This reverts commit d837a64e36019e80ad027282822dc0acacf44bbb.

Commit 24b31d96e9a065eab578b7d8280f811a19768f65 fixes the original problem of "ghost windows" that d837a64 was trying to solve. (I have no idea how or why it does.)

One side-effect of the workaround from d837a64 is that it causes closed windows to be moved to the (0,0) before being destroyed. This means that when creating new windows, they are placed on the left-most display, instead of the display in which the previously closed window was on as you would expect. Reverting this kludge fixes this bug.


kovidgoyal:
This workaround is still needed, I should have tested this more thoroughly before merging. On my Tahoe system with current master, the invisible frame sticks around without this workaround. I have changed the workaround to keep the frame origin unchanged so it should solve your issue. See https://github.com/kovidgoyal/kitty/commit/fb4a3d7f179d96356334d857e9c3bc056b65ce21

Let me know if you still have an issue.

PR #9020, opened 2025-09-27
ssh kitten: add optional password/TOTP auto-fill via ssh.conf

NightWatcher314:
Overview
- Adds optional automation to auto-enter password and generate TOTP when prompted by OpenSSH, driven by ssh.conf per-host configuration. Falls back to kitty’s askpass UI for other prompts.

Motivation
- Some environments disallow or don’t reliably accept one‑way pubkey‑only authentication, or enforce keyboard‑interactive flows (password + TOTP). This provides a lightweight, opt‑in way to reduce repetitive manual entry while preserving the ssh kitten UX and connection latency benefits.

Usage
- In ~/.config/kitty/ssh.conf under the matching hostname block:
  - password <plain-text password>
  - totp_secret <base32 secret>
  - totp_digits <optional, default 6>
  - totp_period <optional, default 30>
- Works with askpass set to native/unless-set. Only login password/OTP prompts are auto‑answered; host key confirmations and key passphrases are not.

Security
- Secrets in ssh.conf are plain text; users should enforce strict permissions (e.g., 600) or avoid storing passwords if unacceptable. TOTP is generated per RFC 6238 (SHA1, base32 secret).

Implementation
- kittens/ssh/auth_config.go: Parse password/totp_* from ssh.conf with same hostname matching semantics as existing ssh kitten config blocks.
- kittens/ssh/config.go: Ignore these new keys in the primary config loader to avoid “bad line” warnings.
- kittens/ssh/main.go: Export KITTY_SSH_ASKPASS_HOST/USER so askpass can resolve the correct host block.
- kittens/ssh/askpass.go: Detect password/OTP prompts; auto‑print configured password or generated TOTP; otherwise fallback to kitty askpass dialog.

Notes
- Non‑breaking and opt‑in; no behavior change unless users add the new keys in ssh.conf.
- Happy to refine prompt detection strings or support alternative secret stores if desired.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -151,7 +151,7 @@ Detailed list of changes
   switching of dark/light mode when setting :opt:`macos_titlebar_color` to a
   arbitrary color (:iss:`9034`)
 
-- goto_session: Add ``--sort-by=alphabetical`` to have the interactive session
+- :ac:`goto_session`: Add ``--sort-by=alphabetical`` to have the interactive session
   picker list the sessions in a fixed order rather than by most recent
   (:disc:`9033`)
 
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -3270,11 +3270,6 @@ GLFWAPI GLFWcocoarenderframefun glfwCocoaSetWindowResizeCallback(GLFWwindow *w, 
 
 @end
 
-static CGFloat
-title_bar_and_tool_bar_height(NSWindow *window) {
-    return window.frame.size.height - [window contentRectForFrameRect:window.frame].size.height;
-}
-
 static
 void clear_title_bar_background_views(NSWindow *window) {
 #define tag @"kitty-for-transparent-titlebar"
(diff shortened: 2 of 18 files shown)


kovidgoyal:
I'll review the code in detail when I have more time, but off the top of my head, make the storage of passwords support multiple backends. For now just keep plain text, but change the format of the setting to be of the form backend:secret with text as the only currently supported backend.

NightWatcher314:
sure, I will do it.

kovidgoyal:
Looks generally OK, some comments:
1. Rather than reparsing the conf file add your new config fields to ssh/main.py, possibly in a new "askpass" section. Avoids duplication of logic and adds the new options to the documentation automatically. It does of course mean the askpass kitten has to parse the full file, but I dont think that is significant overhead.
2. Use range over int in generateTotp also not sure why you are using int64 rather than plain int there?

NightWatcher314:
Okay, I’ve already changed the loop to use range, and I’ve also moved the configuration into main.py.

kovidgoyal:
merged after squashing and with some cleanups.

francislavoie:
FWIW, I don't use `kitten ssh` but I solved this problem for myself with an `expect` script like this, to watch for a password prompt during my interactive session then try to autofill it with 1password:

```expect
#!/usr/bin/expect -f
set timeout -1

set env(TERM) "xterm-256color"
spawn ssh [lindex $argv 0]

# Track when we last autofilled
set last_autofill 0
set cooldown 60

set sudo_re {\[sudo\] password for francis: $}

interact {
    -o -re $sudo_re {
        send_user "\[sudo\] password for francis: "
        set now [clock seconds]

        if {($now - $last_autofill) >= $cooldown} {
            # Autofill password
            set pw [string trim [exec op item get --account <account-id> <item-id> --reveal --field password]]
            send -- "$pw\r"
            send_user "\n\[Password autofilled\]\n"
            set last_autofill $now
        }
    }
}
```

Then I use it like `./ssh-autofill.sh foo`.

I also use `expect` in my `session.conf` to autofill my initial terminal tabs for specific servers with this command so I can just hit enter to login when I'm ready after a reboot or whatever:

```
new_tab foo
cd ~
launch filledprompt "./ssh-autofill.sh foo"
```

The `filledprompt` script is this:

```expect
#!/usr/bin/expect

#trap sigwinch and pass it to the child we spawned
#this allows the gnome-terminal window to be resized
trap {
 set rows [stty rows]
 set cols [stty columns]
 stty rows $rows columns $cols < $spawn_out(slave,name)
} WINCH

set arg1 [lindex $argv 0]

# Get a Bash shell
spawn -noecho bash

# Wait for a prompt
expect "$ "

# Type something
send $arg1

# Hand over control to the user
interact

exit
```

NightWatcher314:
I used to rely on expect scripts for automatic login as well, but I ran into a few issues:
	•	Sessions started with expect’s interact sometimes failed to properly render Nerd Fonts.
	•	I couldn’t use a unified syntax to connect to all machines, which added extra cognitive load. For example, I had to run a dedicated expect script for hosts requiring TOTP, while using plain ssh for others.

My workflow back then was to first use an expect script to establish a reusable session, and then connect again with ssh. This way, I could reuse the existing connection without re-authentication, and also avoid the issues mentioned above.

Recently I discovered that Kitty has reimplemented its own SSH, and the code is much easier to follow compared to OpenSSH. So I’ve been experimenting with adding support for TOTP, passwords, and other features directly in Kitty’s implementation.

PR #9112, opened 2025-10-13
Add draw_borders_when_focused option

### LINKED ISSUE #9109: global watcher add on_layout_change

I have a global watch script that changes the border colour when focus changes. But I also want the border to show when there is a single window, e.g. switch from `split` to `stack`. However, in this case, the `on_focus_change` is not fired. Neither is `on_resize`

I think an `on_layout_change` callback make sense to be in the watcher. Alternatively, the `on_resize` should be fired.

There doesn't seem to be any workaround.

**Additional context**

N/A

jackielii:
## Summary

This PR adds a new configuration option `draw_borders_when_focused` that allows borders to be drawn around windows when the OS window is focused, regardless of the layout or number of windows.

## Motivation

Currently, borders are only shown in certain layouts (like splits) when there are multiple windows. This makes it difficult to see if a terminal window has focus when:
- Using the stack layout (which normally shows no borders)
- Having only a single window in any layout
- Switching between different applications

## Changes

- **New config option**: `draw_borders_when_focused` (default: `no`)
  - When enabled, borders are always drawn when the OS window is focused
  - Uses `active_border_color` when focused
  - Uses `inactive_border_color` when unfocused
  
- **Implementation**:
  - Modified `tabs.py` to enable border drawing when option is set
  - Modified `borders.py` to check OS window focus state (only when option is enabled)
  - Modified `boss.py` to trigger border redraw on focus changes
  
- **Backward compatibility**: The option defaults to `no`, preserving existing behavior

## Use Cases

1. **Stack layout visibility**: See which window has focus in stack layout
2. **Single window indicator**: Know if your terminal is active even with one window
3. **OS window focus feedback**: Clear visual indicator of terminal focus state

## Testing

Tested with:
- Single window in splits layout ✅
- Single window in stack layout ✅  
- Multiple windows in splits layout ✅
- Multiple windows in stack layout ✅
- Focus changes between applications ✅
- Color changes (active → inactive → active) ✅

## Example Configuration

```conf
window_border_width 4.0pt
active_border_color #ebcb8b
inactive_border_color #4c566a
window_margin_width 1
single_window_margin_width 5
draw_borders_when_focused yes
```

or with
```shell
kitty/launcher/kitty --config=NONE \
  -o font_size=14 \
  -o background=#1e1e1e \
  -o foreground=#d4d4d4 \
  -o window_border_width=4.0pt \
  -o active_border_color=#ebcb8b \
  -o inactive_border_color=#4c566a \
  -o window_margin_width=1 \
  -o single_window_margin_width=4 \
  -o draw_borders_when_focused=yes \
  -o "enabled_layouts=splits,stack" \
  -o "map cmd+f toggle_layout stack" \
  -o "map cmd+d launch --location=vsplit --cwd=current" \
  -o "map cmd+shift+d launch --location=hsplit --cwd=current"
```

Fixes #9109

### DIFF
--- a/kitty/tabs.py
+++ b/kitty/tabs.py
@@ -419,8 +419,6 @@ class Tab:  # {{{
         if tm is not None:
             ly = self.current_layout
             opts = get_options()
-            # Draw borders if: normal conditions OR new option enabled
-            # The new option makes kitty behave like draw_minimal_borders is true
             draw_borders = (
                 (ly.needs_window_borders and self.windows.num_visble_groups > 1) or ly.must_draw_borders
                 or opts.draw_window_borders_for_single_window


kovidgoyal:
Isn't it the job of the OS to indicate which OS window has focus? Why
should the application be doing that. And if you really want kitty to do
this, why not use inactive_text_alpha?

jackielii:
When I have a full screen window, and a floating window, it's very hard to know who has focus. I tried all kinds of different ways:

1. [JankyBorders](https://github.com/FelixKratz/JankyBorders) with yabai
2. normal_window_opacity vs active_window_opacity in yabai
3. inactive_text_alpha in kitty

The problem with the above is 1) the transparency makes the text not viewable, most of the time you want both window to be fully visible to not be distracting. 2) and active border on the overall windows doesn't indicate that that what kitty window I had focus.

Bottom line is the visual feedback for me is: a yellow rect around kitty -> I start typing, no rect means wrong window. I got it wrong so many times with other options, that's why I arrived at this solution. I tried to work around it with the watcher, but there is always something that's not right.

Please could you consider this. It is fully compatible with existing configuration, no breaking change. If there are any improvements you need, please let me know. Thanks and Regards,

kovidgoyal:
OK, but this PR needs significant refactoring.

1) The option should be draw_window_borders_for_single_window
2) When this option is true kitty must behave like draw_minimal_borders is
true
3) There is no need to do anything with os window focus. Basically, when
this option is set window borders are always drawn, inactive ones when
the windows are not focused and active ones otherwise, OS window focus
state is irrelevant.

jackielii:
> The option should be draw_window_borders_for_single_window

done

> When this option is true kitty must behave like draw_minimal_borders is true

done

> There is no need to do anything with os window focus. Basically, when
> this option is set window borders are always drawn, inactive ones when
> the windows are not focused and active ones otherwise, OS window focus
> state is irrelevant.

Hmmm, how can I display inactive border colour when OS window focus is lost?

kovidgoyal:
On Tue, Oct 14, 2025 at 10:05:41AM -0700, Jackie Li wrote:
> jackielii left a comment (kovidgoyal/kitty#9112)
>
> > The option should be draw_window_borders_for_single_window
>
> done
>
> > When this option is true kitty must behave like draw_minimal_borders is true
>
> done
>
> > There is no need to do anything with os window focus. Basically, when
> > this option is set window borders are always drawn, inactive ones when
> > the windows are not focused and active ones otherwise, OS window focus
> > state is irrelevant.
>
> Hmmm, how can I display inactive border colour when OS window focus is lost?

You can special case the case of display only one window and draw its
border as inactive when its parent os window is not focused.

When there is more than one window, it should continue to use active
border color, so that it is clear which window will get focus when the
OS window gains focus next.

kovidgoyal:
On Tue, Oct 14, 2025 at 10:05:41AM -0700, Jackie Li wrote:
> jackielii left a comment (kovidgoyal/kitty#9112)
>
> > The option should be draw_window_borders_for_single_window
>
> done
>
> > When this option is true kitty must behave like draw_minimal_borders is true
>
> done

Oh and that was a typo I meant false. Or more precisely it should act
like draw_minimal_borders is False when there is only a single visible window
in the layout.

jackielii:
> You can special case the case of display only one window and draw its
> border as inactive when its parent os window is not focused.
> 
> When there is more than one window, it should continue to use active
> border color, so that it is clear which window will get focus when the
> OS window gains focus next.

I think I found my solution: I can use a watcher script to set the active border colour to my desired value. something like:

```python
def on_focus_change(boss: Boss, window, data):
    focused = data['focused']

    if focused:
        boss.call_remote_control(window, ('set-colors', '--configured', 'active_border_color=#ebcb8b')) # my active color
    else:
        boss.call_remote_control(window, ('set-colors', '--configured', 'active_border_color=#4c566a')) # my inactive color

```

jackielii:
> You can special case the case of display only one window and draw its
> border as inactive when its parent os window is not focused.
> 
> When there is more than one window, it should continue to use active
> border color, so that it is clear which window will get focus when the
> OS window gains focus next.

Just to confirm, you want:

single window + has focus -> active border
single window + no focus -> inactive border
multiple window + regardless of focus -> active border

is this correct?

kovidgoyal:
On Tue, Oct 14, 2025 at 10:25:39AM -0700, Jackie Li wrote:
> jackielii left a comment (kovidgoyal/kitty#9112)
>
> > You can special case the case of display only one window and draw its
> > border as inactive when its parent os window is not focused.
> >
> > When there is more than one window, it should continue to use active
> > border color, so that it is clear which window will get focus when the
> > OS window gains focus next.
>
> Just to confirm, you want:

If you have already found a solution, I dont really want anything :)

Here are the cases:

a) option set:
1) Single *visible* window in the layout - always draw a border
   border color = active if os_window is focused else inactive

2) multiple visible windows in the layout - no change in existing
behavior, in particular draw_minimal_borders should be respected as
before.

b) option unset - no changes from present behavior

option must default to unset.

jackielii:
All done. Although I still suggest using inactive border colour when "focus" is lost i.e. by either moving to another kitty window or os window. But if this is what you expect, my watcher workaround works fine too! Thanks again!

kovidgoyal:
that's what I said in case a.1 - the border color should be inactive when the os window is unfocused

jackielii:
> that's what I said in case a.1 - the border color should be inactive when the os window is unfocused

I meant for all cases 1 window or many windows. That's ok. I got it working for my case which is it draws a border for stack view. Good enough for me

The code is ready to be reviewed. Let me know if any comments. Thanks

kovidgoyal:
You need to also do something about draw_minimal_borders in
layout/base.py. layout geometry calculations depend on
lgd.draw_minimal_borders. Currently it is static, you will instead need
to update it based on the current number of visible groups and the new
setting in Layout._set_dimensions.

jackielii:
done, updated `_set_dimensions` to respect number of windows , draw_minimal_borders, and also draw_window_borders_for_single_window

jackielii:
the test is failing, I think I need to set_option in these tests? fill fix

jackielii:
fixed test. should be good to go. thanks

PR #9117, opened 2025-10-16
panel: use an anchor of 0 for --edge=center-sized

### LINKED ISSUE #9116: `quick-access-terminal` anchors
## Summary

The `quick-access-terminal` kitten uses the `panel` kitten internally, inheriting its `edge` option. This causes unexpected behavior on the `niri` Wayland compositor (and likely all smithay-based compositors).

## Problem Description

I want to center the terminal by using `edge center-sized` option. However, instead of being centered, the terminal window is anchored to the top-left corner while keeping the specified size.

After some testing, I discovered that:
- When the anchor is set to `0`, the behavior is correct — the window appears centered as expected.
- When setting `edge right`, for example, the window is anchored to the right edge and stretched vertically, as the anchors were set to `top + bottom + right`.

It seems that when all anchors are set, window should be stretched to the entire screen and the window’s size should no longer affect the final layout.

## Proposed Solution

Introduce a platform-specific option (e.g. `anchor` or `anchor_left`, `anchor_top`, etc.) that allows users to explicitly control anchor values or disable them entirely.

This would make it possible to achieve consistent positioning behavior across different Wayland compositors.

alex-huff:
Fixes #9116

### DIFF
--- a/glfw/wl_window.c
+++ b/glfw/wl_window.c
@@ -1054,6 +1054,7 @@ layer_set_properties(const _GLFWwindow *window, bool during_creation, uint32_t w
                 case GLFW_EDGE_CENTER:
                     break;
                 case GLFW_EDGE_CENTER_SIZED:
+                    which_anchor = 0;
                     panel_width = width; panel_height = height;
                     break;
                 case GLFW_EDGE_NONE:


kovidgoyal:
Does this work on other non-smithay compositors?

alex-huff:
> Does this work on other non-smithay compositors?

I've tested on sway, anvil, niri but I'll test some more in a bit.

kovidgoyal:
I tested it on a couple more, hyprland and KDE, it is anyway spec
compliant so good enough.

alex-huff:
> I tested it on a couple more, hyprland and KDE, it is anyway spec compliant so good enough.

Thanks, I'll just verify it works for the other compositors in the compat matrix.

PR #9128, opened 2025-10-18
Accumulate VALUE120 high-res discrete scrolling events

evan-goode:
Fixes a bug where VALUE120 scroll events with value < 1 line would always scroll a full line, leading to strange-feeling scrolling behavior on smooth-scrolling input devices.

Explanation:

```
        // VALUE120 events are "discrete" in that they scroll lines, not pixels.
        // But each VALUE120 event of value `v` represents a fraction of a
        // line, specifically `v` 120ths of a line.

        // GLFWscrollfun does not handle non-highres scroll events with
        // non-integer value. We could use `highres` for VALUE120 events, but
        // that would be wrong since the units should be interpreted as
        // fractional lines, not pixels.

        // So instead, we accumulate discrete events until we have at least one
        // full line to scroll, and then call the GLFWscrollfun to scroll entire
        // lines.
```

### DIFF
--- a/glfw/wl_init.c
+++ b/glfw/wl_init.c
@@ -202,21 +202,23 @@ pointer_handle_frame(void *data UNUSED, struct wl_pointer *pointer UNUSED) {
     _GLFWwindow* window = _glfw.wl.pointerFocus;
     if (!window) return;
     float x = 0, y = 0;
-    int highres = 0;
+    const int HIGHRES = 1;
+    const int VALUE120 = 1 << 4;
+    int flags = 0;
 
     if (info.discrete.y_axis_type != AXIS_EVENT_UNKNOWN) {
         y = info.discrete.y;
-        if (info.discrete.y_axis_type == AXIS_EVENT_VALUE120) y /= 120.f;
+        if (info.discrete.y_axis_type == AXIS_EVENT_VALUE120) flags |= VALUE120;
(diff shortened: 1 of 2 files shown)


kovidgoyal:
The correct fix for this would be to pass the fact that these are 120th of line units in the flags parameter of GLFWscrollfun() and then change the code in mouse.c that handles them to use that information to actually scroll by 120th of a line. IIRC only the first four bits of flags is currently used so you can use bit 5 for it.

evan-goode:
> The correct fix for this would be to pass the fact that these are 120th of line units in the flags parameter of GLFWscrollfun() and then change the code in mouse.c that handles them to use that information to actually scroll by 120th of a line. IIRC only the first four bits of flags is currently used so you can use bit 5 for it.

Thanks, done. I didn't look at first and I assumed the GLFWscrollfun was some library code.

I realized that we currently do not handle the case when the two scroll axes are different types, e.g. one is continuous and another is discrete. But that would only be a problem with some very small minority of pointing devices.

kovidgoyal:
I dont quite follow the logic. In mouse.c you are treating a VALUE120
value as pixels (the logic is identical for high res and value120
values? For value120 value shouldnt this be

s = offset * cell_height / 120

rather than

s = offset / cell_height

In other words a value of 120 should cause a scroll by one line.

evan-goode:
You're right, my bad. How's that?

PR #9148, opened 2025-10-24
Fix: Make portals.conf fallback (Fixes #9144)

### LINKED ISSUE #9144: desktop-ui portal should set gtk as fallback Settings portal
**Describe the bug**

Currently `` creates this `*-portals.conf` file:

```ini
[preferred]
org.freedesktop.impl.portal.Settings=kitty
org.freedesktop.impl.portal.FileChooser=kitty
```

However GTK4 now [fully relies on the `Settings` portal for theming](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/issues/25#note_1324304), [using fallback values if the portal does not provide them](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/wayland/gdksettings-wayland.c#L269). Therefore making `kitty` the only `Settings` portal breaks GTK4 theming.

The correct config would be (as mentioned in first link above GTK4 reads config from all configured portals):

```ini
[preferred]
org.freedesktop.impl.portal.Settings=kitty;gtk
org.freedesktop.impl.portal.FileChooser=kitty
```

**To Reproduce**
1. `kitten desktop-ui enable-portal`
2. Start a GTK4 application
3. Notice theming is incorrect

azertyfun:


kovidgoyal:
There are two places in that function you need to patch and ;* should be
added only for the Settings interface.

azertyfun:
Oops, I missed that.

I'm not convinced that this should only be done for the Settings interface. `kitty;*` for the `FileChooser` interface means that the kitty file chooser will be used if available, only falling back to other file chooser(s) if kitty is not able to open the file chooser. IMO that is better UX.

The only downside I can see is that if kitty's implementation is not available, `xdg-desktop-portal` is not particularly verbose in explaining why it was not used. But IMO that's a problem to be solved in `xdg-desktop-portal`, not by forcing the inconvenient behavior of completely breaking the file chooser even though the GTK or Qt one would otherwise be available.

PR #9152, opened 2025-10-25
Add focus_tab command to session files

### LINKED ISSUE #9151: create focus_tab for kitty session

When restoring a session file, there is currently no way to specify which tab should be active. The `focus` command only focuses the last window in the last tab, but doesn't allow specifying a particular tab to be active when the session is loaded. This means that when saving and restoring sessions, the active tab is not preserved - instead, focus is determined by the most recently focused window (via `last_focused_at`), which may not be the desired tab.

Add a new `focus_tab` command to session files that accepts a tab index (0-based) to specify which tab should be active when the session is loaded.

Example usage in a session file:

```
new_tab "Config"
launch /bin/bash

new_tab "Development"
launch /bin/bash

new_tab "Logs"
launch /bin/bash

Focus the first tab (index 0)

focus_tab 0
```

This would allow session files to preserve the active tab state when saving and restoring sessions.

**Implementation details:**

The implementation requires:
1. A new `focus_tab(idx: int)` method in the `Session` class that sets `active_tab_idx`
2. Parsing support for the `focus_tab` command in `parse_session()`
3. A fix to `goto_session()` to avoid overriding the `focus_tab` command with `switch_to_session()` when creating new OS
windows

I have a working implementation that adds ~15 lines to `kitty/session.py` and is fully backward compatible. Happy to
submit a PR if this feature would be accepted.

- Using `focus_matching_window` with complex window matching - this is cumbersome and doesn't directly target tabs
- Modifying window `last_focused_at` timestamps during serialization - fragile and hacky
- Living with the incorrect tab focus on session restore - not ideal for workflow preservation

**Additional context**

This feature is particularly useful when:
- Saving sessions with a specific "main" tab that should always be focused on restore
- Creating predefined session files for different workflows where a particular tab should be the entry point
- Using session management tools that save/restore the current OS window state

The implementation maintains backward compatibility - sessions without `focus_tab` work exactly as before.

jackielii:
**Summary**

fix #9151

This PR adds a new `focus_tab` command to session files that allows specifying which tab should be active when a session is loaded.

**Motivation**

Currently, when restoring a session file, there is no way to specify which tab should be active. The `focus` command only focuses the last window in the last tab. When sessions are loaded, the active tab is determined by the most recently focused window (via `last_focused_at`), which may not be the desired tab.

This makes it impossible to preserve the active tab state when saving and restoring sessions.

**Implementation details**

- Tab index is 0-based
- The command is fully backward compatible - sessions without `focus_tab` work exactly as before
- When a session creates a new OS window, `focus_tab` is respected
- When tabs are added to an existing OS window, the original behavior is preserved

### DIFF
--- a/docs/sessions.rst
+++ b/docs/sessions.rst
@@ -159,6 +159,7 @@ all the major keywords you can use in kitty session files:
     launch tail -f /var/log/syslog
 
     # Focus the first tab (index 0) when the session loads
+    # You can also use a match expression like: focus_tab title:logs
     focus_tab 0
 
     # Create a complex layout using multiple splits. Creates two columns of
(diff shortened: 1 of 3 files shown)


jackielii:
Thanks @kovidgoyal for the explanation in the issue. As I said, my workflow maybe different, but I get to use a custom kitten to tweak it. The only missing piece is the `focus_tab`. This change should backward compatible.

This PR should be ok for you to review. Thanks!

PR #9156, opened 2025-10-26
implement --base-dir in save_as_session

jackielii:
based on issue comment: https://github.com/kovidgoyal/kitty/issues/9151#issuecomment-3448397304

This is on top of https://github.com/kovidgoyal/kitty/pull/9152

- `--base-dir` : if relative path is specified, it concats with base-dir
- `--focus-tab`: if specified, it appends `focus_tab` to the end of session file

### DIFF
--- a/kitty/session.py
+++ b/kitty/session.py
@@ -642,15 +642,11 @@ def save_as_session_part2(boss: BossType, opts: SaveAsSessionOptions, path: str)
     if not path:
         return
     from .config import atomic_save
-    # Handle --base-dir option
     if opts.base_dir and not os.path.isabs(path):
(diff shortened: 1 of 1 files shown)


jackielii:
@kovidgoyal is you're ok, I can add doc and change log as well

kovidgoyal:
There should be no need for --focus-tab that behavior should be automatic
already. When you save a session, the currently active tab and window
should become the currently active tab and window when loading that
session. If that is not currently the case, then it is a bug that should
be fixed, preferably in a separate PR. --base-dir is fine.

jackielii:
this is ready for review

PR #9163, opened 2025-10-27
add focus_tab to restore tab focus by default

jackielii:
I think previously the code was trying to sort tabs by activity history (least recently used → most recently used).  But this didn't preserve the order of the tabs as they appear. With the `focus_tab` addition, we don't need to do that any more. We just need to save them as they appear and add `focus_tab` at the end to restore the focus

### DIFF
--- a/kitty/tabs.py
+++ b/kitty/tabs.py
@@ -1312,13 +1312,8 @@ class TabManager:  # {{{
         is_first: bool = False
     ) -> list[str]:
         ans = []
-        hmap = {tab_id: i for i, tab_id in enumerate(self.active_tab_history)}
-        if (at := self.active_tab) is not None:
-            hmap[at.id] = len(self.active_tab_history) + 1
-        def skey(tab: Tab) -> int:
-            return hmap.get(tab.id, -1)
         active_tab_index = -1
-        for i, tab in enumerate(sorted(self, key=skey)):
+        for i, tab in enumerate(self.tabs):
             if tab is self.active_tab:
                 active_tab_index = i
             ans.extend(tab.serialize_state_as_session(session_path, matched_windows, ser_opts))


kovidgoyal:
The reason it was done that way was to preserve the history. If you want to preserve visual position instead then the activity history gets lost. I dont know if that's a worthwhile tradeoff, it should probably be an option for save_as_session. Or better would be to preserve both, possibly by some new syntax in the session file, say some JSON data for the new_tab command similar to how it is done for restoring window layout data.

jackielii:
> The reason it was done that way was to preserve the history. If you want to preserve visual position instead then the activity history gets lost. I dont know if that's a worthwhile tradeoff, it should probably be an option for save_as_session. Or better would be to preserve both, possibly by some new syntax in the session file, say some JSON data for the new_tab command similar to how it is done for restoring window layout data.

Thanks for the explanation, I have suspected it was done for a reason. 

By "history" do you mean the order the tabs are created? I don't think they (visual order & history) are inclusive of one the  other, i.e. not possible to preserve both.

I'm quite certain my workflow is different than your workflow at this point. So I'll explain mine:

1. in the beginning, I have no sessions
2. I open a kitty os window
3. I add some tabs then some windows
4. I reorder them because related windows I put them together for visual cue to jump with cmd+{1-9}
5. I save session and give it a name
6. I close it for forget about it
7. I restore / switch it using fuzzy find

IIUC, step 4 is what you didn't expect?

I'm happy with just adding another option. Or just keep it as is, a custom kitty is also straightforward. Downside is relying on some internal API. I'm mostly fine with it. When I upgrade, I expect some part to be broken and most of the time, they're easy fixes.

jackielii:
Just had a look at the `active_tab_history` related code. It's for jumping back in tab history right? e.g. go to n'th last tab?

kovidgoyal:
Yes, it is for jumping back.

jackielii:
> Yes, it is for jumping back.

We probably can't use the order of serialisation for both tab history and visual history. I see two options:

1. have sequence of tabs as one and add tab_history or visual_order command to record
2. have both `tab_history [idx, ...]` and `visual_order ...]` command

PR #9185, opened 2025-11-04
Add action to copy the last command output

rgalhos:
Adds the `copy_last_command_output` keyboard shortcut, which copies the output of the last shell command to the clipboard.


kovidgoyal:
Adding the action is sufficient. This isnt something that's going to
have a default map and definitely not a map for F1

rgalhos:
Sorry, I thought had to add a map to `definition.py` to generate the docs. Removed it.

PR #9219, opened 2025-11-12
Allow browsing a specific directory with sessions

akvachan:
Implementation:
  - Interpret a directory argument by listing only *.kitty-session and similar files.
  - Reuse the existing sorting logic for the directory chooser and document the workflow.
  
Flow:
1. User saves all or some sessions to a specific dire, let's say ~/.config/kitty/session.
2. User binds goto_session ~/.config/kitty/session.
3. User invokes the binding and sees a picker, just like the one he would see without providing any arguments in goto_session.


akvachan:
Screenshot 2025-11-12 at 18 13 48

PR #9243, opened 2025-11-24
changes to bash scripts in graphics protocol documentation

joshuarayton:
Improve bash performance and use best practice. 

Changing the shebang to `/usr/bin/env bash` is what I think is best practice.

No need to use an external invocation to `stty` to get size when there is `$COLUMNS` and `$LINES`. 

Prefixing shell builtins with the `builtin` builtin isn't necessary as they take precedence over external binaries and we aren't making our own functions with the same names. 

Shouldn't unset `$IFS` after changing it's value because it changes back to it's default value of `$' \t\n'` after. 

`echo` statement split to fit 80 columns and improves readability. 

Completely changed the `transmit-png` function. Let me know if I've misimplemented the protocol but it works on a few images I've tried.
Major performance increase comes from not needing to remove all `[[:space:]]`.

### DIFF
--- a/docs/graphics-protocol.rst
+++ b/docs/graphics-protocol.rst
@@ -136,15 +136,19 @@ code to demonstrate its use
 
     .. code-block:: sh
 
-        #!/bin/bash
+        #!/usr/bin/env bash
 
         # This uses the kitten standalone binary from kitty to get the pixel sizes
         # since we can't do IOCTLs directly. Fortunately, kitten is a static exe
         # pre-built for every Unix like OS under the sun.
 
-        builtin read -r rows cols < <(command stty size)
-        IFS=x builtin read -r width height < <(command kitten icat --print-window-size); builtin unset IFS
-        builtin echo "number of rows: $rows number of columns: $cols screen width: $width screen height: $height"
+        shopt -s checkwinsize
+        (:)  # set $COLUMNS and $LINES
+
+        IFS=x read -r width height < <(kitten icat --print-window-size)
+
+        echo "number of rows: $LINES number of columns: $COLUMNS"
+        echo "screen width: $width screen height: $height"
 
 
 Note that some terminals return ``0`` for the width and height values. Such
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On Sun, Nov 23, 2025 at 04:31:04PM -0800, Joshua Rayton wrote:
> Improve bash performance and use best practice.
>
> Changing the shebang to `/usr/bin/env bash` is what I think is best practice.

Not really, dont know of any system where that's actually necessary,
but beyond the pointless overhead of calling env, no harm, so OK.

>
> No need to use an external invocation to `stty` to get size when there is `$COLUMNS` and `$LINES`.

COLUMNS and LINES are not reliable. And checkwinsize is a bashism. I
tried to keep the script as close to POSIX sh as possible, so it can be
ported to other shells relative easily. That said the script is
obviously not POSIX sh compatible currently.

>
> Prefixing shell builtins with the `builtin` builtin isn&#39;t necessary as they take precedence over external binaries and we aren&#39;t making our own functions with the same names.

If only that were true. People override builtins in their bashrc files all
the time.

>
> Shouldn&#39;t unset `$IFS` after changing it&#39;s value because it changes back to it&#39;s default value of `$&#39; \t\n&#39;` after.

That's fine, though I prefer being explicit.

>
> `echo` statement split to fit 80 columns and improves readability.
>
> Completely changed the `transmit-png` function. Let me know if I&#39;ve misimplemented the protocol but it works on a few images I&#39;ve tried.
> Major performance increase comes from not needing to remove all `[[:space:]]`.

base64 -w is not portable, that's why [[:space:]] is used. -w doesnt
exist on macOS and various BSDs for instance.

kovidgoyal:
I have modified the scripts to be POSIX sh compatible so bash is no longer required.

joshuarayton:
> Not really, dont know of any system where that's actually necessary, but beyond the pointless overhead of calling env, no harm, so OK.

On a macOS system, `/bin/bash` is an outdated version. Users may choose to install a newer version of bash with something like homebrew which I think would be located in `/opt/homebrew/bin/bash` (not sure, I don't have a mac). `/bin/bash` on a mac doesn't have features like associative arrays (I think, once again: I don't have a mac). `#!/usr/bin/env bash` gets the user's choice of bash on their system, not whatever came in `/bin/bash`.

> COLUMNS and LINES are not reliable. And checkwinsize is a bashism. I tried to keep the script as close to POSIX sh as possible, so it can be ported to other shells relative easily. That said the script is obviously not POSIX sh compatible currently.

Not sure how `$COLUMNS` and `$LINES` aren't reliable in bash if you set the `checkwinsize` shell option and execute a subshell, but if you'd like to convert it to a POSIX shell script then okay.

> If only that were true. People override builtins in their bashrc files all the time.

`.bashrc` only gets sourced in an interactive shell context, not scripts. It is possible to override the functions in scripts via `$BASH_ENV` but that would have to be really intentionally done, not by arguably "accidentally" overriding the functions in their configuration. 

> That's fine, though I prefer being explicit.

It's not being explicit, it's different behaviour. `$IFS` is `$' \t\n'` by default, then during the read it's `"x"`, then it goes back to `$' \t\n'`. If it's unset, then it goes to `""` when read. 

> base64 -w is not portable, that's why [[:space:]] is used. -w doesnt exist on macOS and various BSDs for instance.

I did check for portability on [macOS](https://ss64.com/mac/base64.html) and [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=base64) which `-w` does seem to be on, but not sure if `-w 0` would work the same. 

> I have modified the scripts to be POSIX sh compatible so bash is no longer required.

Alright. I would probably opt for this based on your new implementation.

```sh
#!/bin/sh

transmit_png() (
	case "$(base64 --help)" in
		*"--wrap"*) flag="-w";;  # Linux (GNU coreutils)
		*"--break"*) flag="-b";;  # macOS/BSD
		*)
			echo "Unknown base64 command: cannot set line width" >&2
			exit 1
			;;
	esac

	printf "\e_Ga=T,f=100,m=1;\e\\"

	base64 "$flag" 4096 "$1" |
	while IFS= read -r chunk; do
		printf "\e_Gm=1;%s\e\\" "$chunk"
	done

	printf "\e_Gm=0;\e\\"
)

transmit_png "$1"

```

This runs in a subshell (use of `()` instead of `{}`) which doesn't persist the variables set inside.

Uses a case statement and one invocation to get the usage of `base64`. 

Bring the first escape codes out of the loop, and use `$'\e\\'` instead of `$'\a'`? This might not be according to the specification so up to you. 

Thank you for considering these changes.

kovidgoyal:
The current code simply runs base64 -w first which will work in most
places and fallsback to -b andthen finally openssl base64 for OpenBSD
which has no base64 executable at all.

No need to rely on --help output which is fragile.

PR #9270, opened 2025-12-05
Update `symbol_map` setting for nerd fonts

david:
Updates Nerd Font codepoints from https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points.


kovidgoyal:
Only PUA unicode codepoints should be used with symbol_map by default as
is noted in the FAQ.
For non PUA codepoints there is no particular reason to use NERD fonts.

PR #9303, opened 2025-12-19
Add MRU (Most Recently Used) tabs remote control command

dmgerman:
First of all, thank you for the great kitty.

I have been using it for a while and one of the things I really appreciate is how you make the information of the tabs available for other tools to be able to integrate with it.  I think that's a completely new way of thinking.

I have been building extensions to both Emacs and Hammerspoon (MacOS only) to jump to a given tab by providing a list of tabs. For example:
image

but I quickly discovered that I needed to jump to the most-recently-used tab. So I implemented a new command.

This PR implements a new `kitty @ mru` remote control command that returns tabs sorted by most recently used order. This provides external tools with an way to query tab access order.

Changes:
- Add `last_visited_at` timestamp tracking to Tab class
- Update timestamp when tabs become active
- Add `get_mru_tabs()` method to TabManager
- Add `get_mru_tabs_all_os_windows()` method to Boss
- Implement new MRU remote control command in kitty/rc/mru.py
- Add comprehensive automated tests in kitty_tests/mru.py
- Support --all-os-windows flag to query across all OS windows

The MRU command returns JSON with tab ID, last_visited_at timestamp, and os_window_id for each tab, sorted with most recent first.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -137,6 +137,9 @@ Detailed list of changes
 0.45.0 [future]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+- Add a new remote control command :ref:`at-mru` to list tabs in most recently used (MRU) order.
+  This provides external tools with an efficient way to query tab access history.
+
 - Add support for the `paste events protocol <https://rockorager.dev/misc/bracketed-paste-mime/>`__ (:iss:`9183`)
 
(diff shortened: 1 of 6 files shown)


dmgerman:
Hi Kovid, does this means you don't think the feature should be merged? Or there is something wrong with it that I can improve?

kovidgoyal:
No, no, see the commit message, I have implemented it in a
different way. You now get the history of tab activation as part of existing
kitten @ ls output.

dmgerman:
Sorry I didn't paid attention.   Your solution is much better than mine, cleaner and simpler.
Thank you very much.

PR #9325, opened 2025-12-25
Add wide gamut color support with OKLCH and LAB formats

jokull:
## Summary

Adds support for modern wide gamut color formats (OKLCH and CIE LAB) with CSS Color Module Level 4 gamut mapping.

## Motivation

Provides proper wide gamut color specification with perceptually uniform color spaces for theme creation.

## Implementation

**Color formats:**
```conf
foreground oklch(0.9 0.05 140)    # Perceptually uniform
color1     lab(70 50 -30)         # Device-independent
```

**Key features:**
- Parse-time conversion to sRGB
- CSS Color 4 gamut mapping for out-of-gamut colors
- Inline comment support in config files
- Full backward compatibility

**Files changed:**
- Python: `kitty/rgb.py`, `kitty_tests/datatypes.py` (+250 lines)
- Go: `tools/utils/style/colorspaces.go`, `wrapper.go` (+350 lines, new tests)
- Docs: `docs/wide-gamut-colors.rst` (moved from inline docs per feedback)

## Addressing PR Feedback

**1. Go Implementation ✓**
- Added complete Go implementation in `tools/utils/style/`
- Supports OKLCH and LAB with gamut mapping
- Comprehensive test coverage (all tests passing)

**2. Performance ✓**
- Benchmarked on Apple M1 Pro:
  - OKLCH: ~4.6 µs/op
  - LAB: ~1.5 µs/op
  - 9 mixed colors: ~11 µs total
- Performance is acceptable; C implementation not needed

**3. Documentation ✓**
- Moved detailed color space docs to separate `docs/wide-gamut-colors.rst`
- Configuration docs now just link to detailed documentation
- Reduces size of main configuration documentation

## Gamut Mapping

Implements CSS Color Module Level 4 gamut mapping algorithm:
- Binary search chroma reduction
- Preserves lightness and hue
- Uses deltaE OK (JND threshold: 0.02) for perceptual accuracy
- Ensures graceful degradation on sRGB displays

**Reference:**
- [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#css-gamut-mapping)
- [OKLCH Color Space](https://bottosson.github.io/posts/oklab/)

## Testing

```bash
./test.py to_color           # Python tests pass
go test ./tools/utils/style/ # Go tests pass
```

### DIFF
--- a/docs/wide-gamut-colors.rst
+++ b/docs/wide-gamut-colors.rst
@@ -1,0 +1,82 @@
+Wide gamut color formats
+=========================
+
+kitty supports modern wide gamut color formats for precise color specification.
+These formats can be used anywhere a color value is accepted in the configuration
+(foreground, background, color0-color255, etc.).
+
+OKLCH Colors
+------------
+
+OKLCH is a perceptually uniform color space, ideal for creating color themes.
+The format is::
+
+    foreground oklch(0.9 0.05 140)
+    color1     oklch(0.7 0.25 25)
+
+Parameters:
+
+- **L** (Lightness): 0 to 1, where 0 is black and 1 is white
+- **C** (Chroma): 0 to approximately 0.4, represents color saturation
+- **H** (Hue): 0 to 360 degrees (0=red, 120=green, 240=blue)
+
+Benefits:
+
+- Perceptually uniform - equal changes produce equal perceived differences
+- Adjusting lightness preserves hue (unlike HSL)
+- Industry standard for modern color design
+
+Example::
+
+    foreground oklch(0.9 0.05 140)
+    color1     oklch(0.65 0.25 29)    # Vibrant red-orange
+    color2     oklch(0.65 0.25 142)   # Vibrant green
+    color3     oklch(0.70 0.19 90)    # Warm yellow
+
+CIE LAB Colors
+--------------
+
+CIE LAB is a device-independent color space designed to approximate human vision.
+
+The format is::
+
+    background lab(20 5 -10)
+    color4     lab(50 0 -50)
+
+Parameters:
+
+- **L**: Lightness, 0 to 100 (0 = black, 100 = white)
+- **a**: Green (-) to red (+), typically -100 to +100
+- **b**: Blue (-) to yellow (+), typically -100 to +100
+
+Example::
+
+    background lab(10 0 0)           # Very dark neutral gray
+    foreground lab(90 0 0)           # Very light neutral gray
+    color1     lab(50 60 40)         # Red
+    color4     lab(50 0 -50)         # Blue
+
+Gamut Mapping
+-------------
+
+When you specify colors in OKLCH or CIE LAB formats that are outside your display's
+color gamut, kitty automatically converts them using the CSS Color Module Level 4
+gamut mapping algorithm:
+
+- Preserves the original lightness and hue as much as possible
+- Reduces chroma (saturation) until the color fits within the displayable range
+- Uses perceptual color difference (deltaE OK) to minimize visible changes
+- Maximizes color saturation while staying in gamut
+
+This ensures that wide gamut colors gracefully degrade on standard sRGB displays while
+taking full advantage of wide gamut displays when available. The mapping happens
+automatically - you don't need to do anything special.
+
+For example, :code:`oklch(0.7 0.4 25)` might be too saturated for sRGB but will be
+automatically adjusted to fit while preserving the perceived hue and lightness.
+
+References
+----------
+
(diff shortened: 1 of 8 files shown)


kovidgoyal:
How would this work for out of gamut colors (i.e. colors not
representable in sRGB?)

jokull:
Let me see how browser are handling it and come back.

kovidgoyal:
In principle, I am OK with this. Some comments:
1) Conf files are also parsed by kittens such as the diff kitten which allows users to specify colors for UI elements. So you would need to add support for this to the relevant parsing code in Go as well (kittens are written in Go) IIRC in tools/utils/styles/

2) I am concerned about performance of parsing conf files which directly affects user visible metrics like startup time and config reload time. There are of course only a few hundred colors to parse in even the most pathological scenarios so it *should* not be a problem, but it would be good to either have some metrics demonstrating it is not a problem or move the parsing code into C which will make it as performant as it is likely to get. If you wish to pursue the latter approach and you have questions about how to do it, feel free to ask. The Color object is already implemented in C for performance, so the basics are there. I would recommend making to_color a class method of the Color type with METH_CLASS.

3) Might be better to move the docs for PU color space into a different document (rst file) and just link to it from the conf docs as the conf docs are already rather giant.

jokull:
Thanks for the detailed feedback! I've addressed all three points:

## 1. Go Implementation ✓

Added complete Go implementation in `tools/utils/style/`:
- OKLCH and LAB parsing with gamut mapping
- All tests passing
- Follows same structure as Python implementation

## 2. Performance Benchmarks ✓

Ran benchmarks on Apple M1 Pro to verify parsing performance is acceptable:

```
BenchmarkParseOklch-10           229826    4.6 µs/op
BenchmarkParseLab-10             796572    1.5 µs/op
BenchmarkParseManyColors-10       98880   13.0 µs/op  (10 colors)
```

For typical config files (~50 colors), total parsing time is under 0.5ms, which should be negligible for startup/reload. Happy to implement in C if you feel the Python overhead is still a concern for kitty's performance goals.

## 3. Documentation Reorganization ✓

Moved detailed color space documentation to `docs/wide-gamut-colors.rst` as suggested. The conf docs now just link to it, keeping them concise.

---

## Removed Display P3 Support

Also removed Display P3 support to simplify the implementation. The main reasons:

1. **OKLCH already handles wide-gamut colors**: OKLCH is device-independent and can represent any color including P3, Rec2020, and beyond. Users who want P3 colors can specify them directly in OKLCH with higher chroma values.

2. **Better readability**: `oklch(0.7 0.26 145)` is more readable than `color(display-p3 0.5 0.9 0.6)` - you can immediately understand lightness, saturation, and hue.

3. **Simpler maintenance**: Two color formats (OKLCH + LAB) instead of three. Both use similar gamut mapping approach.

Reference: [Evil Martians - OKLCH in CSS: why we moved from RGB and HSL](https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl) has a good explanation of why OKLCH supersedes format-specific approaches like P3.

Let me know if you'd like any changes!

jokull:
For funsies I created a [theme editor](https://www.solberg.is/kitty?theme=1) with GitHub login, forking and sharing

kovidgoyal:
The benchmarks are for the Go implementation, if I am not mistaken? Go should be fast enough at this, it's the python I am concerned about, these kinds of numerical computation/parsing heavy workloads tend to be 5-20x slower in python compared to AOT compiled code.

Your theme editor looks nice, feel free to send a PR to the kitty-themes repo to add a link to it in the README and possibly mention it in docs/kittens/themes.rst. You might consider adding the basic kitty specific colors to it as well such as the tab bar colors and the border colors.

jokull:
I created realistic configs with 50+ colors (typical for a themed config) and measured complete `load_config()` time:

- Full config file with foreground, background, cursor colors, tab colors, borders, and 32 palette colors
- Parsed 100 times to get reliable averages
- Measured actual `load_config()` calls, not just isolated `to_color()` calls

**Results:**
```
Hex (baseline):   0.644ms avg, 0.606ms min
OKLCH (new):      0.720ms avg, 0.675ms min

Overhead:         0.076ms (11.9% slower)
```

If you prefer a C implementation for architectural consistency, I'm happy to do it.

Re: the theme editor - thank you! I'll send a PR to kitty-themes to add it - just need to work on it a bit.

kovidgoyal:
OK I will review and merge this when I next have some free time. Sadly
am in the middle of giant calibre feature right now so very limited time
for kitty.

PR #9361, opened 2026-01-07
Added C examples for kitty keyboard protocol

LaDorade:
Added example because I was struggling getting things to work...

### DIFF
--- a/docs/keyboard-protocol.rst
+++ b/docs/keyboard-protocol.rst
@@ -82,6 +82,7 @@ without too many changes, do the following:
 #. Emit the escape code ``CSI > 1 u`` at application startup if using the main
    screen or when entering alternate screen mode, if using the alternate
    screen.
+    #. In C for example: ``printf("\x1B[<1u")``
 #. All key events will now be sent in only a few forms to your application,
    that are easy to parse unambiguously.
 #. Emit the escape sequence ``CSI < u`` at application exit if using the main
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm afraid language specific examples are not appropriate here.

LaDorade:
> I'm afraid language specific examples are not appropriate here. 

I understand, so is there a means to have clearer examples/implementations ?

kovidgoyal:
What's the implementation here? You simply have to write an escape code
to the tty. The escape code is clearly described. How the user does it
in their language of choice is not something for kitty to document.

LaDorade:
> What's the implementation here? You simply have to write an escape code
> to the tty. The escape code is clearly described. How the user does it
> in their language of choice is not something for kitty to document.
> 

I guess it's just lack of knowledge on my part, but I really struggled and wanted to share an example with direct use of the sequence.
That's my bad

kovidgoyal:
On Wed, Jan 07, 2026 at 11:34:48PM -0800, Dorade wrote:
> LaDorade left a comment (kovidgoyal/kitty#9361)
>
> > What's the implementation here? You simply have to write an escape code
> > to the tty. The escape code is clearly described. How the user does it
> > in their language of choice is not something for kitty to document.
> >
>
> I guess it's just lack of knowledge on my part, but I really struggled and wanted to share an example with direct use of the sequence.

I sympathise, but adding such examples makes the documentation longer
which is worse for everyone else. The docs need to stay focused on
the things specific to the protocol, not general tips on how to program
for terminals.

PR #9472, opened 2026-02-04
Add JPEG XL (JXL) image format support

hjanuschka:
## Summary

Add support for decoding JPEG XL images using the libjxl library.

## Features

- Native libjxl decoding with multi-threaded support via libjxl_threads
- Works with `kitten icat` for displaying JXL images in the terminal
- Works with kitty graphics protocol (format=101)
- Outputs RGBA (4 bytes per pixel) format
- Color profile handling via libjxl's built-in support

## Dependencies

Requires libjxl and libjxl_threads to be installed:
- **macOS**: `brew install jpeg-xl`
- **Linux**: Install `libjxl-dev` (Debian/Ubuntu) or equivalent

## Testing

Tested locally with various JXL images. The following all work:
- `kitten icat image.jxl` - displays JXL images
- `kitten __convert_image__ RGBA < image.jxl` - converts JXL to RGBA
- Python: `load_jxl_data(data)` - decodes JXL bytes to RGBA

image


kovidgoyal:
icat already supports display jxl images if you have ImageMagick
installed. And the icat kitten cannot depend on native code since it is
meant to be a standalone binary.

Issue #9469, opened 2026-02-03
edit-in-kitty: detect if SSH connection has dropped before attempting to update remote file


waresnew:
**Is your feature request related to a problem? Please describe.**
Situation: When I use edit-in-kitty to edit a file over SSH and my internet cuts out, the SSH connection will close while I still have my local editor (eg. neovim) open with the file.

Trying to save the file after the disconnect makes the current shell session unusable (a notable symptom is that pressing keys like `j` will instead send 6;;106u). I'm guessing this is because kitty ends up directly sending the file transfer data to the local shell instead of sending it to the remote server.

Furthermore, since I may not notice that the SSH connection is gone (eg. because neovim fills the entire terminal), if I try to "save and quit", it will lead to the loss of unsaved changes because the data was not actually saved. The temporary local file gets deleted on quit too in this situation, so I can't recover my changes.

**Describe the solution you'd like**

I'm willing to try implementing a solution myself, but I just wanted to get your opinion on my ideas first.

When kitty tries to update the remote file, check if the connection has been closed. If so, then don't delete the temporary local file on quit (because saving after disconnecting still updates the local file properly). Then when the user quits, notify the user of that temporary local file's location. This way, the user won't lose any unsaved changes.

**Describe alternatives you've considered**

An alternative is to modify my own neovim config and detect if the current buffer is stored in a temporary directory (meaning that it's likely a file opened in edit-in-kitty), and then implementing some sort of backup system before attempting to quit the editor. I don't think it's easy to detect the SSH connection dropping from neovim alone, so this would have to activate based on the file location.

**Additional Context**

This is what's outputted when I try to edit a file, disconnect internet, save, then quit:

```
Waiting for editing to be completed, press Esc to abort...
Shared connection to [redacted] closed.
> KITTY_DATA_START
zsh: command not found: KITTY_DATA_START
> DONE
zsh: command not found: DONE
> KITTY_DATA_END
zsh: command not found: KITTY_DATA_END
```

kovidgoyal:
Feel free to try though from what I recall of the implementation, you wont find it easy to detect if the SSH connection has dropped robustly. Possibly the best way would be to record the pid of the current active child in the target window and check if that has changed when updating. You will need to experiment. Closing, feel free to follow up if you have more questions, or send a PR.

waresnew:
[...]

PR #9480, opened 2026-02-06
edit-in-kitty: Warn if SSH dropped + ssh: Restore keyboard flags

waresnew:
See #9469 for context

Notes:
- I ended up not comparing pid's directly because on further inspection, there is already a `child_is_remote` property which makes more sense to use
- Despite blocking any further `send_data()` calls after the ssh connection drops, the terminal window is still broken (nothing happens when typing anything) after SSH closes. My guess is that it's because the edit-in-kitty kitten doesn't get to send the `abort_msg`. After some experimenting, sending `\x1b[>0u` and a SIGINT will fix the window. I had to use `parse_bytes` from `kitty_tests` because writing the control sequence to stdout didn't work, probably because the window was in a broken state.

### DIFF
--- a/kittens/ssh/main.go
+++ b/kittens/ssh/main.go
@@ -736,12 +736,12 @@ func run_ssh(ssh_args, server_args, found_extra_args []string) (rc int, err erro
 	cd.hostname_for_match, cd.username = hostname_for_match, uname
 	escape_codes_to_set_colors, err := change_colors(cd.host_opts.Color_scheme)
 	if err == nil {
-		err = term.WriteAllString(escape_codes_to_set_colors + loop.SAVE_PRIVATE_MODE_VALUES + loop.HANDLE_TERMIOS_SIGNALS.EscapeCodeToSet())
+		err = term.WriteAllString(escape_codes_to_set_colors + loop.SAVE_PRIVATE_MODE_VALUES + loop.PUSH_KEY_FLAGS + loop.HANDLE_TERMIOS_SIGNALS.EscapeCodeToSet())
 	}
 	if err != nil {
 		return 1, err
 	}
-	restore_escape_codes := loop.RESTORE_PRIVATE_MODE_VALUES + loop.HANDLE_TERMIOS_SIGNALS.EscapeCodeToReset()
+	restore_escape_codes := loop.RESTORE_PRIVATE_MODE_VALUES + loop.POP_KEY_FLAGS + loop.HANDLE_TERMIOS_SIGNALS.EscapeCodeToReset()
 	if escape_codes_to_set_colors != "" {
 		restore_escape_codes += "\x1b[#Q"
 	}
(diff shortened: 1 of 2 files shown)


kovidgoyal:
That seems rather random. You should instead use the
clear_screen() method on the window with reset=True
and then send either SIGWINCH or SIGINT to have the shell redraw its
prompt.

waresnew:
I did try clear_screen(reset=True) before, but that clears the terminal scrollback as well, so the outputs of past commands in that window will be gone. Is that fine?

kovidgoyal:
Ah right, forgot about that. In that case use reset_termios_state() and
then restore modes via escape code,
https://sw.kovidgoyal.net/kitty/misc-protocol/#simple-save-restore-of-all-terminal-modes
since the edit-in-kitty kitten uses that escape code to save modes,
restoring them and termios state should be all that's needed. Although,
the ssh kitten probably is already doing this on connection drop, need
to check.

waresnew:
I tried resetting termios state and using parse_bytes to send `\x1b[?r` to restore modes, but it didn't fix the terminal.

Also just to clarify, the reason why I used parse_bytes is because I don't know how to send terminal control sequences to kitty otherwise, because sys.stdout.write just writes to the kitty process's own stdout.
 
I tested and the ssh kitten does send the restore modes sequence when the connection drops: https://github.com/kovidgoyal/kitty/blob/a7b8e880c91613907c655550669dbad4c04153ac/kittens/ssh/main.go#L751-L757 but that seems to not be enough in this case.

waresnew:
The window's key encoding flags are 1+4+8+16 when edit-in-kitty is running. If the ssh connection drops, it stays at 29. Since this includes flag 8, the user can't type properly. In normal usage, if you abort (press esc), it goes back to 0. Popping the stack once with `\x1b[<u` in my testing fixed it too. Are the key encoding flags also supposed to be updated when restoring terminal modes? Because in my testing, parse_bytes with `\x1b[?r` does trigger screen_restore_modes in the C code, but it doesn't fix the terminal and the key encoding flags remain as 29.

kovidgoyal:
On Sat, Feb 07, 2026 at 12:21:35PM -0800, Eric Liu wrote:
> waresnew left a comment (kovidgoyal/kitty#9480)
>
> The window's key encoding flags are 1+4+8+16 when edit-in-kitty is running. If the ssh connection drops, it stays at 29. Since this includes flag 8, the user can't type properly. In normal usage, if you abort (press esc), it goes back to 0. Popping the stack once with `\x1b[<u` in my testing fixed it too. Are the key encoding flags also supposed to be updated when restoring terminal modes? Because in my testing, parse_bytes with `\x1b[?r` does trigger screen_restore_modes in the C code, but it doesn't fix the terminal.

No it doesn't restore keyboard mode. The best solution for that is to
have the ssh kitten emit `\x1b[>u` before invoking ssh and `x1b[<u`
along with `x1b[?r` at exit. That does prevent the user from running
ssh with a keyboard mode other than legacy but this should not be a
problem since ssh knows only the legacy keyboard mode anyway.

waresnew:
That works, I made the changes now 👍 

A SIGINT is not needed anymore so I removed that part

PR #9485, opened 2026-02-07
Add Metal support for Apple Silicon Macs

NripeshN:
OpenGL is deprecated on MacOS. This PR attempts to port the existing codebase to metal


kovidgoyal:
This is on my TODO list as part of a more general refactor to support multiple GPU backends: Vulkan, OpenGL and Metal. But it will be done slowly over time. The first step will be movinf the shaders to slang so that they can be backend independent.

PR #9486, opened 2026-02-08
Support hide_window_decorations titlebar-only on Wayland

antoinecellerier:
Hide the CSD titlebar subsurface while keeping shadow borders for resizing. On SSD compositors (GNOME), forces CSD mode to draw kitty's own shadows without a titlebar.

Manually tested on the following:
   - GNOME Shell: 49.3
   - Wayland: 1.24.0 (libwayland-client)

This PR was authored by GitHub Copilot  / Opus 4.6. I'm not familiar with the kitty code base nor wayland window compositing but reading through the changes they feel reasonable. If you feel this is garbage, please let me know :)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -214,6 +214,11 @@ Detailed list of changes
 
 - edit-in-kitty: Handle connection drop more gracefully (:pull:`9480`)
 
+- Wayland: Add support for :code:`titlebar-only` in
+  :opt:`hide_window_decorations` to hide the titlebar while keeping shadow
+  borders for resizing. On compositors that use server-side decorations (such as
+  GNOME), this forces client-side decoration mode (:pull:`9486`)
+
 
 0.45.0 [2025-12-24]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/glfw/glfw.py
+++ b/glfw/glfw.py
@@ -323,6 +323,7 @@ def generate_wrappers(glfw_header: str) -> None:
     const char* glfwWaylandMissingCapabilities(void)
     void glfwWaylandRunWithActivationToken(GLFWwindow *handle, GLFWactivationcallback cb, void *cb_data)
     bool glfwWaylandSetTitlebarColor(GLFWwindow *handle, uint32_t color, bool use_system_color)
+    void glfwWaylandSetTitlebarHidden(GLFWwindow *handle, bool hidden)
     void glfwWaylandRedrawCSDWindowTitle(GLFWwindow *handle)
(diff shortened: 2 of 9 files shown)


kovidgoyal:
What manual testing did you do? In particular the main source of issues
with CSD is switching between full screen/docked/normal window modes.
Also what happens with right clicks in the shadow area?
Did you test what happens on an SSD supporting compositor?

If a value of titlebar-only disables SSD then a value of yes should
disable SSD too, if it doesnt already, dont remember. And the docs of
the option should mention that.

Finally update the changelog with an entry for this.

I am rather busy at the moment, with a whole queue of AI generated PRs
to review so it will be some time before I can actually look at the code
in your PR.

antoinecellerier:
I've mostly tested keyboard based resizing/tiling, mouse based resizing, changing the config vague and reloading (no effect). This is the list of scenarios copilot had come up with for me to validate:
   1. Shadows visible around the window edges with titlebar-only
   2. No titlebar — no title text, no close/maximize/minimize buttons
   3. Resize handles work — drag the shadow edges to resize
   4. Alt+drag (or Super+drag on GNOME) still moves the window
   5. Fullscreen (Ctrl+Shift+F11) hides shadows, exiting restores them

My understanding is that my Gnome / Wayland setup uses SSD but I'm trusting AI on this one ...

Let me check the other items and add missing docs / changelog entries to the pr.

antoinecellerier:
I've validated right clicking in the shadow area doesn't do anything & added extra doc and changelog entries. Happy to run any other tests if useful. thanks

kovidgoyal:
I'm confused. GNOME is almost unique amongst Wayland compositors for *not* supporting SSD. This needs to be tested on something like kwin or niri or sway or hyprland that do support SSD. Does using CSD there via this options cause any issues? Does setting this option to yes also force usage of CSD or does that happen for titlebar-only?

antoinecellerier:
Got it. maybe Opus was misrepresenting things to me. Will test it one one of those other options and report back.

antoinecellerier:
So just tested in KDE under kwin / wayland, everything looks good there too. Borders clearly show, resize handles work, right clicks are captured as one would expect. Anything else you'd find useful to test?

stewartmatheson:
This feature is something that I have been looking for. Thanks for taking the time to do this. I have built your branch and been testing it for the last few days under gnome. I have not seen any odd behavior in regard to resizing outside of  things gnome seems to do with all windows not just kitty. 

I also spent a bit of time building sway and I was able to test your branch in that WM. Not exactly sure what to be looking for but again nothing to me seems broken. It's worth noting however I have spent very little time in sway testing.

Screenshot from 2026-02-23 08-49-01

Screenshot from 2026-02-23 08-50-30

Issue #9514, opened 2026-02-16
Incorrect display of colors


MRC019:
**Describe the bug**
Incorrect display of colors

**To Reproduce**
1. Run kitty with custom colors configured
2. Press Ctrl+Shift+F6 
3. Check the colors against their codes with the color picker

**Screenshots**

<img width="1301" height="692" alt="Image" src="https://github.com/user-attachments/assets/61a372cf-4281-4249-8f55-a4e721ca2fd4" />

<img width="1920" height="1046" alt="Image" src="https://github.com/user-attachments/assets/29cb1bd1-f55a-475b-bad8-d48d5df4c86f" />
Same themes in ghostty (left) and in kitty (right)

**Environment details**
```
kitty 0.45.0 created by Kovid Goyal
Linux irbis 6.18.8-zen2-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Fri, 06 Feb 2026 06:47:23 +0000 x86_64
Arch Linux 6.18.8-zen2-1-zen (/dev/tty)
Running under: Wayland (Hyprland v0.53.3,) missing: blur icon
OpenGL: '4.6 (Core Profile) Mesa 25.3.5-arch1.1' Detected version: 4.6
Frozen: False
Fonts:
  medium: NotoSansMono-Regular: /usr/share/fonts/noto/NotoSansMono-Regular.ttf:0
          Features: ()
    bold: NotoSansMono-Bold: /usr/share/fonts/noto/NotoSansMono-Bold.ttf:0
          Features: ()
  italic: NotoSansMono-Regular: /usr/share/fonts/noto/NotoSansMono-Regular.ttf:0
          Features: ()
      bi: NotoSansMono-Bold: /usr/share/fonts/noto/NotoSansMono-Bold.ttf:0
          Features: ()
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /usr/bin/fish
System color scheme: dark. Applied color theme type: none
Loaded config files:
  /home/arch/.config/kitty/kitty.conf

Config options different from defaults:
background_opacity      0.5
Colors:
	active_border_color     #7aa2f7   
	active_tab_background   #7aa2f7   
	active_tab_foreground   #1f2335   
	background              #24283b   
	color0                  #1d202f   
	color1                  #f7768e   
	color10                 #9fe044   
	color11                 #faba4a   
	color12                 #8db0ff   
[...]

PR #9516, opened 2026-02-16
Fix: Increase sRGB LUT precision to 8 decimal places for more accurat…

vishalpatil-45:
Increased precision in srgb_lut.py from [X] to 8 decimal places. This ensures that the generated srgb_gamma.h provides more accurate color transformations, resolving the color mismatch reported in #9514.


kovidgoyal:
#9514 has absolutely nothing to do with this, it's a bug In Hyprland. https://github.com/hyprwm/Hyprland/discussions/12788  That said I see no harm in increasing precision here.

PR #9520, opened 2026-02-17
fix: blank screen on macOS resume from sleep

jorgemmsilva:


kovidgoyal:
Would it work to just not draw (aka return early) in the out of sequence render function when the framebuffer is not ready. Presumably if we do that, eventually macOS will repair the framebuffer? If not then it should use the context creation code in nsgl_context.m so as to avoid code duplication and ensure the new context is created with all the correct properties.

jorgemmsilva:
do you mean something like this?

```patch
diff --git a/kitty/glfw.c b/kitty/glfw.c
index 846e51900..4ae5d3b7e 100644
--- a/kitty/glfw.c
+++ b/kitty/glfw.c
@@ -346,6 +346,12 @@ window_iconify_callback(GLFWwindow *window, int iconified) {
 static void
 cocoa_out_of_sequence_render(OSWindow *window) {
     make_os_window_context_current(window);
+    GLenum fb_status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
+    if (fb_status != GL_FRAMEBUFFER_COMPLETE) {
+        window->needs_render = true;
+        request_tick_callback();
+        return;
+    }
     window->needs_render = true;
     bool rendered = false;
     if (window->fonts_data->sprite_map) rendered = render_os_window(window, monotonic(), true);
```

This was my first attempt, but that didn't work unfortunately :/

jorgemmsilva:
>If not then it should use the context creation code in nsgl_context.m so as to avoid code duplication and ensure the new context is created with all the correct properties.

👀 I'll try this

kovidgoyal:
On Wed, Feb 18, 2026 at 03:09:55AM -0800, Jorge Silva wrote:
> This was my first attempt, but that didn't work unfortunately :/

Yes, that's what I meant. OK if it doesnt work then your approach looks
OK, modulo the point about re-using as much existing context creation code
as possible.

Issue #9471, opened 2026-02-04
[bug] Rendering issue when scaling \uf444 with Menlo and mapped Symbol Nerd Fonts


broeknbytes:
**Describe the bug**
When rendering the glyph `\uf444` and centering vertically the dot gets chopped in half. Issue is not there when aligned top/bottom, and also issue is not there when starting with `--config NONE`. I'm hoping this is an error on the user part that can be easily solved. 

**To Reproduce**
Print `\uf444` vertically aligned top, bottom, middle:

```bash
printf "\e]66;n=6:d=7:v=0:w=1;\uf444\a\e]66;n=6:d=7:v=1:w=1;\uf444\a\e]66;n=6:d=7:v=2:w=1;\uf444\a\n"
```

**Screenshots**
<img width="1128" height="38" alt="Image" src="https://github.com/user-attachments/assets/b34f32e6-32d9-4653-aa32-f28a8496ec27" />

**Environment details**
```
kitty 0.45.0 (c26b770530) created by Kovid Goy

<img width="837" height="25" alt="Image" src="https://github.com/user-attachments/assets/cf53e112-232b-4c10-aa3e-50a645fdd704" />

al
Darwin MacBookProM1.local 24.6.0 Darwin Kernel Version 24.6.0: Wed Nov  5 21:33:58 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6000 arm64
ProductName:            macOS ProductVersion:           15.7.3 BuildVersion:            24G419
OpenGL: '4.1 Metal - 89.4' Detected version: 4.1
Frozen: True
Fonts:
  medium: Menlo-Regular: /System/Library/Fonts/Menlo.ttc
          Features: ()
    bold: Menlo-Bold: /System/Library/Fonts/Menlo.ttc
          Features: ()
  italic: Menlo-Italic: /System/Library/Fonts/Menlo.ttc
          Features: ()
      bi: Menlo-BoldItalic: /System/Library/Fonts/Menlo.ttc
          Features: ()
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /bin/zsh
System color scheme: light. Applied color theme type: none
Loaded config files:
  /Users/me/.config/kitty/kitty.conf

Config options different from defaults:
[...]

PR #9537, opened 2026-02-21
Fix offset cell overflow check

maxdexh:
Similar to #9535, this fixes a check that mistakenly uses the scaled cell width to check if a symbol offset would cause a cell to be overflown.

This at least fixes the cropping seen in #9471. I would not consider that issue fully fixed by this though, since the symbols are still being offset too far, which can be seen with `printf "X\e]66;n=5:d=7:h=0;\uf130\aX\n"` vs `printf "X\e]66;n=5:d=7:h=1;\uf130\aX\n"`.

### DIFF
--- a/kitty/core_text.m
+++ b/kitty/core_text.m
@@ -1046,7 +1046,8 @@ do_render(CTFontRef ct_font, unsigned int units_per_em, bool bold, bool italic, 
 }
 
 bool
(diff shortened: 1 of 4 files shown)


kovidgoyal:
render_glyphs_in_cells() has both freetype and core_text implementations. If you are changing the signature then you need to change the signature in the core text backend as well.

maxdexh:
Right, this is not rust, I can't just put an underscore at the start of the name to make the error go away ^^

maxdexh:
I hope just ignoring the parameter is okay. I do not know how the other module works.

maxdexh:
Why is the canvas size being upscaled if there is nothing written to it? Or is there?

kovidgoyal:
Well to fix the original issue the coretext backend will need to be
changed as well, I will take it a look at it when I have some time.

maxdexh:
I don't think that this is enough to fix the issue, since the offset is still too large to begin with. This just fixes the cropping caused by that. I have not yet looked into how the scaling works exactly and I cannot see a clear reason why this would be, since it seems to get smaller like it should, but not enough?

PR #9562, opened 2026-02-26
Allow n=d in the text sizing protocol

maxdexh:
As discussed before, the existing implementation already works for `n=d`. 
This PR just relaxes the recently added checks and updates the docs for the TSP to only require `n <= d`. 
I also took the liberty to make the other checks for the scaling parameters more consistent.

### DIFF
--- a/docs/text-sizing-protocol.rst
+++ b/docs/text-sizing-protocol.rst
@@ -75,11 +75,11 @@ There are only a handful of metadata keys, defined in the table below:
 
     "n", "Integer from 0 to 15", "0", "The numerator for the fractional scale."
 
-    "d", "Integer from 0 to 15", "0", "The denominator for the fractional scale. Must be ``> n`` when non-zero."
+    "d", "Integer from 0 to 15", "0", "The denominator for the fractional scale. Must be ``>= n`` when non-zero."
 
-    "v", "Integer from 0 to 2",  "0", "The vertical alignment to use for fractionally scaled text (n < d). ``0`` - top, ``1`` - bottom, ``2`` - centered"
+    "v", "Integer from 0 to 2",  "0", "The vertical alignment to use for fractionally scaled text (``n <= d``). ``0`` - top, ``1`` - bottom, ``2`` - centered"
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I dont see how this is supposed to work. As per the updated text sizing protocol docs alignment is not text alignment. It is alignment of the subscaled canvas inside the full size canvas. When n=d the two are the same, so alignment is a no-op. This was always the intended behavior, now the implementation matches it and the spec has been clarified to describe the behavior more closely. Closing but feel free to discuss further if you wish.

maxdexh:
When specifying w, the symbol is aligned inside the multicell, which is useful to align symbols in a tui, but it requires specifying d and n. Without support for this, I will have to switch to using `d=15:n=14` in my project for the next kitty version, which makes me sad.

kovidgoyal:
Again, alignment does not apply to symbols (or text), only to the *scaled area* in which they are rendered. Within that area alignment remains unaffected. What is aligned is the actual scaled area they are rendered in. That scaled area is the same size as the the full canvas when n==d. Therefore alignment cannot apply to it. Now in the actual implementation you get the illusion that it applies because what it does is use the MIN of the rendered width and scaled width rather than the scaled width which is in contravention to the spec, but present for historical reasons. This is pretty fragile behavior in any case as the "rendered_width" really depends on font and font engine. In FreeType for example it is the advance calculated from metrics in the glyph. The advance can be wildly inaccurate for this purpose as its semantic meaning is how much to move the pen forward when rendering another character after this one from the same font without an intervening space. But what we are doing is using this as a symbol and mixing text from different fonts. To make this use case robust you would need to have a whole different "symbol render mode" for the protocol. As I said before simply use w=2 for your mic and muted mic characters. Dont try to align them, aligning doesnt do what you think it does, and is not suitable for this use case.

maxdexh:
Hm, I was not aware that it is not intended/guaranteed behavior. I also thought I read that alignment applies within the cells the text occupies, but the docs don't actually state that.

I'll probably just go back to using an image.

Issue #9620, opened 2026-03-07
Segfault when dragging tab off-screen with `window_title_bar_min_windows 1`


mcrmck:
Dragging a tab off-screen to detach it causes a segfault on master (0a52972f5).

**Config**
```
window_title_bar top
window_title _bar_min_windows 1
```

**Repro (on macos)**
1. Open kitty with at least 2 tabs
2. Drag a tab outside the window to detach
3. Crash

**Backtrace (truncated)**
```
EXC_BAD_ACCESS (code=1, address=0x0)

frame #3: call_cell_program at shaders.c:1106
frame #5: draw_cells(srd=..., window=0x0000000000000000) at shaders.c:1211
frame #6: render_prepared_os_window(..., num_visible_windows=1) at child-monitor.c:882
```

The crash is at the title bar `draw_cells` call in `render_prepared_os_window` (line 882). During tab detach, t he source window drops to 1 visible window, but `trd->screen` is still non-NULL and the geometry check passes -- so it attempts to render with stale title bar data and hits a NULL dref in the GL driver.

PR #9621, opened 2026-03-07
Fix segfault when dragging tab to detach with window title bars enabled

mcrmck:
## Summary

Fixes a segfault when dragging a tab off-screen to detach it into a new OS window while `window_title_bar_min_windows 1` is set.

## Root cause

Two bugs in `set_window_title_bar_render_data` (`state.c`):

1. **Uninitialized `WindowGeometry.spaces`** — The local `WindowGeometry g` was declared without initialization. Only `left/top/right/bottom` are parsed from Python args; the `spaces` field contained stack garbage (observed values like `593040`, `11747456` vs expected single-digit padding values).

2. **Missing `reload_all_gpu_data` flag** — When a window is transferred between OS windows via tab detach, `attach_window` creates a new (empty) VAO for the title bar and clears `trd->screen`. Python's relayout then calls `set_window_title_bar_render_data` to reassign the screen, but without marking `screen->reload_all_gpu_data = true`, `send_cell_data_to_gpu` skips the upload (the screen's dirty flags reflect the old context). The render loop then calls `draw_cells` on an empty VAO buffer, causing a NULL deref in `gleRunVertexSubmitImmediate`. This matches the existing pattern for the main window screen in `attach_window` (line 450).

## Repro

```conf
window_title_bar top
window_title_bar_min_windows 1
```

1. Open kitty, create 2 tabs
2. Drag a tab outside the window to detach
3. Segfault

## Fix

- Zero-initialize `WindowGeometry g = {0}`
- Set `screen->reload_all_gpu_data = true` after `init_window_render_data`, matching the existing pattern in `attach_window`

### DIFF
--- a/kitty/state.c
+++ b/kitty/state.c
@@ -865,12 +865,13 @@ PYWRAP1(set_tab_bar_render_data) {
 }
 
 PYWRAP1(set_window_title_bar_render_data) {
-    WindowGeometry g;
+    WindowGeometry g = {0};
     id_type os_window_id, tab_id, window_id;
     Screen *screen;
     PA("KKKOIIII", &os_window_id, &tab_id, &window_id, &screen, &g.left, &g.top, &g.right, &g.bottom);
     WITH_WINDOW(os_window_id, tab_id, window_id)
         init_window_render_data(&window->window_title_render_data, g, screen);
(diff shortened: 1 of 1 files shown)


mcrmck:
Fixes #9620

Issue #9619, opened 2026-03-07
Window Title Bar Dragging


mcrmck:
Following up on @kovidgoyal 's comments in #9450 about draggable window title bars

> Yes draggable window title bars would be an important feature, possibly with a mappable action that when triggered shows the title bars and then auto hides them after the drag operation is completed... However, this belongs in a separate PR after this one is merged. It will require per layout code to translate drop locations into positions in the layout.

I'd like to work on this but want to get your guidance before going down the wrong path. A few questions:

**Per-layout drop logic:** This seems like the main complexity. For tabs, drop position is linear - for window splits, the user could mean "insert left/right/above/below." Do you have a preference for how this should work? Some options:
- Drop onto a window's title bar --> swap positions
- Drop onto edges of a window --> insert in that direction (splits-style)
- Someting simpler to start?

**Layout scope:** Should this work across all layouts from the start, or is it reasonable to start with splits and extend later?

**Show-then-hide UX:** You mentioned a mappable action that temporarily shows the title bars for a drag, then auto-hides. Should this be the primary interaction model, or should it also work when title bars are permanently visible?

**Cross-tab / cross-OS-window:** Tab dragging supports moving tabs between OS windows. Should window dragging support moving a window to a different tab, or is within-tab reordering sufficient for a first pass?

**MIME type / drag data:** I assume this would just follow the same pattern as tab dragging (`application/net.kovidgoyal.kitty-window-{PID}` or similar) -- anything else you'd want in the payload?

Happy to put in the work, just want to make sure the approach aligns with your vision. No rush if this is not a priority, or if there is more work to be done on tab dragging first.

kovidgoyal:
On Sat, Mar 07, 2026 at 10:53:16AM -0800, mcrmck wrote:
> mcrmck created an issue (kovidgoyal/kitty#9619)
>
> Following up on @kovidgoyal 's comments in #9450 about draggable window title bars
>
> > Yes draggable window title bars would be an important feature, possibly with a mappable action that when triggered shows the title bars and then auto hides them after the drag operation is completed... However, this belongs in a separate PR after this one is merged. It will require per layout code to translate drop locations into positions in the layout.
>
> I'd like to work on this but want to get your guidance before going down the wrong path. A few questions:
>
> **Per-layout drop logic:** This seems like the main complexity. For tabs, drop position is linear - for window splits, the user could mean "insert left/right/above/below." Do you have a preference for how this should work? Some options:
[...]

PR #9626, opened 2026-03-09
Add draggable window title bars

mcrmck:
Follow-up to #9450 and #9619. Implements drag-to-reorder for window title bars, as discussed with @kovidgoyal in those threads.

### What this does
- **Drag a window title bar** to reorder or move windows:
    - Drop on another window's title bar --> swap positions
    - Drop on another window's body quadrant (left/right/top/bottom half) --> insert as a split in that direction
    - Drop on a tab bar tab --> move the window into its active tab
    - Drop outside kitty --> detach into a new OS window
- **Tab highlight during drag** - when hovering over the tab bar while dragging a window, the target tab is highlighted using active-tab styling, mirroring how the destination window for the title bar is highlighted.
- **`toggle_window_title_bars` action** - a mappalbe action that temporarily force-shows title bars to allow drag-to-reorder, then autho-hides them when the drag completes. Useful when `window_title_bar_min_windows` means bars are normally hidden.
- **`window_title_bar_drag_threshold`** - pixel distance the mouse must travel before a drag is initiated (default 5). Set to 0 to disable dragging.

### Implementation notes
The approach mirrors tab dragging closely:
- `window_being_dragged` C state struct mirrors `tab_being_dragged`
- Same MIME type convetion: `application/net.kovidgoyal.kitty-window-{PID}`
- Same threshold-based drag start, same thumbnail rendering, same `on_drop` / `on_drag_source_finished` split for local vs. cross-window drops

`Splits` layout has a dedicated `insert_window_next_to()` that properly repositions a window in the pair tree. Other layouts fall back to `move_window_to_group`.

### Known limitations + areas I'd welcome feedback on
- The directional insert fallback fro non-Splits layouts (Stack, Tall, Fat, etc.) uses `move_window_to_group`, which moves to the target's group position but doesn't create a true split. I wasn't sure how to express "insert to the left of this window" generically across all layouts -- would welcome guidance on whether a per-layout method is the right approach or if there's a better abstraction.
- Cross-OS-window drops rely on the destination OS window receiving an `on_drop` event. The fallback in `on_drag_source_finished` creates a new OS window when the WBD state hasn't been cleared by successful 'on_drop', which should cover the "dropped outside kitty" case.
- The `toggle_window_title_bars` action currently affects all tabs in the active tab manager. It may be more correct to scope it to only the active tab.

I tried my bbest here but this is a genuinely complex feature and I expect there are things you'll want to adjust. Happy to rework any part of this based on your feedback.

### Test Plan

  - [ ] Drag window title bar, drop on another title bar → windows swap
  - [ ] Drag window title bar, drop on left/right/top/bottom body quadrant → directional split insert
  - [ ] Drag to tab bar, hover over different tab → tab highlights; release → window moves to that tab
  - [ ] Drag to another OS window → window moves into that window's active tab
  - [ ] Drag outside kitty → new OS window created
  - [ ] `toggle_window_title_bars` shows/hides bars; drag completes; bars auto-hide
  - [ ] `window_title_bar_drag_threshold = 0` disables dragging
  - [ ] 1-row windows: no title bar shown, no crash
  - [ ] Config reload while title bars visible: correct relayout

Ref: #9619

### DIFF
--- a/kitty/layout/splits.py
+++ b/kitty/layout/splits.py
@@ -2,7 +2,7 @@
 # License: GPLv3 Copyright: 2020, Kovid Goyal <kovid at kovidgoyal.net>
 
 from collections.abc import Collection, Generator, Iterator, Sequence
-from typing import Any, ClassVar, Literal, Optional, TypedDict, Union
+from typing import Any, Optional, TypedDict, Union
 
 from kitty.borders import BorderColor
 from kitty.fast_data_types import BOTTOM_EDGE, LEFT_EDGE, RIGHT_EDGE, TOP_EDGE
--- a/kitty/layout/tall.py
+++ b/kitty/layout/tall.py
@@ -4,7 +4,7 @@
 import sys
 from collections.abc import Generator, Iterator, Sequence
 from itertools import islice, repeat
-from typing import Any
+from typing import Any, ClassVar, Literal
 
 from kitty.borders import BorderColor
 from kitty.conf.utils import to_bool
@@ -136,7 +136,7 @@ class Tall(Layout):
     name = 'tall'
     main_is_horizontal = True
     no_minimal_window_borders = True
-    drag_overlay_mode = 'axis_y'
+    drag_overlay_mode: ClassVar[Literal['full', 'axis_y', 'axis_x', 'free']] = 'axis_y'
     layout_opts = TallLayoutOpts({})
     main_axis_layout = Layout.xlayout
     perp_axis_layout = Layout.ylayout
--- a/kitty/layout/vertical.py
+++ b/kitty/layout/vertical.py
@@ -2,7 +2,7 @@
 # License: GPLv3 Copyright: 2020, Kovid Goyal <kovid at kovidgoyal.net>
 
 from collections.abc import Generator, Iterable
-from typing import Any
+from typing import Any, ClassVar, Literal
 
 from kitty.borders import BorderColor
 from kitty.types import Edges, NeighborsMap, WindowMapper
@@ -64,7 +64,7 @@ class Vertical(Layout):
     name = 'vertical'
     main_is_horizontal = False
     no_minimal_window_borders = True
-    drag_overlay_mode = 'axis_y'
+    drag_overlay_mode: ClassVar[Literal['full', 'axis_y', 'axis_x', 'free']] = 'axis_y'
     main_axis_layout = Layout.ylayout
     perp_axis_layout = Layout.xlayout
 


kovidgoyal:
Approach sounds more or less correct, as you say it is a complex feature, so will review code when I have a chunk of free time.

dannyfiresnake:
Just tried this patch out. Amazing @mcrmck!

Bug report: I could not drop on another window's title bar to swap positions.

Enhancement idea 1: It could use some overlays on the window being dragged to so you know where the split is ending up before you drop.

Enhancement idea 2: Dragging onto the tab bar outside the tabs would move the window to a newly created tab.

mcrmck:
Thanks for testing @dannydulai , and for the detailed feedback

**Bug fix:** I believe I've fixed the title bar swap. The root cause was that `start_window_drag` was missing the code to force-show title bars on destination windows during the drag -- the cleanup to _clear_ `force_show_title_bars` already existed (in both `on_window_drop` and `on_drag_source_finished`) but the corresponding code to _set_ it at drag start was never written. So unless title bars were already naturally visible via `window_title_bar_min_windows`, there was nothing to drop onto. I also fixed a related UX issue  where the title bar highlight was appearing whenever the cursor was anywhere over the destination window, not just its title bar row - which gave misleading feedback. Please let me know if the swap is still not working, I'm happy to give it a more thorough look.

**Enhancement 2 implemented:** Dropping a window onto the empty space in the tab bar (past the last tab) now creates a new tab and moves the window into it.

**Enhancement 1 -- seeking guidance before implementing:** I agree that a drop preview overlay (showing which quadrant a body drop will split into) would be highly valuable. Looking at the codebase, the rendering pieces are actually all there -- there's an existing `ROUNDED_RECT_PROGRAM` shader that supports semi-transparent drawing with alpha blending, and the `set_border_rects` pattern shows a proven Python --> C --> GPU pipeline for passing geometry. The quadrant detection logic already exists in `on_window_drop_move`. So the main question isn't whether its feasible, it's which architectural papproach is preferred: reusing the existing borders infrastructure, add a lightweight dedicated overlay path, or call `ROUNDED_RECT_PROGRAM` directly from the drag render pass. I didn't want to go down one of these paths and have it rejected on architectural ground, so -- @kovidgoyal , welcome your input on path forward.

kovidgoyal:
Not sure what you mean by re-using the borders infrastructure? You want to draw a solid colored rectangle over the target window? IMO the visual_bell infrastructure is better suited for this, you could even have the highlight be animated fading in as the drag enters the window. I dont recall the details of this so it might need a bit of plumbing but not very much, I think.

mcrmck:
You're right, that makes much more sense -- `draw_visual_bell` is already doing exactly this per-window in the render pass via `TINT_PROGRAM`, just with the edges uniform covering the full window. For the drag preview I'd constrain edges to the target quadrant and hold the intensity while hovering (rather than timing out). I'll take a look at what plumbing is needed on the Screen struct side and give it a go.

mcrmck:
Implemented Enhancement 1 -- semi-transparent tint overlay onthe destination window during a drag. Followed the `visual_bell` path as suggested: reusing `TINT_PROGRAM` in `shaders.c`, two new `Screen` fields (`start_drag_overlay_at`, `drag_overlay_quadrant`), and mirroring the `child-monitor.c` `needs_render` / `set_maximum_waits` hooks that visual bell already uses to keep frames firing during animation. Fades in over 150ms, clears immediately on drop or exit.

One thing I noticed: the right overaly shape depends on the layout. For most layouts (Tall, Stack, etc.) a body drop is a reorder/swap, so we highlight the full window. For the Splits layout it's a real directional insert, so we show just the target half. I'm using `hasattr(layout, 'insert_window_next_to')` to distinguish them since that's the same gate `_insert_window_in_direction` uses internally -- happy to hear if there's a cleaner way to introspect that.

Performance-wise: when no drag is active, `draw_drag_preview_overlay` is a two-field read early-return per window frame, and the `child-monitor.c` additions are two integer comparisons sitting right next to the identical `start_visual_bell_at` checks. Active-drag cost is identical to visual bell.

dannyfiresnake:
Tried these out, the overlays make the whole thing feel so much better. Great work!

I did run into a minor issue: when I am in 'splits' layout, and I drag from one window to the title of another, the swap works properly, but the highlight looks like its going to split to the top. I would think it would be appropriate to go full covered highlight in this situation.

The drag to tab-bar to create a new tab works well, but it lacks a visual indicator of what it's going to do. @kovidgoyal, feelings on what that might look like? A temporary new tab maybe while hovering?

kovidgoyal:
On Thu, Mar 19, 2026 at 06:50:30PM -0700, Danny wrote:
> The drag to tab-bar to create a new tab works well, but it lacks a visual indicator of what it's going to do. @kovidgoyal, feelings on what that might look like? A temporary new tab maybe while hovering?


There are two cases. Dropping onto an existing tab, where there are two
options:

1) Disallow the drop.
2) add the window to the tab at some default position and give it focus.

If we opt for (2) then there needs to be some indicator, possibly just
temporarily change the title of the tab to "Add to tab" or maybe drawing
a drop indicator emoji before the rest of the tab title, or changing the
tab background color. However all of these suffer from the issue that
when the user is using custom tab drawing we cant really control what
gets drawn there.

As for dropping in empty, one could draw a fake tab though that will be
more work, simply drawing some text or drop indicator emoji (like a
fancy down arrow or a fancy plus) after the last tab should be enough.

dannyfiresnake:
If you drop onto a different existing tab, it highlights while being hovered over. I felt this this made sense and works well.

If you drop onto he same tab the window was being dragged from, it does nothing on hover. Dropping reorders the windows on that tab (potentially). I would be ok with it doing nothing.

Dragging on to the bar itself showing a "+" indicator would be enough feedback, but a new color needs to be introduced for that indicator. I couldn't find a foreground color property for the non-tabbed part of the tab bar. It's possible I missed it.

If you add a new fake tab, it could easily have a title of + as well, and that'd look pretty great.

Looking forward to seeing what you implement @mcrmck.

kovidgoyal:
On Thu, Mar 19, 2026 at 08:18:37PM -0700, Danny wrote:
> dannydulai left a comment (kovidgoyal/kitty#9626)
>
> If you drop onto a different existing tab, it highlights while being hovered over. I felt this this made sense and works well.
>
> If you drop onto he same tab the window was being dragged from, it does nothing on hover. Dropping reorders the windows on that tab (potentially). I would be ok with it doing nothing.
>
> Dragging on to the bar itself showing a "+" indicator would be enough feedback, but a new color needs to be introduced for that indicator. I couldn't find a foreground color property for the non-tabbed part of the tab bar. It's possible I missed it.

You can use any colored emoji like 💧 or a colored plus sign or similar.

mcrmck:
I drafted an implementation of the "+" indicator for the empty tab bar space -- it shows up as a synthetic tab (using the existing inactive/active tab styling) as soon as the drag threshold is crossed and dissappears immediately on drop.

On the existing-tab drop question: right now hovering over an existing tab highlights it and dropping it moves the window into that tab. No extra indicator beyond the highlight.

My inclination is to leave it there and call it done for now. The highlight is already meaningful feedback, and any text-mutation approach (changing the tab title to "Add to tab", prepending an emoji, etc.) breaks with custom tab renderers. The empty-bar drop covers the primary "move to a different tab" use case anyway.

That said, if you'd prefer we disallow existing-tab drops entirely, that's equally clean. I'm also open to additional ideas if there are any.

Cheers.

dannyfiresnake:
Screenshot 2026-03-22 at 12 20 45 AM

I'm seeing a "? +" because my `tab_title_template` has a single char to show the layout, and it shows "?" before the title because there is an unrecognized `layout_name`. Any way to ignore the `tab_title_template` for that synthetic tab?

While we are at it, any way to keep that synthetic tab on the screen always? It'd a really nice tab for clicking to add a new window on a new tab.

mcrmck:
Thanks for the feedback, made both those tweaks in 412cd30.

dannyfiresnake:
Looks great!

I'm very happy with the state of this now. I'll report back if I find any other issues.

Even if I like it, I think forcing the + tab is not great for kitty users as a whole... probably want to put that on a new setting, like `tab_bar_show_add`... @kovidgoyal ?

If user has it hidden, you can just show it when dragging a tab around.

kovidgoyal:
On Sun, Mar 22, 2026 at 01:29:00PM -0700, Danny wrote:
> Even if I like it, I think forcing the + tab is not great for kitty users as a whole... probably want to put that on a new setting, like `tab_bar_show_add`... @kovidgoyal ?

Yes, this shouldn't default to on.

mcrmck:
`tab_bar_show_new_tab_button` is implemented, and defaults to off. Thanks for the suggestions @dannydulai . @kovidgoyal please let me know if there's anything else I can improve.

kovidgoyal:
Looks fine to me I just need to find the chunk of time to review it and merge.

kovidgoyal:
Issues I noticed:

1) When two OS windows are present, dragging a window over another in the first OS Window highlights both the target window and the active window in the second OS Window. Dragging it over the second OS Window causes highlighting to fail not showing up over any window

2) Similarly the + tab does not show up in the tab bar of the second OS Window

3) This one is a little tricky to reproduce but is happening consistently for me. Dragging a window in one continuous movement over the + tab bar and then into a second OS Window causes a new window to be spawned in the first OS Window, even before the drop is performed.

All these test were done in Linux Wayland with --config=NONE -o 'map f1 toggle_window_title_bars'

dragging was 

There were various other miscellaneous misbehaviors I noticed but likely they have the same underlying cause of confusion when there are multiple OS Windows. Will test again once that is fixed.

mcrmck:
Ok I believe I found (at least part of) the issue here:

In `tabs.py` if a drag was initiated while `tab_bar_show_new_tab_button` is enabled (or any window drag was active), the synthetic `+` tab (`tab_id = -1`) would be included in `tab_being_dropped.tab_ids`. The next `tab_bar_data` call would then `KeyError: -1` when looking up `tmap[-1]`, since `tmap` only contains real tabs.

Additionally, in `boss.py` `on_drop_move`, all tab managers were receiving `is_dest=True` with the drag coordiantes from whatever OS window was being hovered. Since every OS window uses window-local coordinates starting at `(0,0)`,  `(x,y)` from Window 1 maps into the exact same position inside OS Window 2's viewport. `_find_window_at` and the tab bar check both uses this local coordinate space -- so a drag at `(100, 200)` in Window 1 would find and highlight a window at `(100, 200)` in Window 2 as well. Fix: `is_dest=True` is now only passed to the tab manager that owns the OS window receiving the drag event; all others get `is_dest=False` and clear their state.

This should hopefully address bugs 1 and 2. I believe it may also resolve bug 3, since the prematuer new-window dcreation was most likely caused by incorrect drop-target state bleeding into OS Window 2's tab manager -- but please let me know.

Thanks for giving this a look.

kovidgoyal:
#3 is still occurring for me.

Also there needs to be the ability to drop in the top/bottom or
left/right halves of a window in the vertical/horizontal and tall/fat
layouts. And in the grid layout it should be upper/lower halves I think.

kovidgoyal:
I wouldnt worry about 3 I traced it and discovered its caused by a bug in wayland, not related to this PR.

kovidgoyal:
3 is fixed by: https://github.com/kovidgoyal/kitty/commit/215bbf68de2dbabee955f5ab721d285a279ef470

mcrmck:
I gave this a stab in 6c37c1c.

**Architecture:** Added a `drag_overlay_mode` class variable to the `Layout` base class (`'full'` | `'axis_y'` | `'axis_x'` | `'free'`). Each layout declares its axis in one line; `tabs.py` and `boss.py` dispatch on it — no isinstance checks anywhere.

**Two new primitives:**
- `WindowList.insert_window_group_next_to(target_group_id, after)` — positional insert (pop + re-insert), which is what the non-Splits layouts actually needed. The existing move_window_group only swaps.
- `Layout.insert_window_next_to(...)` — base implementation that uses the new primitive for `axis_x`/`axis_y` layouts and falls back to swap for `'full'` (Stack unchanged). Splits' tree-based override is untouched.

`_insert_window_in_direction` in `boss.py` goes from a 7-line `hasattr` branch to one line.

Direction assignments:
- Vertical, Tall, Grid → top/bottom (axis_y)
- Horizontal, Fat → left/right (axis_x)
- Splits → free 4-way (unchanged)
- Stack → full-window swap (unchanged)

Please let me know what you think, both from an implementation and user-experience point of view.

Issue #9381, opened 2026-01-14
None English layout shortcuts problem


Robanni:
**Describe the bug**  
When pressing Ctrl+Shift (or similar shortcuts) with Russian keyboard layout active, instead of expected shortcut behavior or Russian characters, kitty outputs terminal keycodes like `84;6u`, `89;6u`, `96;6u`.

**To Reproduce**  
1. Switch to Russian keyboard layout  
2. Press Ctrl + Russian letter keys (e.g., Ctrl+shift+я → `84;6u`, Ctrl+shift+ё → `89;6u`, Ctrl+shift+у → `96;6u`)  
3. See raw keycodes instead of shortcut processing or Russian characters  

<img width="326" height="55" alt="Image" src="https://github.com/user-attachments/assets/4236178e-4ff2-4ca7-bfc9-82690ea8ef35" />

**Environment details**  
```
kitty 0.45.0 created by Kovid Goyal
Linux localhost.localdomain 6.18.4-1-default #1 SMP PREEMPT_DYNAMIC Thu Jan  8 09:54:45 UTC 2026 (5b8e683) x86_64
Running under: X11
OpenGL: '4.6 (Core Profile) Mesa 25.3.3' Detected version: 4.6
Frozen: False
Fonts:
  medium: SourceCodePro-Regular: /usr/share/fonts/truetype/SourceCodePro-Regular.otf:0
          Features: ()
    bold: SourceCodePro-Semibold: /usr/share/fonts/truetype/SourceCodePro-Semibold.otf:0
          Features: ()
  italic: SourceCodePro-It: /usr/share/fonts/truetype/SourceCodePro-It.otf:0
          Features: ()
      bi: SourceCodePro-SemiboldIt: /usr/share/fonts/truetype/SourceCodePro-SemiboldIt.otf:0
          Features: ()
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib64/kitty
  extensions dir: /usr/lib64/kitty/kitty
  system shell: /usr/bin/zsh
System color scheme: dark. Applied color theme type: none
Loaded config files:
  /home/robanni/.config/kitty/kitty.conf

Config options different from defaults:

Important environment variables seen by the kitty process:
	PATH                                /home/robanni/.local/bin:/usr/local/bin:/usr/bin:/bin
	LANG                                ru_RU.UTF-8
	SHELL                               /usr/bin/zsh
	DISPLAY                             :0
	USER                                robanni
[...]

PR #9671, opened 2026-03-16
Add per-mapping --allow-fallback for layout-independent shortcuts

keplian:
## Summary

Add `--allow-fallback` option to the `map` command for per-mapping control over keyboard shortcut fallback behavior. When a non-Latin layout (Russian, Arabic, etc.) is active, shortcuts like `ctrl+c` stop working. This option enables fallback to the physical key position (`alternate_key` from the keyboard protocol).

The option controls both fallback types:
- `shifted` — shifted_key fallback (default, preserves existing behavior)
- `ascii` — alternate_key fallback with non-ASCII guard (new)
- `shifted,ascii` — both (used by default kitty bindings)

Default kitty bindings updated with `--allow-fallback=shifted,ascii`. User custom mappings default to `shifted` only (backward compatible). The ascii fallback only triggers when the current layout key is non-ASCII (codepoint > 127), so Dvorak/Colemak users are unaffected.

## Changes

**Python:**
- `allow_fallback` field on `KeyMapOptions` with validation
- Per-mapping speculative lookup in `get_shortcut()` and `shortcut_matches()`
- `alternate_key` support in `KeyEvent.matches()` for Python kittens
- All default bindings updated with `--allow-fallback=shifted,ascii`

**Go:**
- `AllowFallback` on `KeyAction` and `ParsedShortcut`
- `ParseMap()` handles `--allow-fallback=value` flag parsing
- Two matching modes: `MatchesParsedShortcutWithFallback()` (per-mapping for ShortcutTracker) and `MatchesParsedShortcut()` (unconditional with non-ASCII guard for hardcoded shortcuts)
- Code generators updated to emit `AllowFallback` values

**Kitten migrations:**
- `themes` and `command_palette` migrated to configurable shortcuts via ShortcutTracker
- `diff` and `choose_files` bindings updated with `--allow-fallback=shifted,ascii`

## Test plan
- [x] Python tests: 5 new test methods covering per-mapping fallback, parsing, non-ASCII guard
- [x] Go tests: ParseMap with --allow-fallback + AlternateKey matching
- [x] `./dev.sh build` passes
- [x] Manual testing with Russian layout: kitty shortcuts and kitten shortcuts work
- [x] Dvorak safety: non-ASCII guard prevents fallback for ASCII keys

Relates to #9381

### DIFF
--- a/docs/mapping.rst
+++ b/docs/mapping.rst
@@ -274,6 +274,55 @@ for copying to clipboard.
    and you cannot have mappings with and without conditions applying to multi-keys
    with the same first key.
 
+Non-Latin keyboard layout support
+--------------------------------------
+
+When using a non-Latin keyboard layout (e.g. Russian, Arabic, Greek), letter-key
+shortcuts like :kbd:`Ctrl+C` stop working because the key produces a non-Latin
+character. kitty solves this with the ``--allow-fallback`` option on the ``map``
+directive, which controls how shortcuts fall back to the physical key position.
+
+The ``--allow-fallback`` option accepts a comma-separated list of fallback types:
+
+``shifted``
+    Fall back to the *shifted key* — the character produced when Shift is held
+    with the key. This is the default for all mappings and preserves the existing
+    behavior.
+
+``ascii``
+    Fall back to the *alternate key* — the character that the physical key would
+    produce in a standard US layout. This only triggers when the key produces a
+    non-ASCII character, so it has no effect on Latin-based layouts like Dvorak
+    or Colemak.
+
+``none``
+    Disable all fallback matching. The mapping will only match the exact key
+    specified, ignoring both shifted and alternate key positions.
+
+For example::
+
+    # Enable both shifted and ASCII fallback (used by default kitty shortcuts)
+    map --allow-fallback=shifted,ascii kitty_mod+c copy_to_clipboard
+
+    # Only ASCII fallback, no shifted key fallback
+    map --allow-fallback=ascii ctrl+s save_something
+
+    # Disable all fallback (neither shifted nor alternate key matching)
+    map --allow-fallback=none ctrl+x some_action
+
+All default kitty shortcuts use ``--allow-fallback=shifted,ascii``, so they work
+out of the box with non-Latin layouts. Custom mappings without an explicit
+``--allow-fallback`` get the default value of ``shifted``, which preserves
+backward compatibility.
+
+.. note::
+
+   The ``ascii`` fallback uses a non-ASCII guard: it only activates when
+   the key produces a character with a Unicode code point above 127. This means
+   alternative Latin layouts (Dvorak, Colemak, etc.) are never affected by the
+   ``ascii`` fallback — only non-Latin layouts trigger it.
+
+
 Sending arbitrary text or keys to the program running in kitty
 --------------------------------------------------------------------------------
 
--- a/kittens/choose_files/main.py
+++ b/kittens/choose_files/main.py
@@ -156,15 +156,15 @@ map('Change to root directory', 'cd_root ctrl+/ cd /')
 map('Change to home directory', 'cd_home ctrl+~ cd ~')
 map('Change to home directory', 'cd_home ctrl+` cd ~')
 map('Change to home directory', 'cd_home ctrl+shift+` cd ~')
-map('Change to temp directory', 'cd_tmp ctrl+t cd /tmp')
+map('Change to temp directory', 'cd_tmp --allow-fallback=shifted,ascii ctrl+t cd /tmp')
 
-map('Next filter', 'next_filter ctrl+f 1')
-map('Previous filter', 'prev_filter alt+f -1')
+map('Next filter', 'next_filter --allow-fallback=shifted,ascii ctrl+f 1')
+map('Previous filter', 'prev_filter --allow-fallback=shifted,ascii alt+f -1')
 
-map('Toggle showing dotfiles', 'toggle_dotfiles alt+h toggle dotfiles')
-map('Toggle showing ignored files', 'toggle_ignorefiles alt+i toggle ignorefiles')
-map('Toggle sorting by dates', 'toggle_sort_by_dates alt+d toggle sort_by_dates')
-map('Toggle showing preview', 'toggle_preview alt+p toggle preview')
+map('Toggle showing dotfiles', 'toggle_dotfiles --allow-fallback=shifted,ascii alt+h toggle dotfiles')
+map('Toggle showing ignored files', 'toggle_ignorefiles --allow-fallback=shifted,ascii alt+i toggle ignorefiles')
+map('Toggle sorting by dates', 'toggle_sort_by_dates --allow-fallback=shifted,ascii alt+d toggle sort_by_dates')
+map('Toggle showing preview', 'toggle_preview --allow-fallback=shifted,ascii alt+p toggle preview')
 
 egr()  # }}}
 
--- a/kittens/command_palette/main.go
+++ b/kittens/command_palette/main.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/kovidgoyal/kitty/tools/cli"
+	"github.com/kovidgoyal/kitty/tools/config"
 	"github.com/kovidgoyal/kitty/tools/tty"
 	"github.com/kovidgoyal/kitty/tools/tui"
 	"github.com/kovidgoyal/kitty/tools/tui/loop"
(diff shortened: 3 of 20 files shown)


kovidgoyal:
I am somewhat sceptical of this, for keyboard layouts that remap keys
like dvorak or colemak or similar this can have very unexpected side
effects, pressing ctrl+x could instead match ctrl+c if there is no
mapping for ctrl+x and the layout maps c to x.

To make this safer it should probably only apply when the key sent by
the layout is non-ascii. In such cases it should be fairly safe to
fallback to the base layout key when the current layout does not match.

Furthermore to make this fully robust, the map command should get a
--allow-fallback=shifted,ascii or similarly named option so individual mappings
can be controlled to have fallback or not. The default can be
shifted. And the default bindings can be adjusted to use shifted,ascii.

Then we dont need a global kitty.conf option and users with non-ascii
keyboard layouts can have a somewhat better experience, though really
there are no guarantees that doing this will work well across
languages/layouts.

keplian:
Good point about dvorak/colemak. I've been thinking about this more and
  I think the simplest solution is to just always fallback to alternate_key
  when the current layout key is non-ascii, without any config option at
  all. The C code in key_encoding.c already does exactly this for legacy
  encoding unconditionally, so it makes sense to mirror that in the python
  shortcut matching.

  This way it just works out of the box for non-latin layout users, its
  completely safe for dvorak/colemak since their keys are all ascii and the
  guard prevents fallback, and there is no new config option to maintain.
  Direct key matches still take priority, so if someone explicitly maps
  ctrl+в to a different action it will still work.

  One question on the non-ascii boundary — should this use the same
  check as is_legacy_ascii_key() in key_encoding.c, or something broader?
  For example french layout produces é on the 2 key, which is > 127,
  so the fallback would trigger there. Is that desirable or should it
  be limited to clearly non-latin scripts?

  Regarding per-mapping --allow-fallback, I see the value for fine-grained
  control, but the go side currently has no infrastructure for per-mapping
  options — ParseMap() does naive string splitting, KeyAction has no
  option fields, and MatchesParsedShortcut ignores AlternateKey entirely.
  Plus 13+ hardcoded go shortcuts (ctrl+c, ctrl+d etc) cant benefit from
  per-mapping flags at all. This could be a good follow-up if the
  granularity turns out to be needed.

kovidgoyal:
On Mon, Mar 16, 2026 at 05:17:13AM -0700, Pavel Meshalkin wrote:
> keplian left a comment (kovidgoyal/kitty#9671)
>
> Good point about dvorak/colemak. I've been thinking about this more and
>   I think the simplest solution is to just always fallback to alternate_key
>   when the current layout key is non-ascii, without any config option at
>   all. The C code in key_encoding.c already does exactly this for legacy
>   encoding unconditionally, so it makes sense to mirror that in the python
>   shortcut matching.
>
>   This way it just works out of the box for non-latin layout users, its
>   completely safe for dvorak/colemak since their keys are all ascii and the
>   guard prevents fallback, and there is no new config option to maintain.
>   Direct key matches still take priority, so if someone explicitly maps
>   ctrl+в to a different action it will still work.

Yes, that is basically what I was proposing, except I want it
configurable for individual mappings. This is needed to mitigate risk
that we cause breakage. If some particular shortcut is problematic, we
can simply have it not fallback easily with this infrastructure. I dont
want to be scrambling building the infrastructure after a problem is
discovered. Also IMO user specified shortcuts should not fallback by
default, since users will specify the actual key they want matched. In
order to make that happen, the scheme I proposed above is needed.

>
>   One question on the non-ascii boundary — should this use the same
>   check as is_legacy_ascii_key() in key_encoding.c, or something broader?
>   For example french layout produces é on the 2 key, which is > 127,
>   so the fallback would trigger there. Is that desirable or should it
>   be limited to clearly non-latin scripts?

Not sure I follow. If the key event is ctrl+é it should fallback to
matching ctrl+2, but ctrl+2 should not fallback to matching ctrl+x.

>
>   Regarding per-mapping --allow-fallback, I see the value for fine-grained
>   control, but the go side currently has no infrastructure for per-mapping
>   options — ParseMap() does naive string splitting, KeyAction has no
>   option fields, and MatchesParsedShortcut ignores AlternateKey entirely.
>   Plus 13+ hardcoded go shortcuts (ctrl+c, ctrl+d etc) cant benefit from
>   per-mapping flags at all. This could be a good follow-up if the
>   granularity turns out to be needed.

IMO, this infrastructure should be added *first* and the hardcoded
shortcuts should be migrated to use proper matching.

This should be fairly trivially implementable using an LLM.

keplian:
Reworked the implementation per your feedback. The changes:

The global match_physical_keys option is gone, replaced with per-mapping
--allow-fallback on the map command. The option controls both shifted
and ascii (alternate_key) fallback:

  map --allow-fallback=shifted,ascii cmd+c copy_to_clipboard

The default value is "shifted" which preserves existing shifted_key
behavior for all mappings. Default kitty bindings are updated to use
"shifted,ascii". User custom mappings get "shifted" by default so they
dont fallback via alternate_key unless they explicitly opt in.

The ascii fallback only triggers when the current layout key is
non-ascii (codepoint > 127), so dvorak/colemak users are unaffected
even if they happen to enable it.

On the go side, added two matching modes: MatchesParsedShortcutWithFallback
for ShortcutTracker (respects per-mapping AllowFallback from KeyAction),
and the existing MatchesParsedShortcut which passes "shifted,ascii"
unconditionally for hardcoded shortcuts (safe due to the non-ascii guard).

ParseMap now handles --allow-fallback=value and --allow-fallback value
forms. ShortcutTracker.Match passes the per-mapping fallback to matching.

Migrated themes and command_palette kittens to use ShortcutTracker with
configurable map entries. Dynamic shortcuts in themes (category first
letter) stay hardcoded but benefit from the matcher fix automatically.

Updated diff and choose_files kitten bindings with --allow-fallback=shifted,ascii.
Tests added for python (5 test methods) and go (ParseMap + key matching).

kovidgoyal:
Looks good I'll review in detail when I have some time.

kovidgoyal:
Please fix the conflict with current master, in command_palette.go
other than that this looks mostly good to go.

keplian:
Rebased on current master. Resolved the conflict in command_palette/main.go — kept the new word-level search refactor and re-applied the ShortcutTracker migration on top of it.

kovidgoyal:
On a more careful reading, few more issues:

1) --allow-fallback should be parsed in parse_map() in options/utils.py
not added ad hoc in conf/types.py

1.5) Why is __init__/__repr__ method added to KeyMapOptions? This should be
automatically handled by dataclass decorator

2) Matching the alternate key currently operates for key code > 127
I am not sure if we should not further restrict it to exclude functional
keys. So that fallback happens only for actual unicode text characters.
May not be necessary but on the other hand is less likely to cause
bugs/regressions. Or maybe I am over thinking it.

3) convert_allow_fallback can be made more performant rather than
constructing a frozenset on ever invocation either just use direct
equality checks which for two strings will prob be faster than using a
set or move the set of allowed values to global level

4) Havent verified this but just eyeballing the code in the Go code
paths it looks like --allow-fallback= is invalid and there is no way to
have no fallbacks in the Go code paths

Maybe rather than using --allow-fallback= have it be
--allow-fallback=none

kovidgoyal:
And of course fix the paths to the docs that CI is failing on.

keplian:
All points addressed:

1) Moved flag parsing into parse_options_for_map() in options/utils.py,
removed the ad hoc handling from conf/types.py and generate.py.

1.5) Removed custom __init__/__repr__, dataclass(frozen=True) with
LiteralField descriptors handles it correctly.

2) Added < 0xE000 guard so fallback only triggers for actual unicode
text characters, not PUA-range functional keys. Applied in both
Python (keys.py, key_encoding.py) and Go (key-encoding.go).

3) Extracted frozenset to module-level _allowed_fallback_values constant.

4) Added --allow-fallback=none support in both Python and Go paths.
Empty string value means no fallback at all.

keplian:
Fixed. Removed the standalone mapping.rst, docs live in definition.py
option text now. Also fixed kitten_link role to handle flags in
parseable_text.

kovidgoyal:
Why is there a change to dnd.c, seems to be completely unrelated to this PR?

Also, in conf/types.py for ShortcutMapping use parse_options_for_map in
__init__ to set the key and action_defn string properties rather than
the ad hoc _skip_flags. You might need to delay import it to avoid a
circular import.

_convert_allow_fallback should not sort the allowed values. The order of
values should be preserved and fallback should happen in that order. In
other words for --allow-fallback=shifted,ascii first the shifted key
should be tried then the ascii key and vice versa for --allowed-fallback=ascii,shifted

keplian:
Addressed all points, squashed into a single commit:

1) Removed the dnd.c change, was left over from a bad rebase. Sorry
2) ShortcutMapping uses parse_options_for_map() in __init__, key and
action_def are plain attributes.
3) _convert_allow_fallback preserves user-specified order, no sorting.

Issue #9676, opened 2026-03-16
Non-native fullscreen doesn't respect NSView.safeAreaInsets on displays with a notch


dannydulai:
## Bug description

  When using `macos_traditional_fullscreen yes`, kitty renders content behind the display's safe area inset (i.e., behind the notch on MacBook models that  have one). The content is physically obscured.

  ## Expected behavior

  macOS provides [`NSView.safeAreaInsets`](https://developer.apple.com/documentation/appkit/nsview/safeareainsets) (available since macOS 12.0) which reports the exact insets needed to avoid hardware obstructions like the notch. In non-native fullscreen, the top inset is non-zero on notched displays and zero on everything else.

  The content area should be offset by `safeAreaInsets.top` so that no text is rendered behind the notch.

  ## Reproduction

  1. Use a MacBook with a notch (2021+)
  2. Set `macos_traditional_fullscreen yes` in kitty.conf
  3. Toggle fullscreen — top lines of content are hidden behind the notch

  ## Environment

  - kitty 0.46.1
  - macOS 15 (Sequoia) on a MacBook with a notch

dannyfiresnake:
I'd also like to offer a workaround to anyone who comes across this. If you save this kitten below and map it to Cmd-Enter, it'll do the right thing by dynamically changing the padding. It's ugly since it compiles a little swift program on first activation, but it does the job even if you have multi-monitor setups with some screens having notches and some without notches.


```python
"""Kitten to toggle fullscreen and adjust padding for MacBook Pro notch."""

import os
import subprocess
import tempfile

KITTY_CONFIG_DIR = os.path.join(os.path.expanduser('~'), '.config', 'kitty')
NOTCH_DETECT_BIN = os.path.join(KITTY_CONFIG_DIR, '__native_notch_detect')
DEFAULT_PADDING = 4

SWIFT_SOURCE = '''\
import AppKit

let args = CommandLine.arguments
let wx = Double(args[1])!
let wy = Double(args[2])!

let primaryHeight = NSScreen.screens.first?.frame.height ?? 0
let point = NSPoint(x: wx, y: primaryHeight - wy)

var targetScreen = NSScreen.main ?? NSScreen.screens.first!
for screen in NSScreen.screens {
    if screen.frame.contains(point) {
        targetScreen = screen
        break
    }
}

let frame = targetScreen.frame
print(Int(frame.size.width))
print(Int(frame.size.height))
if #available(macOS 12.0, *) {
[...]

PR #9678, opened 2026-03-16
Add safe area inset padding for macOS traditional fullscreen

dannydulai:
## Summary
- On MacBooks with a notch, traditional (non-native) fullscreen renders content behind the camera housing. This adds padding to respect `NSScreen.safeAreaInsets` so terminal text and the tab bar avoid the notch area, while the background color fills behind it.
- Queries `NSScreen.safeAreaInsets` (macOS 12+) when entering/exiting traditional fullscreen, plumbs the insets through GLFW to kitty's layout system, and fills the padding area using the existing border rect rendering.

## Test plan
- [ ] On a MacBook with a notch, enable `macos_traditional_fullscreen yes` and toggle fullscreen — content should be padded below the notch
- [ ] Tab bar at top and bottom positions should both work correctly
- [ ] Exiting fullscreen should restore normal layout (no extra padding)
- [ ] On a Mac without a notch or external display, behavior should be unchanged (insets are zero)
- [ ] On Linux/X11/Wayland, stub implementations return zero — no effect

Fixes https://github.com/kovidgoyal/kitty/issues/9676

### DIFF
--- a/glfw/cocoa_platform.h
+++ b/glfw/cocoa_platform.h
@@ -141,6 +141,9 @@ typedef struct _GLFWwindowNS
     bool            in_traditional_fullscreen;
     bool            in_fullscreen_transition;
     bool            suppress_frame_constraints;
+    id              notch_cover_window;
+    unsigned int    notch_cover_color;
+    float           notch_cover_opacity;
     bool            titlebar_hidden;
     unsigned long   pre_full_screen_style_mask;
     CGRect          pre_traditional_fullscreen_frame;
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -595,6 +595,7 @@ static const NSRange kEmptyRange = { NSNotFound, 0 };
 @end
 
 static void update_titlebar_button_visibility_after_fullscreen_transition(_GLFWwindow*, bool, bool);
+static void _glfwUpdateNotchCover(_GLFWwindow*);
 
 @implementation GLFWWindowDelegate
 
@@ -814,6 +815,7 @@ static void update_titlebar_button_visibility_after_fullscreen_transition(_GLFWw
             NSWindowStyleMask savedMask = window->ns.pre_full_screen_style_mask;
             CGRect savedFrame = window->ns.pre_traditional_fullscreen_frame;
             window->ns.in_traditional_fullscreen = false;
+            _glfwUpdateNotchCover(window);
             window->ns.suppress_frame_constraints = true;
             dispatch_async(dispatch_get_main_queue(), ^{
                 _GLFWwindow *w = NULL;
@@ -2432,6 +2434,12 @@ void _glfwPlatformDestroyWindow(_GLFWwindow* window)
     if (_glfw.ns.disabledCursorWindow == window)
         _glfw.ns.disabledCursorWindow = NULL;
     free_drop_data(window);
+    if (window->ns.notch_cover_window) {
+        [w removeChildWindow:window->ns.notch_cover_window];
+        [window->ns.notch_cover_window close];
+        [window->ns.notch_cover_window release];
+        window->ns.notch_cover_window = nil;
+    }
 
     [w orderOut:nil];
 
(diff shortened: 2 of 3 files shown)


kovidgoyal:
Is this PR ready for review? You have a bunch of items under test plan that dont seem to be checked.

dannyfiresnake:
It is ready. I did all those items (except the last since I lack the graphical Linux setup) and I've been using the implementation all day on a Mac. Very happy with it.

I was unable to test on a Mac without a notch (no hardware), but I was able to test on external notchless display.

Thank you for the quick response time. It's a pretty simple PR, but let's see what you think.

kovidgoyal:
Some questions from reading the code:

1) Why is this needed only in traditional fullscreen but not regular
fullscreen?

2) what happens if you set the global menubar to auto hide and drag a
kitty window so that it is under the notch

The design of this feels suspect to me. IMO a better design would be to
have a callback that glfw uses to inform the application that the safe
area insets have changed, similar to the window resize callback.

That way it deals with this more generally, if in the future (1) or (2)
above become a concern.

dannyfiresnake:
Mac OS has an option to automatically hide and show the menu bar in: full screen only, on desktop only, always, never.

If you hide the menu bar, you can't put the windows behind the notch. If you drag a window there, it just bumps it down a little bit so it doesn't obscure the notch row.

If you show the menu bar, there's a menu bar there and you can't put the window there. 

When you go to native full screen, it follows the above rules because it obeys safe area insets.

The weird thing that non-native full screen on Mac does is assume the screen goes from top to bottom fully. IOS, and then macOS, introduced the idea of safe area insets to deal with this. The only two places is really breaks is when a developer decides to go full screen without using the Mac OS full screen feature. In my experience that's been some games and some terminals.

All we are doing is making non-native full screen follow the rules of MacOS screens. 

Safe area insets don't really change, at least not on any hardware today. This API is really made for notches on iOS and MacOS, which are defined by the hardware.

It's worth noting that on iOS, this is been true for almost 9 years. On MacOS, it's been nearly 5.

I think you are assuming that this is about graphical layout. This API is about hardware, which doesn't change. My patch already handles mixed hardware setups (with and without insets).

kovidgoyal:
OK, in that case the correct solution is simply to have traditional
fullscreen respect the safe areas when calculating the full screen
window size. A much simpler patch that will be about 10 lines of code in
one file. It means that it will behave like regular full screen (leave
some space at the top for the notch. Correct?

dannyfiresnake:
Isn't that what I did?

Unfortunately there is one step I couldn't figure out easily. If you just resize the window to be below the notch, the desktop background image will show around the notch, and the menu bar will too.

The window needs to be full screen to obscure that and start drawing the top tabs / terminal area down below the notch. We need to decide what to put around the notch. There are two choices: all black (to obscure the existence of the notch) or some type of background. MacOS makes no rules about what shows. Some games render game space there, but don't put any UI up there. That's also what iterm2, ghosty, and others do. So I did the same by drawing background there. Originally I drew the background in the opengl code, but I moved it to the python in case Kitty wanted to put something up there in the future. 

Remember, native fullscren does more than position the window. It also draws black bars up there, and it hides the native MacOS title bar on the window, and it makes that show in a slide down method when you move your mouse up there. Even if you don't have a notch, your Mac should do this title bar thing.

kovidgoyal:
On Mon, Mar 16, 2026 at 10:22:49PM -0700, Danny wrote:
> dannydulai left a comment (kovidgoyal/kitty#9678)
>
> Isn't that what I did?

No I am suggesting you simply position the OS Window below the notch
when transitioning to full screen.

Let the desktop background or whatever else show there, similar to
whatever shows in regular fullscreen.

>
> Unfortunately there is one step I couldn't figure out easily. If you just resize the window to be below the notch, the background image will show around the notch, and the menu bar will too.

Yes, why is that a problem?

>
> The window needs to be full screen to obscure that and start drawing the top tabs / terminal area down below the notch. We need to decide what to put around the notch. There are two choices: all black or some type of background. MacOS makes no rules about what shows. Some games render game space there, but don't put any UI up there. That's also what iterm2, ghosty, and others do. So I did the same by drawing background there.

Yes, I understand what you did, but I am asking why not simply leave it
blank and let whatever is underneath show up. This even allows the user
to interact with those elements which is an unexpected side bonus.

If you really want to draw black, aka background color over that area,
do it in the cocoa backend itself by creating a subview and positioning
it above the window. Create the subview on full screen transition.
Remove it when leaving full screen.

The color for this is already present in the cocoa backend via
glfwCocoaSetWindowChrome which kitty already maintains so that setting
the titlebar to the background color works.

Or another idea is to have a blur effect on the subview and not use the
background color.

dannyfiresnake:
The entire point of full screen is to not see other elements.

I want the window to be full screen, I just don't want important UI to be under the notch. I can already get what you are suggesting, it's called "maximizing the window." I also want it to feel the same as when I fullscren on non-notch monitor.

kovidgoyal:
No, the entire point of fullscreen is to utilise all available screen
space, which maximize does not do. And in any case, as I said, if for some
reason you are allergic to seeing a menubar above your full screen
window, then use a subview. I am not willing to accept this PR which
contains unnecessary changes and complexity to support such a niche use
case. There are changes to about 10 files, a whole extra API,
storage of completely unnecessary data (the extents) in both cocoa and
kitty layers etc etc. And all this just to draw a black feature-less and
interaction less rectangle. Which can be done with about 10 lines of
code in glfw/cocoa_window.m itself. I suggest you feed my approach to
your LLM and see what it comes up with rather than arguing about it.

dannyfiresnake:
I've updated the PR. Very simple now, and responds to the background color changes. It positions the OpenGL window down below the notch and creates a background window to cover the menubar.

Thank you for the feedback and suggestion to keep this simple, while still preserving my intent. Much appreciated.

kovidgoyal:
Much better thanks. Couple of minor points.

1) If you are setting setReleasedWhenClosed to NO then IIRC you need to
release it manually after closing. I may be misremembering though, I
can never keep those cocoa rules straight.

2) I dont recall if background_opacity applies in traditional fullscreen
mode or not. If it does, then you would need to take that into account
when setting the color of the notch window

3) You should close/release the notch window in the glfw window
destructor as well to handle the case when the user closes the window
without leaving fullscreen.

dannyfiresnake:
1. I released manually.

2. background opacity does apply in traditional full screen, and I added support for it it. JFYI, I had to add a set_os_window_chrome(os_window) to PYWRAP1(change_background_opacity) otherwise it wouldn't update when you reloaded the config.

3. the close/release was done in _glfwPlatformDestroyWindow() already. I assume that's the destructor you speak of.

kovidgoyal:
I just realised I dont need hardware to test this, I can just use a hardcoded safe area inset. Doing so I see the colors of the child window and the main window dont quite match. This is easily visible running kitty as

kitty -o background=darkgray -o macos_traditional_fullscreen=y

kovidgoyal:
Its because the parent window has a shadow. Needs to be removed and restored on returning from full screen.

kovidgoyal:
Easier fix is to just put the child window above the parent window, committed.

PR #9709, opened 2026-03-19
Unmark IME pre-edit text on Ctrl+H

kdnk:
## Summary

- On macOS, when using an IME (such as the Japanese one), deleting uncommitted (marked) text using `Ctrl+H` would leave the last character remaining on the screen. This is because `Ctrl+H` was not triggering `unmarkText`, unlike the physical Backspace (`0x33`) or Escape (`0x35`) keys.
- This PR adds a check for `Ctrl+H` (keycode `0x04` with the `Control` modifier) to the list of keys that call `unmarkText` in `glfw/cocoa_window.m`. This ensures that `Ctrl+H` correctly clears the IME pre-edit state, matching the behavior of the Backspace key on macOS.
- Fixes the issue where uncommitted characters remain when deleting with `Ctrl+H` in Japanese IME on macOS.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -182,6 +182,10 @@ Detailed list of changes
 
 - macOS: When using :opt:`macos_traditional_fullscreen` do not render content under the notch (:pull:`9678`)
 
+- X11: Fix massive scroll when switching focus between kitty and another application (:iss:`9703`)
+
+- Markers: Fix marking not working for multicell characters (:iss:`9705`)
+
 0.46.1 [2026-03-16]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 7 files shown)


kovidgoyal:
That is matching any modifier combination that includes control. And why should ctrl+H act like backspace?

kdnk:
> why should ctrl+H act like backspace?

Because pressing Ctrl+H at the same time functions as backspace by default on macOS.
When using an IME (e.g., for Japanese input), we’ve seen cases where it doesn’t work correctly, so I’d like to make it work with IME.

kdnk:
> That is matching any modifier combination that includes control. 

Sorry, I'll fix it 🙇‍♂️

kovidgoyal:
merged

PR #9741, opened 2026-03-24
Show hyperlink target URL on hover with modifier/always/never

zzhaolei:
## Summary
Add `show_hyperlink_targets=ctrl|cmd|shift|always|never` so the hover URL bar is shown only when the specified modifier is held (or always), defaulting to `never`.

## Behavior
- `ctrl`, `cmd`, `shift`: show only while the corresponding key is pressed during hover.
- `always`: always show.
- `never`: never show (default).

## Pros
- Reduces visual noise by requiring an explicit modifier.
- Preserves existing hyperlink detection and rendering scope.

## Repro
1. `./kitty/launcher/kitty -o show_hyperlink_targets=cmd`
2. Hover an OSC 8 hyperlink while holding Cmd.
3. The hover URL bar appears; releasing Cmd hides it.

## AI Disclosure
- `AI-assisted: Yes (OpenAI Codex)`

### DIFF
--- a/kitty/glfw.c
+++ b/kitty/glfw.c
@@ -457,8 +457,6 @@ refresh_callback(GLFWwindow *w) {
     request_tick_callback();
 }
 
-static int mods_at_last_key_or_button_event = 0;
-
 #ifndef __APPLE__
 typedef struct modifier_key_state {
     bool left, right;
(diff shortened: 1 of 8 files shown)


kovidgoyal:
1) You need to handle boolean value sin the option for legacy compatibility
when parsing it.

2) There is no need to have another global to track current modifiers.
Simply move the existing mods_at_last_key_or_button_event into
global_state. And remove the update from mouse_event() in mouse.c that
is wrong since mouse_event() is also called for motion events that have
modifiers set to -1.

3) Use super (cmd) in the docs when referring to and mark them using
:kbd:``

4) Clarify the behavior when multiple modifiers are pressed.

5) In show hyperlink targets with modifiers there is no need to mask
against lock mask since you are & with single bit constants anyway

zzhaolei:
## Summary
Adjust `show_hyperlink_targets` handling per review feedback.

## Changes
- Treat legacy boolean values as `yes` → `always`, `no` → `never` when parsing `show_hyperlink_targets`.
- Move `mods_at_last_key_or_button_event` into `global_state` and remove the incorrect modifier update in `mouse_event()` (motion events can pass `-1`).
- Clarify docs to use :kbd:`Super` (macOS :kbd:`Cmd`) and explain behavior when multiple modifiers are pressed.
- Remove the unnecessary lock-mask operation when checking modifiers.

## Reviewer Note
All requested changes have been applied based on your feedback.

Issue #9755, opened 2026-03-26
a mouse_selection option like line_from_begin


mc36:
**Is your feature request related to a problem? Please describe.**
selecting line with mouse should have an option to keep the whitespaces at the beginning of the line

**Describe the solution you'd like**
in mouse.c 

```
        case MOUSE_SELECTION_LINE_BEGIN:
            if (!screen_selection_range_for_line(screen, w->mouse_pos.cell_y, &start, &end)) break;
            start = 0;
            S(EXTEND_LINE);
            break;

```
**Describe alternatives you've considered**
change my muscle memory... :)

**Additional context**
configuring cisco (alike) routers needs this:
```

sid(cfg-subif)#ipv4 secondary-address 3.3.3.3                                                                         
sid(cfg-subif)#ipv4 secondary-address 3.3.3.4                                                                         
sid(cfg-subif)#ipv4 secondary-address 3.3.3.5                                                                         
sid(cfg-subif)#show config-differences this                                                                            
sid(cfg-subif)#                                                                                                        
interface bundle1.227
 ipv4 secondary-address 3.3.3.3
 ipv4 secondary-address 3.3.3.4
 ipv4 secondary-address 3.3.3.5
 exit

and at this point one just types "no" and copy-pastes back the wrong line

sid(cfg-subif)#no ipv4 secondary-address 3.3.3.4                                                                       
sid(cfg-subif)#                                                                                                        
sid(cfg-subif)#show config-differences this                                                                            
interface bundle1.227
 ipv4 secondary-address 3.3.3.3
 ipv4 secondary-address 3.3.3.5
 exit


```

kovidgoyal:
Not something I am interested in implementing. In such cases do a drag
select instead of a line select. Or even a box (rectangle) select.

mc36:
im more than happy to  add the surroundings if you're willing to review/merge?

it's pretty convenient imho that one just double-tap the wrong line and ready to go, and, maybe python devvies also would honor the idea, not just the infra guys... :)

kovidgoyal:
I am fine to review, merging will depend on the overheads in terms of
code, performance, complexity, etc.

mc36:
sure thing, challenge accepted!  thanks in advance! :)

PR #9758, opened 2026-03-27
adding mouse action line_from_begin see #9755

mc36:
as per discussed in https://github.com/kovidgoyal/kitty/issues/9755 here we go


mc36:
thank you! :)

PR #9779, opened 2026-03-31
Add an experimental interactive search kitten

zzhaolei:
The goal of this PR is to provide a lightweight, zero-configuration, and deeply integrated search experience within Kitty. While users can still rely on external tools like `less` or `neovim` as alternatives, this kitten aims to offer a more idiomatic and "native" default search.

Users can experience it via the default `cmd+f`.

## Features
- **Interactive Search:** A dedicated input box for dynamic filtering.
- **Result Navigation:** Supports cycling through matches (Previous/Next).
- **Scrollable Content:** Smoothly scroll through the buffer within the kitten UI.
- **Keybinding Hints:** On-screen tips for better discoverability.

## TODOs & Known Issues
- [ ] **Wrapped lines & ANSI escapes:** Currently, long lines are truncated/missing after wrapping. This likely needs better handling of escape sequences or screen line logic, which I'm still figuring out.
- [ ] **UI Polish:** Adding borders to the search box and the content area (similar to the `choose_files` kitten) to improve the visual hierarchy.
- [ ] **Streaming updates:** It would be ideal to read from the scrollback buffer in a streaming/real-time fashion, so the search stays up-to-date as the buffer grows. 
- [ ] **Image**

https://github.com/user-attachments/assets/c3ff13af-491c-48b6-b3ed-d98903e11ee5

### DIFF
--- a/kittens/search/main.go
+++ b/kittens/search/main.go
@@ -1,0 +1,429 @@
+package search
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/kovidgoyal/kitty"
+	"github.com/kovidgoyal/kitty/tools/cli"
+	"github.com/kovidgoyal/kitty/tools/config"
+	"github.com/kovidgoyal/kitty/tools/tty"
+	"github.com/kovidgoyal/kitty/tools/tui/loop"
+	"github.com/kovidgoyal/kitty/tools/utils"
+	"github.com/kovidgoyal/kitty/tools/wcswidth"
+)
+
+var debugPrintln = tty.DebugPrintln
+var _ = debugPrintln
+var _ = fmt.Print
+
+type DisplayLine struct {
+	raw    string
+	matchs []*Match
+}
+
+type Search struct {
+	text         string
+	currentMatch *Match
+}
+
+type Match struct {
+	line int
+
+	start   int
+	end     int
+	current bool
+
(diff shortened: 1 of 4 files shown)


kovidgoyal:
There is already an in progress pager kitten. IMO, this should just be a "start in search mode" command line flag for that. If you want live updates, then you will need to basically port vt-parser.c (at least for SGR and OSC 8 and OSC 66 escape codes) and port the LineBuf/CPUCell/GPUCell structs to Go. That will allow this functionality to be  robust. However for a proper pager kitten there is a lot more work to consider. You will need to implement transferrence of graphics protocol images, all the usual pager keybindings and so on. You also need to implement fast scroll to specified line using SIMD that work is mostly done already.

zzhaolei:
I’m currently trying to implement this incrementally, so I don’t have a concrete ETA at the moment.

While working on it, I ran into an issue: the scrollback text I can retrieve does not seem to contain any image-related information (e.g. Kitty graphics protocol / OSC 1337 sequences).

Could you please advise on what logic I should adjust in order to capture these image payloads?
Alternatively, would it be possible for the scrollback output to include image-related escape sequences or metadata as well?

scrollback ansi text:
```
[mbash-3.2$ kitten icat ~/Downloads/test.png 



[mbash-3.2$ 
```

kovidgoyal:
Yes currently the scrollback doesnt have this. It requires changes to
kitty. I suggest you leave that for later. Getting to a replacement for
less is the first step.

PR #9802, opened 2026-04-03
Fix timing, memory safety, and tar extraction vulnerabilities

z3rco:
## Summary

Found and fixed several security issues across the C, Python, and Go code during a source audit. All changes are minimal and surgical.

### Timing-safe comparisons
- **crypto.c** — `Secret.__eq__` used `memcmp` (not constant-time) and compared only `min(len_a, len_b)` bytes, so secrets of different lengths with a shared prefix compared equal. Switched to `CRYPTO_memcmp` with a length-equality check first.
- **remote_control.py** — Remote control passwords were looked up via `dict.get()`, which leaks timing through Python's hash+compare. Replaced with a constant-time linear scan using `hmac.compare_digest`.
- **file_transmission.py** — Bypass token comparison in `check_bypass` used `==` for both the `kitty-1` and `sha256` protocols. Switched to `hmac.compare_digest`.

### Memory safety (C)
- **child-monitor.c** — `write_to_peer` had an inverted condition (`n > used` instead of `n < used`). Since `send()` never returns more than `used`, the `memmove` never fired, so partial writes resent stale data from the buffer head.
- **ibus_glfw.c** — `IBUS_ADDRESS` env var copied via `memcpy` without a null terminator. When `strlen(addr) >= PATH_MAX`, the static buffer was returned unterminated.
- **x11_window.c** — Two `realloc` calls (clipboard + DnD chunked writer) didn't check for `NULL`, leading to a use-after-free of the old pointer and a `memcpy` into `NULL + offset`.
- **dnd.c** — `accepted_mimes` buffer had no size cap (unlike `registered_mimes` which caps at 1MB) and the `realloc` pattern leaked the old pointer on failure. Added the same 1MB cap and safe realloc.
- **png-reader.c** — `rowbytes * height` was computed as `png_uint_32` arithmetic, which can overflow on 32-bit before assigning to `size_t`. Added explicit casts.

### Secrets hygiene
- **disk-cache.c** — `CacheValue` contains a 64-byte encryption key that was left in memory after `free()`. Added `explicit_bzero` before freeing.

### Tar extraction hardening
- **tar.go** — Hardlink targets were not validated against the destination prefix the way `dest` and symlinks were, allowing a malicious tar to hardlink outside the extraction directory then overwrite the target via a later entry. Also stripped setuid/setgid/sticky bits from extracted file modes.

## Test plan
- [x] Full build (`make`) passes with zero errors/warnings
- [x] Full test suite (`make test`) passes — 2 pre-existing font-name failures unrelated to these changes
- [x] Python files pass `py_compile` and `ast.parse`
- [x] Go files pass `gofmt -e`
- [ ] Upstream CI

### DIFF
--- a/glfw/ibus_glfw.c
+++ b/glfw/ibus_glfw.c
@@ -287,7 +287,9 @@ get_ibus_address_file_name(void) {
     addr = getenv("IBUS_ADDRESS");
     int offset = 0;
     if (addr && addr[0]) {
-        memcpy(ans, addr, GLFW_MIN(strlen(addr), sizeof(ans)));
+        size_t len = GLFW_MIN(strlen(addr), sizeof(ans) - 1);
+        memcpy(ans, addr, len);
+        ans[len] = '\0';
         return ans;
     }
     const char* disp_num = NULL;
--- a/glfw/x11_window.c
+++ b/glfw/x11_window.c
@@ -937,7 +937,9 @@ get_clipboard_data(const _GLFWClipboardData *cd, const char *mime, char **data) 
         if (!chunk.sz) break;
         if (cap < sz + chunk.sz) {
             cap = MAX(cap * 2, sz + 4 * chunk.sz);
-            buf = realloc(buf, cap * sizeof(buf[0]));
+            char *new_buf = realloc(buf, cap * sizeof(buf[0]));
+            if (!new_buf) { free(buf); *data = NULL; cd->get_data(NULL, iter, cd->ctype); return 0; }
+            buf = new_buf;
         }
         memcpy(buf + sz, chunk.data, chunk.sz);
         sz += chunk.sz;
@@ -3636,7 +3638,9 @@ write_chunk(void *object, const char *data, size_t sz) {
     if (data) {
         if (cw->cap < cw->sz + sz) {
             cw->cap = MAX(cw->cap * 2, cw->sz + 8*sz);
-            cw->buf = realloc(cw->buf, cw->cap * sizeof(cw->buf[0]));
+            char *new_buf = realloc(cw->buf, cw->cap * sizeof(cw->buf[0]));
+            if (!new_buf) return false;
+            cw->buf = new_buf;
         }
         memcpy(cw->buf + cw->sz, data, sz);
         cw->sz += sz;
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -1940,7 +1940,7 @@ write_to_peer(Peer *peer) {
     else if (n < 0) {
         if (errno != EINTR) { log_error("write() to peer socket failed with error: %s", strerror(errno)); peer->write.used = 0; peer->write.failed = true; }
     } else {
-        if ((size_t)n > peer->write.used) memmove(peer->write.data, peer->write.data + n, peer->write.used - n);
+        if ((size_t)n < peer->write.used) memmove(peer->write.data, peer->write.data + n, peer->write.used - n);
         peer->write.used -= n;
     }
     talk_mutex(unlock);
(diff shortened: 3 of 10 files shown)


kovidgoyal:
On Fri, Apr 03, 2026 at 08:12:37AM -0700, Z3rco wrote:
> ## Summary

Thanks for the review.

>
> Found and fixed several security issues across the C, Python, and Go code during a source audit. All changes are minimal and surgical — no refactoring, no behavior changes beyond the fixes themselves.
>
> ### Timing-safe comparisons
> - **crypto.c** — `Secret.__eq__` used `memcmp` (not constant-time) and compared only `min(len_a, len_b)` bytes, so secrets of different lengths with a shared prefix compared equal. Switched to `CRYPTO_memcmp` with a length-equality check first.
> - **remote_control.py** — Remote control passwords were looked up via `dict.get()`, which leaks timing through Python&#39;s hash+compare. Replaced with a constant-time linear scan using `hmac.compare_digest`.

This breaks type checking.

> - **file_transmission.py** — Bypass token comparison in `check_bypass` used `==` for both the `kitty-1` and `sha256` protocols. Switched to `hmac.compare_digest`.
>
> ### Memory safety (C)
> - **child-monitor.c** — `write_to_peer` had an inverted condition (`n &gt; used` instead of `n &lt; used`). Since `send()` never returns more than `used`, the `memmove` never fired, so partial writes resent stale data from the buffer head.

This is a bug, sure. Don't see how it's a security issue.

> - **ibus_glfw.c** — `IBUS_ADDRESS` env var copied via `memcpy` without a null terminator. When `strlen(addr) &gt;= PATH_MAX`, the static buffer was returned unterminated.

Again dont see how this is a security issue. If the attacker can control
the env vars the kitty process sees, its game over in a dozen different
ways already.

> - **x11_window.c** — Two `realloc` calls (clipboard + DnD chunked writer) didn&#39;t check for `NULL`, leading to a use-after-free of the old pointer and a `memcpy` into `NULL + offset`.

Again a bug, but a security issue? OOM will almost certainly end up terminating the
process long before this could be exploited.

> - **dnd.c** — `accepted_mimes` buffer had no size cap (unlike `registered_mimes` which caps at 1MB) and the `realloc` pattern leaked the old pointer on failure. Added the same 1MB cap and safe realloc.

Failure is OOM, leak of old memory doesnt really matter there as the
process will die. As for no cap, yes but that's a potential DoS at best.

> - **png-reader.c** — `rowbytes * height` was computed as `png_uint_32` arithmetic, which can overflow on 32-bit before assigning to `size_t`. Added explicit casts.

The expression was sizeof(png_byte) * rowbytes * d->height and sizeof() evaluates to size_t.

>
> ### Secrets hygiene
> - **disk-cache.c** — `CacheValue` contains a 64-byte encryption key that was left in memory after `free()`. Added `explicit_bzero` before freeing.

explicit_bzero() is not portable. You need to use memset_s on macOS, for
instance.

>
> ### Tar extraction hardening
> - **tar.go** — Hardlink targets were not validated against the destination prefix the way `dest` and symlinks were, allowing a malicious tar to hardlink outside the extraction directory then overwrite the target via a later entry. Also stripped setuid/setgid/sticky bits from extracted file modes.

Those bits are preserved *by design*, as should be obvious by the fact
that they are explicitly enumerated. Also, ExtractAllFromTar is only
used with tar files kitty itself creates, so not potentially malicious
ones. That said, no harm in restricting hard links to be within the dest
path.

z3rco:
Fair points. I used "security" loosely. These are correctness/robustness bugs, not exploitable vulnerabilities. The intent was to fix spots where the code didn't behave as written, not to claim active threat scenarios.

kovidgoyal:
Yeah no worries, as I said thanks for the review. I am always happy to harden kitty code.

Issue #5851, opened 2023-01-03
Draggable File Hyperlinks


0xTadash1:
**Is your feature request related to a problem? Please describe.**
It is annoying to launch a GUI file manager every time to upload a file to the web or open in desktop app.

**Describe the solution you'd like**
Make `file://` texts or hyperlinks (e.g. **`ls --hyperlink`**) being able to drag-and-drop.\
For example, when a hyperlink for an image file is dragged from kitty and dropped it to some painting app, the painting app open it.

**Describe alternatives you've considered**

**Additional context**
Since there are tools such as [ripdrag](https://github.com/nik012003/ripdrag) and [dragon](https://github.com/mwh/dragon) that allow drag-and-drop without leaving the terminal as much as possible, I think there is a demand for them, though it may be small.

kovidgoyal:
Not something I am interested in. Simply running open filename is a lot
easier and more efficient than drag and drop.

trygveaa:
> Simply running open filename is a lot easier and more efficient than drag and drop.

That's not the same functionality though. If you want to e.g. upload a file to a web page, you can't run open filename, you have to drag it or open the file picker in the browser and select it. I've even encountered web pages which requires dragging the file, and doesn't have the option to open the file picker.

kovidgoyal:
On Wed, Jan 04, 2023 at 02:21:47AM -0800, Trygve Aaberge wrote:
> > Simply running open filename is a lot easier and more efficient than drag and drop.
> 
> That's not the same functionality though. If you want to e.g. upload a file to a web page, you can't run open filename, you have to drag it or open the file picker in the browser and select it. I've even encountered web pages which requires dragging the file, and doesn't have the option to open the file picker.

Ah, websites trying to be applications, the greatest things since sliced
bread. I'm afraid that's too niche a usecase for me to bother, but I
will accept a PR that adds a mouse action for this.

kovidgoyal:
A mouse action means a configurable action that performs the drag and drop. See https://sw.kovidgoyal.net/kitty/conf/#mouse-actions

kovidgoyal:
Implemented in master.

PR #9804, opened 2026-04-03
Add mouse_drag_url action for drag-and-drop of URLs and hyperlinks

karlb:
I tried to accommodate the requests mentioned in #5851. Does this go into the right direction and have a chance of being merged? If it does, I'll rebase to latest master and give it some more testing and polish.

---

Implements drag-and-drop for URLs and OSC 8 hyperlinks as an opt-in mouse_map action, as requested in #5851. The default mouse mapping is unchanged — left press still starts text selection. To enable URL dragging, users add to kitty.conf:

    mouse_map left press ungrabbed mouse_drag_url

When enabled, pressing over a detected URL or hyperlink initiates a platform drag-and-drop operation after the mouse moves beyond url_drag_threshold pixels. The URL can be dropped into file managers, browsers, or other applications. If no URL is under the mouse, it falls back to normal text selection. Setting url_drag_threshold to 0 disables dragging entirely.

Platform support: macOS (NSPasteboard/NSDraggingSession) and Wayland. X11 XDND is not yet implemented. On Wayland, compositors don't deliver drop events back to the drag source, so self-drops use a short timer workaround to deliver the drop to the window under the mouse.

File paths without a scheme are converted to file:// URLs with proper percent-encoding via urllib.parse.quote(), so paths with spaces and special characters work correctly.

Closes https://github.com/kovidgoyal/kitty/issues/5851

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -2005,6 +2005,13 @@ class Boss:
             url = 'file://' + quote(os.path.abspath(url))
         start_drag_with_data(os_window_id, {'text/uri-list': (url + '\r\n').encode()}, (), operations=GLFW_DRAG_OPERATION_COPY)
 
+    def _wayland_self_drop(self, data: dict[str, bytes], timer_id: int | None = None) -> None:
+        from .fast_data_types import current_focused_os_window_id, get_os_window_mouse_pos
+        os_window_id = current_focused_os_window_id()
+        if os_window_id and self.os_window_map.get(os_window_id):
+            x, y = get_os_window_mouse_pos(os_window_id)
+            self.on_drop(os_window_id, data, True, int(x), int(y))
+
     def on_drag_source_finished(
         self, was_dropped: bool, was_canceled: bool, accepted_mime_type: str, action: int, data: dict[str, bytes] | None,
         needs_toplevel_on_wayland: bool
@@ -2042,7 +2049,12 @@ class Boss:
                 tm.on_tab_drop_move()
             if was_dropped:  # detach tab into new OS Window
                 self._move_tab_to(tab)
-
+        elif is_wayland() and was_dropped and not was_canceled and data and 'text/uri-list' in data:
+            # On Wayland, compositors don't send drop events back to the drag
+            # source surface, so self-drops never reach on_drop. Use a short
+            # timer to let the pointer ungrab so the OS window's mouse position
+            # updates, then route through on_drop for proper hit-testing.
+            add_timer(partial(self._wayland_self_drop, data), 0.05, False)
 
     @ac('win', '''
         Focus the nth OS window if positive or the previously active OS windows if negative. When the number is larger
--- a/kitty/fast_data_types.pyi
+++ b/kitty/fast_data_types.pyi
@@ -1819,6 +1819,9 @@ class MousePosition(TypedDict):
     cell_y: int
     in_left_half_of_cell: bool
 
+def get_os_window_mouse_pos(os_window_id: int) -> tuple[float, float]: ...
+
+
 def get_mouse_data_for_window(os_window_id: int, tab_id: int, window_id: int) -> Optional[MousePosition]: ...
 
 
--- a/kitty/state.c
+++ b/kitty/state.c
@@ -1563,6 +1563,18 @@ os_window_focus_counters(PyObject *self UNUSED, PyObject *args UNUSED) {
 
 
 static PyObject*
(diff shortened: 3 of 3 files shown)


kovidgoyal:
There is a single drag_threshold setting in master now. Use that. And not sure why you need any platform specific code here at all? dropping of text/uri-list onto kitty windows already works, you can drag and drop from a file manager to kitty and it will paste the URLs, so why do you need special drop handling code?

karlb:
> There is a single drag_threshold setting in master now. Use that.
Done.

> And not sure why you need any platform specific code here at all? dropping of text/uri-list onto kitty windows already works, you can drag and drop from a file manager to kitty and it will paste the URLs, so why do you need special drop handling code?

No platform specific code is strictly needed. It was just a collection of tweaks that made certain cases nicer on the respective platform. To keep this PR focused, I dropped all of them but one: dropping a link into the same kitty wayland-window where it was dragged from, which unfortunately does not work without changes on wayland. This is a case I consider very useful for
* getting a full path for a file where only the basename is displayed (drag file and drop right away)
* dropping into a different kitty split in the same wayland-window
* dropping into a different kitty tab in the same wayland-window

Do you see a better way to get this working?

I've split the wayland self-drop fix into a separate commit to have a clean commit just for the base feature.

kovidgoyal:
I dont follow. You can drag a tab around in kitty under wayland and drop
it onto the same window it works fine I dont see why dragging a
text/uri-list around would be any different.

Also, if you are going to allow dragging of URLs then you should
probably also allow dragging of selected text as text/plain

karlb:
> I dont follow. You can drag a tab around in kitty under wayland and drop it onto the same window it works fine I dont see why dragging a text/uri-list around would be any different.

Tab drag works for same-window reorder because it tracks position in real-time via on_drop_move events. For the link-dragging rely on the on_drop event, which is never delivered to the drag source surface on wayland from what I can see (I'm new to wayland dnd, so I might be wrong, or maybe it is limited to mutter rather than wayland in general). I can try to find out more about why this is the case.

> Also, if you are going to allow dragging of URLs then you should probably also allow dragging of selected text as text/plain

I don't have a use for that myself, but I can try to add that if you think it should be part of this PR.

karlb:
https://github.com/kovidgoyal/kitty/commit/c88adfba98cf3184a255e6b71d2bf3f0b631ca17 works for dragging links from kitty into other applications.

But dropping URIs into kitty does not work for me on current master even before c88adfba98cf3184a255e6b71d2bf3f0b631ca17 (tested with gnome/wayland) even from other applications (e.g. nautlius) where it used to work. LLM explanation and patch that makes it work again in my setup:

**Root cause**: The DnD refactor (commit e2b0208bd) bumped `wl_data_device_manager` to version 3 (needed for drag source support), but never added `wl_data_offer_set_actions()` calls. Version 3 of the Wayland protocol requires the client to call both `wl_data_offer_accept()` AND `wl_data_offer_set_actions()` during drag enter/motion for the compositor to consider the drop valid. Without `set_actions`, the compositor sends a LEAVE instead of a DROP when the user releases.

**Fix**: Added `wl_data_offer_set_actions(d->id, actions, actions)` in `update_drop_state()` alongside the existing `wl_data_offer_accept()` call.

```
diff --git a/glfw/wl_window.c b/glfw/wl_window.c
index ea0bc5983..7cdc2adc9 100644
--- a/glfw/wl_window.c
+++ b/glfw/wl_window.c
@@ -2531,6 +2531,9 @@ update_drop_state(_GLFWWaylandDataOffer *d, _GLFWwindow* window UNUSED, size_t a
         d->drag_accepted = accepted;
         d->mime_for_drop = new_preferred_mime;
         wl_data_offer_accept(d->id, d->serial, d->mime_for_drop);
+        // Version 3 of wl_data_device_manager requires set_actions for drops to work
+        uint32_t actions = accepted ? WL_DATA_DEVICE_MANAGER_DND_ACTION_COPY : WL_DATA_DEVICE_MANAGER_DND_ACTION_NONE;
+        wl_data_offer_set_actions(d->id, actions, actions);
     }
 }
```

---

But even that does not make dropping URIs from kitty into kitty work for me.

Is this feedback helpful?

kovidgoyal:
I added wl_data_offer_set_actions() and it works fine in GNOME dropping
to same window and different window. What isnt working for some reason
is the drag image is not showing up. No clue why. Works fine in other
Wayland compositors.

karlb:
Works well for me, including the drag image (the full file URI `file:///...`). Thanks!

PR #9808, opened 2026-04-04
feat(borders): add optional border ring behavior across kitty OS windows

oxysoft:
## Summary
Add an optional "border ring" behavior for multi-window focus retention.

When enabled and multiple kitty OS windows exist, the last-focused kitty OS window keeps its active border styling even after focus moves to a non-kitty application. This makes it easier to visually track the actual active pane across a single-instance kitty session spread over multiple top-level windows.

## New options
- `border_ring_behavior`
- `unfocused_active_border_color`
- `unfocused_inactive_border_color`
- `unfocused_border_brighten_factor`

## Behavior
- focus moves between kitty windows: normal active/inactive border behavior
- focus moves from kitty to a non-kitty app, with multiple kitty OS windows open:
  the last-focused kitty OS window keeps the active ring
- focus moves away with only a single kitty OS window open:
  borders behave as inactive as before

## Implementation notes
- expose `num_os_windows()` from C state
- use `last_focused_os_window_id()` plus `current_focused_os_window_id()` to decide the ring holder
- redraw borders across OS windows when focus changes
- add fallback unfocused border styling in shader/color plumbing

## Verification
- Python syntax checks pass for the touched Python modules
- patch rebased cleanly onto current `upstream/master`

I am not a C developer, so I would especially value critique on the implementation shape here. If the feature is useful but the plumbing is wrong, guidance on the preferred architecture would be very helpful.

### DIFF
--- a/kitty/border_vertex.glsl
+++ b/kitty/border_vertex.glsl
@@ -9,7 +9,9 @@
 #define TAB_BAR_MARGIN_COLOR 6
 #define TAB_BAR_EDGE_LEFT_COLOR 7
 #define TAB_BAR_EDGE_RIGHT_COLOR 8
-uniform uint colors[9];
+#define UNFOCUSED_ACTIVE_BORDER_COLOR 9
+#define UNFOCUSED_INACTIVE_BORDER_COLOR 10
+uniform uint colors[11];
 uniform float background_opacity;
 uniform float gamma_lut[256];
 
--- a/kitty/borders.py
+++ b/kitty/borders.py
@@ -6,7 +6,7 @@ from enum import IntFlag
 from functools import partial
 from typing import NamedTuple
 
-from .fast_data_types import BORDERS_PROGRAM, current_focused_os_window_id, get_options, init_borders_program, set_borders_rects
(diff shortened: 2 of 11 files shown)


oxysoft:
I would especially appreciate critique/guidance on the implementation shape here. I am not a C developer, so if the feature is useful but this is the wrong plumbing or the wrong abstraction boundary, direction on the preferred approach would be very helpful.

kovidgoyal:
What's the use case for this? Why is it useful to know what the active kitty window is when it doesn't actually have focus? And there is already inactive_text_alpha. Why not use that instead of another setting to brighten background?

kovidgoyal:
no followup

oxysoft:
With terminals across two monitors, the focus border stays active on the unfocused terminal. This means that at a glance the active pane is not obvious, mentally it's two hops. The active pane hooks my attention first, but that is not enough I have to spot which window is active. The goal here is to introduce options to make the active pane border highlight unique across terminals in single instance mode. The setting to brighten background is so that the inactive terminal still faintly telegraphs the last active pane upon returning focus to the window. Ring buffer is maybe over engineered, I just thought if the user closes windows it would be nice to cycle backwards in focus history. This because the active pane highlight should not be necessarily bound to window focus, as we would like the last focused terminal to still have the active border highlight for its active pane when it is unfocused. It's a separate focus history that is update by window focus but not on unfocus, and the active pane is peek on top.

PR #9814, opened 2026-04-05
font: use configured font family as preferred fallback

MiterMe:
## Summary

When a character is not found in the configured font, the fallback mechanism was using a hardcoded `"monospace"` family. This caused issues when the user configured a specific font that didn't contain certain characters (e.g., CJK characters in a Latin-only italic font).

For example, with configuration:
```
italic_font family="Monaspace Xenon"
```

When rendering Chinese italic text, kitty would fallback to a generic monospace font instead of trying fonts related to the configured italic font family.

## Solution

This fix adds the base font's family name as the preferred fallback before `"monospace"`, allowing fontconfig to first try fonts in the same family before falling back to system defaults.

The pattern now becomes:
1. First try the configured font's family (e.g., "Monaspace Xenon")
2. Then try "monospace" as generic fallback
3. Fontconfig automatically finds the best match for the character

## Changes

- Add `family_name_for_face()` function to get font family name from a Face object
- Modify `create_fallback_face()` to use base_face's family as first choice
- Implement for both fontconfig (Linux) and CoreText (macOS)

## Testing

Tested with a font configuration where the italic font doesn't contain CJK characters. The fallback now correctly uses fontconfig to find appropriate fallback fonts while maintaining the style context.

## Notes

This is a more targeted fix than adding full font fallback chain support. It improves the most common case where users configure a single font per style and expect sensible fallback behavior.

### DIFF
--- a/kitty/core_text.m
+++ b/kitty/core_text.m
@@ -1165,6 +1165,13 @@ postscript_name_for_face(const PyObject *face_) {
     return "";
 }
 
+const char*
+family_name_for_face(const PyObject *face_) {
+    const CTFace *self = (const CTFace*)face_;
+    if (self->family_name) return PyUnicode_AsUTF8(self->family_name);
+    return "";
+}
+
 
 static PyObject *
 repr(CTFace *self) {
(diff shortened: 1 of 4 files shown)


kovidgoyal:
This doesnt make sense to me. fallback is fontconfigs responsibility not the applications. Your suggested patch will cause font fallback in fontconfig that the user configures to be overridden by kitty. If you fond monospace doesnt give you your preferred fallback, you should configure fontconfig to use your preferred families in whatever fallback order you like.

PR #9819, opened 2026-04-06
Add alias and combine support to the command palette

dmgerman:
## Summary

- Action aliases and kitten aliases are now discoverable in the command palette, each in their own dedicated section
- Combine keybindings get their own "Combined actions" section instead of landing in "Miscellaneous"
- Alias names are searchable in the TUI

## Refactoring

Broke up the 200-line `collect_keys_data` into single-responsibility functions and reduced nesting depth across both Python and Go to improve readability and maintainability.

### DIFF
--- a/kittens/command_palette/main.py
+++ b/kittens/command_palette/main.py
@@ -265,7 +265,9 @@ def add_alias_sections(
                 target_list.append(build_alias_entry(opts.alias_map, display, expansion))
 
     default_mode_cats = modes.setdefault('', {})
-    sort_key = lambda b: (b['key'] == '', b['key'] or b['action_display'])
+    def sort_key(b: dict[str, str]) -> tuple[bool, str]:
+        return (b['key'] == '', b['key'] or b['action_display'])
+
     for section_name, entries in (('Action aliases', action_alias_entries), ('Kitten aliases', kitten_alias_entries)):
         if entries:
             existing = default_mode_cats.get(section_name, [])


kovidgoyal:
If i type win mov it correctly matches move_window but if i type winmov it matches window.

dmgerman:
A space splits keywords. Each keyword can match an item. In this particular case, winmov is one keyword. Since it is not a direct match, it uses Levenshtein distance to match the closest match (in this case window).  

Would you prefer a different type of matching?

by the way, this search method was added in the previous PR.

dmgerman:
One more thing, the advantage of splitting keywords at space is that order in which they are match does matter. So "win mov" and "mov win" will match move_window (useful when you do not know the actual name of the command).

kovidgoyal:
No no I am fine with splitting at spaces. I was not sure if the winmov
match was intended behavior or not. Does the levenshtein distance match
apply only when there are no exact matches or does it always apply?

dmgerman:
See #9822 for a description of how the search algorithm works.

dmgerman:
@kovidgoyal any feedback on why the PR was not merged?

I find it very useful to be able to execute commands by name and was missing the ability to run combined actions this way too.

kovidgoyal:
Oh sorry closed it accidentally. havent got around to reviewing it yet.

dmgerman:
@kovidgoyal Thank you.

PR #9841, opened 2026-04-09
Add macos_fullscreen_ignore_safe_area_insets option

pdobrev:
Adds a configuration option `macos_fullscreen_ignore_safe_area_insets` to allow ignoring the safe area insets when running in traditional full-screen mode on MacOS.

The motivation for this is that I personally run Kitty as my terminal emulator on Mac primarily because of its full-screen mode. A recent change in https://github.com/kovidgoyal/kitty/pull/9678 introduced respecting the safe area insets by shifting the window down, so the macbook notch doesn't overlap content in the terminal, which is a very reasonable default. I personally prefer to have the full-screen mode fill the entire screen and use all of the available screen real estate and accept that the notch will cover part of the tab bar. I think the proper handling of the notch should be content-aware (i.e. the content in the terminal to be shifted, so that the notch doesn't overlap it, but still have it be full screen), however that's a lot more work, so I added an option to opt out of that behavior.

I've tried to keep the diff minimal and most of the changes are just piping the configuration option to the cocoa window.

Not sure if there's appetite for such a feature, but maybe other folks also have this preference, so I thought I'd open a PR. If not, happy to run this as a personal fork, too.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -3419,6 +3419,7 @@ bool _glfwPlatformToggleFullscreen(_GLFWwindow* w, unsigned int flags) {
     NSWindow *window = w->ns.object;
     bool made_fullscreen = true;
     bool traditional = !(flags & 1);
+    bool ignore_safe_area_insets = flags & 2;
     NSWindowStyleMask sm = [window styleMask];
     if (traditional) {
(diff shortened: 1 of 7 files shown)


diocletiann:
thank you!

PR #9873, opened 2026-04-15
Fix focus_follows_mouse switching active window on desktop return

lxcode:
When returning to a desktop/space, enter_event fires and detects the window under the cursor, triggering focus_follows_mouse to switch focus even though the mouse didn't cross a window boundary. Add a focus_follows parameter to set_currently_hovered_window so enter_event can update hover state without triggering focus switching. Normal mouse motion still switches focus on boundary crossings as before.


kovidgoyal:
There is no guarantee that enter_event() is not from a mouse motion.
Granted, typically if it is mouse motion you will get subsequent motion
events most of the time, but I am not at all sure that is always the
case.

lxcode:
Gotcha, thanks. Yeah, i can't think of a way to address that without including more complexity than it's probably worth, given it's a tiny issue. Feel free to close out.

PR #9874, opened 2026-04-15
Expose window activity state in the `kitty @ ls`

muchzill4:
## Summary

Adds two new fields to the kitty @ ls output at ~both~ the window ~and tab~ level:
- `needs_attention`
- `has_activity_since_last_focus`

~At the tab level, these are aggregated from all child windows.~

## Motivation

The `kitty @ ls` command currently provides no way to determine which windows or tabs have unseen activity. This information is already tracked internally (used for tab bar rendering, bell handling, etc.) but is not exposed to remote control consumers. Scripts and extensions that build custom UIs or status lines need this data to highlight tabs/windows that need attention.

### DIFF
--- a/kitty/tabs.py
+++ b/kitty/tabs.py
@@ -129,6 +129,8 @@ class TabDict(TypedDict):
     is_active: bool
     title: str
     title_overridden: bool
+    needs_attention: bool
+    has_activity_since_last_focus: bool
     layout: str
     layout_state: dict[str, Any]
     layout_opts: dict[str, Any]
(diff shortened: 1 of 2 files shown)


kovidgoyal:
There is no need to replicate the data in the tb dict if it is already
present in the window dict. clients can preform the loop to deduce the
tabs values if they want to, otherwise all clients pay that cost when
most wont need that information.

muchzill4:
> There is no need to replicate the data in the tb dict if it is already
> present in the window dict

https://github.com/kovidgoyal/kitty/pull/9874/commits/0aa5d32f5f0169e910605b3f547ce0d1574a9636

PR #9890, opened 2026-04-18
macos: explicitly enable modern window corners on macOS 26

zzhaolei:
## Summary

This change explicitly enables the modern window corner style on macOS 26 while preserving the existing behavior on older macOS versions.

## Motivation

On newer macOS releases, the system can present updated window corner styling. However, builds targeting older SDKs or older deployment environments do not use the newer appearance automatically, even when running on macOS 26.

To keep compatibility with older macOS versions, this change applies the new corner styling explicitly only on macOS 26 and later, while leaving earlier versions unchanged.

## Changes

- Add a runtime macOS version check for macOS 26+
- Explicitly apply the newer window corner curve on supported systems
- Keep behavior unchanged on macOS versions below 26
- Link against `QuartzCore` for the required corner curve symbol

## Compatibility

This change is intentionally conservative:

- macOS 26 and later: use the modern corner style explicitly
- macOS 25 and earlier: no behavior change

This ensures we support the new appearance on macOS 26 without affecting users on older releases.

before:
before

after:
after

diff:
diff

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -3918,6 +3918,18 @@ apply_titlebar_color_settings(_GLFWwindow *window) {
 #undef tc
 }
 
+static void
+apply_window_corner_curve(_GLFWwindow *window) {
+    if (!window || !window->decorated) return;
+    if (@available(macOS 26.0, *)) {
+        GLFWWindow *nsw = window->ns.object;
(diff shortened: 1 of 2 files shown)


zzhaolei:
Tested with `hide_window_decorations titlebar-only` and `hide_window_decorations yes`.

PR #9891, opened 2026-04-18
kitten diff: add support for sticky header

pdobrev:
This PR adds optional support for sticky headers in the diff kitten.

The motivation for this is to always know which file you're currently looking the diff for, which is something I find useful when doing big git diffs.

Here's an example of it running for the current diff:

https://github.com/user-attachments/assets/4a7e5d15-e467-4420-b68b-7c8c8a8db9c3

Implementation notes:
- Keep track of where the title lines are located in the local lines array and store that in a sorted array
- From the current position, do a binary search in that array to find the closest title and render it always on top

There's an alternatively approach where we do a full lookup table, so we don't need to use binary search, but the added memory overhead didn't seem meaningful. However, it is a tad simpler in terms of code (we don't need to import sort, no need for helper functions to find title index, etc), so if that's the preferred style, happy to change to it, if there's interest in the feature at all.

Enabled by setting `sticky_header` in the config.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -251,7 +251,7 @@ Detailed list of changes
 
 - macOS: Fix occasional phantom cursors being drawn on screen (:iss:`9725`)
 
-- diff kitten: A new option :opt:`sticky_header` to pin the current file's name to the top of the screen as you scroll through a multi-file diff (:pull:`9891`)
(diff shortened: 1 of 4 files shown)


kovidgoyal:
When scrolling down and no lines are left in the first file the header
of the second file gets hidden by the sticky header of the first file.
See attached screenshot. Also add an entry to the docs/changelog.rst for
this. 
screenshot

pdobrev:
Added a fix for the overlap with the next file's header and an entry in the changelog


https://github.com/user-attachments/assets/86272518-39de-4e86-9968-05d9da66a247

PR #9900, opened 2026-04-20
feat: respawn_window

SharzyL:
Add a `respawn_window` action that restarts the shell (or any child process) in the currently active window without changing its position in the layout.

### Usage

```conf
map f5 respawn_window
map ctrl+f5 respawn_window --cwd current
map shift+f5 respawn_window --cwd /some/path
map alt+f5 respawn_window --env FOO=bar /bin/zsh
```

Supports `--cwd` (including `current`, `last_reported`, `oldest`, `root`), `--env`, `--copy-env`, `--hold`, `--hold-after-ssh`, and positional command arguments. Works correctly in ssh kitten sessions — `respawn_window --cwd current` in an ssh window will re-establish the ssh connection with the remote cwd preserved.

### Motivation

Sometimes a terminal window becomes unresponsive — a hung process, a broken pipe, or a stale ssh session. Currently the only option is to close the window and open a new one, which loses the window's position in the layout and requires navigating back to the desired working directory. `respawn_window` solves this by replacing the child process in-place: the window, its layout position, and its id all remain unchanged. Tmux has a `respawn-pane` action that does exact this thing.

### Implementation notes

The core challenge is that kitty's io_thread continuously `poll()`s on child file descriptors in a separate thread. Normal launch/kill operations update the fd list safely because the changes happen inside the io_loop itself (via `add_queue`/`remove_children`), before the next `poll()` call. But for respawn, using `remove_children` + `add_children` would trigger `death_notify`, which runs the normal window close flow — exactly what we want to avoid. Adding a new "replace queue" would require more invasive changes to the io_loop. Instead, we use `dup2` to swap the underlying pty while keeping the fd number unchanged, so the io_thread transitions seamlessly without any queue mechanism:

1. **Fork a new child process**, which creates a new pty pair. The new child is blocked on `terminal_ready_fd` — it cannot produce output until `mark_terminal_ready()` is called during relayout.
2. **`dup2(new_fd, old_fd)`** — redirect the old fd (which the io_thread is polling) to the new pty. The fd number stays the same, so the io_thread transitions seamlessly. Safe because the new child is still blocked.
3. **`replace_child()` in C** — atomically swap the pid, screen, and clear `needs_removal` in the child monitor.
4. **Relayout** triggers `mark_terminal_ready()`, unblocking the new child. Output goes directly into the new screen.
5. **`killpg(old_pgid, SIGHUP)`** — kill the old process group, matching the normal window close path (`hangup()` in `child-monitor.c`).

The `terminal_ready_fd` mechanism guarantees the new child cannot produce output before the screen swap, so there is no race between dup2 and replace_child.

**Argv preservation for ssh sessions:** When a window is launched with `--cwd=current` from an ssh kitten session, `Child.__init__` rewrites `self.argv` from the original shell (e.g., `['/bin/zsh']`) to the ssh kitten command line via `modify_argv_for_launch_with_cwd`. This rewrite is in-place and destructive. Without care, `respawn()` would pass the already-rewritten ssh command line back to `modify_argv_for_launch_with_cwd`, causing double-rewriting and a corrupted command. To prevent this, `Child` saves `argv_before_cwd_rewrite` before the rewrite in `__init__`, and `respawn()` always resets `self.argv` from it before calling `modify_argv_for_launch_with_cwd` again.

**Files changed:**

| File | Purpose |
|------|---------|
| `kitty/boss.py` | `respawn_window` action, argument parsing consistent with `launch` |
| `kitty/child.py` | `Child.respawn()` — forks new process, supports cwd/argv/env/cwd_from/hold; restores state on fork failure |
| `kitty/child-monitor.c` | `replace_child()` — atomically swaps pid/screen/fd in the child monitor |
| `kitty/window.py` | `Window.respawn_child()` — orchestrates dup2, replace_child, and cleanup |
| `kitty/options/utils.py` | Register `respawn_window` for `shlex_parse` argument handling |
| `docs/changelog.rst` | Changelog entry under 0.47.0 |
| `kitty_tests/respawn.py` | Tests: basic respawn, env propagation, hold, repeated respawn, not-forked guard, dup2 fd reuse |

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -168,6 +168,8 @@ Detailed list of changes
 0.47.0 [future]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+- A new action :ac:`respawn_window` to restart the shell in the currently active window without changing its position in the layout. Supports ``--cwd``, ``--env``, ``--hold``, and command arguments.
+
 - A new option :opt:`palette_generate` to automatically generate the 256 color palette from the first 16 colors (:pull:`9426`)
 
 - For builtin key mappings automatically :ref:`fallback <mapping-fallback>` to matching the US-PC layout key when the pressed key has no matches and is a non-English character (:pull:`9671`)
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -3085,6 +3085,53 @@ class Boss:
             opts.next_to = opts.next_to or f'id:{self.window_for_dispatch.id}'
         launch(self, opts, args_)
 
+    @ac('win', '''
+Respawn the currently active window
+
+Restart the shell in the currently active window without changing its position
+in the layout. The terminal screen will be cleared and a new shell will start.
+Supports :code:`--cwd`, :code:`--env`, :code:`--hold`, and command arguments::
+
+    map f5 respawn_window
+    map ctrl+f5 respawn_window --cwd current
+    map shift+f5 respawn_window --cwd /some/path
+    map alt+f5 respawn_window --env FOO=bar /bin/zsh
+''')
+    def respawn_window(self, *args: str) -> None:
+        from kitty.launch import get_env, parse_launch_args
+        from kitty.window import CwdRequest, CwdRequestType
+
+        window = self.window_for_dispatch or self.active_window
+        if not window:
+            return
+
+        try:
+            opts, cmd_args = parse_launch_args(args)
+        except ValueError as e:
+            self.show_error(_('Invalid arguments'), str(e))
+            return
+
+        cwd: str | None = None
+        cwd_from: CwdRequest | None = None
+        if opts.cwd:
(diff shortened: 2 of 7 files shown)


kovidgoyal:
If all you want to do is create a new window that occupies the same spot
in the layout and runs the same command you can simply spawn an overlay
window, then close the original window. This should be a trivial five
lines of python code. You can probably even do it purely using existing
actions. Something like

```
map f1 combine | launch --cwd=current --type=overlay-main | remote_control close-window --match recent:1
```

SharzyL:
> map f1 combine : launch --cwd=current --type=overlay-main : remote_control close-window --match recent:1

Oh it works but with a little modification

```
map f1 combine | launch --cwd=current --type=overlay-main | remote_control close-window --match state:overlay_parent
```
Use `|` to avoid the last `:` interpreted as a separator.

PR #9919, opened 2026-04-24
Add `--extension` flag to `save_as_session` action

kamalsacranie:
The `--extension` flag allows for a fallback extension to be applied to the session path if no extension is specified. This is useful in combination with `--base-dir` as it allows specifying only the name of the session when prompted.


kovidgoyal:
IMO it would be better to simply always add an extension if missing. No
need for an option.

kamalsacranie:
Omg perfect thanks for adding

PR #9916, opened 2026-04-24
fix ligature

zzhaolei:
## Summary

Fix ligature grouping for fonts that use Iosevka-style `.join-*` glyph names but do not expose `==` / `===` ligatures.

Some fonts, such as Comfy Code, use glyph names like `.join-r`, `.join-m`, and `.join-l` for composable ligature pieces. kitty previously detected this spacer strategy by shaping `==`, which works for Iosevka but fails for fonts where `==` does not form a ligature. As a result, each join glyph was grouped and rendered as an independent cell, causing broken or scrambled ligatures in the main terminal renderer while font preview rendering looked correct.

This change expands spacer strategy detection to shape additional common ligature samples:

- `->`
- `<-`
- `<<=`
- `<==>`

If any of those shaped glyphs use `.join-l`, `.join-m`, or `.join-r`, the font is treated as using the Iosevka-style spacer strategy.

It also handles `.liga + placeholder + final glyph` cases, such as Cascadia-style output like `less_less_equal.liga`, `LIG`, `equal`, so those glyphs stay in a single render group instead of being split incorrectly.

## Details

- Add helper detection for shaped glyph names ending in `.join-l`, `.join-m`, or `.join-r`.
- Reuse a small helper to populate ASCII probe cells for spacer strategy detection.
- Extend spacer strategy probing beyond `==` / `===`.
- Add grouping logic for `.liga` glyphs followed by placeholder glyphs and a matching final component glyph.

## Verification

Tested locally with Comfy Code:

- `<<=` groups as `(3 cells, 3 glyphs)`
- `<===>` groups as `(5 cells, 5 glyphs)`
- `->` groups as `(2 cells, 2 glyphs)`
- `<-->` groups as `(4 cells, 4 glyphs)`

Also ran:

```sh
python3 test.py --module fonts
```

Result:
```
Ran 11 tests

OK (skipped=1)
```

before:
[comfy code](https://github.com/zzhaolei/comfy-code)(custom build iosevka):
image

after:
[comfy code](https://github.com/zzhaolei/comfy-code)(custom build iosevka):
image

cascadia code:
image

iosevka:
image

iosevka term:
image

source code pro(no ligature):
image

PR #9923, opened 2026-04-26
Add tests for glyph_matches_dot_liga_final_component

zzhaolei:
Add focused test coverage for `glyph_matches_dot_liga_final_component()`.

The shaping assertions added in #9916 verified the resulting groups, but they did not actually exercise the code path where `glyph_matches_dot_liga_final_component()` returns `true`. This follow-up adds a small test-only entry point and uses it to verify the helper directly.

### DIFF
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -1489,35 +1489,6 @@ ligature_type_for_glyph(hb_font_t *hbf, glyph_index glyph_id, SpacerStrategy str
     return ligature_type_from_glyph_name(glyph_name, strategy);
 }
 
-static bool
-dot_liga_final_component(const char *glyph_name, char *output, size_t output_sz) {
-    static const char suffix[] = ".liga";
-    static const size_t suffix_len = sizeof(suffix) - 1;
-    size_t len = strlen(glyph_name);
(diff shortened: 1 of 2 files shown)


kovidgoyal:
adding a printf to glyph_matches_dot_liga_final_component shows that
function is not even called for those two tests

kovidgoyal:
That's a pretty artificial test are there no actual strings that exercise
this code path? And if that is the case then why have it at all?

zzhaolei:
I dug into the real shaping output for actual ligature strings and this helper turned out to be speculative code rather than something exercised by the current shaping path.

For example, with real Fira Code ligatures such as `#_(` and `<*>>`, the shaped glyph stream does not match the assumptions behind `glyph_matches_dot_liga_final_component()`. In particular, the branch assumed that the first glyph in the current group would be the `.liga` glyph, but in the real output the group starts with spacer glyphs and the `.liga` glyph appears later. As a result, this branch is not reached for the real ligature cases I checked.

Given that I could not find a real string/font combination in the current test setup that naturally exercises this path, I removed both the speculative branch and the artificial helper test. After removing it, ligatures still shape and group correctly, and the real ligature tests continue to pass.

kovidgoyal:
Did you forget to push, because I still see the test and branch

zzhaolei:
I’ve pushed the latest changes now.

PR #9924, opened 2026-04-26
support up to 15 markers

corpix:
Hi!

Personally I found 3 markers are usually not enough, what do you think about extending it to 15? Should be enough for everyone :smile:


kovidgoyal:
I dont think the overhead is worth it. Spending another two bits per cell
and adding 13 new config options for this use case is not worth it.

PR #9935, opened 2026-04-27
Guard line buffer copy bounds

Idate96:
## Summary
- reject out-of-range `LineBuf.copy_line_to()` row indices before indexing line buffer storage
- add regression coverage for the Python-facing out-of-bounds copy path

## Motivation / Bug
`LineBuf.copy_line_to()` accepts a row index from Python and previously used it directly to index `line_attrs`, `line_map`, and the backing cell storage. If a caller passed `y >= self->ynum`, the C extension could read past the allocated line buffer arrays.

This mirrors the bounds behavior already present in `create_line_copy()`, which rejects an out-of-range row before accessing the same line buffer storage.

A minimal reproducer for the unsafe path is:

```python
from kitty.fast_data_types import LineBuf

lb = LineBuf(2, 3)
line = lb.line(0)
lb.copy_line_to(99, line)
```

Before this change, that call reached the C indexing path with an invalid row. After this change, it raises `ValueError: Out of bounds` before touching the backing arrays.

While I originally found this while investigating a local Kitty 0.46.2 crash in `kitty.fast_data_types.so`, I checked the call sites after review feedback and `LineBuf.copy_line_to()` appears to be used only by tests. So this PR is not meant to claim it fixes that runtime crash; it is a narrow bounds fix for the exposed helper path.

## Validation
- `PATH=$HOME/.local/opt/go1.26.2/bin:$PATH PYTHONNOUSERSITE=1 python3 setup.py build --skip-building-kitten`
- `PATH=$HOME/.local/opt/go1.26.2/bin:$PATH PYTHONNOUSERSITE=1 PYTHONPATH=$PWD python3 -m unittest kitty_tests.datatypes`
- manual reproducer now raises `ValueError: Out of bounds`

### DIFF
--- a/kitty/line-buf.c
+++ b/kitty/line-buf.c
@@ -292,6 +292,7 @@ copy_line_to(LineBuf *self, PyObject *args) {
     unsigned int y;
     Line src, *dest;
     if (!PyArg_ParseTuple(args, "IO!", &y, &Line_Type, &dest)) return NULL;
+    if (y >= self->ynum) { PyErr_SetString(PyExc_ValueError, "Out of bounds"); return NULL; }
     src.xnum = self->xnum; dest->xnum = self->xnum;
     dest->ynum = y;
     dest->attrs = self->line_attrs[y];
--- a/kitty_tests/datatypes.py
+++ b/kitty_tests/datatypes.py
@@ -279,6 +279,8 @@ class TestDataTypes(BaseTest):
         l2 = lb.create_line_copy(2)
         lb.copy_line_to(1, l2)
         self.ae(l2, lb2.line(2))
+        with self.assertRaises(ValueError):
+            lb.copy_line_to(lb.ynum, l2)
         lb.clear_line(0)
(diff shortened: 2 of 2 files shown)


kovidgoyal:
Sticking range checks in hot paths is defensive, true, but it has negative performance characteristics. I prefer sanitizing data when it enters the system rather than checking it everywhere along the code path. Also as far as I know copy_line_to is only used in tests, which is why it doesnt bother with range checks. Are you using some kitten that uses copy_line_to? In any case your PR is fine minus the defensive check in linebuf_init_line, remove that and I will merge.

Idate96:
Thanks, that makes sense. I removed the defensive check from `linebuf_init_line_at()` and kept the bounds check only at the Python-facing `copy_line_to()` entry point.

I also checked the call sites and you are right: `LineBuf.copy_line_to()` appears to be used only by tests. So this PR does not fix the local runtime crash I was investigating. That crash was on 0.46.2 and, after looking at current master, it now looks more likely to have been covered already by the newer paused-rendering/overlay line fixes. I updated the PR description to frame this as the narrow helper bounds fix only.

PR #9983, opened 2026-05-07
Change momentum_scroll value from 0.96 to 0.35

beastr45:
I made this pull request because by default, I think that 0.96 is far too high for momentum_scroll and a smaller value would be a more sane default for all users. Initially I though this was a bug. What happens is that when you scroll it keeps on going long after touchpad scrolling stops. Im not sure what the reasoning for such a high value is because a commit titled ... changed it from 0.04 (good) to 0.94.  I decided to move it to 0.35 because its still there but it does not make using a text editor in the terminal difficult. Since conf.rst does not have any of the configuration code I am assuming its autogenerated and excluded a documentation update from this commit, however I did not test that.

Im don't want to test this, but: This is happening on xorg, so the momentum might have a higher effect on xorg compared to wayland? since it is high enough it seems to be a bug and not a feature. 

Generally I think that momentum scrolling is only really desirable on the web, and not the terminal since it is mainly UI polish and not a good feature for text editors.

### DIFF
--- a/kitty/options/types.py
+++ b/kitty/options/types.py
@@ -619,7 +619,7 @@ class Options:
     mark2_foreground: Color = Color(0, 0, 0)
     mark3_background: Color = Color(242, 116, 188)
     mark3_foreground: Color = Color(0, 0, 0)
-    momentum_scroll: float = 0.96
+    momentum_scroll: float = 0.5
     mouse_hide_wait: MouseHideWait = MouseHideWait(hide_wait=0.0, show_wait=0.0, show_threshold=40, scroll_show=True) if is_macos else MouseHideWait(hide_wait=3.0, show_wait=0.0, show_threshold=40, scroll_show=True)
     notify_on_cmd_finish: NotifyOnCmdFinish = NotifyOnCmdFinish(when='never', duration=5.0, action='notify', cmdline=(), clear_on=('focus', 'next'))
     open_url_with: list[str] = ['default']


kovidgoyal:
I disagree and the commit you reference changed the meaning of the value
from x to 1 - x. Your issues are almost certainly caused by X11, which
has an utterly broken input subsystem.

beastr45:
Regardless, I don't think a terminal should have momentum scrolling at all, the scroll back value I just chose to be lower as a middle ground. No idea what x-1 refers to. Will test on Wayland later, like I said, to me any scrolling momentum is unexpected so since I didn't actually test on Wayland, the default behavior might actually be the same, I just am over exaggerating because I don't like scrolling momentum very much. Scrolling momentum mentioned as Wayland only feature so there is bug where feature is applied to x11 windows when it shouldn't?

X11 input isn't really broken because most compositors use the same input library as X11 but I may be mistaken there.

kovidgoyal:
On Wed, May 06, 2026 at 08:42:44PM -0700, Bear wrote:
> beastr45 left a comment (kovidgoyal/kitty#9983)
>
> Regardless, I don't think a terminal should have momentum scrolling at all, the scroll back value I just chose to be lower as a middle ground. No idea what x-1 refers to. Will test on Wayland later, like I said, to me any scrolling momentum is unexpected so since I didn't actually test on Wayland, the default behavior might actually be the same, I just am over exaggerating because I don't like scrolling momentum very much. Scrolling momentum mentioned as Wayland only feature so there is bug where feature is applied to x11 windows when it shouldn't?

That's YOUR opinion, if you want your opinion to set the defaults, you
need to create your own project.

>
> X11 input isn't really broken because most compositors use the same input library as X11 but I may be mistaken there.

Yes, you are mistaken.

beastr45:
ok, my bad. Good luck :)

Issue #9857, opened 2026-04-12
Add fake slant support for CJK fonts


0And1Story:
**Is your feature request related to a problem? Please describe.**

I'm trying to configure Chinese fonts in kitty. I prefer to use JetBrains Mono in English and Noto Sans CJK SC in Chinese.

To enable italic in Chinese fonts, I use `matrix` in fontconfig. But it does not take effects at all.

```xml
  <!-- italic (fake slant) -->
  <match target="font">
    <test name="family"><string>Noto Sans CJK SC</string></test>
    <test name="slant" compare="more_eq"><const>italic</const></test>
    <edit name="matrix" mode="assign">
      <matrix>
        <double>1</double><double>0.2</double>
        <double>0</double><double>1</double>
      </matrix>
    </edit>
    <edit name="slant" mode="assign"><const>italic</const></edit>
    <edit name="embeddedbitmap" mode="assign"><bool>false</bool></edit>
  </match>
```

Since most CJK fonts do no provide italic glyphs, it is very inconvenient.

(Well, I know that fake slant is not a good feature in font designing, but it is special and important in CJK fonts.)

More information at discussion [How to enable fake slant feature in kitty?](https://github.com/kovidgoyal/kitty/discussions/9700).

**Describe the solution you'd like**

Handle matrix settings of fontconfig in kitty.

**Describe alternatives you've considered**

Add more features related to font fallbacks.

kovidgoyal:
Not something I wish to spend time on. PR is welcome.

PR #9990, opened 2026-05-09
Read FC_MATRIX from fontconfig

Strykar:
Refs #9857, #9700.

## What

`pattern_as_dict()` in `kitty/fontconfig.c` extracts file path, family,
weight, slant, hinting, and so on, but never called
`FcPatternGetMatrix`. Any per-font matrix set by a user's fontconfig
rule was silently dropped before reaching FreeType.

Most users hit this without realising it. fontconfig ships
`/etc/fonts/conf.d/90-synthetic.conf` (default-enabled in every distro)
that applies a slant matrix `(1, 0.2, 0, 1)` to any roman-only font
when italic is requested. So italic Chinese, Japanese, Korean, and any
other language with no italic-style font files has been rendering
upright in kitty for as long as kitty has supported CJK.

## Changes

1. **`kitty/fontconfig.c`**: read `FC_MATRIX` and add it to the
   descriptor dict as `(xx, xy, yx, yy)`. The key is only present when
   fontconfig actually sets a matrix.
2. **`kitty/freetype.c`**: in `face_from_descriptor()`, read the tuple,
   convert to 16.16 fixed point, store on `Face`, call
   `FT_Set_Transform` once if non-identity. Identity is filtered out so
   faces without a transform incur no extra work.
3. **`kitty/freetype.c`**: also call `hb_font_set_synthetic_slant`
   (slant = `xy/xx`) and `hb_ft_font_changed` so HarfBuzz shaping (mark
   positioning, cluster boundaries) reflects the slanted rendering. The
   `synthetic_slant` call is gated on `HB_VERSION_ATLEAST(4,0,0)` since
   `setup.py` allows down to 1.5.0.
4. **`kitty/freetype.c`**: extend `face_equals_descriptor()` to compare
   the matrix. Without this, the per-FontGroup fallback cache returns
   an unmatrixed face for a matrixed descriptor (or vice versa)
   whenever upright and italic share a font file - the common case for
   CJK, where one .ttc serves both slants and only the matrix differs.

A small `read_matrix_from_descriptor` helper is shared between
`face_from_descriptor` and `face_equals_descriptor` so the
identity-skip, tuple-parsing, and `isfinite()` validation stay in one
place.

## Why this is safe

`FT_Set_Transform` is sticky on the FT face: set once at init,
inherited by every later `FT_Load_Glyph`, no per-call overhead. The
per-`Face` glyph atlas cache stays correct because the matrix never
changes after init.

`FT_Set_Transform` only affects per-glyph outline transforms before
rasterization. Face-level metrics from `FT_Set_Char_Size` are
untouched, so cell width and height calculation is unchanged. For pure
shears (`xx=1, yy=1`) horizontal advance is preserved and monospace
layout is intact.

## Verification

Built and tested locally on Arch with `noto-fonts-cjk` installed and
default fontconfig (no custom rule):

- `kitty.fonts.fontconfig.fc_match('Noto Sans CJK SC', italic=True)`
  returns a descriptor containing `'matrix': (1.0, 0.2, 0.0, 1.0)`,
  applied by the stock 90-synthetic rule.
- Same query for upright leaves matrix unset.
- Rendering U+4E2D (中) from the resolved face: 23 px wide upright, 25
  px wide italic. Both descriptors point at the same `.ttc` file,
  exercising the `face_equals_descriptor` cache equality path.
- All 329 existing kitty tests pass.

## Caveats

- Underline and strikethrough position come from face-level metrics
  (`face->size->metrics`) and are not affected by `FT_Set_Transform`,
  so a synthetic-italic glyph keeps an upright underline. This matches
  per-cell underline rendering in every terminal.
- Matrix is honoured on the fallback path (which goes through
  `FcFontMatch` and runs `target="font"` rules) but not on the primary
  `font_family` resolution path (`find_best_match` walks the
  `FcFontList` cache which does not). Affects users who set
  `font_family` directly to a CJK font; doesn't affect the asker's use
  case in #9700 (CJK as fallback in italic context). The fix is
  Python-side: run `FcConfigSubstitute(FcMatchFont)` on font-map
  results in `kitty/fonts/fontconfig.py`. Worth a follow-up issue.
- Color glyph rendering (COLR/CBDT/SVG) goes through cairo, which
  unconditionally calls `FT_Set_Transform` on the FT face from its own
  scale matrix on every render
  (`_cairo_ft_unscaled_font_set_scale` in `cairo-ft-font.c`), so the
  matrix is not honoured on that path. Stock fontconfig rules don't
  apply `FC_MATRIX` to color fonts, so this is a hand-built config
  edge case. A correct fix would route the matrix through cairo via
  `cairo_set_font_matrix` instead of `cairo_set_font_size`.
- `get_glyph_width` uses `FT_LOAD_DEFAULT` (no rasterization) and
  returns `M.width` from the glyph metrics, which `FT_Set_Transform`
  does not update. For pure shears (`xx=1`) this is the unsheared bbox
  width and equals the unsheared advance, which is the right answer
  for monospace cell-counting. Non-shear matrices on monospace fonts
  are out of scope.

## Scope

Linux only. macOS kitty uses CoreText, not fontconfig, so there is no
`FC_MATRIX` to read on that platform.

### DIFF
--- a/kitty/fontconfig.c
+++ b/kitty/fontconfig.c
@@ -41,6 +41,7 @@ static struct {PyObject *face, *descriptor;} builtin_nerd_font = {0};
 #define FcPatternAddInteger dynamically_loaded_fc_symbol.PatternAddInteger
 #define FcPatternCreate dynamically_loaded_fc_symbol.PatternCreate
 #define FcPatternGetBool dynamically_loaded_fc_symbol.PatternGetBool
+#define FcPatternGetMatrix dynamically_loaded_fc_symbol.PatternGetMatrix
 #define FcPatternAddCharSet dynamically_loaded_fc_symbol.PatternAddCharSet
 #define FcConfigAppFontAddFile dynamically_loaded_fc_symbol.ConfigAppFontAddFile
 
@@ -66,6 +67,7 @@ static struct {
     FcBool (*PatternAddInteger) (FcPattern *p, const char *object, int i);
     FcPattern * (*PatternCreate) (void);
     FcResult (*PatternGetBool) (const FcPattern *p, const char *object, int n, FcBool *b);
+    FcResult (*PatternGetMatrix) (const FcPattern *p, const char *object, int n, FcMatrix **m);
     FcBool (*PatternAddCharSet) (FcPattern *p, const char *object, const FcCharSet *c);
     FcBool (*ConfigAppFontAddFile) (FcConfig *config, const FcChar8 *file);
 } dynamically_loaded_fc_symbol = {0};
@@ -117,6 +119,7 @@ load_fontconfig_lib(void) {
         LOAD_FUNC(PatternAddInteger);
         LOAD_FUNC(PatternCreate);
         LOAD_FUNC(PatternGetBool);
+        LOAD_FUNC(PatternGetMatrix);
         LOAD_FUNC(PatternAddCharSet);
         LOAD_FUNC(ConfigAppFontAddFile);
 }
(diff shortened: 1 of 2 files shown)


0And1Story:
2026-05-12-084225_hyprshot<br>

2026-05-12-085028_hyprshot<br>

That's exactly what I want. lgtm

PR #9991, opened 2026-05-10
Honor background image overrides for single instance windows

oxysoft:
When kitty is invoked with --single-instance and an existing instance handles the request, the new invocation's parsed options are used for sizing and background opacity, but background_image/background_image_layout overrides are ignored. The new OS window therefore inherits the running instance's configured background and callers have to repaint it after launch.\n\nThis applies an explicit background_image override to the newly created OS window during the single-instance handoff, without changing the configured background for existing or future windows. It also stores the requested background layout on per-window background overrides so a one-shot launch can honor e.g. background_image_layout=cscaled instead of depending on the process-global layout.

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -977,8 +977,16 @@ class Boss:
                     override_title=args.title or None, window_state=wstate, x=pos_x, y=pos_y)
                 if session.focus_os_window:
                     focused_os_window = os_window_id
-                if opts.background_opacity != get_options().background_opacity:
+                global_opts = get_options()
+                if opts.background_opacity != global_opts.background_opacity:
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Not sure I follow, why do you need to explicitly check if the
option is overriden? Simply check if the value is different
between opts and global_opts()

oxysoft:
Good point, fixed. I removed the raw CLI override check and now compare the parsed incoming opts against the running global opts before applying a per-window background image.

I kept the small per-window layout storage because `set_background_image(..., configured=False, layout=...)` already accepts a layout for the target OS window, but the renderer was still reading the process-global `background_image_layout` for per-window background image overrides. Without that, the new single-instance window gets the requested image but still renders it using the existing instance's layout.

PR #10007, opened 2026-05-13
fontconfig: apply target="font" rules to primary font_family path

Strykar:
`find_best_match()` picks candidates from the `FcFontList` cache populated by `fc_list()`. `FcFontList` skips both `fontconfig` substitution phases, so `target="font"` rules never run and per-font edits like `FC_MATRIX` never attach to the chosen descriptor. The fallback path in the same module goes through `fc_match() -> FcFontMatch()`, which does run `MatchFont`, so `fontconfig.fc_match()` in #9990 returned the matrix correctly. The two paths diverged for any user setting font_family directly to a CJK font.

Bind `FcFontRenderPrepare` and expose `fc_render_prepare(family, bold, italic, spacing, candidate)` at the Python boundary. It is the same call `FcFontMatch` makes internally to combine a request pattern with a listed font, including the `MatchFont` substitution pass.

Wrap the three `FcFontList`-derived returns in `find_best_match()` with `fc_render_prepare()` so the descriptor that reaches Face construction carries the matrix.

`dict_to_font_pattern` forwards only intrinsic file-level properties. User-config fields (hinting, hint_style, subpixel, lcdfilter) are omitted on purpose: `pattern_as_dict` defaults missing booleans to False and missing ints to 0, so we cannot distinguish "fontconfig set this" from "the default leaked through." Forwarding them would let stale defaults preempt `MatchFont` substitution for rules using `mode="add"` or `mode="append"`. The `fc_match` fallback path leaves them alone too.

Verified locally on Arch with stock fontconfig and noto-fonts-cjk:

  >>> from kitty.fast_data_types import fc_list, fc_render_prepare
  >>> cjk = [c for c in fc_list(allow_bitmapped_fonts=True)
  ...        if c.get('family') == 'Noto Sans CJK SC']
  >>> for c in cjk[:2]:
  ...     p = fc_render_prepare('Noto Sans CJK SC', False, True, -1, c)
  ...     assert c.get('spacing') == p.get('spacing')  # no field regressed
  >>> from kitty.fonts.fontconfig import find_best_match
  >>> find_best_match('Noto Sans CJK SC', italic=True, monospaced=False)
  ... .get('matrix')
  (1.0, 0.2, 0.0, 1.0)
  >>> find_best_match('Noto Sans CJK SC', italic=False, monospaced=False)
  ... .get('matrix')
  None

Refs: #9857, #9990

### DIFF
--- a/kitty/fast_data_types.pyi
+++ b/kitty/fast_data_types.pyi
@@ -457,6 +457,13 @@ def fc_match_postscript_name(
 
 def add_font_file(path: str) -> bool: ...
 def set_builtin_nerd_font(path: str) -> Union[CoreTextFont, FontConfigPattern]: ...
+def fc_render_prepare(
+    family: str,
+    bold: bool,
+    italic: bool,
+    spacing: int,
+    candidate: FontConfigPattern,
+) -> FontConfigPattern: ...
 
 
 class FeatureData(TypedDict):
--- a/kitty/fontconfig.c
+++ b/kitty/fontconfig.c
@@ -44,6 +44,7 @@ static struct {PyObject *face, *descriptor;} builtin_nerd_font = {0};
 #define FcPatternGetMatrix dynamically_loaded_fc_symbol.PatternGetMatrix
 #define FcPatternAddCharSet dynamically_loaded_fc_symbol.PatternAddCharSet
 #define FcConfigAppFontAddFile dynamically_loaded_fc_symbol.ConfigAppFontAddFile
+#define FcFontRenderPrepare dynamically_loaded_fc_symbol.FontRenderPrepare
 
 static struct {
     FcBool(*Init)(void);
@@ -70,6 +71,7 @@ static struct {
     FcResult (*PatternGetMatrix) (const FcPattern *p, const char *object, int n, FcMatrix **m);
     FcBool (*PatternAddCharSet) (FcPattern *p, const char *object, const FcCharSet *c);
     FcBool (*ConfigAppFontAddFile) (FcConfig *config, const FcChar8 *file);
+    FcPattern * (*FontRenderPrepare) (FcConfig *config, FcPattern *pat, FcPattern *font);
 } dynamically_loaded_fc_symbol = {0};
 #define LOAD_FUNC(name) {\
     *(void **) (&dynamically_loaded_fc_symbol.name) = dlsym(libfontconfig_handle, "Fc" #name); \
@@ -122,6 +124,7 @@ load_fontconfig_lib(void) {
         LOAD_FUNC(PatternGetMatrix);
         LOAD_FUNC(PatternAddCharSet);
         LOAD_FUNC(ConfigAppFontAddFile);
+        LOAD_FUNC(FontRenderPrepare);
 }
 #undef LOAD_FUNC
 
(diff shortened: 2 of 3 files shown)


kovidgoyal:
This is failing a test on my Linux system with
```
======================================================================
FAIL: test_font_selection (kitty_tests.fonts.Selection.test_font_selection) (spec='consolas')
  File "/home/kovid/work/kitty/kitty/launcher/../../kitty_tests/fonts.py", line 62, in s
    self.ae(expected, actual)
    ~~~~~~~^^^^^^^^^^^^^^^^^^
    actual = ('Consolas', 'Consolas', 'Consolas', 'Consolas')
    alternate = None
    expected = ('Consolas', 'Consolas-Bold', 'Consolas-Italic', 'Consolas-BoldItalic')
    family = 'consolas'
    opts = <kitty.options.types.Options object at 0x7f23b5f40190>
    self = <kitty_tests.fonts.Selection testMethod=test_font_selection>
    x = 'Consolas'
AssertionError: Tuples differ: ('Consolas', 'Consolas-Bold', 'Consolas-Italic', 'Consolas-BoldItalic') != ('Consolas', 'Consolas', 'Consolas', 'Consolas')

First differing element 1:
'Consolas-Bold'
'Consolas'

- ('Consolas', 'Consolas-Bold', 'Consolas-Italic', 'Consolas-BoldItalic')
?                       -----            -------            -----------

+ ('Consolas', 'Consolas', 'Consolas', 'Consolas')

Ran 1 test in 0.042s
```

Selection of bold/italic/bold-italic variants of Consolas is breaking.

kovidgoyal:
More information. It fails when the spec is "consolas" but not when the spec is family=consolas. The difference in the two is the former returns hint_style=1 as a result of get_font_from_spec() with your PR and the latter returns hint_style=0. Without your PR both return hint_style=0. Presumably some matching rule in fontconfig that applies in renderprepare but not in list is causing it. Which generally means this PR is likely to break font selection.

kovidgoyal:
The following diff illustrates the issue:
```
diff --git a/kitty/fonts/fontconfig.py b/kitty/fonts/fontconfig.py
index ecfbef93c..c2fb84187 100644
--- a/kitty/fonts/fontconfig.py
+++ b/kitty/fonts/fontconfig.py
@@ -204,6 +204,10 @@ def prepare(p: FontConfigPattern) -> FontConfigPattern:
                 continue  # IBM Plex Mono has fullname of regular face == family_name under fontconfig
             exact_match = find_best_match_in_candidates(cq, scorer, is_medium_face, ignore_face=ignore_face)
             if exact_match:
+                if 'consolas' in family.lower():
+                    if prepare(exact_match) != exact_match:
+                        import deepdiff
+                        print(33333, bold, italic, deepdiff.DeepDiff(exact_match, prepare(exact_match)))
                 return prepare(exact_match)
```

For bold=True and italic=False we see the dict has changed as follows:
```
{'values_changed': {"root['weight']": {'new_value': 200, 'old_value': 80}, "root['hint_style']": {'new_value': 1, 'old_value': 0}, "root['lcdfilter']": {'new_value': 1, 'old_value': 0}, "root['hinting']": {'new_value': True, 'old_value': False}}}
```

In other words fc_render_prepare is changing not just hinting/lcdfilter but also font weight.

kovidgoyal:
Most likely the correct approach is to apply matching rules when the descriptor is turned into an actual face for rendering not during listing. But it's been a awhile since I looked at this code so...

PR #10009, opened 2026-05-13
Resolve characters eagerly for Dvorak - QWERTY ⌘ layout

### LINKED ISSUE #10008: cmd+n does not create new window when using "Dvorak - QWERTY ⌘" layout
**Describe the bug**
I upgraded to Tahoe 26.5 recently and ran into an issue with the default cmd+n keybinding passing the text to the terminal window when using the "Dvorak - QWERTY ⌘" layout.

**To Reproduce**
1. Install Tahoe 26.5 and latest kitty version (0.46.2).
2. Switch the keyboard layout to `Dvorak - QWERTY ⌘`
3. Run `/Applications/kitty.app/Contents/MacOS/kitty --config=NONE`
4. Type `aoeu` (left home row) then cmd+N. (At least a few characters are necessary, it doesn't always work with zero or one, four seems to work reliably.)

Instead of opening a new window, the `8;9u` characters are typed to the terminal window. Other cmd combinations are also broken (e.g. cmd+q).

**Screenshots**

Image

**Environment details**

```
kitty 0.46.2 (4483a6c110) created by Kovid Goyal
Darwin hostname 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:15 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6041 arm64
ProductName:		macOS ProductVersion:		26.5 BuildVersion:		25F71
OpenGL: '4.1 Metal - 90.5' Detected version: 4.1
Frozen: True
Fonts:
  medium: Menlo-Regular: /System/Library/Fonts/Menlo.ttc
          Features: ()
    bold: Menlo-Bold: /System/Library/Fonts/Menlo.ttc
          Features: ()
  italic: Menlo-Italic: /System/Library/Fonts/Menlo.ttc
          Features: ()
      bi: Menlo-BoldItalic: /System/Library/Fonts/Menlo.ttc
          Features: ()
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /opt/homebrew/bin/bash
System color scheme: dark. Applied color theme type: none

Changed shortcuts:
	cmd+k →  clear_terminal_and_scrollback
	cmd+l →  clear_last_command
	ctrl+cmd+, →  reload_config
	ctrl+cmd+l →  clear_screen
	opt+cmd+k →  clear_scrollback
	opt+cmd+r →  reset_terminal
	shift+cmd+/ →  open_kitty_website

Important environment variables seen by the kitty process:
	PATH                                /Applications/kitty.app/Contents/MacOS:usr/local/sbin:/usr/sbin:/usr/bin:/sbin:/bin
	LANG                                en_US.UTF-8
	EDITOR                              vim
	SHELL                               /opt/homebrew/bin/bash
	USER                                user
	LC_COLLATE                          C
```

**Additional context**
Reproducible with `kitty --config NONE`

jvstein:
This appears to resolve #10008 for me.

~This helps with cmd+n, but doesn't resolve all the issues. cmd+v still appears broken.~

EDIT: Nevermind, this was my config. CMD+V is fine.

### DIFF
--- a/glfw/cocoa_init.m
+++ b/glfw/cocoa_init.m
@@ -930,6 +930,10 @@ int _glfwPlatformInit(bool *supports_window_occlusion)
         debug_key("---------------- key down -------------------\n");
         debug_key("%s\n", [[event description] UTF8String]);
         if (!_glfw.ignoreOSKeyboardProcessing && !_glfw.keyboard_grabbed) {
+            // On macOS Tahoe, performKeyEquivalent fails to match key
+            // equivalents on remapped layouts like "Dvorak - QWERTY ⌘"
+            // unless the event's characters property has been accessed first.
+            (void)[event characters];
             // first check if there is a global menu bar shortcut
             if ([[NSApp mainMenu] performKeyEquivalent:event]) {
                 debug_key("keyDown triggered global menu bar action ignoring\n");


kovidgoyal:
I am rather sceptical about this. I dont see why calling that function
should have any effect, and if it actually does then it's quite likely
that whatever effect that might be, might break something else down the
line.

jvstein:
There's quite likely some sort of timing aspect to it. Running with `--debug-keyboard` does not replicate the issue. It's possible calling the function just adds enough of a delay to never trigger the issue for me.

Without it, kitty is pretty unusable without the cmd shortcuts. No other applications are affected.

With the change, I haven't replicated the issue in a day of heavy usage.

jvstein:
For additional context, when it happens I can typically use cmd+` once and then the cmd shortcuts will work for at least one invocation. That initially had me suspecting a hidden window was capturing the keystroke.

kovidgoyal:
I dont see what possible timing issue there could be since this is all
single threaded code, on the GUI thread. But easily check by using a
sleep.

jvstein:
> I dont see what possible timing issue there could be since this is all single threaded code, on the GUI thread. But easily check by using a sleep.

No, that's fair. It's fully reproducible with a sleep replacing the `characters` call.

jvstein:
I think I tracked it back to my mouse driver. ☹️

Sorry for the noise.

jvstein:
No, that was a one-off fluke. I turned off all my HID drivers and am still reproducing it. It's only started since the upgrade, so I'm sure something has changed under the hoods regarding the way keystrokes are translated for this particular keyboard layout. I can't ever reproduce it on the plain Dvorak or QWERTY keyboards.

kovidgoyal:
> > I dont see what possible timing issue there could be since this is all single threaded code, on the GUI thread. But easily check by using a sleep.
> 
> No, that's fair. It's fully reproducible with a sleep replacing the `characters` call.

Do you mean the bug happens even with a sleep or that it doesnt happen with a sleep?

jvstein:
> Do you mean the bug happens even with a sleep or that it doesnt happen with a sleep?

The bug happens even with a sleep. I tried 50ms and 500ms.

jvstein:
I just tried rebooting in safe mode and was still able to reproduce it. I'm upgrading another laptop I have to see if it's reproducible on that one.

kovidgoyal:
> > Do you mean the bug happens even with a sleep or that it doesnt happen with a sleep?
> 
> The bug happens even with a sleep. I tried 50ms and 500ms.

That implies that line is having some side effect, I am definitely not comfortable doing that without understanding what it does as key handling is a particularly fragile/messy area because of macOS' rather crazy system where individual apps are responsible for handling global shortcuts not the OS. Closing pending more information, feel free to update but as is I am not willing to make this change.

jvstein:
I agree that it's likely a side effect and understand the hesitance. It feel like a regression in the OS, not something the app should have to deal with.

After upgrading the OS to 26.5, I was able to reproduce the issue on my old "13-inch, M1, 2020" MacBook Pro, as well.

kovidgoyal:
Yeah I have a 2020 M1 as well (used purely for testing kitty and calibre as
I dont use macs), but doesnt repro for me.

jvstein:
After some additional digging, I'm pretty sure this is a lazy evaluation bug in macOS specific to this keyboard layout. I've sent the feedback their way.

jvstein:
Thank you for looking. I mostly only use macOS for work. kitty on Linux has always been super solid.

I'll live with a custom build including this patch for now. I'll report back if it's fixed in a later patch version.

Thanks again for all you do with both kitty and calibre! ❤️

jvstein:
I updated to the latest 0.49.0 version of kitty on two different macs, one on macOS 26.6.2 and the other on macOS 26.7.0.

I'm no longer able to reproduce the issue on either machine, so presumably this was fixed in the OS and I no longer need my custom build with this change.

PR #10035, opened 2026-05-19
Quote SSH bootstrap symlink moves

M-Hassan-Raza:
This fixes SSH copy path handling so generated tar paths stay under the expected home/root archive roots.

It avoids false HOME prefix matches like /home/me vs /home/me2, rejects copy destinations containing .., matches exclude patterns against remote copy names, and quotes symlink moves safely in the remote bootstrap script.

Tested with go test ./kittens/ssh, ./test.py --module ssh, python3 .github/workflows/ci.py gofmt, and git diff --check master...HEAD.

### DIFF
--- a/kitty_tests/ssh.py
+++ b/kitty_tests/ssh.py
@@ -81,6 +81,9 @@ print(' '.join(map(str, buf)))'''), lines=13, cols=77)
                 os.mkdir(f'{local_home}/d1/r')
                 touch('d1/r/noooo')
                 os.symlink('d2/x', f'{local_home}/d1/y')
+                bad_link_name = '$(touch SHOULD_NOT_EXIST)'
+                os.symlink('d2/x', os.path.join(local_home, 'd1', bad_link_name))
+                os.symlink('-dash-target', os.path.join(local_home, 'd1', 'dash-target-link'))
                 os.symlink('simple-file', f'{local_home}/s1')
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Quoting symlink moves is fine, but copy is not supposed to be contained
in HOME. Quoting the docs for it:

Copy files and directories from local to remote hosts. The specified files are assumed to be relative to the HOME directory and copied to the HOME on the remote. Directories are copied recursively. If absolute paths are used, they are copied as is.


So they are *relative* to HOME but not *contained* to HOME. Indeed even
absolute paths are allowed, so there is no question of containment.

M-Hassan-Raza:
> Quoting symlink moves is fine, but copy is not supposed to be contained in HOME. Quoting the docs for it:
> 
> Copy files and directories from local to remote hosts. The specified files are assumed to be relative to the HOME directory and copied to the HOME on the remote. Directories are copied recursively. If absolute paths are used, they are copied as is.
> 
> So they are _relative_ to HOME but not _contained_ to HOME. Indeed even absolute paths are allowed, so there is no question of containment.

Yeah fair point, I conflated the internal archive layout with the remote copy path contract :(

I narrowed the PR to only the symlink move quoting fix. PLease lmk if it's solid now or any further changes are needed (i can trim, change etc further as much as needed to get this merged).

PR #10041, opened 2026-05-20
Preserve copy_or_noop passthrough result

M-Hassan-Raza:
This fixes a regression where copy_or_noop no longer passed the mapped key through when there was no active selection.

Window.copy_or_noop() already returns True in the no-selection case, and Boss.dispatch_action() uses exactly True to mean “do not consume this key”. The Boss.copy_or_noop() wrapper was calling the window method but not returning its result, so the True became None and the key was consumed.

I added a focused regression test for both paths: passthrough when the window returns True, and consumed when it returns False. The test is a little longer than the fix because it exercises the dispatcher behavior rather than just calling the wrapper directly. Happy to trim or reshape it if you prefer a smaller test.

Tested with:
./test.py --module keys
python3 -m py_compile kitty/boss.py kitty_tests/keys.py

### DIFF
--- a/kitty/boss.py
+++ b/kitty/boss.py
@@ -3737,6 +3737,6 @@ class Boss:
         if w := self.active_window:
             w.search_scrollback()
 
-    def copy_or_noop(self) -> None:
+    def copy_or_noop(self) -> bool | None:
         if w := self.active_window:
-            w.copy_or_noop()
+            return w.copy_or_noop()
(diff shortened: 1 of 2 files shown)


M-Hassan-Raza:
Hmmm I think the issue has been fixed upstream already? Great! I'll close the PR shortly in that case :D

kovidgoyal:
This PR was actually merged, not sure why github didnt auto close it.

M-Hassan-Raza:
> This PR was actually merged, not sure why github didnt auto close it.

Given how Github is doing these days, I am not shocked at all hahahaha. Thankfully it's not a big deal I'm just glad the uostream got the fixes it needed.

PR #10051, opened 2026-05-21
Add the Orbiton text editor to the documentation

xyproto:


xyproto:
Thanks for creating Kitty, btw, it is excellent and I use it almost every day! :)

kovidgoyal:
You are welcome but please add your tool in only one place, either viewer or editors depending on its primary use case.

xyproto:
I updated the PR. :+1:

PR #10084, opened 2026-05-28
Make first mouse clicks activate and pass through

zzhaolei:
## Summary

Allow first mouse clicks to both activate/focus the target and pass through to the clicked element.

This applies to:
- inactive Cocoa OS windows via `acceptsFirstMouse:`
- inactive kitty windows inside an already focused OS window

## Details

The original Cocoa-only change would have made inactive OS windows pass first clicks through, while inactive kitty windows inside a focused OS window still treated the first click as focus-only.

This patch updates the kitty-internal behavior as well, so both levels now follow the same rule: the first click focuses the target and continues through the normal mouse handling path.

In particular, clicking an inactive kitty window inside a focused OS window still switches kitty focus to the clicked window, but no longer suppresses forwarding of the initial left click and matching release. This allows the click to be handled normally, including by kitty mouse mappings and mouse-tracking applications.

This change only adjusts Cocoa OS-window behavior and kitty's own internal window behavior.

## Testing

- `./dev.sh build`
- `./test.py --module mouse`
- `./test.py --module glfw`
- `./test.py --module layout`

Implemented with assistance from Codex.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1015,7 +1015,7 @@ static void _glfwUpdateNotchCover(_GLFWwindow*);
 - (BOOL)acceptsFirstMouse:(NSEvent *)event
 {
     (void)event;
-    return NO;  // changed by Kovid, to follow cocoa platform conventions
+    return YES;  // follow cocoa platform conventions
 }
 
 - (void)mouseDown:(NSEvent *)event


kovidgoyal:
This change is not that simple, if you make OS Windows behave this way
then you need to make kitty windows behave this way too, currently
clicking on an unfocused kitty window in an OS window that has focus
only switches kitty window focus only but does not pass through. The
behavior needs to be consistent. And one has to check if the linux
platform conventions are the same, though with linux there probably are
no platform conventions.

zzhaolei:
Thanks for the feedback. You're right — this change needs more thought to ensure consistency between OS window-level and kitty internal window-level click behavior. Closing for now while I rework the approach.

Issue #10090, opened 2026-05-29
Recommended way to avoid graphics disk-cache writes for high-frequency image updates


Mekann2904:
**Is your feature request related to a problem? Please describe.**

I am experimenting with using kitty's graphics protocol as a rendering target for a browser display.

The application renders the current contents of a web page as an image and sends it to kitty repeatedly.

The workload repeatedly sends full-frame images to kitty at a high frequency. POSIX shared memory transfer (`t=s`) works correctly as a client-to-kitty transfer mechanism, but kitty still appears to write the received image data into the graphics disk-cache.

The main concern is SSD write traffic caused by repeatedly writing short-lived frame-sized images to the graphics disk-cache.

In my current test case:

* each full-frame image update is roughly 10 MB;
* updates happen frequently;
* each image is usually replaced by the next update shortly afterwards;
* old frames are not intended to be reused long-term;
* the workload is closer to a transient frame buffer than to persistent image display;
* POSIX shared memory transfer (`t=s`) appears to work correctly.

On macOS, kitty appears to perform large writes to its graphics disk-cache during these updates.

Observed with `fs_usage` on macOS:

```text
kitty ... REG ... /Users/<user>/Library/Caches/kitty/disk-cache-...
kitty ... pwrite F=26 B=0x9ef7b4
```

`0x9ef7b4` is about 10 MB, which roughly matches the size of one frame in this workload. Since the frame is updated frequently, this can produce substantial SSD write traffic even though the client-to-kitty transfer itself uses shared memory.

My understanding is that `t=s` only controls the transmission mechanism from the client to kitty. It does not change how kitty stores the received image data internally.

Looking at the source code, `GraphicsManager` appears to create a disk cache in `kitty/graphics.c`. Image data then appears to go through `add_to_cache()` and `add_to_disk_cache()`. In `kitty/disk-cache.c`, the cache writer appears to write data to the cache file using `pwrite()`.

So my current understanding is:

```text
`t=s` controls the client -> kitty transfer mechanism
↓
kitty reads the POSIX shared memory object
↓
kitty stores the received image data in its internal graphics disk-cache
↓
the disk-cache writes the image data to the cache file
```

Is this understanding correct?

**Describe the solution you'd like**

I would like to know whether there is a supported way to avoid graphics disk-cache writes for images that are expected to be replaced very soon.

[...]

PR #10092, opened 2026-05-30
graphics: add memory-only storage for graphics data

Mekann2904:
Add a new graphics protocol key, N=1, to request that transmitted image/frame data is kept only in memory and not written to the graphics disk cache file.

This is useful for transient high-frequency updates such as video-like streams, where the latest frame is the only useful data and persisting each frame to the disk cache causes unnecessary write traffic.

The implementation keeps the existing graphics cache abstraction intact: memory-only entries can still be read back by animation, composition, and frame coalescing paths. Only persistence to the disk cache file is skipped.

The default behavior is unchanged when N is omitted or set to zero.

Related to #10090

### DIFF
--- a/docs/graphics-protocol.rst
+++ b/docs/graphics-protocol.rst
@@ -1051,8 +1051,16 @@ Key      Value                 Default    Description
 ``o``    Single character.     ``null``   The type of data compression.
          ``only z``
 ``m``    zero or one           ``0``      Whether there is more chunked data available.
-``N``    zero or one           ``0``      If set to ``1``, keep the transmitted image or frame data in memory only,
-                                          without writing it to the graphics disk cache.
+``N``    bitmask               ``0``      Usage hints from the client to the terminal about the intended use of
+                                          the image. Only one hint is currently defined, the ``1`` bit which means
+                                          *transient*. The terminal is free to assume that an image with this hint
+                                          will be used for only a short time, and so may, for example, evict its
+                                          data before other images when the image is soft deleted, has no visible
+                                          placements and the terminal is under storage pressure, or skip writing
+                                          its data to disk. The terminal is also free to ignore the hint. If an
+                                          animation frame with the *transient* hint is composited onto another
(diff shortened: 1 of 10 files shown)


kovidgoyal:
The graphics protocol isnt meant for video or display output, a protocol
for these would be very different in design. That said, I am not
necessarily opposed to adding a lifetime specifier to the protocol,
however it requires some careful thought about use cases. For instance
it might be better to have a cache strategy rather than a simple boolean.
Or a timeout for automatic deletion or some combination and so on.

I will address this in due time when I have the bandwidth for it.

kovidgoyal:
OK I spent some time looking at this. Feedback:

Protocol: Rather than a boolean this should be a bitmask with something like usage hints. For now only a single usage hint will be supported "transient". The protocol should not specify that i means "no disk caching" rather just that it is a hint from the client to the terminal of intended use case. The protocol should just say that the terminal is free to assume that images with this hint will be used for only a short time and therefore evict its data before other images when the image is soft deleted and has no visible placements and the terminal is under storage pressure  and perform optimisations like not cache its data to disk. The protocol should also mention that if an animation frame with the transient flag is composited onto another frame if any of the involved frames are transient the resulting composited frame becomes transient.

Implementation: Mostly OK as far as I can see, though it has been a while since I looked at this part of the code.

Mekann2904:
Thanks for the feedback. I updated the patch to make `N` a usage-hints bitmask rather than a boolean.

For now, the only defined hint is `transient` (`1 << 0`, value `1`). The protocol text describes it as an intended-use hint from the client to the terminal, rather than specifying it as “no disk caching”. The terminal is free to use this hint for optimizations such as evicting the image data earlier under storage pressure, or skipping disk cache writes.

I also updated the implementation so the transient hint is propagated through frame coalescing/composition: if any contributing frame is transient, the resulting composed frame is transient as well.

Tested with:

```sh
./test.py --module parser
./test.py --module graphics
```

PR #10116, opened 2026-06-07
feat: add topmost functionality for MacOS

agmitron:
_Disclaimer: the code is generated by AI. This is more a proof of concept than a production-ready solution. Discussion is needed._

Adds a new `--use-physical-screen-frame` option for MacOS.

It allows to render kitty panels topmost, like many menu bars (`waybar`, `sketchybar`, etc). 

### Use cases

I need this feature because I want to replace the Menu Bar. There are tools like [SketchyBar](https://github.com/FelixKratz/SketchyBar), but I want to use kitty panels because: 

1. I love kitty very much and want to use it more.

2. I already use Kitty Quick Access Terminal windows as a replacement for app launchers, it's very convenient that I can reuse tools like `rg`, `fzf` for everything; would be great to do the same about the menu bar.

3. I have to use both Linux and MacOS (can't choose only one platform) and want this experience to be as portable as possible, so I don't need to use different tooling for the same things (two configs for two menu bars for Linux and MacOS).  

So my plan is to reuse this bar on MacOS and Linux instead of using different bars. On Linux (Wayland), the `topmost` behavior seems to be already implemented.

#### Before

`kitty +kitten panel --edge top ...`

image

#### After

`kitty +kitten panel --edge top --use-physical-screen-frame ...`

image

Just showing that it overlays the menu bar like [SketchyBar](https://github.com/FelixKratz/SketchyBar) does. To look better, I'm going to make it taller then.

### To be discussed

- [x] Does somebody, except for me, need this? Is it worth contributing this? 
- [x] Probably, it has already been implemented and I didn't notice? 
- [x] How properly is it implemented?


kovidgoyal:
Seems easier to just hide the global menubar using macos settings and rather than using physical frame which will likely break with multiscreen setups add "highest" option to --layer that on linux is a synonym for overlay but on macOS sets the level to the highest possible.

agmitron:
> Seems easier to just hide the global menubar using macos settings and rather than using physical frame

Below is the screenshot taken without this new `--use-physical-screen-frame` option.

image

I tried this and figured out that the apps should support rendering outside of "safe zone"*, so global MacOS settings are not enough — this is why I opened this MR. 

Tools like [SketchyBar](https://github.com/FelixKratz/SketchyBar) seem to ignore the "safe zone"* in their code.

*`safe zone height = screen height - menu bar height`. 

> "highest" option to --layer that on linux is a synonym for overlay but on macOS sets the level to the highest possible

Do you mean it's already implemented? Could you please give an example of a command? 

Does "highest" possible mean "the highest by z-axis" or "the highest be y-axis"?  I mean the `y` axis.

I tried this: 

```
kitty +kitten panel \
  --edge=top \
  --layer=overlay \
  --lines=10 \
  zsh
```
  
Got: 

image

I have no idea how this can be moved upper (by `y` axis) without "breaking" the safe zone :( 

My `Automatically hide and show the menu bar` param is set to `Always`: 

image

kovidgoyal:
Ah I assumed hiding the global menubar means apps are free to use the area of the erstwhile menubar. Are you saying that's not the case? If so then yes you will need a dedicated setting for it. You also need to consider the case of menubar at other edges than the top. And --layer controls z-index which you need to set pretty high to use as a menubar replacement otherwise you will get overlaps hiding the menubar etc.

agmitron:
> Ah I assumed hiding the global menubar means apps are free to use the area of the erstwhile menubar. Are you saying that's not the case?

Yes, exactly. Unfortunately this is a MacOS quirk, and the apps don't automatically stretch to fill the whole screen height when the menu bar is hidden (or, better said, the "safe zone" is not adjusted when the menu bar hides).

> You also need to consider the case of menubar at other edges than the top

Sure, will test it. 

> And --layer controls z-index which you need to set pretty high to use as a menubar replacement otherwise you will get overlaps hiding the menubar etc.

Sure, will also try to add and test it.

agmitron:
Hi, I've got some updates: 

## New layer: `shell`

> And --layer controls z-index which you need to set pretty high to use as a menubar replacement otherwise you will get overlaps hiding the menubar etc.

I have added `--layer=shell` option (naming is discussable). On MacOS, it's under the native Menu Bar by z-axis, so if you hover your cursor at the top edge, you will still see the native MacOS menu bar: 

shell

The difference with `--layer=overlay` is that `--overlay` is above Menu Bar by `z`-axis. 

`--layer=shell` is also below Mission Control:

image


... while `--layer=overlay` will be rendered on top of it:

image


So, to use Kitty as a Menu bar replacement on MacOS, you are going to need two options: 

1. `--use-physical-screen-frame` — to place `kitty` on top of the screen by `y`-axis

2. `--layer=shell` — "almost" on top by `z`-axis like `--layer=overlay` but still allows you to reach your native MacOS Menu Bar (you usually want this)

My example of Launch Agent on MacOS (to run at a startup):

```nix
  launchd.user.agents.kitty-menubar.serviceConfig = {
    ProgramArguments = [
      localKitty
      "+kitten"
      "panel"
      "--edge"
      "top"
      "--use-physical-screen-frame"
      "--margin-left"
      "0"
      "--margin-top"
      "0"
      "--columns"
      "400"
      "--layer"
      "shell"
      "--lines"
      "26px"
      "--focus-policy"
      "not-allowed"
      "--config"
      "/Users/${user}/.config/kitty/menubar.conf"
      "/Users/${user}/.config/kitty/scripts/menubar.sh"
    ];
    RunAtLoad = true;
    StandardOutPath = "/tmp/kitty-menubar-launchd.log";
    StandardErrorPath = "/tmp/kitty-menubar-launchd.err";
  };
```

agmitron:
Also tested this:

> You also need to consider the case of menubar at other edges than the top

My verdict: makes most sense on `top`, makes some sense at `right` and `left`, doesn't make any sense on MacOS if used at `bottom`, due to window management limitations. 

## bottom

In short: not usable on MacOS anyway, and the same result could be achieved without my changes just by using `--layer=overlay`.

1. `--use-physical-screen-frame` doesn't make any noticeable effect.

2. `--layer=shell`: hides `kitty` below the window by `z`-axis. Doesn't make any sense for the menu bar at the bottom. 

For me personally, this is not even usable on MacOS, because I still have some empty space left at the top, and I don't need two menu bars: 

image

However, as I mentioned before, `--layer=overlay` will stick Kitty even above Mission Control: 

image

I don't have clear ideas how to make this really usable for bottom-edge menu bars on MacOS specifically. 

It should either "reserve" some space at the bottom, like the top Menu Bar does (to prevent overlaps with other windows), or be hidden/shown automatically (like MacOS Dock). This seems to be out of the scope anyway. 

## left / right

`--use-physical-screen-frame` makes sense here to fill the full height. 

Might be usable if you: 

a) setup toggle (show/hide) for it to not to overlap other windows

b) are okay if it's always on top

c) manage windows width in a way that it doesn't overlap with such a menu bar.

image

image

kovidgoyal:
What does layer = shell do on Linux? I think we need to add another macOS specific setting call it ns_window_layer defaulting to unset. It should accept NSWIndowLayer constant names and also simple arithmetic expressions and integer values that way users can choose exactly which layer they want the panel to be on. See #10172 for a use case for this.

agmitron:
I agree in overall — this is too specific for MacOS. Makes zero sense on Linux, no need to reuse this API.

> What does layer = shell do on Linux?

Wayland: falls back to `layer=overlay`

X11: also falls back to `layer=overlay`

> I think we need to add another macOS specific setting call it `ns_window_layer`

Do you mean a config setting, not a `kitty +kitten panel` flag (like in the current implementation)?

I think it should be defined in the config with the `macos_` prefix: 

```conf
# proposed options
macos_ns_window_layer unset # or integers
macos_use_physical_screen_frame yes # not a CLI flag of kitty panel kitten, can be used for any Kitty instance

# ...
macos_option_as_alt both
macos_traditional_fullscreen yes
macos_fullscreen_ignore_safe_area_insets yes
# ...
```

> See https://github.com/kovidgoyal/kitty/issues/10172 for a use case for this.

Makes more sense that this `NSWIndowLayer` should be configured in `kitty.conf`, not as a flag for `kitty +kitten panel` (current PR)

kovidgoyal:
On Sun, Jun 21, 2026 at 02:20:01AM -0700, Aleksei Gmitron wrote:
> agmitron left a comment (kovidgoyal/kitty#10116)
>
> I agree in overall — this is too specific for MacOS.
>
> > What does layer = shell do on Linux?
>
> Wayland: falls back to `layer=overlay`
>
> X11: also falls back to `layer=overlay`
>
> > I think we need to add another macOS specific setting call it ns_window_layer
>
> Do you mean a config option, not a flag (like in the current implementation)? Or both?

Both should be config options. That way they can be used from the
command line via --override as well as in the config file for things
like the quick access terminal.

glfw/cocoa_window.m
```diff
@@ -2577,6 +2577,71 @@ void _glfwPlatformDestroyWindow(_GLFWwindow* window)
...
+        char *end = NULL;
+        long v = strtol(s, &end, 0);
+        if (end == s || errno) return false;
+        *val = v; *spec = end;
+        return true;
+    }
+    char name[96]; size_t i = 0;
+    while ((isalnum(*s) || *s == '_') && i + 1 < sizeof(name)) name[i++] = *s++;
+    name[i] = '\0';
+    if (!i || !ns_window_level_constant(name, val)) return false;
```
agmitron, line 2612: Looks ugly. I am not familiar with Objective-C. Need feedback about such guards & checks.

agmitron:
@KeqiZeng hi, if you can build my branch from the source, could you please test your use-case? 

In your config, set the new option `macos_ns_window_layer` to one of NS constants you mentioned [here](https://github.com/kovidgoyal/kitty/issues/10172#issuecomment-4757151103).

Supported values: 

1. MacOS constants

```
C(NSNormalWindowLevel); C(NSFloatingWindowLevel); C(NSSubmenuWindowLevel); C(NSTornOffMenuWindowLevel);
    C(NSMainMenuWindowLevel); C(NSStatusWindowLevel); C(NSModalPanelWindowLevel); C(NSPopUpMenuWindowLevel);
    C(NSScreenSaverWindowLevel);
    C(kCGBaseWindowLevel); C(kCGMinimumWindowLevel); C(kCGDesktopWindowLevel); C(kCGBackstopMenuLevel);
    C(kCGNormalWindowLevel); C(kCGFloatingWindowLevel); C(kCGTornOffMenuWindowLevel); C(kCGDockWindowLevel);
    C(kCGMainMenuWindowLevel); C(kCGStatusWindowLevel); C(kCGModalPanelWindowLevel); C(kCGPopUpMenuWindowLevel);
    C(kCGDraggingWindowLevel); C(kCGScreenSaverWindowLevel); C(kCGMaximumWindowLevel); C(kCGOverlayWindowLevel);
    C(kCGHelpWindowLevel); C(kCGUtilityWindowLevel); C(kCGDesktopIconWindowLevel); C(kCGAssistiveTechHighWindowLevel);
    C(kCGCursorWindowLevel); C(kCGNumberOfWindowLevelKeys);
```

2. Integers (both negative & positive & 0) 

3. Simple expressions with Apple constants (e.g. `kCGNormalWindowLevel - 1`).

glfw/cocoa_window.m
KeqiZeng: Hi @agmitron, thanks for your implementation! I tested this PR on my machine, it works fine for me. Here is a video.

https://github.com/user-attachments/assets/87d2b947-9f2b-4db0-8fc5-920ed7a85aa2

PR #10128, opened 2026-06-10
fix(choose-fonts): guard preview rendering against partial terminal color query

noctilust:
From: Forrest <noctilust@users.noreply.github.com>
Date: Wed, 10 Jun 2026
Subject: [PATCH] fix(choose-fonts): guard preview rendering against partial terminal color query

---
Disclaimer: This analysis and code fix was produced by an AI coding agent
(Hermes Agent, Nous Research) based on the reporter's bug investigation.
The author reviewed and verified the root cause and the fix before submission.
---

## Cause Analysis

The choose-fonts kitten queries the terminal for the user's foreground and
background colors at startup by sending a DCS request:

    h.lp.QueryTerminal("font_size", "dpi_x", "dpi_y", "foreground", "background")
    // in kittens/choose_fonts/ui.go, initialize(), line 81

The terminal responds with separate DCS response strings — one per field —
which arrive asynchronously and are dispatched through on_query_response:

    func (h *handler) on_query_response(key, val string, valid bool) error {
        // ... parse val into the corresponding text_style field
        switch key {
        case "foreground":
            h.text_style.Foreground = val
        case "background":
            h.text_style.Background = val
            return h.draw_screen()   // <-- bug: unconditional redraw
        }
        return nil
    }

The race condition: there is no guaranteed ordering between the DCS responses.
The foreground response may arrive before, after, or interleaved with the
background response — kitty's terminal stack processes them as it receives
the serialised data from the PTY, and the data may be split across read
boundaries.

If the background response arrives before the foreground response, the
sequence is:

1. on_query_response("background", "#1a1b26", true) is called
2. h.text_style.Background is set to "#1a1b26"
3. h.draw_screen() is called immediately
4. h.text_style.Foreground is still "" (Go zero-value string)

The call chain from draw_screen() reaches the font list preview:

    draw_screen() → h.current_pane.draw_screen() → FontList.draw_screen()
    → FontList.draw_family_summary() → checks Background != "" → true
    → FontList.draw_preview() → sends "render_family_samples" to backend
      with text_style = { foreground: "", background: "#1a1b26", ... }

The Python backend's opts_from_cmd() then calls:

    opts.foreground = to_color(ts['foreground'])   # to_color("")
    ValueError: Invalid color name: ''

The guard in draw_family_summary only checked Background, not Foreground:

    if self.handler.text_style.Background != "" {    // insufficient
        return self.draw_preview(...)
    }

The faces pane has no guard at all — it unconditionally launches a goroutine
that sends render_family_samples:

    if !found {
        self.preview_cache[key] = make(map[string]RenderedSampleTransmit)
        go func() {
            // ...
            kitty_font_backend.query("render_family_samples", map[string]any{
                "text_style": self.handler.text_style,  // may have empty fields
                ...
            })
        }()
        return
    }

## Reproduction

Confirmed by the reporter (noctilust) on kitty 0.47.2 / macOS:

- Running `kitten choose-fonts` crashed immediately with:
  `ValueError: Invalid color name: ''`

- Setting explicit foreground and background in kitty.conf (which populates
  opts through load_config() instead of the DCS query path) eliminated the
  crash, confirming the race-condition theory.

- Removing the workaround reproduced the crash again.

The race is timing-sensitive — it depends on how the PTY delivers the DCS
response data. It reproduces reliably in some environments and sporadically
in others.

## Fix (3 files in kittens/choose_fonts/)

1. **ui.go** — Only trigger draw_screen() from the foreground or background
   handler when the counterpart field is already populated:

   ```
   case "foreground":
       h.text_style.Foreground = val
       if h.text_style.Background != "" {
           return h.draw_screen()
       }
   case "background":
       h.text_style.Background = val
       if h.text_style.Foreground != "" {
           return h.draw_screen()
       }
   ```

   This ensures the UI never renders with a partial TextStyle. Whichever
   response arrives second will find the counterpart already populated and
   trigger the draw.

2. **list.go** — Require both Foreground and Background to be non-empty
   before rendering the font preview:

   ```
   -if self.handler.text_style.Background != "" {
   +if self.handler.text_style.Foreground != "" && self.handler.text_style.Background != "" {
   ```

   This is a belt-and-suspenders guard: even if draw_screen() is called
   through another path (on_wakeup, on_resize) before the DCS responses
   arrive, the preview won't send incomplete colors to the backend.

3. **faces.go** — Skip the render_family_samples goroutine if either color
   field is still empty:

   ```
   if !found {
   +    if self.handler.text_style.Foreground == "" || self.handler.text_style.Background == "" {
   +        return
   +    }
       self.preview_cache[key] = make(map[string]RenderedSampleTransmit)
       go func() { ... }()
       return
   }
   ```

   The preview cache remains empty, so the next redraw (triggered when the
   counterpart response arrives or by on_wakeup) will retry with full colors.

Note: font_size, dpi_x, and dpi_y are populated via the same DCS queries,
but their Go zero values (0.0) don't cause crashes — only the empty string
color values do.

## Impact

All choose-fonts users are susceptible to this crash. The trigger is the
ordering of DCS query responses from the terminal — a race that depends on
PTY buffering, terminal timing, and the kernel scheduler. Some users see it
reliably, others never, but the code is incorrect regardless.

Signed-off-by: Forrest <noctilust@users.noreply.github.com>

diff --git a/kittens/choose_fonts/ui.go b/kittens/choose_fonts/ui.go
index ... 100644
--- a/kittens/choose_fonts/ui.go
+++ b/kittens/choose_fonts/ui.go
@@ -138,8 +138,14 @@ func (h *handler) on_query_response(key, val string, valid bool) error {
 		}
 	case "foreground":
 		h.text_style.Foreground = val
+		if h.text_style.Background != "" {
+			return h.draw_screen()
+		}
 	case "background":
 		h.text_style.Background = val
+		if h.text_style.Foreground != "" {
+			return h.draw_screen()
+		}
 	}
 	return nil
 }
diff --git a/kittens/choose_fonts/list.go b/kittens/choose_fonts/list.go
index ... 100644
--- a/kittens/choose_fonts/list.go
+++ b/kittens/choose_fonts/list.go
@@ -130,7 +130,7 @@ func (self *FontList) draw_family_summary(start_x int, sz loop.ScreenSize) (err
 		lp.QueueWriteString(line)
 		y++
 	}
-	if self.handler.text_style.Background != "" {
+	if self.handler.text_style.Foreground != "" && self.handler.text_style.Background != "" {
 		return self.draw_preview(start_x, y, sz)
 	}
 	return
diff --git a/kittens/choose_fonts/faces.go b/kittens/choose_fonts/faces.go
index ... 100644
--- a/kittens/choose_fonts/faces.go
+++ b/kittens/choose_fonts/faces.go
@@ -55,6 +55,9 @@ func (self *faces) draw_screen() (err error) {
 	previews, found := self.preview_cache[key]
 	if !found {
 		self.preview_cache[key] = make(map[string]RenderedSampleTransmit)
+		if self.handler.text_style.Foreground == "" || self.handler.text_style.Background == "" {
+			return
+		}
 		go func() {
 			var r map[string]RenderedSampleTransmit
 			s := key.settings

### DIFF
--- a/kittens/choose_fonts/faces.go
+++ b/kittens/choose_fonts/faces.go
@@ -52,6 +52,9 @@ func (self *faces) draw_screen() (err error) {
 	previews, found := self.preview_cache[key]
 	if !found {
 		self.preview_cache[key] = make(map[string]RenderedSampleTransmit)
+		if self.handler.text_style.Foreground == "" || self.handler.text_style.Background == "" {
+			return
+		}
 		go func() {
 			var r map[string]RenderedSampleTransmit
 			s := key.settings
--- a/kittens/choose_fonts/list.go
+++ b/kittens/choose_fonts/list.go
@@ -122,7 +122,7 @@ func (self *FontList) draw_family_summary(start_x int, sz loop.ScreenSize) (err 
 		lp.QueueWriteString(line)
 		y++
 	}
-	if self.handler.text_style.Background != "" {
+	if self.handler.text_style.Foreground != "" && self.handler.text_style.Background != "" {
 		return self.draw_preview(start_x, y, sz)
 	}
 	return
--- a/kittens/choose_fonts/ui.go
+++ b/kittens/choose_fonts/ui.go
@@ -139,9 +139,14 @@ func (h *handler) on_query_response(key, val string, valid bool) error {
 		}
 	case "foreground":
 		h.text_style.Foreground = val
+		if h.text_style.Background != "" {
+			return h.draw_screen()
+		}
 	case "background":
 		h.text_style.Background = val
-		return h.draw_screen()
+		if h.text_style.Foreground != "" {
+			return h.draw_screen()
+		}
 	}
 	return nil
 }


noctilust:
Full exception traceback
```
➜ kitten choose-fonts
  File "lib/python3.14/runpy.py", line 203, in _run_module_as_main
  File "lib/python3.14/runpy.py", line 88, in _run_code
  File "lib/python3.14/kitty_main.py", line 7, in <module>
  File "lib/python3.14/kitty/entry_points.py", line 142, in main
  File "lib/python3.14/kitty/entry_points.py", line 100, in namespaced
  File "lib/python3.14/kitty/entry_points.py", line 23, in runpy
  File "<string>", line 1, in <module>
  File "lib/python3.14/kittens/choose_fonts/backend.py", line 219, in main
  File "lib/python3.14/kittens/choose_fonts/backend.py", line 77, in opts_from_cmd
  File "lib/python3.14/kitty/conf/utils.py", line 56, in to_color
  File "lib/python3.14/kitty/rgb.py", line 37, in to_color
ValueError: Invalid color name: ''
Error: exit status 1
```

Issue #10135, opened 2026-06-11
Mouse input permanently stops working in all windows after an aborted tab drag on Wayland


ttys3:
**Describe the bug**

On Linux under Wayland, clicking a tab in the tab bar and accidentally flicking the pointer upward (a quick press → tiny drag → release gesture) can permanently break mouse handling in kitty: from that moment on, no window in any tab or OS window responds to mouse clicks or text selection. Keyboard input keeps working, and switching tabs via the tab bar still partially works, but everything inside the windows is dead to the mouse. The only recovery is restarting kitty.

The same gesture can also unexpectedly detach the tab into a new OS window ("the tab flies away"). The detach itself is the intended drag-tab-to-detach feature introduced in v0.46.0 (and `drag_threshold` is only 5px, so it is easy to trigger accidentally), but when the gesture is fast enough the drag aborts half way and leaves kitty's internal drag state permanently stuck, which is the actual bug.

**To Reproduce**

There are two ways to trigger it. Method A is deterministic; method B is the
race as it happens during normal use (confirmed reproducible).

Method A — deterministic, via the state leak on release outside a tab:

1. `kitty --config NONE --debug-input -o drag_threshold=20` (the larger threshold just widens the press-to-release window from ~5px to ~20px to make the steps easy; the leak itself also exists with the default of 5px when the press lands within 5px of a tab edge)
2. Open a second tab with `Ctrl+Shift+T`
3. Press the left mouse button on a tab title, keep it pressed, move right by ~10-15px so the pointer crosses the tab boundary (staying under 20px), and release on the empty tab-bar area / separator next to the tab
4. Now just move the mouse anywhere by more than 20px (no button pressed)
5. Mouse input is now dead in every kitty window. With `--debug-input` every subsequent click logs `handled by tab bar` instead of being dispatched to the window. See error.

Method B — the race during normal use (a fast "click and flick" on a tab):

1. Run kitty under Wayland with at least two tabs (default config is enough)
2. To widen the race window, create some render load first, e.g. run `yes` in one tab so the screen is scrolling
3. On another tab's title, press the left button, flick the pointer upward past `drag_threshold` (5px) and release, all in one fast motion (well under ~100 ms) — the over-threshold motion and the release must arrive in the same event batch, before the next frame renders
4. The drag never visibly starts, and from then on mouse clicks and text selection stop working in all kitty windows. May take a few attempts since it is timing-dependent; the render load from step 2 makes it land much more reliably.

[...]

PR #10136, opened 2026-06-11
fix(tabs): mouse handling stuck after aborted tab drag on Wayland

ttys3:
resolve issue https://github.com/kovidgoyal/kitty/issues/10135

A quick click-and-flick on a tab could leave all of kitty with mouse input permanently redirected to the tab bar, making every window unclickable and text selection impossible.

Starting a tab drag is asynchronous: the drag thumbnail is rendered on the next frame before glfwStartDrag is called. If the button is released in that window, wl_data_device_start_drag is sent with a stale serial that no longer matches an active pointer implicit grab, so the compositor silently ignores it. The wl_data_source then never receives any event, on_drag_source_finished never runs, and the tab_being_dragged state is stuck forever, hijacking all mouse events.

Fix in layers:
- glfw/Wayland: track the implicit grab (serial of the first button press and pressed-button count), use that serial for start_drag and refuse with EAGAIN when there is no active implicit grab instead of letting the compositor silently drop the request
- mouse.c: a left button release arriving while a tab drag is marked started but no system DND is active means the drag never launched (an active DND consumes the release on all platforms), so clear the drag state instead of waiting for DND events that will never come
- tabs.py: handle OSError from start_drag_with_data for tab drags the same way window drags already do; clear the potential-drag state when the release lands on the new-tab button or empty tab bar area
- tabs.py/boss.py: clear drag state on drag finish/drop even when the dragged tab has already been closed

### DIFF
--- a/glfw/wl_init.c
+++ b/glfw/wl_init.c
@@ -102,6 +102,11 @@ pointerHandleEnter(
 
 static void
 pointerHandleLeave(void* data UNUSED, struct wl_pointer* pointer UNUSED, uint32_t serial, struct wl_surface* surface) {
+    // The pointer never leaves the surface during an implicit grab, so a
+    // leave event means any implicit grab is over (e.g. the compositor took
+    // over the pointer for drag-and-drop). The matching button releases will
+    // never be delivered to us.
+    _glfw.wl.pointer_button_count = 0;
     _GLFWwindow* window = _glfw.wl.pointerFocus;
     if (!window) return;
     _glfw.wl.serial = serial;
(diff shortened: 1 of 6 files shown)


ttys3:
@kovidgoyal Thanks for merging the PR so quickly.

I installed the build containing this PR and tested it for a day; I found that the likelihood of the issue occurring has dropped significantly, though I did accidentally trigger it once today (I use Kitty for essentially eight hours during work).
This suggests there are still aspects the previous fix didn't fully account for. I will try to come up with reproduction steps and then submit another PR.

kovidgoyal:
Note that this breaks drag and drop on niri, #10271

PR #10138, opened 2026-06-12
fix(wayland): detect start_drag silently ignored by the compositor

ttys3:
The pointer_button_count check added in dc36e2165 uses the client side view of the implicit grab, which is stale when the button release is still in flight: glfwStartDrag passes the check, but by the time the compositor processes wl_data_device.start_drag the implicit grab is gone and the request is silently dropped. The data source then never receives any event, so the tab drag state leaks forever, hijacking mouse handling again, and the already created xdg_toplevel_drag toplevel gets mapped as a stray undecorated window showing the drag thumbnail that nothing ever destroys.

Detect this deterministically using protocol ordering: issue a wl_display.sync right after start_drag. An accepted start_drag synchronously produces events (wl_pointer.leave from the DND grab taking over, wl_data_device.enter, wl_data_source events) that are ordered before the sync callback. If the callback fires with none of them seen, the request was dropped: cancel the drag, which notifies the application (clearing the tab drag state) and destroys the drag toplevel and data source.

Also defer mapping the drag toplevel until the session is confirmed, so the stray window can never appear, not even for one frame.

Verified against headless GNOME Shell 50.2 (mutter) with injected pointer timing scans: the unpatched build deadlocks with an orphaned drag toplevel within ~13 fast flicks, the patched build catches every race hit and cancels cleanly, with no deadlock, no stray window, and normal slow drag/reorder/detach behaviour preserved.

### DIFF
--- a/glfw/wl_init.c
+++ b/glfw/wl_init.c
@@ -107,6 +107,9 @@ pointerHandleLeave(void* data UNUSED, struct wl_pointer* pointer UNUSED, uint32_
     // over the pointer for drag-and-drop). The matching button releases will
     // never be delivered to us.
     _glfw.wl.pointer_button_count = 0;
+    // A DND grab taking over the pointer sends a leave, making this the
+    // earliest proof that a just requested start_drag was accepted.
+    _glfwWaylandConfirmDragSession();
     _GLFWwindow* window = _glfw.wl.pointerFocus;
     if (!window) return;
     _glfw.wl.serial = serial;
--- a/glfw/wl_platform.h
+++ b/glfw/wl_platform.h
@@ -434,6 +434,20 @@ typedef struct _GLFWlibraryWayland
         struct xdg_surface *toplevel_xdg_surface;
         struct xdg_toplevel *toplevel_xdg_toplevel;
         struct wl_buffer *toplevel_buffer;
+        // wl_data_device.start_drag is silently ignored by compositors when
+        // its serial does not match an active pointer implicit grab, which
+        // can happen as drags are started asynchronously and the client side
+        // view of the grab can be stale. A wl_display.sync issued right after
+        // start_drag detects this: any compositor event proving the DND
+        // session is live (wl_pointer.leave, wl_data_device.enter, any
+        // wl_data_source event) is ordered before the sync callback, so if
+        // the callback fires first the start_drag was dropped.
+        struct wl_callback *start_confirmation;
+        bool session_confirmed;
+        // The drag toplevel was configured before the session was confirmed,
+        // mapping it was deferred so it cannot end up as a stray regular
+        // window if start_drag was silently ignored.
+        bool toplevel_map_deferred;
         struct {
             const char *mime_type;
             int fd;
@@ -487,6 +501,7 @@ void animateCursorImage(id_type timer_id, void *data);
 struct wl_cursor* _glfwLoadCursor(GLFWCursorShape, struct wl_cursor_theme*);
 void destroy_data_offer(_GLFWWaylandDataOffer*);
 const char* _glfwWaylandCompositorName(void);
+void _glfwWaylandConfirmDragSession(void);
 
 typedef struct wayland_cursor_shape {
     int which; const char *name;
--- a/glfw/wl_window.c
+++ b/glfw/wl_window.c
@@ -2561,6 +2561,8 @@ update_drop_source_actions(_GLFWwindow *window, _GLFWWaylandDataOffer *offer) {
 static void
 drag_enter(void *data UNUSED, struct wl_data_device *wl_data_device UNUSED, uint32_t serial, struct wl_surface *surface, wl_fixed_t x, wl_fixed_t y, struct wl_data_offer *id) {
     debug_input("Drop entered\n");
+    // an enter for a drag we started means the DND session is live
+    _glfwWaylandConfirmDragSession();
     _GLFWWaylandDataOffer *offer = &_glfw.wl.drop_data_offer;
     mark_data_offer(offer, id);
     if (!offer->id) return;
@@ -3164,9 +3166,7 @@ GLFWAPI bool glfwWaylandBeep(GLFWwindow *handle) {
 // Drag source {{{
 
 static void
-drag_toplevel_xdg_surface_configure(void *data UNUSED, struct xdg_surface *surface, uint32_t serial) {
-    debug_input("Drag toplevel surface configured\n");
-    xdg_surface_ack_configure(surface, serial);
+map_drag_toplevel(void) {
     struct wl_buffer *buf = _glfw.wl.drag.toplevel_buffer;
     if (buf) {
         wl_surface_attach(_glfw.wl.drag.drag_icon, buf, 0, 0);
@@ -3176,6 +3176,23 @@ drag_toplevel_xdg_surface_configure(void *data UNUSED, struct xdg_surface *surfa
     }
     if (_glfw.wl.drag.drag_icon) wl_surface_commit(_glfw.wl.drag.drag_icon);
     if (buf) wl_buffer_destroy(buf);
+}
+
+static void
+drag_toplevel_xdg_surface_configure(void *data UNUSED, struct xdg_surface *surface, uint32_t serial) {
+    debug_input("Drag toplevel surface configured\n");
+    xdg_surface_ack_configure(surface, serial);
+    if (_glfw.wl.drag.start_confirmation && !_glfw.wl.drag.session_confirmed) {
+        // The compositor has not yet proven that it accepted start_drag. If
+        // it silently ignored it (stale implicit grab serial), committing a
+        // buffer now would map the toplevel as a stray regular window that
+        // nothing ever destroys. Defer mapping until the session is
+        // confirmed; if it never is, the toplevel is destroyed unmapped.
+        debug_input("Deferring drag toplevel map until drag session is confirmed\n");
+        _glfw.wl.drag.toplevel_map_deferred = true;
+        return;
+    }
+    map_drag_toplevel();
 }
 
 static const struct xdg_surface_listener drag_toplevel_xdg_surface_listener = {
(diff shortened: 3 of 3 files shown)


ttys3:
Follow-up to #10136 (issue #10135).

## Summary

#10136 made the aborted-tab-drag deadlock much rarer, but as mentioned in [my comment there](https://github.com/kovidgoyal/kitty/pull/10136), it still triggered once during a day of normal use. The remaining hole is a time-of-check/time-of-use race that **no client-side precondition check can close** — only the compositor knows whether the implicit grab is still alive when it processes `start_drag`. This PR detects the failure deterministically at the protocol level and recovers cleanly. Reproduction steps below, as promised.

When the race hits, there are two user-visible symptoms:

- mouse handling wedges exactly as in #10135: `tab_being_dragged` leaks forever and every mouse event in every kitty window is redirected to the tab bar handler, making all windows unclickable until kitty is restarted
- on compositors supporting `xdg_toplevel_drag` (mutter, kwin), the toplevel created to follow the pointer during the drag gets mapped as a stray undecorated zombie window showing the rendered tab title, which nothing ever destroys — and if it ends up under the pointer, kitty stops receiving pointer events entirely

## Why the check from #10136 is not enough

#10136 tracks the pointer implicit grab client side (pressed-button count plus the grab serial) and makes `glfwStartDrag` fail with `EAGAIN` when the client already knows there is no grab. But the client's view is necessarily stale: the button release may already have happened on the compositor side — with the `wl_pointer.button` release event still sitting unread in the socket — at the moment `glfwStartDrag` runs (the drag thumbnail is rendered one frame after the threshold is crossed, so there is always a window). The check passes, `start_drag` is issued with the press serial, and the compositor silently ignores it: per the protocol there is no error event for this. The `wl_data_source` never receives `cancelled` / `dnd_finished` / `dnd_drop_performed`, so `Boss.on_drag_source_finished` never runs and the drag state is never cleared. Meanwhile the drag toplevel has already been created and gets mapped on its first `configure`, becoming the zombie window.

The full sequence:

```mermaid
sequenceDiagram
    participant I as Input device
    participant C as Compositor
    participant E as kitty event loop
    participant F as kitty frame callback

    I->>C: button press on tab
    C->>E: wl_pointer.button press (serial S)
    Note over C: implicit grab S starts
    E->>E: dispatch press: grab serial = S, button count = 1
    I->>C: motion past drag_threshold
    C->>E: wl_pointer.motion
    E->>E: dispatch motion: tab drag marked started,<br/>thumbnail render queued for next frame
    Note over F: next frame
    F->>F: start_tab_drag() runs<br/>(main thread busy, socket not read)
    I->>C: button release
    Note over C: implicit grab S ends here
    C--)E: wl_pointer.button release<br/>(queued in the socket, not yet dispatched)
    F->>C: create xdg_toplevel + xdg_toplevel_drag for the drag image
    F->>F: client-side grab check from #10136:<br/>button count still 1 -> passes (STALE)
    F->>C: wl_data_device.start_drag(serial = S)
    Note over C: S no longer matches an active implicit grab<br/>-> request silently ignored,<br/>no error event exists in the protocol
    C->>E: xdg_surface.configure (drag toplevel)
    E->>E: attach buffer + commit -><br/>toplevel MAPPED as a regular window (zombie)
    E->>E: dispatch release: tab drag marked started and<br/>a data source object exists -> swallowed,<br/>"the DND session will clean up"
    Note over E,C: but the DND session never existed:<br/>no wl_data_source event will ever arrive,<br/>tab_being_dragged leaks forever (all mouse input hijacked),<br/>the zombie toplevel is never destroyed
```

(The recovery path from #10136 in `mouse.c` does not fire either: by the time the release is dispatched, the data source object already exists and looks active, so the release is attributed to the — nonexistent — DND session.)

## The fix

Detect the silent drop deterministically via protocol ordering: issue a `wl_display.sync` right after `start_drag`. An accepted `start_drag` synchronously produces compositor events — `wl_pointer.leave` from the DND grab taking over, `wl_data_device.enter`, `wl_data_source` events — that are guaranteed to be ordered before the sync callback. If the callback fires with none of them seen, the request was dropped: cancel the drag through the normal `GLFW_DRAG_CANCELLED` path (which clears the tab drag state in the application) and destroy the data source and drag toplevel.

Additionally, mapping of the `xdg_toplevel_drag` toplevel is deferred until the session is confirmed, so the zombie window can never appear, not even for one frame: if the session is never confirmed, the toplevel is destroyed unmapped.

```mermaid
sequenceDiagram
    participant C as Compositor
    participant K as kitty (glfw Wayland backend)

    K->>C: wl_data_device.start_drag(serial = S)
    K->>C: wl_display.sync -> callback CB
    Note over K: drag toplevel mapping deferred<br/>until the session is confirmed
    alt start_drag accepted by the compositor
        C->>K: wl_pointer.leave / wl_data_device.enter / wl_data_source.*
        K->>K: session confirmed -> map the drag toplevel
        C->>K: CB done (ordered after the events above)
        Note over K: normal drag proceeds
    else start_drag silently dropped (stale serial S)
        C->>K: CB done, with no DND event seen before it
        K->>K: cancel: GLFW_DRAG_CANCELLED to the app,<br/>destroy data source + drag toplevel (never mapped)
        Note over K: tab drag state cleared,<br/>no deadlock, no zombie window
    end
```

## How to reproduce (deterministic, 100%)

The race window between "the client checks its stale view of the implicit grab" and "the compositor processes `start_drag`" is normally a few milliseconds wide, so hitting it from real input is timing-dependent (with current master it takes on the order of a dozen fast flicks on a loaded session, see Testing). The window can be widened arbitrarily **without changing any logic** — the delay below simply simulates a busy main loop / slow frame at the worst possible moment, which is exactly what render load does in practice. Apply this helper to `kitty/tabs.py`, right before the `start_drag_with_data` call in `start_tab_drag`:

```diff
                 try:
+                    if (d := os.environ.get('KITTY_TAB_DRAG_DELAY', '')):
+                        import time
+                        time.sleep(float(d))  # widen the race window deterministically
                     start_drag_with_data(self.os_window_id, drag_data, thumbnails)
```

Then, on a Wayland session (mutter or kwin to also see the zombie window):

1. `KITTY_TAB_DRAG_DELAY=0.8 kitty --config NONE --debug-input`, open a second tab with `Ctrl+Shift+T`
2. Press the left button on a tab title and drag horizontally past `drag_threshold` (more than 5px), keeping the button held — the drag is now marked started, and on the next frame `start_tab_drag` enters the 0.8s delay, during which the event loop is blocked and the Wayland socket is not read
3. Release the button within those 0.8s (trivial to do by hand)

On current master (with #10136), **every single time**: the compositor ends the implicit grab on the release while the queued release event is never seen by the client before it issues `start_drag` with the now-stale press serial; the compositor silently ignores the request, no `wl_data_source` event ever arrives, and

- mouse handling wedges (every subsequent click logs `handled by tab bar` and no window responds to clicks or selection), and
- on mutter/kwin the drag toplevel appears as a stray zombie window showing the tab title — and if it sits under the pointer, kitty receives no pointer events at all.

With this PR, **every single time**: `--debug-input` logs `Wayland: start_drag was silently ignored by the compositor, cancelling drag`, the drag state is cleaned up, the toplevel is never mapped, and the mouse keeps working normally.

## Testing

- `python3 test.py --module dnd`: 143 tests pass (directly covers the drag-and-drop paths touched here); `--module mouse` / `--module glfw` also pass; the full suite shows only pre-existing environment-specific failures unrelated to this change
- deterministic verification using the reproducer above (pointer injection via the `org.gnome.Mutter.RemoteDesktop` API against a headless GNOME Shell, 3 isolated rounds per build, fresh kitty each round): without this PR 3/3 rounds break (drag toplevel mapped as a zombie window, zero data-source events ever, mouse broken — twice with the zombie window intercepting the pointer entirely, once with the classic tab-bar hijack); with it 3/3 rounds recover cleanly (silently-ignored `start_drag` detected and cancelled, toplevel never mapped, mouse fully functional)
- timing-scan verification on the same headless setup (no helper patch, scanning press→motion→release timings around the race window): a build with only #10136 deadlocks with an orphaned drag toplevel within ~13 flicks; with this PR every race hit is caught with no deadlock and no stray windows over hundreds of flicks, and deferred-then-confirmed toplevel mapping is observed on the drags that genuinely start
- manual verification on a live Wayland session: slow tab drags (reorder, drop on another OS window, detach) keep working; a fast flick now degrades gracefully into a click/no-op instead of a deadlock

kovidgoyal:
You are welcome! If we are anyway syncing with the compositor do we
still need client side tracking of drag state from your previous PR?

ttys3:
> You are welcome! If we are anyway syncing with the compositor do we still need client side tracking of drag state from your previous PR?

Good question.  in short:
the answer is Yes. prev PR should keep.

--------

  The sync replaces the *precondition check*, but not the *serial selection* — they're doing different jobs:

   **`pointer_grab_serial` — keep.** This isn't detection; it's issuing `start_drag` with the right serial in the first place. `pointer_serial` is overwritten on `enter` and on every button event (press *and* release), so it's only incidentally the implicit-grab serial. If a drag is started
  while a second button is held — or after any stray pointer event — `pointer_serial` is a non-grab serial, the compositor silently drops the request, and the sync then *cancels* a drag that should have worked. The sync recovers from failure; it can't turn a mis-issued request into a success.
  (For the plain single-button tab gesture the two serials coincide, so there it's belt-and-suspenders.) It's also latched via the button count and reset on `leave`, so the count tracking has to stay to keep the serial correct.

  **`pointer_button_count == 0 → EPERM` — strictly redundant now.** The sync is a superset: every silently-dropped `start_drag`, the no-grab case included, is caught, so this is no longer needed to avoid the deadlock. I'd still keep it as a synchronous fast-path — when the client already knows
  there is no grab (e.g. the button-less motion in method A) it bails before creating a data source + `xdg_toplevel_drag` and doing a full round-trip, and returns the error synchronously so `start_tab_drag` clears state without waiting for the async cancel. It's ~3 lines and the one piece you
  could drop without reintroducing the bug. (`pointerHandleLeave` also reuses the count reset as the earliest confirmation that the session went live, so it's wired into the sync path too.)

  The kitty-side bits stay regardless: the `tabs.py` early-return cleanup fixes the method-A leak at its source (a plain logic bug, not Wayland-specific), and the `mouse.c` release-recovery is complementary to the sync — it catches the release-before-`start_tab_drag`-runs sub-case, while the
  sync catches the issued-then-dropped one (by which point a data source already exists and looks active, so `mouse.c` can't).

Issue #10123, opened 2026-06-09
Prompt Glypy Rendering Issue


jemabaris:
**Describe the bug**
Out of nowhere the size of how the glyphs of my powerlevel10k prompt are rendered has changed (They are suddenly way smaller). See the attached screenshot; The small one is kitty and the normal one is Ghostty (where it's still rendered as I'm used to and how it's supposed to be). The fact that it's still okay in Ghostty lead me to the assumption that could be some kitty related issue.

EDIT: Just noticed that all glyphs (like for example in neovim) are affected and not just the ones in the prompt.

**Screenshots**

<img width="1923" height="402" alt="Image" src="https://github.com/user-attachments/assets/49cfe083-d2ec-49ea-80b6-89d434bdb350" />

**Environment details**
```
kitty 0.47.2 (c126e227d3) created by Kovid Goyal
Linux janis-arch 7.0.11-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 02 Jun 2026 18:26:58 +0000 x86_64
EndeavourOS 7.0.11-arch1-1 (/dev/tty)
DISTRIB_ID="EndeavourOS"
DISTRIB_RELEASE="rolling"
DISTRIB_DESCRIPTION="EndeavourOS Linux"
DISTRIB_CODENAME="rolling"
Running under: Wayland (Hyprland v0.55.3,) missing: blur icon key-repeat top_level_drag
OpenGL: '3.1.0 NVIDIA 610.43.02' Detected version: 3.1
Frozen: False
Fonts:
  medium: MonoLisa: /usr/share/fonts/MonoLisa/MonoLisa-normal.ttf:0
          Features: ()
    bold: MonoLisaNormal-SemiBold: /usr/share/fonts/MonoLisa/MonoLisa-normal.ttf:393216
          Features: ()
  italic: MonoLisa: /usr/share/fonts/MonoLisa/MonoLisa-italic.ttf:262144
          Features: ()
      bi: MonoLisaItalic-SemiBold: /usr/share/fonts/MonoLisa/MonoLisa-italic.ttf:393216
          Features: ()
Paths:
  kitty: /home/janis/Repos/kitty/kitty/launcher/kitty
  base dir: /home/janis/Repos/kitty
  extensions dir: /home/janis/Repos/kitty/kitty
  system shell: /bin/zsh
System color scheme: dark. Applied color theme type: none
Loaded config files:
  /home/janis/.config/kitty/kitty.conf

[...]

PR #10145, opened 2026-06-14
freetype: do not apply the fontconfig size-fixup matrix to color glyphs

### LINKED ISSUE #10144: 0.47.3: color/bitmap emoji rendered far too small (fontconfig size matrix double-applied)
After updating to 0.47.3, all color emoji render shrunk to a tiny dot in the top-left of their cell. It happens everywhere color emoji appear (shell prompt, neovim, plain `printf`), not just one app.

This is the same visual symptom as #10123 (closed with "use symbol_map"), but I think that workaround misses the actual cause, which is a fontconfig size matrix being applied twice. Details below.

### Repro

```sh
kitty --config NONE sh -c 'printf "\U0001F40D snake\n"; sleep 5'
```

The snake shows as a dot instead of a full-cell emoji.

### Bisect

0.47.2 renders it correctly, 0.47.3 does not. The only font/glyph change in `v0.47.2..v0.47.3` is [ee937bdd1b](https://github.com/kovidgoyal/kitty/commit/ee937bdd1b86a23df4bc1698f01128ea979b753c) ("freetype: route FC_MATRIX through cairo for color glyphs"). Reverting just that commit on top of 0.47.3 restores correct rendering.

### Root cause

Noto Color Emoji is a fixed-size CBDT bitmap font, so fontconfig hands every consumer a downscale matrix to fit it to the requested pixel size. On this system:

```
$ fc-match -v 'Noto Color Emoji' | grep matrix
	matrix: [0.114679 0; 0 0.114679](w)
```

kitty already does its own cell-fit scaling for color glyphs (`fit_cairo_glyph`). Before `ee937bdd1b` it ignored the fontconfig matrix for these. Now `apply_cairo_font_size` feeds that `~0.1147x` matrix into the cairo font matrix as well, so the already-cell-sized emoji gets scaled down by another ~9x. Font selection is correct, only the scale is wrong:

```
$ kitty --debug-font-fallback sh -c 'printf "\U0001F40D\n"'
U+1f40d emoji_presentation Face(family=Noto Color Emoji ... color=True)
```

So the matrix path in `apply_cairo_font_size` probably needs to be skipped for fixed-size / color-bitmap faces, where the size is already handled by `fit_cairo_glyph` rather than the font matrix.

### Environment

```
kitty 0.47.3
Linux 7.0.12-xanmod1 x86_64, Wayland (niri)
Noto Color Emoji 2.051 (CBDT bitmap)
freetype 2.14.2, fontconfig 2.17.1, cairo 1.18.4
```

Strykar:
Fixes the color-emoji-shrunk-to-a-dot regression reported in #10144.

## Problem

Since 0.47.3, color emoji render as a tiny dot in the corner of their cell, everywhere. `ee937bdd1b` started routing `FC_MATRIX` through the cairo font matrix for color glyphs so synthetic slant would reach them. But `FC_MATRIX` is also how fontconfig encodes the pixel-size fixup of fixed-size faces: Noto Color Emoji is a ~109px bitmap strike, matched with a matrix that scales it by `requested_px / strike_px`. `cairo_set_font_size()` already brings the strike to the requested size, so applying that matrix in `apply_cairo_font_size()` scales it down again by the fixup factor, and `fit_cairo_glyph()` only shrinks (never grows), so the emoji stays tiny. The shrink is up to ~9x at small terminal cells.

The original commit's premise that stock fontconfig does not apply `FC_MATRIX` to color fonts is wrong: it does, to the most common color font, for everyone.

#10123 is related. That report was mainly about nerd-font symbols rendering too small, and the reporter later added that emoji had also gone tiny. The emoji part is this bug (worked around there with `symbol_map`, which sidesteps it because those fonts resolve via `FcFontList`, carrying no fixup matrix), so this fix resolves it. The nerd-font symbols are monochrome and never reach the color path this fix touches, so they are unaffected; their cause is separate.

## Fix

Only the shear carries synthetic slant; the diagonal is the size, which `cairo_set_font_size()` + `fit_cairo_glyph()` already own. Apply only the shear (normalize the diagonal to unit scale) and fall through to `cairo_set_font_size()` when there is no shear.

The synthetic-slant case ee937bdd1b added still works: a pure-slant matrix is all shear and is applied unchanged. The one intentional behaviour change is that a non-uniform or rotational hand-built FC_MATRIX is now dropped on the color path (only the shear is carried); stock config never produces that, since the size fixup is a uniform scale.

## Testing

- New `test_color_emoji_not_shrunk` renders a color emoji through the real pipeline and checks it fills its cells. It fails on the unpatched build (~0.28 coverage) and passes on the patched build (~0.84), so it actually guards the regression. Skipped unless a fixed-size color font with a fontconfig fixup matrix is present.
- Verified across 13 fonts including MonoLisa (the font in #10123): emoji coverage goes from ~0.28 to ~0.84-1.00 with the fix, on every main font; text rendering is unchanged.
- Monochrome bitmap fonts are unaffected: fontconfig gives them no fixup matrix, and freetype ignores `FT_Set_Transform` on embedded bitmaps.

Fixes #10144

### DIFF
--- a/kitty/freetype.c
+++ b/kitty/freetype.c
@@ -878,18 +878,26 @@ apply_cairo_font_size(Face *self, unsigned sz_px) {
     // on self->face in face_from_descriptor. cairo owns FT_Set_Transform on
     // its face and derives it from the font matrix on every render
     // (_cairo_ft_unscaled_font_set_scale in cairo-ft-font.c), so the only
-    // channel that reaches glyph rasterization is the cairo font matrix
-    // itself. Encode FC_MATRIX there.
-    // FT_Matrix is xx,xy,yx,yy (row-major); cairo_matrix_init takes
-    // xx,yx,xy,yy. Same matrix, transposed argument order.
-    if (!self->has_matrix) { cairo_set_font_size(self->cairo.cr, sz_px); return; }
+    // channel that reaches glyph rasterization is the cairo font matrix.
+    //
+    // FC_MATRIX is overloaded. Besides synthetic slant, fontconfig also encodes
+    // the pixel size fixup of fixed-size faces here, as a pure diagonal scale
+    // (Noto Color Emoji is matched with [0.1147 0; 0 0.1147]). The color path
+    // already sizes glyphs with cairo_set_font_size() + fit_cairo_glyph(), and
+    // fit_cairo_glyph() only shrinks, so feeding the fixup scale in here shrinks
+    // color emoji with no way to recover (#10144). Honor only the shear that
+    // carries synthetic slant and leave the size to cairo_set_font_size().
+    if (!self->has_matrix || self->matrix.xx == 0 || self->matrix.yy == 0) {
+        cairo_set_font_size(self->cairo.cr, sz_px); return;
+    }
+    double shear_xy = (double)self->matrix.xy / (double)self->matrix.xx;
+    double shear_yx = (double)self->matrix.yx / (double)self->matrix.yy;
+    if (shear_xy == 0 && shear_yx == 0) { cairo_set_font_size(self->cairo.cr, sz_px); return; }
+    // FT_Matrix is xx,xy,yx,yy (row-major); cairo_matrix_init takes xx,yx,xy,yy.
+    // The diagonal is unit scale (size is handled above); apply only the shear.
     double s = (double)sz_px;
-    double xx = self->matrix.xx / 65536.0;
-    double xy = self->matrix.xy / 65536.0;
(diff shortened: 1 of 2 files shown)


kovidgoyal:
@Strykar your color emoji test is failing in CI though not locally, Maybe make it a bit more robust.

Strykar:
@kovidgoyal You pushed your fix just before me, please review https://github.com/kovidgoyal/kitty/pull/10147 it is a robust test for this. I am waiting on the CI to go green.

PR #10147, opened 2026-06-14
fonts: make the color emoji regression test environment-independent

Strykar:
The test added in fbd4e5ad1 asserted an absolute cell-coverage threshold, which depends on the cell aspect ratio and the installed Noto Color Emoji build: a correctly sized emoji covers ~0.84 of its cell on one box but ~0.39 in CI, so the 0.5 threshold failed in CI.

Render the same font two ways at one size instead: via its fontconfig descriptor (which carries the size-fixup matrix) and via its file path (which does not). They must come out the same size; the bug shrinks the descriptor one. Only the matrix differs, so the check no longer depends on the environment or the emoji artwork.

A more robust test for the regression.

### DIFF
--- a/kitty_tests/fonts.py
+++ b/kitty_tests/fonts.py
@@ -392,28 +392,27 @@ class Rendering(FontBaseTest):
 
     def test_color_emoji_not_shrunk(self):
         # Regression test for https://github.com/kovidgoyal/kitty/issues/10144.
-        # fontconfig encodes the pixel-size fixup of fixed-size color faces (such
-        # as Noto Color Emoji, a ~109px bitmap strike) as FC_MATRIX. That scale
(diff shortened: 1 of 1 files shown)


Strykar:
Rebased onto d4106ef2d. This replaces the coverage threshold (lowered to 0.3 there to get CI green) with an environment-independent check: render the same font via its fontconfig descriptor (which carries the size-fixup matrix) and via its file path (which does not), and assert the same size. The buggy render comes out ~8x smaller, so the margin is large and fixed rather than the thin, font/cell-dependent margin of a coverage threshold (the buggy render is ~0.28 here vs the 0.3 cutoff). Verified locally with the sanitized build and full test suite.

Issue #10161, opened 2026-06-17
kitten dnd doesn't work with telegram


koolkhel:
**Describe the bug**
It's not possible to drop a file dragged from remote kitten dnd to Telegram Desktop app chat (and probably, elsewhere).

**To Reproduce**
1. on a remote server, do 'kitten dnd file.txt'
2. drag the window to telegram
3. icon changes
4. but telegram desktop doesn't accept it to chat, so I have to relay the file in tmp dir

**Screenshots**
Can't screenshot that :-(

**Environment details**
```
kitty 0.47.4 (a685af8c35) created by Kovid Goyal
Darwin mbp2024.local 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:15 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6041 arm64
ProductName:		macOS ProductVersion:		26.5.1 BuildVersion:		25F80
OpenGL: '4.1 Metal - 90.5' Detected version: 4.1
Frozen: True
Fonts:
  medium: Iosevka-Koolkhel-Term-Serif: /Users/bitterman/Library/Fonts/IosevkaKoolkhel.ttc
          Features: ()
    bold: Iosevka-Koolkhel-Term-Serif-Semibold: /Users/bitterman/Library/Fonts/IosevkaKoolkhel.ttc
          Features: ()
  italic: Iosevka-Koolkhel-Term-Serif-Italic: /Users/bitterman/Library/Fonts/IosevkaKoolkhel.ttc
          Features: ()
      bi: Iosevka-Koolkhel-Term-Serif-Semibold-Italic: /Users/bitterman/Library/Fonts/IosevkaKoolkhel.ttc
          Features: ()
Paths:
  kitty: /Applications/kitty.app/Contents/MacOS/kitty
  base dir: /Applications/kitty.app/Contents/Resources/kitty
  extensions dir: /Applications/kitty.app/Contents/Resources/Python/lib/kitty-extensions
  system shell: /opt/homebrew/bin/bash
System color scheme: dark. Applied color theme type: none
Loaded config files:
  /Users/bitterman/.config/kitty/kitty.conf

Config options different from defaults:
allow_remote_control    yes
confirm_os_window_close (0, False)
cursor_blink_interval   (0.0, kitty.options.utils.EasingFunction(), kitty.options.utils.EasingFunction())
cursor_text_color       None
enabled_layouts         ['grid']
[...]

PR #10166, opened 2026-06-18
dnd: eagerly stage remote drags so non-promise apps accept them

koolkhel:
On macOS the native drag and drop hands the receiving application a real file:// URL that is read synchronously at drop time. Apps that do not implement file promises (Firefox, Telegram, most Chromium/Qt/Electron apps) therefore could not accept remote drags from the dnd kitten, which relied on lazy file promises. Finder, which does support promises, worked.

Make remote drags work everywhere by downloading (staging) the remote files into a local temp dir before the OS drag is started, then placing real file:// URLs on the pasteboard instead of promises.

The staging is triggered by the drag gesture (not proactively): on mouse-down the kitten builds the offer and sends a new t=P:x=-2 "prepare" request; the terminal stages all uri-list entries via the existing t=k machinery and reports progress with a new t=s status message. The kitten shows "Transferring files, please wait..." until staging completes, then starts the drag with real URLs. If the button is released mid-transfer the staged data is kept and "OK to drag" is shown so the retry is instant.

Local drags and non-macOS platforms report ready immediately, so there is no perceptible delay and lazy transfer behaviour is unchanged there.

See #10161.

### DIFF
--- a/docs/dnd-protocol.rst
+++ b/docs/dnd-protocol.rst
@@ -438,6 +438,51 @@ inform the client of it with::
 The error code for too many resources is ``EMFILE``,
 for IO errors is ``EIO`` and so on.
 
+Eager staging for platforms without lazy file transfer
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+On some platforms (notably macOS) the native drag and drop machinery requires
+that, by the time a drop happens, the dragged files already exist on the local
+filesystem (it hands the receiving application a real ``file://`` URL that is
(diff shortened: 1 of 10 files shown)


kovidgoyal:
yes, that's the obvious solution but its not something that should be
done IMO. The terminal cannot know in advance what files the drop
destination would want. Many applications make multiple versions or
large number of files available which means drag and drop would be very
slow and inefficient as the terminal would need to transfer large amounts of useless
data before even starting the drag. The correct solution here is for
drop destinations to implement support for file promises, this API has
existed in macOS for over a decade and is used by native macOS apps.

kovidgoyal:
Yes, native apps tend to support native APIs properly, browsers not so
much.

PR #10193, opened 2026-06-26
Fix color render path for non-emoji codepoints on color-capable faces

kyzabuilds:
**the fix:**

we were checking `has_emoji_presentation()` for all codepoints when deciding on the render path, but that only applies to emoji with variation selectors (VS15/VS16). now non-emoji codepoints use `face_has_color()` instead, and we set `*was_colored` based on what actually got rendered rather than guessing ahead of time.

image

**what was happening:**

when using colored emoji characters for icons or folders, kitty was rendering just the outline instead of the full colored version. we needed to extend the logic to handle color rendering for these characters properly. they work great now.

let me know if this needs any adjustments. thanks for maintaining such a solid terminal emulator!

### DIFF
--- a/kitty/core_text.m
+++ b/kitty/core_text.m
@@ -522,6 +522,11 @@ face_apply_scaling(PyObject *f, const FONTS_DATA_HANDLE fg) {
     return set_size_for_face(f, 0, false, fg);
 }
 
+bool
+face_has_color(PyObject *f) {
+    return CTFontSupportsColorGlyphs(((CTFace*)f)->ct_font);
+}
+
 static PyObject*
 set_size(CTFace *self, PyObject *args) {
     double font_sz_in_pts, dpi_x, dpi_y;
--- a/kitty/fonts.c
+++ b/kitty/fonts.c
@@ -1200,7 +1200,13 @@ render_group(
         is_only_filled_boxes = true;
         for (unsigned i = 1; i < num_glyphs && is_only_filled_boxes; i++) if (global_glyph_render_scratch.glyphs[i] != box_glyph_id) is_only_filled_boxes = false;
     }
-    bool was_colored = !is_only_filled_boxes && has_emoji_presentation(cpu_cells, global_glyph_render_scratch.lc);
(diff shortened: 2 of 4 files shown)


kovidgoyal:
That is deliberate. Whether a character is colored or not is not purely font dependent. Making it so means that now UIs that use NERD font symbols and expect them to match the foreground color will break. In terminal land only glyphs that have emoji presentation (which is a unicode database property and not solely from the variation selectors) are allowed to be rendered in color.

kyzabuilds:
I hadn't even thought of that, thanks for the feedback! I'd still like to see this kind of thing added. The primary use case is just nicer icons in the terminal. But I see it's not as simple of a change as the pull request assumed. Is there a balance we can strike so that color capable characters render without breaking the terminal?

kovidgoyal:
Well for this to work one needs either a new standard or a new protocol.
A new standard would specify which PUA unicode codepoints have emoji
presentation then they can have consistent appearance. Or alternately,
a less disruptive change would be to give PUA characters Emoji
Presentation Base status so that if they are followed by VS16 they get
emoji presentation.

A new protocol would be an escape code to force a PUA character to have
emoji presentation.

Either way, these would require changes to client applications, not just
terminals.

ef3d0c3e:
> I hadn't even thought of that, thanks for the feedback! I'd still like to see this kind of thing added. The primary use case is just nicer icons in the terminal. But I see it's not as simple of a change as the pull request assumed. Is there a balance we can strike so that color capable characters render without breaking the terminal?

My apologies for bumping a closed PR, but I found a workaround for custom COLRv1 glyphs in kitty.

The solution is to build a font with a gsub table that contains ligatures that start with an emoji: 
```
feature rlig {
    sub uni1f340 VS01 by my-colrv1-glyph;
} rlig
```
Here, uni1f340 is the four-leaf-clover emoji and VS01 is the variation selector 1, i.e U+fe00.

As long as the first codepoint of the substitution is an emoji, kitty will happily render colrv1 glyphs.

I know that this feels like a 'hack', but it seems to be working properly for me.

Issue #10195, opened 2026-06-26
Add option to use prefix-free hints so confirmation is not needed.


tellezhector:
## **Current problem: Inconsistent flow when a hint is a prefix of another**

When using the `hints` kitten ([live documentation](https://sw.kovidgoyal.net/kitty/kittens/hints/)/ documentation at the moment of writing [part 1](https://github.com/kovidgoyal/kitty/blob/c48111d92cba307f869dc1fd4ad0249b50502874/docs/kittens/hints.rst), [part 2](https://github.com/kovidgoyal/kitty/blob/c48111d92cba307f869dc1fd4ad0249b50502874/kittens/hints/main.py#L75)), the keyboard flow is disrupted when there is prefix ambiguity.

Some hints are prefixes of other longer ones (for example, code `1` prefixes `10`, `11`, `1a`, etc.), 

If your collection of active hints has, for example: `[1, ..., 10, 11, ..., 1a, ..., a]` and so far you have already typed `1`, the terminal cannot immediately trigger the action that matches `1`, because there is still ambiguity on whether you intend to match `1` or something that starts with `1`. You to press a confirmation key (`Space`/`Enter` ) to resolve the ambiguity.

In contrast, typing a code that is not a prefix of any other code (like `10`) always triggers immediately without confirmation.

This disrupts the flow, in a multi-match scenario, with the hints collection described above,  you may have intended to match `1` and you are ready to start typing the next hint (let's say `a`), so you type `a` again and you now have matched `1a` instead of `1`, and `a`, as you intended. The flow is disrupted in that you have to remember to hit the confirmation key (but not all the time, because if there is no prefix, then the action triggers as soon as you finish the hint).

Single match scenarios are also disrupted by requiring confirmation on cases where you have typed a prefix of another hint (you have to press one more key that was not explicitly shown in the hint initially)

## **Describe the solution you'd like**

If hints where prefix-free confirmation would no longer be necessary and all matches would trigger as soon as you are done typing any hint, making the flow more consistent (today, you sometimes needs to confirm and sometimes you don't).

### On how to generate prefix-free hints:

Note: I arrived at a solution, but it seemed a bit cryptic so I made my best effort to explain how I arrived to it. All of what is said in this section very likely already known and it probably has a name, but I had fun discovering it on my own, there is a TL;DR at the end of the section and working code if you want to skip all the text.

[...]

PR #10210, opened 2026-07-01
Implements option to generate prefix-free hints

tellezhector:
- Declares --prefix-free CLI flag in main.py.
- Implements dynamic index offset shifting in marks.go to prevent
  generating hints that prefix other hints.
- Integrates prefix-free encoding in main.go.
- Includes tests.

Implements #10195


tellezhector:
Here goes attempt one. I went for a new file (prefix_free_hints.go) to keep things a bit separated but they two functions in there could be easily merged into main.go if that's preferred.

The test "// Test prefix-free property verification" is a bit brute-force-cy, for two different alphabets it generates sets of 2, 4, 8, 16, ..., 1024 hints  and ensures that, within each set, all pairs (pair by pair) are prefix-free (quadratic). I couldn't think of a better way to ensure the behavior survives future edits. I at least made sure that there isn't much of a performance penalty, the tests in that file from beginning to end take from 0.015 to 0.030 seconds on my not so beefy computer.

PR #10225, opened 2026-07-05
[IMP] choose-files: make ctrl+backspace delete a word

Mr7F:
At the moment, ctrl+backspace does nothing in choose-files. With the commit, ctrl+backspace will delete the previous word.

### DIFF
--- a/kittens/choose_files/search-bar.go
+++ b/kittens/choose_files/search-bar.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/kovidgoyal/kitty/tools/tui/loop"
 	"github.com/kovidgoyal/kitty/tools/utils"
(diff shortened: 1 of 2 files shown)


Mr7F:
I added test this time, not sure it's worth, let me know what you think

Mr7F:
@kovidgoyal b90e2124e0e4afe01bfcab574abc9b994433a4d3 won't solve the issue for the kitten `choose-files` no?

When I test your change on the latest master, the bug is still there (it uses custom code for the search bar in `search-bar.go`, with even custom code for the `backspace` key, so `ctrl+backspace` should be there too I think)

kovidgoyal:
Ah then IMO the better fix is to change the kitten to use readline I dont know why I didnt do that in the first place. Maybe I was being lazy. If you want to do that yourself let me know, otherwise I will look at it when I have some time.

Mr7F:
@kovidgoyal I guess it was also because arrow key are used to select files instead of moving the cursor in the text input?

That being said, it would be really nice too to be able to move the cursor in the search bar with arrow key / home / end (if we make a mistake, or we have to many result and want to filter more based on the parent folders), that's something that annoy me a bit sometimes when using that kitten.

Those keys are already used by the kitten (home: `first_on_screen`, end: `last_on_screen`, arrow for movement in the result panel), so maybe for backward compatibility we can add new commands in the `choose-files.conf` (to remap `home` to beginning of line instead of `first_on_screen`, left/right to move the cursor, etc), and not use it by default?

(and so one could still do `alt+home` for `first_on_screen` eg)

What do you think?

kovidgoyal:
https://github.com/kovidgoyal/kitty/commit/011ef4d319546f39c3524c339cae5a98dfc9ac3e

Mr7F:
@kovidgoyal Thanks! It look very nice 🙂

I just spoted a small visual bug, we might want to reduce by one char to not overflow the search bar, see:
image

Also it's minor, but some chars take more place, eg: (for mojo programming language, file extension can end with `🔥`)
image

Other than that, it's perfect, thanks again 🙂

kovidgoyal:
Fixed by 5ca4805

PR #10243, opened 2026-07-09
Fix IME popup position after text_input re-enters on Wayland

lsdhophora:
When a kitty window regains text-input focus (e.g. switching back to a workspace in sway), the zwp_text_input_v3::enter event is fired without a corresponding wl_keyboard.enter. text_input_enter() only called enable + set_content_type, leaving the cursor rectangle at (0,0) and causing the IME popup to appear at the top-left corner.

Restore the last known cursor rectangle (saved during normal input) in text_input_enter() so the popup immediately appears at the correct position.

Tested on sway 1.12 + fcitx5.

### DIFF
--- a/glfw/wl_text_input.c
+++ b/glfw/wl_text_input.c
@@ -35,6 +35,15 @@ text_input_enter(void *data UNUSED, struct zwp_text_input_v3 *txt_input, struct 
         ime_focused = true;
         zwp_text_input_v3_enable(txt_input);
         zwp_text_input_v3_set_content_type(txt_input, ZWP_TEXT_INPUT_V3_CONTENT_HINT_NONE, ZWP_TEXT_INPUT_V3_CONTENT_PURPOSE_TERMINAL);
+	if (last_cursor_width > 0) {
+            zwp_text_input_v3_set_cursor_rectangle(
+                txt_input,
+                last_cursor_left,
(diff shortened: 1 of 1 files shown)


kovidgoyal:
What happens if there are two kitty OS Windows and the enter event is
for a different OS window than the last rectangle is stored for?

lsdhophora:
> What happens if there are two kitty OS Windows and the enter event is for a different OS window than the last rectangle is stored for?

On sway (and likely other compositors), wl_keyboard.enter always fires before zwp_text_input_v3.enter when switching between OS windows. So by the time text_input_enter() runs, last_cursor_* has already been updated to the new window's cursor rectangle via GLFW_IME_UPDATE_CURSOR_POSITION.
I tested with two OS windows in sway — switching back and forth with IME active — and the popup always appeared at the correct position. No race condition was observed.
Additionally, zwp_text_input_v3 is a per-seat singleton shared across all surfaces, so global state for last_cursor_* is not inherently problematic: at any given moment there's only one focused surface, and the cursor rect always corresponds to the currently focused window.
If you're concerned about a hypothetical race where text_input.enter fires before keyboard.enter, the worst case is a one-frame incorrect position, which is still a significant improvement over the current behavior (IME popup stuck at (0,0) permanently until the user types).

PR #10250, opened 2026-07-11
macOS: disable verification code autofill in windows

alexasa79:
## Summary

Explicitly make the Cocoa content view an NSTextContent provider on macOS 11 and newer and report a nil content type. This tells AppKit that kitty terminal input has no semantic text classification, preventing verification-code AutoFill suggestions from appearing in terminal windows.

The protocol conformance and methods are guarded by MAC_OS_X_VERSION_MAX_ALLOWED so builds using older SDKs are unchanged.

## Testing

- git diff --check
- clang syntax-only compilation of glfw/cocoa_window.m against the installed macOS SDK

A full debug build was not run because the system Python is 3.9.6 while kitty requires Python 3.12 or newer.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -216,6 +216,8 @@ Detailed list of changes
 
 - choose-file kitten: Use a full readline editor for the search box. Also allow remapping the up/down/home/end keys to use for editing instead of navigating the file list (:pull:`10225`)
 
+- macOS: Prevent verification code AutoFill suggestions from appearing in kitty terminal windows
+
 
 0.47.4 [2026-06-15]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -896,7 +896,11 @@ static void _glfwUpdateNotchCover(_GLFWwindow*);
 
 // Content view class for the GLFW window {{{
 
+#if MAC_OS_X_VERSION_MAX_ALLOWED >= 110000
+@interface GLFWContentView : NSView <NSTextInputClient, NSTextContent>
+#else
 @interface GLFWContentView : NSView <NSTextInputClient>
+#endif
 {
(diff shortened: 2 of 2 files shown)


kovidgoyal:
What does one have to do to have a verification code autofill suggestion
pop up in kitty? And doing this will almost certainly have other
implications, like say breaking dictation or IME or something.

alexasa79:
> What does one have to do to have a verification code autofill suggestion pop up in kitty? And doing this will almost certainly have other implications, like say breaking dictation or IME or something.

On macOS, the Secure Code Autofill feature (introduced in Mojave and refined in later versions) automatically detects six-digit one-time verification codes received via SMS in the Messages app and offers to autofill them into active text input fields. This is very convenient in web browsers for 2FA login forms, where a suggestion bar or popup appears above the field. 

For some reason this mechanism also works in Kitty terminal, and it is disruptive and annoying. This behavior is specific to macOS 26: Apple now offers verification codes detected by Messages or Mail in every active text-input experience for up to three minutes. 

You were right to question the implications. After checking Apple’s current documentation, my NSTextContent/nil approach is not the correct opt-out. Any thoughts on how to fix this? Perhaps NSAutoFillRequiresTextContentTypeForOneTimeCodeOnMac (https://developer.apple.com/documentation/bundleresources/information-property-list/nsautofillrequirestextcontenttypeforonetimecodeonmac).

Thanks

kovidgoyal:
That sounds more reasonable, although if this is system wide behavior I
dont think we should be opting out of it by default. Does it happen in
Terminal.app for instance? I can certainly imagine someone running a TUI
that requires one time codes.

alexasa79:
I don't know if Terminal.app has this problem. It's tricky to reproduce. It looks like Ghostty had this problem and applied the NSAutoFillRequiresTextContentTypeForOneTimeCodeOnMac fix: 
https://github.com/ghostty-org/ghostty/pull/11351

Perhaps make it configurable?

Issue #10249, opened 2026-07-11
Anonymous heap grows without bound in proportion to terminal output (≥1.6 B retained per byte processed), freed only on window close


Sinity:
**Describe the bug**

A kitty instance hosting high-output full-screen TUI applications retains anonymous heap roughly *proportionally to the output it processes*, far beyond any configured buffer, releasing it only when the window closes. Over a 35 h session one instance grew from 81 MB to **3094 MB** (PSS+swap); the retained pages go cold and get swapped out, never to be touched again, so the machine slowly fills swap with dead terminal heap.

The numbers come from a per-minute whole-system sampler (PSS, swap, and per-process I/O deltas recorded to SQLite), so the growth curve and its correlation with activity are precise, not eyeballed.

**Evidence (kitty 0.47.4, Linux 6.18.38, Wayland/Hyprland, NixOS, glibc)**

Instance A (hosted two full-screen TUI coding agents, high repaint volume, 35 h):

| metric | value |
|---|---|
| PSS+swap at start → end | 81 MB → 3094 MB (**86 MB/h avg**) |
| hourly growth during heavy TUI activity | 130–180 MB/h |
| hourly growth overnight (idle) | 0–7 MB/h |
| total `read()` volume of the process over the session (all fds, i.e. an *upper bound* on PTY input) | **1.87 GiB** |
| heap retained | **3.0 GiB** → ≥ 1.6 bytes retained per byte processed |
| composition (smaps) | private anon only; `Pss_Shmem` = 0; file mappings flat |
| scrollback ceiling (`scrollback_lines 10000`, 3 windows) | ~50–75 MB/window ⇒ observed retention is ~40× beyond it |

Two sibling instances with the same config but low activity grew 8–10 MB/h over the same period.

Instance B (different session, earlier): 526 MB → 2165 MB over 19 h while hourly PTY throughput was as low as 0–5 MiB — and **~900 MB was released at once** when a tab was closed, so the retention is per-window and freeable, not a process-lifetime leak.

**Excluded so far**

- `cursor_trail`: was `0` for the entire 35 h session of instance A (we A/B-tested it specifically; growth rate unchanged vs `cursor_trail 3`).
- Scrollback: see ceiling math above; growth also blows past it with the alternate screen in use most of the time.
- GPU/shmem: composition is pure private anonymous heap.
[...]

PR #10251, opened 2026-07-11
Garbage collect the text cache so unique cell texts do not grow memory without bound

Sinity:
## Problem

The per-Screen `TextCache` interns every unique multi-codepoint cell text (base + combining marks, variation selectors, multicell text) for the lifetime of the window, with no eviction. A stream of unique texts therefore grows memory without bound — the reproducer in https://github.com/kovidgoyal/kitty/issues/10249#issuecomment-4946967736 shows several hundred MB/hour at moderate throughput, and a hostile program can exhaust memory outright.

## Approach

This reuses the design kitty already has for the same shape of problem: the hyperlink pool GC (`screen_garbage_collect_hyperlink_pool`). After 8192 newly interned entries, `PREPARE_FOR_DRAW_TEXT` initiates collection at the next text-drawing boundary:

1. `tc_gc_begin()` steals the cache's entry array, arena and hash map, leaving a fresh empty cache. Allocation failure leaves the cache untouched and skips collection.
2. The Screen remaps `ch_or_idx` in every live `CPUCell` via `tc_gc_map_index()`, lazily re-interning only entries still referenced by a cell. It walks the history buffer, main and alt line buffers, paused-rendering snapshot, and both overlay-line cell arrays.
3. `tc_gc_end()` resets the additions counter and frees the stolen entries, including everything whose last reference scrolled out of history.

The collection check lives only in `PREPARE_FOR_DRAW_TEXT`; combining-character, fixed-width multicell, and tab handling do not independently initiate GC. Re-interning during GC does not increment the additions counter, preventing a live screen full of unique cells from causing collection thrashing.

The alternative hard cap has a user-visible failure mode: long-lived sessions accumulate toward it even with a bounded live working set, after which unseen cell texts must render incompletely. Collection instead bounds memory while preserving cell contents.

## Benchmark

Command, rendered with 1000 repetitions in the branch-built kitty:

`kitty --hold kitten __benchmark__ --repetitions=1000 --render ascii unicode`

| workload | before review revision | after review revision | change |
|---|---:|---:|---:|
| ASCII | 111.3 MB/s | 110.8 MB/s | -0.4% |
| Unicode | 119.5 MB/s | 119.6 MB/s | +0.1% |

The benchmark kitten now also exposes `unique_unicode`, which generates 262,144 distinct base-plus-three-combining-mark cells per repetition. At 1000 rendered repetitions it sustains **36.7 MB/s**, exercising repeated TextCache collections instead of collapsing into cache hits.

## Verification

- `test_text_cache_garbage_collection` covers explicit preservation of referenced entries, collection of scrolled-out entries, index remapping without text changes, and bounded periodic collection under a continuous unique-text stream.
- `python3 test.py --module screen text_cache_garbage_collection`: 1 test passed.
- `python3 test.py`: 350 tests run; the sole failure is the pre-existing `ShellIntegrationWithKitten.test_bash_integration` cursor-shape timeout, previously reproduced on pristine master; 5 skipped. All Go tests pass.
- Live combining-mark A/B: stock RSS rose 167 MB → 266 MB and continued climbing; the GC build stayed flat at 280–282 MB for the entire second half of the five-minute run.

## Edge behavior

- All `tc_*_at_index` lookups bounds-check and degrade to empty text, so stale indices cannot read out of bounds.
- Detached `Line` copies are not remapped, matching the hyperlink GC contract; these are Python-API conveniences rather than render-path buffers.
- Remapping rewrites cells in place without reallocating buffers, so held `CPUCell*` pointers remain valid.

### DIFF
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -905,8 +905,48 @@ set_active_hyperlink(Screen *self, char *id, char *url) {
     }
 }
 
+static void
+text_cache_gc_process_cells(TextCache *tc, TextCacheGCData *gc, CPUCell *cells, size_t count) {
+    for (size_t i = 0; i < count; i++) {
+        CPUCell *c = cells + i;
+        if (!c->ch_is_idx) continue;
+        char_type new_idx;
+        if (tc_gc_map_index(tc, gc, c->ch_or_idx, &new_idx)) c->ch_or_idx = new_idx;
+        else cell_set_char(c, 0);  // stale index, should not happen
+    }
+}
+
+static void
+text_cache_gc_process_linebuf(TextCache *tc, TextCacheGCData *gc, LineBuf *lb) {
+    if (lb) text_cache_gc_process_cells(tc, gc, lb->cpu_cell_buf, (size_t)lb->ynum * lb->xnum);
+}
+
+void
+screen_garbage_collect_text_cache(Screen *self) {
+    // The TextCache interns unique cell texts forever; remap every live cell
+    // index onto a fresh cache so entries that scrolled out of the history
+    // buffer are freed. Mirrors screen_garbage_collect_hyperlink_pool().
+    TextCacheGCData *gc = tc_gc_begin(self->text_cache);
+    if (!gc) return;  // allocation failure, cache left unchanged
+    if (self->historybuf->count) {
+        for (index_type y = self->historybuf->count; y-- > 0;) {
+            CPUCell *cells = historybuf_cpu_cells(self->historybuf, y);
+            text_cache_gc_process_cells(self->text_cache, gc, cells, self->historybuf->xnum);
+        }
+    }
+    text_cache_gc_process_linebuf(self->text_cache, gc, self->main_linebuf);
+    text_cache_gc_process_linebuf(self->text_cache, gc, self->alt_linebuf);
+    text_cache_gc_process_linebuf(self->text_cache, gc, self->paused_rendering.linebuf);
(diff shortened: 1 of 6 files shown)


kovidgoyal:
Some comments:
1) The integration point for checking if gc is needed should be PREPARE_FOR_DRAW_TEXT. Not the three random points claude seems to have chosen. screen_tab() definitely shouldnt be triggering gc
2) I would like to see benchmark results from running:
`make && kitty --hold kitten __benchmark__ --repetitions=1000 --render ascii unicode`
before and after the change
3) adds_since_last_gc needs to be set to zero in end() not begin()
4) Related to (2) it would be good to have a benachmark added to the benchmark kitten to benchmark performance under large numbers of unique unicode codepoints

Sinity:
Addressed all four points in b8da204 and ed51a3e. The GC check now lives only in PREPARE_FOR_DRAW_TEXT; the combining-character, fixed-width multicell, and screen_tab triggers are gone. adds_since_last_gc is reset in tc_gc_end() after the remap completes. I also added the unique_unicode benchmark (262,144 distinct base-plus-three-combining-mark cells per repetition). Requested rendered 1000-repetition results: ASCII 111.3 -> 110.8 MB/s (-0.4%); Unicode 119.5 -> 119.6 MB/s (+0.1%). The new unique workload sustains 36.7 MB/s at 1000 repetitions. Focused GC test passes; full suite is unchanged apart from the known pristine-baseline shell-integration cursor timeout, and all Go tests pass. I updated the PR body to match the revised integration point and include the benchmark details.

PR #10262, opened 2026-07-14
Add `focus_bias` option for vertical and horizontal layouts

devgertschi:
## Summary

Add `focus_bias`, an opt-in option for the existing `vertical` and `horizontal` layouts.

It gives the focused window at least an equal share, or a larger configurable share, of the layout's main axis. For example, with `focus_bias=60`, the focused window gets 60% of the height/width and the remaining windows split the other 40% equally.

```conf
enabled_layouts vertical:focus_bias=60,horizontal:focus_bias=60
```

The bias can also be changed at runtime:

```conf
map ctrl+. layout_action focus_bias 60 70 80
```

## Behavior

The larger slot follows focus automatically when `focus_bias` is enabled. Without it, the existing vertical/horizontal behavior is unchanged.

Keyboard focus changes relayout immediately. Mouse focus changes defer relayout until button release, so the press/release pair is handled against stable geometry. This avoids accidental text selection caused by the target pane expanding between mouse down and mouse up.

The deferred mouse relayout path is opt-in and only used by layouts that need relayout on focus change.

## Motivation

This is useful for portrait/tall windows with vertically stacked terminals, where one pane is the active working context and the others remain visible for logs or background tasks.

## Tests

- `./test.py --module layout`
- Manual mouse testing with multiple panes in `vertical:focus_bias=60`

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -190,6 +190,8 @@ Detailed list of changes
 
 - kitten @ get-text: Add support for :code:`alternate` and :code:`alternate_scrollback` extents to fetch text from the alternate screen buffer (:iss:`10165`)
 
+- Add a new ``verticalfocus`` layout that gives the focused window a configurable larger share of the vertical space.
+
 - Wayland: Fix first OS window being a few cells too small when ``initial_window_width/initial_window_height`` are set in cells and a fractional display scale is in use (:iss:`10146`)
 
 - macOS: New option :opt:`macos_ns_window_layer` to more precisely control what
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -190,7 +190,7 @@ Detailed list of changes
 
 - kitten @ get-text: Add support for :code:`alternate` and :code:`alternate_scrollback` extents to fetch text from the alternate screen buffer (:iss:`10165`)
 
-- Add a new ``verticalfocus`` layout that gives the focused window a configurable larger share of the vertical space.
+- Add a ``focus_bias`` option to the vertical and horizontal layouts that gives the focused window at least an equal share, or a configurable larger share, of the available space.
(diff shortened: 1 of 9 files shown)


kovidgoyal:
Why not just make this an option on the existing vertical/horizontal layouts?

devgertschi:
> Why not just make this an option on the existing vertical/horizontal layouts?

I tried this with the horizontal layout too, but changing the width on focus felt poor UX because it caused frequent relayouts. The existing manual resizing worked better there.

The vertical layout felt different, since changing height was much less disruptive. I kept it separate so I would not change the current vertical and horizontal layouts more than needed.

That said, I can make it an option on the existing layouts if you prefer.

kovidgoyal:
Yes, please make it an option on the existing layouts.

kovidgoyal:
Some comments:

1) how does this interact with resizing of windows. From what I can tell
focus_bias basically disables resizing? Why not implement it using the
existing bias slot mechanism with a reshuffle of the slots on focus
change?

2) It's a bit weird that left clicking will do proper bias changing
focus switch but any other button will do non biased changing focus
switch

devgertschi:
> Some comments:
> 
> 1. how does this interact with resizing of windows. From what I can tell
>    focus_bias basically disables resizing? Why not implement it using the
>    existing bias slot mechanism with a reshuffle of the slots on focus
>    change?
> 2. It's a bit weird that left clicking will do proper bias changing
>    focus switch but any other button will do non biased changing focus
>    switch

Thanks. The mouse-button inconsistency is fixed. I’ve started reworking the bias handling, but preserving manual resizing while moving the focused bias between slots isn’t entirely straightforward. I’ve moved the PR back to draft while I work through it.

PR #10295, opened 2026-07-28
Add terminal visibility reports

rockorager:
## Summary

- implement DEC private mode 2033 for terminal visibility reports
- support one-shot visibility queries with `CSI ? 998 n`
- report visibility changes for tabs, panes, moved views, hidden/minimized windows, macOS occlusion, and Wayland suspended surfaces
- conservatively report potentially visible when visibility is unknown

Specification: https://rockorager.dev/misc/visibility-reports/

## Testing

- `make debug`
- `./test.py` (353 tests passed, 7 skipped; all Go tests passed)


kovidgoyal:
In report_current_visibility() why are we updating visibility_state instead of just reporting it?

kovidgoyal:
Merged changing report visibility to only report not update state. If there is some reason for it to update state then let me know.

Issue #8351, opened 2025-02-19
`shift+space` is recognized as `space` in terminal app


phanen:
**Describe the bug**
`shift+space` is recognized as `space` in terminal app (`nvim`, `kitty`).

**To Reproduce**
1. launch a nvim in kitty
```sh
kitty --config NONE -- nvim -u NONE +'map <s-space> <cmd>echo "shift+space"<cr>' +'map <space> <cmd>echo "space"<cr>'
```
2. in nvim, type `<s-space>`, message say `space` (Expected: `shift+space`).

Here's a workaround for me but I wonder why it didn't work as expected by default:
```sh
kitty --config NONE -o map='shift+space send_key shift+space' -- nvim -u NONE +'map <s-space> <cmd>echo "shift+space"<cr>' +'map <space> <cmd>echo "space"<cr>'
```

kovidgoyal:
This is a bug in nvim.

kitty --config NONE kitten show-key -m kitty

and pressing shift+space shows the key being recognized as shift+space
as expected. shift+space has associated text, the space character.
nvim is recognizing the key based on the associated text instead of the
key code.

Try pressing shift+space in any GUI app and you will see it types the
space character, thereby showing that kitty is encoding the event
correctly.

Your workaround works because it is remapping shift+space to be a key
press via send-key which means it no longer has associated text.

phanen:
`shift+space` is also recognized as `space` in fish-shell.

both nvim/fish-shell enable "disambiguate" only (no associated text)
https://github.com/neovim/neovim/blob/446b9d8c5589f931a71904cd5bf384fec56ebf2b/src/nvim/tui/tui.c#L320-L324.
https://github.com/fish-shell/fish-shell/blob/2a28107be769cd79b28793889f440ba19da78c3e/src/terminal.rs#L310.

`kitten show-key -m kitty` work because it seems enabling all flags.

kovidgoyal:
Yes, since shift+space and space both generate the same *text* they cannot be disambiguated with only disambiguate, they need report associated text as well. If that were not so shift+space would no longer generate text which would be unexpected.

krobelus:
I'm not sure how an app would implement this without breaking at least some macOS scenarios.

To disambiguate shift-space from space, I enable "report associated text".
This makes shift-space send

    CSI 32 ; 2 ; 32 u

My app handles this as follows:
1. first try to find a shortcut for the key (shift-space), ignoring associated text
[...]

PR #10296, opened 2026-07-29
fix(key-encoding): emit CSI u for shift+space in disambiguate mode

phanen:
Emit CSI u for shift+space in disambiguate mode

https://github.com/kovidgoyal/kitty/issues/8351#issuecomment-5113564951


kovidgoyal:
see your issue for why this should not be done.

PR #10303, opened 2026-08-01
Vertical tabs: support multi-line and wrapped tab titles

### LINKED ISSUE #10302: Support for multiline/line-wrapped tab titles (for: vertical tabs)
The new vertical tabs feature is great, and I'm having decent success with it, but my tab titles are too long, even with custom titling, which makes it difficult to use as-is. This is actually even true with horizontal tabs too, but to a lesser degree.

I have my shell configured to write my titles to try to remediate this, and have it currently displaying my current layout, a shortened cwd, my local branch name (which is often quite long), along with the current program opened.

Here's that with horizontal tabs:

Image

And vertical tabs:

Image

I think what would be useful to handle this is support for multi-line tabs with line-wrapping. I'm guessing this is actually a huge pain to do so I can imagine this might not be accepted, but I think it would be really useful.

One reason i think multiline tab titles would work well here is that in my setup, It's about 43 tabs before I reach the bottom of my screen, and that's with an entire empty line between each tab (which I'm trying to figure out how I can remove that too). So there's plenty of vertical room for multi-line tab titles.

I'm currently working on condensing my tab titles, but even just my CWD alone is too long, not even mentioning any of the other things I'd like in my title. 

I've tried setting `tab_title_max_length` but it starts taking too much horizontal space.

I've tried adjusting `tab_title_template` to include newlines, which seems to actually work..kinda. I do get new lines but each line gets indented and after 2 lines my titles start to overlap. I was going ot use a custom `tab_bar.py` if that worked and write my own line wrapping handler but it doesn't work because of how the newlines get handled.

Eg: `tab_title_template "hello\nworld\ncats\ndogs"` results in:

Image

chancez:
Fixes #10302.

Vertical tab titles were limited to a single line, so long titles got
truncated even though the sidebar has plenty of unused vertical space.
Newlines in tab_title_template looked like they almost worked, but each
line was drawn where the previous one ended and titles collided with the
tab below after two lines.

Titles can now span multiple lines, and tabs are packed by the number of
lines each title actually uses rather than a fixed two. The blank line
between tabs is preserved, still inserted while there is room and dropped
automatically once the tabs need the space.

Two new options, both vertical-only and defaulting to current behavior:

- `tab_title_max_lines` (default 1) - how many lines a title may use
- `tab_title_wrap` (default no) - `no`, `yes` to wrap at the tab bar
  width, or a number of cells

Also included as a separate commit: vertical tabs only painted each line
as far as its text reached, leaving a ragged right edge. That predates
this change and is visible with single-line titles on master, but
multi-line titles make it obvious.

One behavior change for existing vertical tab bar users: a title
containing newlines now shows an ellipsis to mark lines dropped by the
line limit.

Four commits, each self-contained. The background fill is separable if
you would rather take it on its own.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -193,6 +193,10 @@ Detailed list of changes
 - Vertical tabs: fill each tab's background for the full width of the tab bar, so
   titles shorter than the sidebar no longer leave a ragged edge
 
+- Vertical tabs: add :opt:`tab_title_wrap` to wrap long tab titles over multiple
+  lines instead of truncating them, either at the width of the tab bar or at a
+  specified number of cells (:iss:`10302`)
+
 
 0.48.2 [2026-07-30]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -190,6 +190,9 @@ Detailed list of changes
 - Vertical tabs: pack tabs according to the number of lines each title actually
   uses, instead of reserving the same number of lines for every tab
 
+- Vertical tabs: fill each tab's background for the full width of the tab bar, so
+  titles shorter than the sidebar no longer leave a ragged edge
+
 
 0.48.2 [2026-07-30]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -180,6 +180,17 @@ Detailed list of changes
 -------------------------------------
 
 
+0.48.3 [future]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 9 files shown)


kovidgoyal:
the blank line is by design, IIRC it should automatically go away when there
is insufficient vertical space, no need for an option for it.

chancez:
You're right, thanks. I've dropped the option and made it automatic again.

The blank line was falling out of every tab reserving two lines, which
stopped being true once a tab occupies only as many lines as its title
needs, so it disappeared as a side effect rather than intentionally. It is
now inserted explicitly between tabs while there is room and dropped once
the tabs need the space, so it still goes away automatically when there is
insufficient vertical space.

Squashed into the commit that changed the packing, so `tab_bar_tab_spacing`
is gone entirely.

Issue #9649, opened 2026-03-12
Scroll issues on 0.46


carlosala:
**Describe the bug**
From 0.46 on, the mouse scroll is too fast. I am using Wayland, nevertheless, I run kitty on x11 (due to decoration bar, it is just ugly on wayland).

I tried to reduce `wheel_scroll_multiplier` with numbers `0 < x < 1` but it doesn't seem to change anything. I'm sure kitty is detecting the touchpad as a `wheel` because if I change the setting to `-1.0` it changes the direction of the scroll.

Besides that, the horizontal scroll does not respect the direction of the scroll, i.e. the vertical scroll is "natural movement" (move fingers up to go down in the document) and the horizontals stays on "non-natural" (move fingers left to go left in the document).

Thanks!

**To Reproduce**
Just happens.

**Environment details**
```
kitty 0.46.0 created by Kovid Goyal
Linux archevo 6.19.6-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Wed, 04 Mar 2026 18:24:56 +0000 x86_64
Arch Linux 6.19.6-zen1-1-zen (/dev/tty)
Running under: X11
OpenGL: '4.6 (Core Profile) Mesa 26.0.1-arch1.1' Detected version: 4.6
Frozen: False
Fonts:
  medium: CascadiaCode-Light: /usr/share/fonts/OTF/CascadiaCode-Light.otf:0
          Features: ()
    bold: CascadiaCode-Bold: /usr/share/fonts/OTF/CascadiaCode-Bold.otf:0
          Features: ()
  italic: CascadiaCode-LightItalic: /usr/share/fonts/OTF/CascadiaCode-LightItalic.otf:0
          Features: ()
      bi: CascadiaCode-BoldItalic: /usr/share/fonts/OTF/CascadiaCode-BoldItalic.otf:0
          Features: ()
Paths:
  kitty: /usr/bin/kitty
  base dir: /usr/lib/kitty
  extensions dir: /usr/lib/kitty/kitty
  system shell: /bin/zsh
System color scheme: dark. Applied color theme type: none
Loaded config files:
  /home/carlo/.config/kitty/kitty.conf

Config options different from defaults:
allow_remote_control    yes
bold_font               Cascadia Code Bold
bold_italic_font        Cascadia Code Bold Italic
close_on_child_death    True
cursor                  None
[...]

PR #10306, opened 2026-08-02
Wayland: do not ignore small movements of the scroll wheel

francislavoie:
### The problem

With a Logitech MX Master 3 on GNOME/Wayland, single notches of the scroll wheel frequently do nothing in terminal programs that have requested mouse events (pagers, editors, TUIs). It feels inconsistent — you nudge the wheel one notch and nothing happens, so you nudge it again. kitty's own scrollback scrolling feels fine, because it scrolls by pixels; only the path that has to synthesise discrete wheel events for the application is affected.

### Why it happens

The wheel is in HID++ high resolution mode, so one physical detent arrives as **several VALUE120 fragments** rather than one event of 120. Crucially, a detent does **not** reliably total 120 units.

Measured with `kitty --debug-input` over 96 detents:

| detent total (v120) | 96 | 104 | 112 | 120 | 128 | 136 | 144 | 152 |
|---|---|---|---|---|---|---|---|---|

96% of detents landed between 0.80 and 1.40 lines. A typical one notch looks like:

```
Scroll type=v120 y: -48   Scroll type=v120 y: -24
Scroll type=v120 y: -24   Scroll type=v120 y: -24     (total -120, over 24ms)
```

`scale_scroll()` carries `pending_scroll_pixels` across events and never resets it. So residual left over from the previous detent — possibly with the opposite sign — has to be cancelled before a line can be emitted, and the detent is consumed doing that.

Correlating what kitty received against what it forwarded, over 96 detents:

| | detents that scrolled nothing |
|---|---|
| after a direction reversal | **47%** (15/32) |
| continuing in the same direction | **14%** (9/64) |

Nothing is lost in aggregate — a swallowed detent is repaid as a double later — but per notch it is unpredictable, which is what makes it feel broken. Direction reversals are worst because the leftover residual is then guaranteed to have the wrong sign.

### The change

Treat an idle gap or a direction reversal as the start of a new detent. Discard the stale residual, and emit a line as soon as the detent is unambiguous instead of waiting for a full line to accumulate. Both decisions are taken when a fragment arrives, never on a timer, so **no latency is added**.

Measured after the change, over 118 detents: **no detent failed to scroll**, in either direction.

### Relationship to #9128

This is the part I would most like a second opinion on.

#9128 deliberately introduced accumulate-until-a-full-line, fixing the opposite complaint: sub-line VALUE120 events each scrolling a whole line felt wrong on smooth-scrolling devices. This PR is not a revert of that, and tries to keep both behaviours:

- Within a **continuous same-direction stream** nothing changes — fragments accumulate exactly as #9128 made them.
- The eager emit only fires at a **detent boundary** (>200ms idle, or a direction change), which a continuously streaming device does not produce mid-scroll.
- Fragments **too small to be a detent still scroll nothing**. The rescue threshold is half a line, comfortably above any single fragment, which is what stops this regressing into per-fragment scrolling.
- Touchpads are unaffected regardless: they arrive as `GLFW_SCROLL_OFFEST_HIGHRES`, and only the V120 branch is touched.

The separation is clean on real hardware — genuine detents opened with a 48–80 unit fragment while an accidental brush of the wheel was 16, and no detent ever totalled between 0.25 and 0.80 lines. But those constants are tuned against one mouse, and I would not be surprised if other hardware wants different ones. Happy to turn them into options, or to take a different approach entirely.

I could not find an existing issue for this. The closest is a comment on #9649 ("I need to scroll a lot ... just to go down one line"), though that was attributed to XWayland.

### Testing

Added `test_v120_scroll_accumulation` in `kitty_tests/mouse.py`, driving the accumulation logic with real captured fragment sequences: detents totalling 96/112/120/152 units each scroll exactly once, reversals are not swallowed, wheel brushes scroll nothing, and rapid multi-detent bursts still scroll the full amount. Writing it caught a genuine bug in an earlier revision of this patch, where a detent could scroll twice.

`./test.py` is otherwise unchanged — the only failures on my machine are two pre-existing `test_font_selection` cases, because this system names the font `UbuntuMonoRoman-*` rather than `UbuntuMono-*`.

### Disclosure

This patch was written with AI assistance (Claude). The diagnosis, the measurements, the policy comparison and the code were produced in that session; I reviewed the result, and verified the before/after numbers on my own hardware. Every figure quoted above comes from instrumenting my actual mouse, not from an estimate. Flagging it since the change is behavioural and tuned to hardware I happen to own — please scrutinise the constants accordingly.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -187,6 +187,10 @@ Detailed list of changes
   often not scrolling, with high resolution wheels such as the Logitech MX
   Master 3 (:pull:`10306`)
 
+- Wayland: Fix a single movement of a high resolution scroll wheel sometimes not
+  scrolling at all, when the wheel reports slightly less than one line for the
+  movement (:pull:`10306`)
+
 - Vertical tabs: Improve handling of multi-line tab titles. Controlled via two new options:
   :opt:`tab_title_max_lines` and :opt:`tab_title_template` (:pull:`10303`)
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -183,6 +183,10 @@ Detailed list of changes
 0.48.3 [future]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+- Wayland: Fix the first movement of the scroll wheel after reversing direction
+  often not scrolling, with high resolution wheels such as the Logitech MX
+  Master 3 (:pull:`10306`)
+
 - Vertical tabs: Improve handling of multi-line tab titles. Controlled via two new options:
   :opt:`tab_title_max_lines` and :opt:`tab_title_template` (:pull:`10303`)
 
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -1406,7 +1406,7 @@ mouse_event(const int button, int modifiers, int action) {
 }
 
 static int
-scale_scroll(MouseTrackingMode mouse_tracking_mode, double offset, GLFWOffsetType offset_type, double *pending_scroll_pixels, int cell_size) {
+scale_scroll(MouseTrackingMode mouse_tracking_mode, double offset, GLFWOffsetType offset_type, double *pending_scroll_pixels, int cell_size, int *last_v120_dir) {
 // scale the scroll by the multiplier unless the mouse is grabbed. If the mouse is grabbed only change direction.
 #define SCALE_SCROLL(which) { double scale = OPT(which); if (mouse_tracking_mode) scale /= fabs(scale); offset *= scale; }
     int s = 0;
@@ -1423,6 +1423,16 @@ scale_scroll(MouseTrackingMode mouse_tracking_mode, double offset, GLFWOffsetTyp
         } break;
         case GLFW_SCROLL_OFFEST_V120: {
             SCALE_SCROLL(wheel_scroll_multiplier);
+            // One physical detent arrives as several VALUE120 events, so the
+            // fractional remainder is carried to the next event. Detents do not
+            // reliably total 120 units, and that carry is what lets an
+            // undersized detent still scroll a line. On a direction change the
+            // residual has the wrong sign, and would be cancelled out by the
+            // first detent of the new direction rather than scrolling it, so
+            // discard it. Scrolling in a single direction is unaffected.
+            const int dir = offset > 0 ? 1 : -1;
+            if (*last_v120_dir && dir != *last_v120_dir) *pending_scroll_pixels = 0;
+            *last_v120_dir = dir;
             const double offset_lines = offset / 120.;
             const double pixels = *pending_scroll_pixels + offset_lines * cell_size;
             if (fabs(pixels) < cell_size) {
@@ -1543,10 +1553,10 @@ scroll_event(const GLFWScrollEvent *ev) {
                 const double offset_lines = (ev->y_offset / 120.) * OPT(wheel_scroll_multiplier);
                 delta_pixels = offset_lines * global_state.callback_os_window->fonts_data->fcm.cell_height;
             }
-            screen->pending_scroll_pixels_y = 0.0;
+            osw->scroll.pending_pixels_y = 0.0;
             if (screen_apply_pixel_scroll(screen, delta_pixels) && screen->selections.in_progress) update_drag(w);
         } else {
-            int s = scale_scroll(screen->modes.mouse_tracking_mode, ev->y_offset, ev->offset_type, &screen->pending_scroll_pixels_y, global_state.callback_os_window->fonts_data->fcm.cell_height);
+            int s = scale_scroll(screen->modes.mouse_tracking_mode, ev->y_offset, ev->offset_type, &osw->scroll.pending_pixels_y, global_state.callback_os_window->fonts_data->fcm.cell_height, &osw->scroll.last_v120_dir_y);
             if (s) {
                 bool upwards = s > 0;
                 if (screen->modes.mouse_tracking_mode) {
@@ -1568,7 +1578,7 @@ scroll_event(const GLFWScrollEvent *ev) {
         }
     }
     if (ev->x_offset != 0.0) {
-        int s = scale_scroll(screen->modes.mouse_tracking_mode, ev->x_offset, ev->offset_type, &screen->pending_scroll_pixels_x, global_state.callback_os_window->fonts_data->fcm.cell_width);
+        int s = scale_scroll(screen->modes.mouse_tracking_mode, ev->x_offset, ev->offset_type, &osw->scroll.pending_pixels_x, global_state.callback_os_window->fonts_data->fcm.cell_width, &osw->scroll.last_v120_dir_x);
         if (s) {
             if (screen->modes.mouse_tracking_mode) {
                 int sz = encode_mouse_scroll(w, s > 0 ? 6 : 7, ev->keyboard_modifiers);
@@ -1659,9 +1669,29 @@ send_mock_mouse_event_to_window(PyObject *self UNUSED, PyObject *args) {
     Py_RETURN_NONE;
 }
 
+static PyObject*
+test_scale_scroll(PyObject *self UNUSED, PyObject *args) {
+    PyObject *values; int cell_size;
+    if (!PyArg_ParseTuple(args, "O!i", &PyList_Type, &values, &cell_size)) return NULL;
+    double pending = 0; int last_dir = 0;
+    PyObject *ans = PyList_New(0);
+    if (!ans) return NULL;
+    for (Py_ssize_t i = 0; i < PyList_GET_SIZE(values); i++) {
+        const double v120 = PyFloat_AsDouble(PyList_GET_ITEM(values, i));
+        if (PyErr_Occurred()) { Py_DECREF(ans); return NULL; }
+        // ANY_MODE so that wheel_scroll_multiplier is reduced to its sign
+        const int s = scale_scroll(ANY_MODE, v120, GLFW_SCROLL_OFFEST_V120, &pending, cell_size, &last_dir);
+        PyObject *t = PyLong_FromLong(s);
+        if (!t || PyList_Append(ans, t) != 0) { Py_XDECREF(t); Py_DECREF(ans); return NULL; }
+        Py_DECREF(t);
+    }
+    return ans;
+}
+
 static PyMethodDef module_methods[] = {
     {"send_mouse_event", (PyCFunction)(void (*) (void))(send_mouse_event), METH_VARARGS | METH_KEYWORDS, NULL},
     METHODB(test_encode_mouse, METH_VARARGS),
+    METHODB(test_scale_scroll, METH_VARARGS),
     METHODB(send_mock_mouse_event_to_window, METH_VARARGS),
     METHODB(mock_mouse_selection, METH_VARARGS),
     {NULL, NULL, 0, NULL}        /* Sentinel */
--- a/kitty/mouse.c
+++ b/kitty/mouse.c
@@ -1433,13 +1433,14 @@ scale_scroll(MouseTrackingMode mouse_tracking_mode, double offset, GLFWOffsetTyp
             const int dir = offset > 0 ? 1 : -1;
             if (*last_v120_dir && dir != *last_v120_dir) *pending_scroll_pixels = 0;
             *last_v120_dir = dir;
-            const double offset_lines = offset / 120.;
-            const double pixels = *pending_scroll_pixels + offset_lines * cell_size;
-            if (fabs(pixels) < cell_size) {
-                *pending_scroll_pixels = pixels;
-                return 0;
-            }
-            s = (int)round(pixels) / cell_size;
+            // Round to the nearest whole line rather than truncating, so that a
+            // detent delivering 0.93 of a line still scrolls. An isolated one
+            // otherwise never would, as there is no carry to make it up.
+            // Rounding is unbiased and movement of under half a line still does
+            // not scroll, so devices that stream many small events only see
+            // their scrolling fire half a line earlier.
+            const double pixels = *pending_scroll_pixels + (offset / 120.) * cell_size;
+            s = (int)round(pixels / cell_size);
             *pending_scroll_pixels = pixels - s * cell_size;
         } break;
         case GLFW_SCROLL_OFFSET_LINES: {
--- a/kitty/screen.h
+++ b/kitty/screen.h
@@ -114,7 +114,6 @@ typedef struct {
     PyObject_HEAD
 
     unsigned int columns, lines, margin_top, margin_bottom, scrolled_by, pixel_scroll_offset_y;
-    double pending_scroll_pixels_x, pending_scroll_pixels_y;
     CellPixelSize cell_size;
     OverlayLine overlay_line;
     id_type window_id;
--- a/kitty/state.h
+++ b/kitty/state.h
@@ -465,6 +465,10 @@ typedef struct OSWindow {
     CloseRequest close_request;
     bool is_layer_shell, hide_on_focus_loss;
     struct { int x, y; } last_drag_event;
+    struct {
+        double pending_pixels_x, pending_pixels_y;
+        int last_v120_dir_x, last_v120_dir_y;
+    } scroll;
 } OSWindow;
 
 static inline float
--- a/kitty_tests/mouse.py
+++ b/kitty_tests/mouse.py
@@ -425,3 +425,42 @@ class TestMouse(BaseTest):
         self.ae(sel(), '')
         multi_click(x=2.4)
         self.ae(sel(), '')
+
+    def test_v120_scroll_direction_reversal(self):
+        # A high resolution wheel delivers one physical detent as several
+        # VALUE120 events, and the fractional remainder is carried between them.
+        # Detents do not reliably total 120 units, so that carry is what lets an
+        # undersized detent still scroll a line. But when the direction changes
+        # the residual has the wrong sign, and the first detent of the new
+        # direction gets spent cancelling it instead of scrolling.
+        # The fragment sequences below were captured from a Logitech MX Master 3.
+        from kitty.fast_data_types import test_scale_scroll
+        self.set_options()   # scale_scroll reads wheel_scroll_multiplier
+        cell = 40
+
+        def run(*v120):
+            # scale_scroll is exercised in mouse tracking mode, which reduces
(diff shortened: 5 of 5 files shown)


kovidgoyal:
Some comments:
1) I dont follow the logic of rescue lines? Is this supposed to convert a sub cell size scroll into a line scroll? Where does the value of 0.5 come from? 
2) The state parameters dont belong on Screen, bu on OSWindow since these events are scoped to OS Windows

francislavoie:
1. Yes, that's right, it's converting a scroll that's smaller than 120 units to count as 120 units. And 0.5 is a threshold (percentage of 120) where scrolls below that are not counted as being enough to scroll a line, to ignore inputs that are less than a full notch of the mouse scroll wheel. From my testing, one scroll-wheel notch emitted anywhere between 96 and 168 units (i.e. 0.8 to 1.4 times the 120 units threshold), so 0.5 is below that 0.8 which would be enough to ignore bad data (like slightly touching the scroll wheel but not enough to scroll it, depending on how sensitive the sensor is), but it accept a good scroll.
2. OSWindow makes sense, but `pending_scroll_pixels_x/y` are on Screen and `v120_accumulate` is manipulating it. Should we also move those to OSWindow as well so it's all in one place, or should it get split up?

kovidgoyal:
OK I'll review the code later, in the meantime moving the pending values to OS Window looks correct to me.

kovidgoyal:
So looking over the code comments describing the algorithm,
I have the following concerns

The basic fix of reset of pending pixels when detecting direction
change or large gap is unambiguously a good thing, though the gap
threshold here should probably be much larger. The justification for a
small gap doesnt really work, see below.

The rest of the algorithm doesnt make much sense to me. If your mouse is
not generating a value of at least 120 for every detent that would be a
bug in the mouse driver. I dont see how we can compensate for that
successfully in application code.

What your algorithm (unless i misunderstood it) does is basically
trigger a scroll for values less than 120 at the start of a burst.
But that then just means that the last click in a burst may not trigger
a scroll, instead of the first. That improves the case of scrolling one
click at time, but it doesnt fix the problem, since scrolling two clicks
at a time will now only scroll once.

The only place this can be reliably fixed is in the mouse driver.

If I have misunderstood something feel free to elaborate.

francislavoie:
I restructured it into two commits that can be reviewed independently. The eager-emit, the rescue and the gap threshold are all gone. (I did try to test the last-click-in-burst prediction and couldn't: during continuous scrolling the VALUE120 stream contains no detent boundaries, so there's no observable "last click".)

**Commit 1 — discard the residual on a direction change.** `pending_scroll_pixels` is signed and never reset, so after a reversal the first detent is spent cancelling it rather than scrolling.

**Commit 2 — round the accumulator to the nearest line instead of truncating.** Detents don't reliably total 120 units: measured on an MX Master 3, sizes ran 96–152 with only about a quarter landing exactly on 120. Truncating means a detent delivering 0.93 of a line scrolls nothing, and an isolated one never will, as there's no carry to make it up. Rounding is unbiased so long scrolls travel the same distance, and movement under half a line still doesn't scroll — a device streaming many small events only sees its scrolling fire half a line earlier.

Measured with `--debug-input` against a program that requests mouse events, segmenting detents by the inter-fragment gap (sharply bimodal, ~8 ms within a detent and >250 ms between):

|             | detents scrolling nothing after a reversal |
| ----------- | ------------------------------------------ |
| before      | 15/32                                      |
| commit 1    | 16/148                                     |
| commits 1+2 | 0/210                                      |

(each row is a separate capture, hence the differing totals)

With both commits, across 407 individual detents every one scrolled, including all 141 that delivered less than a full line. Total motion delivered went from 93.7% with commit 1 alone to 99.4%. The remaining shortfall is 3 of 79 fast multi-detent bursts finishing one line short.

If you're still not persuaded that the sub-120 case is kitty's problem to solve, drop the second commit — the first stands on its own.

**On this being a driver bug:** Firefox and GTK3 apps are fine with the same mouse because they bind `wl_seat` below version 8 and receive `axis_discrete` plus one aggregated event of exactly ±120 per detent — mutter does that accumulation on their behalf. kitty binds version 10 for `key_state.repeated`, so it gets the raw `axis_value120` fragments. There's no extra information at the lower layer; mutter is just doing this work for older clients, so no driver change is needed — only doing the same accumulation correctly here.

kovidgoyal:
On Sun, Aug 02, 2026 at 01:11:35PM -0700, Francis Lavoie wrote:
> If you're still not persuaded that the sub-120 case is kitty's problem to solve, drop the second commit — the first stands on its own.

I'll review it again, it certainly sounds like a better implementation.

>
> **On this being a driver bug:** Firefox and GTK3 apps are fine with the same mouse because they bind `wl_seat` below version 8 and receive `axis_discrete` plus one aggregated event of exactly ±120 per detent — mutter does that accumulation on their behalf. kitty binds version 10 for `key_state.repeated`, so it gets the raw `axis_value120` fragments. There's no extra information at the lower layer; mutter is just doing this work for older clients, so no driver change is needed — only doing the same accumulation correctly here.

No, I meant if the mouse is not generating 120 units per detent that is a
bug that needs to be solved in the mouse hardware/driver not in
compositors/applications. In other words the legacy "GD: wheel" event from
the mouse should match N "GD: High res wheel" events where N is the right
number to add up to 120.

If your mouse isnt doing that, that is very much a bug in the mouse.

That said I am not opposed to having kitty behave better with such
broken mice, provided it can be done so without affecting behavior with
non-broken mice.

kovidgoyal:
On Sun, Aug 02, 2026 at 01:11:35PM -0700, Francis Lavoie wrote:
> francislavoie left a comment (kovidgoyal/kitty#10306)
>
> I restructured it into two commits that can be reviewed independently. The eager-emit, the rescue and the gap threshold are all gone. (I did try to test the last-click-in-burst prediction and couldn't: during continuous scrolling the VALUE120 stream contains no detent boundaries, so there's no observable "last click".)

As I understand it mice firmware send both legacy wheel click events and
high res fractional events, though its quite possible the Linux input stack
mangles/discards them somewhere along the way, so you would need to
observe them at a lower level kernel input events or USB traces. Probably
too difficult.

>
> **Commit 1 — discard the residual on a direction change.** `pending_scroll_pixels` is signed and never reset, so after a reversal the first detent is spent cancelling it rather than scrolling.
>
> **Commit 2 — round the accumulator to the nearest line instead of truncating.** Detents don't reliably total 120 units: measured on an MX Master 3, sizes ran 96–152 with only about a quarter landing exactly on 120. Truncating means a detent delivering 0.93 of a line scrolls nothing, and an isolated one never will, as there's no carry to make it up. Rounding is unbiased so long scrolls travel the same distance, and movement under half a line still doesn't scroll — a device streaming many small events only sees its scrolling fire half a line earlier.
>
> Measured with `--debug-input` against a program that requests mouse events, segmenting detents by the inter-fragment gap (sharply bimodal, ~8 ms within a detent and >250 ms between):
>
> |             | detents scrolling nothing after a reversal |
> | ----------- | ------------------------------------------ |
> | before      | 15/32                                      |
> | commit 1    | 16/148                                     |
> | commits 1+2 | 0/45                                       |
>
> With both commits, across 63 individual detents every one scrolled, including all 10 that delivered less than a full line. Total motion delivered went from 93.7% with commit 1 alone to 99.0%. The remaining shortfall is 3 of 43 fast multi-detent bursts finishing one line short.

How are you measuring the number of detents in a burst? I thought you
said above that there are no detent events in the input stream?

>
> If you're still not persuaded that the sub-120 case is kitty's problem to solve, drop the second commit — the first stands on its own.

For the moment I will merge the first commit and we can discuss further
improvements.

kovidgoyal:
After merging your first commit, the remaining issue is only about
isolated single detent movements, correct? In a multi detent burst the
user isnt going to be able to tell how many detents they have scrolled
and match that up with lines scrolled in kitty.

So addressing the single detent scroll case. I dont think rounding is
the correct solution. That penalises mice that work correctly to
compensate for mice that do not. Indeed, as a user I would *expect*
kitty (or in this case the program running in kitty)
to not scroll a line until I have moved my wheel one full detent.
With rounding and a properly functioning mouse kitty would scroll a line
on a half detent movement. This is clearly wrong.

Now coming to the issue of malfunctioning mice, IMO the correct solution
is a new pointer events protocol that passes scroll events down to the
program running in the terminal with support for high res etc.

This is anyway on my TODO list for other reasons. However, it will be
some time before I get around to implementing this and even longer
before the ecosystem adopts it, so it's not a practical solution.

So lets think about practical ways to solve the single detent on broken
mouse use case, without impacting non-broken mice.

kovidgoyal:
And note that I have closed this PR as it is partially merged, but feel
free to continue the discussion.

francislavoie:
Thanks for merging the first commit!

I disagree that it's a bug or fundamental flaw with my mouse or the driver, it's actually a feature. Logitech mice like mine use a "MagSpeed" ratchet, which uses magnets to do the ratcheting (see https://gizmodo.com/this-new-logitech-mouse-has-a-magnetic-wheel-that-feels-1837751915 for an explainer) and when flicking the mouse it reverses the magnet's polarity to make it disengage the ratcheting to let it spin freely (or I can press a button behind the scroll wheel to switch to free-spinning mode). While it spins freely, you do expect it to report as precisely as possible. This does mean that each ratchet step can land slightly off-center because magnets aren't perfect, or would need to be too strong to force it to the correct position.

I don't have another mouse to test with right now (I've only ever owned Logitech mice for the past 25 years, nothing on the market has ever felt better for me). I'm not sure what negative effect the rounding approach would have on mechanically-locking mice.

This article https://who-t.blogspot.com/2021/08/libinput-and-high-resolution-wheel.html explains that libinput chose 120 due to its nice mathematical properties of being highly divisible, which allows for flexibility in device compatibility. Ultimately, this comes down to the question of how to properly map higher precision inputs onto fixed-sized/chunky lines.

Would you accept a patch to add a `wheel_scroll_round_partial_detents` config option (default off I guess) to enable rounding?

> How are you measuring the number of detents in a burst? I thought you
said above that there are no detent events in the input stream?

I used a script like this that Claude generated, opened in a kitty instance built from source with the code edits:

<details>
<summary>Details</summary>


```python
#!/usr/bin/env python3
"""Probe B -- what kitty itself resolves a notch into."""
import os, re, select, sys, termios, time, tty

BURST_GAP_S = 0.25
SGR = re.compile(rb"\x1b\[<(\d+);(\d+);(\d+)([mM])")
LOG = open(os.environ.get("PROBE_B_LOG", "/tmp/probe_b.log"), "a", buffering=1)
T0 = time.monotonic()

def emit(text):
    sys.stderr.write(text); sys.stderr.flush()
    LOG.write(f"[{time.monotonic() - T0:8.3f}] " + text.replace("\r\n", "\n"))

def main():
    fd = sys.stdin.fileno()
    if not os.isatty(fd):
        print("probe-b: stdin is not a tty -- run this inside kitty", file=sys.stderr); return 1
    old = termios.tcgetattr(fd)
    sys.stdout.write("\x1b[?1000h\x1b[?1006h")   # 1000 = button events, 1006 = SGR coords
    sys.stdout.flush()
    tty.setraw(fd)
    buf, burst, n = b"", [], 0
    try:
        while True:
            r, _, _ = select.select([fd], [], [], 0.05)
            now = time.monotonic()
            if r:
                chunk = os.read(fd, 1024); buf += chunk
                if b"\x03" in buf: break
                for m in SGR.finditer(buf):
                    b_ = int(m.group(1))
                    if b_ in (64, 65, 66, 67):               # wheel up/down/left/right
                        name = {64:"UP",65:"DOWN",66:"LEFT",67:"RIGHT"}[b_]
                        dt = (now - burst[-1][0]) * 1000 if burst else 0.0
                        burst.append((now, name))
                        emit(f"  wheel {name:5s} dt={dt:6.1f}ms\r\n")
                buf = buf[buf.rfind(b"M") + 1:] if b"M" in buf else b""
            elif burst and (now - burst[-1][0]) > BURST_GAP_S:
                n += 1
                span = (burst[-1][0] - burst[0][0]) * 1000
                emit(f"BURST #{n}: kitty forwarded {len(burst)} wheel event(s) "
                     f"[{' '.join(x[1] for x in burst)}] span={span:.1f}ms\r\n" + "-"*60 + "\r\n")
                burst = []
    except KeyboardInterrupt:
        pass
    finally:
        termios.tcsetattr(fd, termios.TCSADRAIN, old)
        sys.stdout.write("\x1b[?1006l\x1b[?1000l"); sys.stdout.flush()
    return 0

sys.exit(main())
```

**How it works:** it enables SGR mouse reporting (`?1000h` + `?1006h`), which makes it a mouse-grabbing TUI. kitty then stops scrolling its own scrollback and instead forwards a discrete wheel button event (SGR button 64/65) for each line it decided to scroll. So **one logged `wheel` line = one line kitty resolved from the fragments** — the exact quantity in dispute.

It was always run paired with `--debug-input` so both sides were captured at once:

```
PROBE_B_LOG=/path/probe.log kitty --debug-input --title "TEST" \
    -o confirm_os_window_close=0 python3 probe_b_kitty.py > /path/debug.txt 2>&1
```

`debug.txt` gets what kitty **received** (`Scroll type=v120 y: -48…`), `probe.log` gets what kitty **forwarded**, both timestamped — that pairing is what every number in this investigation came from.

Two caveats if you reuse it:

- **`BURST_GAP_S = 0.25` is only for the console summary lines.** All the analysis segmented detents from the `--debug-input` fragment timestamps instead, so that constant never affected the results.
- **The buffer trim on line 67 is sloppy** — it discards the buffer at each read and can drop a partial escape sequence split across reads. It never bit us (fragments arrive well inside single reads at 8 ms cadence), but it'd need fixing for a general-purpose tool.

</details>

> Indeed, as a user I would expect kitty (or in this case the program running in kitty) to not scroll a line until I have moved my wheel one full detent. With rounding and a properly functioning mouse kitty would scroll a line on a half detent movement. This is clearly wrong.

I don't think I agree with this. If I showed the intent to want to be scrolling up but the input is effectively ignored, that's what's clearly wrong IMO. a "half detent" is also basically impossible because scroll wheels are designed to settle at either the local minimums, and half would mean sitting at the maximum which is only possible by physically holding the wheel at that position, which serves no purpose whatsoever in practice.

kovidgoyal:
On Mon, Aug 03, 2026 at 02:27:13AM -0700, Francis Lavoie wrote:
> francislavoie left a comment (kovidgoyal/kitty#10306)
>
> I disagree that it's a bug or fundamental flaw with my mouse or the driver, it's actually a feature. Logitech mice like mine use a "MagSpeed" ratchet, which uses magnets to do the ratcheting (see https://gizmodo.com/this-new-logitech-mouse-has-a-magnetic-wheel-that-feels-1837751915 for an explainer) and when flicking the mouse it reverses the magnet's polarity to make it disengage the ratcheting to let it spin freely (or I can press a button behind the scroll wheel to switch to free-spinning mode). While it spins freely, you do expect it to report as precisely as possible. This does mean that each ratchet step can land slightly off-center because magnets aren't perfect, or would need to be too strong to force it to the correct position.

Outputting different values for something that is supposed to represent a fixed amount of scrolling is a bug, clearly. It's obvious your mouse wants to be a high precision scroller but has jury rigged a hack on top to get back quantum scrolling. That hack does not work well. The mouse should decide what it wants to be and be that. If it wants quantum based scrolling it should do real quantum based scrolling, which means outputting a consistent value per quantum.

>
> I don't have another mouse to test with right now (I've only ever owned Logitech mice for the past 25 years, nothing on the market has ever felt better for me). I'm not sure what negative effect the rounding approach would have on mechanically-locking mice.

I explained in my last post. A mouse that actually does generates 120 per line would scroll prematurely with rounding. I am not willing to penalise better behaved mice to accommodate worse behaved ones.

>
> This article https://who-t.blogspot.com/2021/08/libinput-and-high-resolution-wheel.html explains that libinput chose 120 due to its nice mathematical properties of being highly divisible, which allows for flexibility in device compatibility. Ultimately, this comes down to the question of how to properly map higher precision inputs onto fixed-sized/chunky lines.

Eh, no. You mean that's why Microsoft chose it, for the very first wheel
mouse the Microsoft Intellimouse.
https://devblogs.microsoft.com/oldnewthing/20130123-00/?p=5473

>
> Would you accept a patch to add a `wheel_scroll_round_partial_detents` config option (default off I guess) to enable rounding?

No, I'm afraid this is too specific a workaround, sorry. I am not going
to have kitty config for brand of input device specific workarounds.

francislavoie:
> It's obvious your mouse wants to be a high precision scroller but has jury rigged a hack on top to get back quantum scrolling. That hack does not work well.

I don't agree, it is a high precision scroller, as evidenced by scrolling in a browser and having near pixel precise scrolling when I unlock the wheel. It's not a hack, it's just... how magnets work. And it does work very well, in every application I've used in the past 7 years that I've owned one, except for terminals like Kitty (because it inherently requires more chunky scrolling).

> I explained in my last post. A mouse that actually does generates 120 per line would scroll prematurely with rounding.

"prematurely" is not premature IMO because the user showed the intent of wanting to scroll in that direction and waiting when they've already done most of the input to get there seems counterproductive to me. But either way, rounding in that case would do nothing _because_ it already generates 120 per line, so 120 would be rounded to 120 and it's a no-op.

kovidgoyal:
Umm what? You are saying that it works well *as a high precision scroller*. I agree, I am saying its buggy when used as something *other than a high precision scroller*. 

Say I have an actual non buggy mouse, lets call it GoodMouse TM. Now GoodMouse allows me to scroll by fractions of a detent, when I move the mouse by quarter a detent it reports a value of 30. Then if I move it further by a quarter of a detent it reports another 30. Now if I move it a twelfth of a detent it reports 10. For a total of 70, but given that I have not yet moved a full detent, it *does not* click or otherwise indicate a detent has been scrolled.  At this point I have *not* indicated I want to scroll a full line, Indeed I have very specifically indicated I want to scroll a fraction of a line. You are proposing that my specific intention which cannot be represented (currently) in a terminal program become a line scroll, an unexpected action versus no action which is expected given I know my terminal program can only scroll by whole lines and i have explicitly not scrolled one detent. 

Now your particular mouse is buggy and therefore when you scroll one full detent as defined by its magnets clicking what it actually reports you have scrolled is in a wide range. Therefore there is a mismatch, you thought you scrolled one detent but kitty will not scroll a line. This mismatch is caused by your mouse's buggy behavior. If I were you I would get a better mouse or if you like this mouse then use it only as a high precision scroller turn off the magnets that generate the click sounds, and there is no longer any mismatch between your expectations and what kitty does because your expectations are not being falsely set by a buggy detent indicator.

Issue #1077, opened 2018-10-17
OSX ctrl cmd swapped - way to swap in kitty too?


viennaa:
Hi,

since I have ctrl and cmd swapped in the whole system I wonder, if there is the possibility to swap them in the kitty config as well (since I would need to use CMD+C in order to issue a Ctrl+C console behaviour). Haven't found anything so far. 

I tried remapping them directly with the glfw keys (swapped 0x2 and 0x8 here) without success.
on_key_input: glfw key: 347 native_code: 0x36 action: PRESS mods: 0x2 text: ''
on_key_input: glfw key: 341 native_code: 0x3b action: PRESS mods: 0x8 text: ''

Is this somehow possible
viennaa

kovidgoyal:
You cannot remap modifiers in kitty, I'm afraid. The closest you can
come is mapping entire keypresses, which you will have to do for all
keypresses to have that affect. It should also be a fairly simple thing
to atch kitty to do what you need, see keys.c in the kitty source code

sergei-dyshel:
FYI, I've made a simple patch to address this problem, see 056c124f

flowchartsman:
@kovidgoyal, is there any way you would be able to take a look at the latest patch from @sergei-dyshel and see if there's a way to make this configurable? It does exactly what I want, and it would be amazing to offer it to everyone. https://github.com/kovidgoyal/kitty/commit/056c124f0fa3dc4872791998bba4b3fd5ac76e0b

kovidgoyal:
Not something I am interested in personally, sorry.
It might help if you explained the need for it.

sergei-dyshel:
Here's my vision of why one would like to swap `Cmd` and `Ctrl` on Mac: 

On Linux (and Window) `Ctrl` and to lesser extent `Alt` are used as *application* modifiers (i.e. most application keybindings use combination of `Ctrl` and `Alt`), so one could call `Ctrl` the *primary* modifier and `Alt` a *secondary* one. On the other side `Windows`(aka `Super`) is mostly used as *global* modifier, i.e. for OS-level tasks, window management, global shortcuts.

On MacOS the situation is more complex, since `Ctrl` is also used for application-level bindings but there still seem to be a concept of `Cmd` being the *primary* modifier, `Option` (aka `Alt`) being the *secondary* modifier and `Ctrl` is used for *global* tasks but also for application.

[...]

PR #10307, opened 2026-08-02
Add remap_modifier to present one modifier key to kitty as another

bjohas:
Implements the option discussed in #1077.

```conf
remap_modifier <from> <to>
```

Repeatable, off by default. The declarations compose as a single simultaneous
permutation applied once per event, so

```conf
remap_modifier ctrl super
remap_modifier super ctrl
```

is a swap rather than two rewrites collapsing to the identity. Only shift, alt,
ctrl, super, hyper and meta may be named; anything else is rejected with an
error. The last declaration for a given source wins.

## Design

Per the direction in #1077 it is application wide rather than limited to keys
forwarded to the child. It rewrites the event rather than the config, in
`key_callback()` before anything reads kitty.conf, so kitty's own shortcuts,
`kitty_mod` and the bytes sent to the child all see the remapped modifier, and
every mapping in kitty.conf is written in terms of the modifier a key *becomes*.
Position in kitty.conf is not significant.

Also applied in `mouse_button_callback()` and `scroll_callback()`, so `mouse_map`
agrees with `map` and a wheel event agrees with a click. A modifier key's own
press reports the destination key, so a program using the full keyboard protocol
does not see ctrl reported together with `LEFT_HYPER`.

On macOS the remap is inverted in `get_cocoa_key_equivalent()` so a menu
accelerator lands on the physical key that now produces it. That inverse can
fail — two sources onto one modifier, or hyper/meta, which
`glfwGetCocoaKeyEquivalent()` has no flags for and would silently reduce to a key
equivalent with no modifiers at all, turning `map super+q quit` into a bare `q`.
It declines to install the accelerator rather than guess; the shortcut itself
still works.

`--debug-input` reports each remapped event, including when only a modifier key's
own identity was rewritten.

Generated option files are produced by `gen config`, not hand-edited.

## Testing

`kitty_tests/options.py` covers the parser: the empty default, a single remap,
both directions of a swap, last-declaration-wins, multi-modifier destinations,
and that every rejection is reported rather than silently ignored.

Behaviour is covered outside the test suite, by driving a real window with
xdotool and asserting on the encoded bytes — run against the same binary with and
without the option, requiring the encodings to exchange:

```
baseline                             Hyper+A -> CSI 97;17u   Ctrl+A -> CSI 97;5u
remap_modifier ctrl<->hyper          Hyper+A -> CSI 97;5u    Ctrl+A -> CSI 97;17u
remap_modifier ctrl->super->hyper->ctrl                      Ctrl+A -> CSI 97;9u
```

The 3-cycle is the case that distinguishes a simultaneous permutation from a
sequential one, which would collapse to the identity and yield `CSI 97;5u`. That
harness needs X11 and xdotool, so it is not included here.

Not tested: macOS. I have no Mac, so the `get_cocoa_key_equivalent()` path
compiles clean in isolation but has never run — you offered in #1077 to test that
side.


kovidgoyal:
Works OK on macOS at least for basic things like swapping ctrl and
command.

However, I do have some concerns especially with remapping shift.

Because you are doing the mapping at the kitty level rather than at the
glfw level, there is the question of compose keys and key events that
produce text.

One can argue, not without merit, that compose keys and text production
occur before remapping and so should be left alone, however, say you
swap shift and ctrl using this mechanism. Like this

kitty -o 'remap_modifier shift ctrl' -o 'remap_modifier ctrl shift'

now pressing shift+A types capital A pressing ctrl+A types small a.
Similarly if you run

kitten show-key -m kitty

you see shifted key reports and associated text behaving strangely.


IMO it might make more sense to do this remapping at the glfw level
before feeding the key event into the system to get its associated text.
So on linux it would be in xkb_glfw.c when the raw event is received
from the OS, it would first be remapped and then fed into libxkbcommon
and then sent on to application code. But then identifying which key it is for the remapping becomes much harder as you cant lean on libxkbcommon to do it.

On macOS, I believe it would happen in the key_down, flags_changed and key_up blocks in
cocoa_init.m but there is the complication of the global menubar
actions.

Remapping modifiers is more complex than it seems!

bjohas:
You are right, and the shift case is worse than a cosmetic inconsistency — with
`remap_modifier shift ctrl` plus `remap_modifier ctrl shift` I get

```
physical Shift+A   ->  CSI 97;5;65     mods ctrl, text "A"
physical Ctrl+A    ->  CSI 97;2        mods shift, no shifted key, no text
```

so the modifier bits and the text contradict each other, exactly as you describe.

I have moved it below libxkbcommon. The masks are permuted in
`glfw_xkb_update_modifiers()` before `xkb_state_update_mask()`, so the keysym,
the shifted key, compose and the text are all produced *from* the remapped state
and cannot disagree with it. Same repro now:

```
physical Shift+A   ->  mods ctrl, no text, sends 0x1
physical Ctrl+A    ->  mods shift, shifted key 97:65, text "A"
```

Identifying the key turned out not to be needed: this permutes the modifier
*mask*, not keycodes, and `glfw_xkb_update_modifiers()` already receives the raw
masks and already resolves `shiftIdx`/`controlIdx`/`hyperIdx` and friends. It is
about 20 lines. Since the GLFW modifiers are themselves derived from that xkb
state, the old kitty-level remap would now double-apply, so it is compiled out
except on macOS. `x11_window.c` needed the same treatment in `translateState()`,
which only ever sees button and DND events, because X11 pointer events do not go
through xkb; Wayland pointer events already carry xkb-derived modifiers.

Checked that this does not disturb level shifting: with `remap_modifier ctrl
hyper` active, AltGr+e still produces the identical keysym, modifiers and text as
with no remap. Only the named modifiers are touched, and `clean_state` is left
alone.

macOS is done the same way, in the third commit: `remap_cocoa_flags()` at the top
of `keyDown`/`keyUp`/`flagsChanged`, so the flags reaching `UCKeyTranslate()` and
`interpretKeyEvents:` are already remapped. Without it `Option` had the same
desync there that shift had on Linux, since it produces text natively unless
`macos_option_as_alt` is set. Two things I could not avoid: hyper and meta have
no Cocoa flag, so a remap onto them is left unapplied rather than silently
dropping the modifier; and `interpretKeyEvents:` is only approximately corrected,
because the event's `characters` strings were computed by the OS from the
physical flags and cannot be recomputed — the `UCKeyTranslate()` path, normally
the one taken, is exact.

I have no Mac, so that path is written but has never run. What I could check is
that the flag translation compiles and behaves correctly in isolation under the
same flags, including that unrelated flags and CapsLock survive. You offered to
test on yours, which is the part I cannot do.

One thing worth knowing separately: on Wayland `hyper` and `meta` are only
detected when `KITTY_WAYLAND_DETECT_MODIFIERS` is set, so remapping to or from
them does nothing there regardless of any of this. I have added a line to the
option docs.

I have force-pushed rather than stacking the change, since it partly reverts the
previous approach and reads better as one design. The branch is three commits:
the option, its tests, and the macOS part.

kovidgoyal:
That's certainly better, however running:

kitty -o 'remap_modifier ctrl super' -o 'remap_modifier super ctrl' --debug-input kitten show-key -m kitty

we see that pressing ctrl gets reported as ctrl both in the
--debug-input output and more importantly in the show-key output.

Indeed the press event is reported as ctrl and the corresponding release
event as super which is obviously not acceptable.

One minor nit, move glfw_is_loaded declaration after imports.

And yes feel free to rebase with your changes into a sensible commit
structure, I prefer that approach myself.

Once it is fully working on Linux I will test it on macOS

bjohas:
Fixed. `update_modifier_state_on_modifier_key_event()` was being handed the
physical modifier while the event's mods had already been remapped inside glfw,
so the press added the physical bit to an already-remapped mask and the release
cleared a bit that was never set, leaving the remapped one behind. The identity
remap now runs first and yields the destination modifier, which is what the
fixup gets. Your repro now gives:

```
physical ctrl : press LEFT_SUPER/super,  release mods none
physical super: press LEFT_CONTROL/ctrl, release mods none
```

the same shape as unremapped kitty with the two substituted. Added an end-to-end
case for exactly this, asserting the press is `CSI 57444;9u` and the release
`CSI 57444;1:3u` — both of which fail on the old behaviour.

`glfw_is_loaded` moved below the includes. I also moved the `--debug-input` line
for the mask into `glfw_xkb_update_modifiers()`, where the remap actually
happens: from kitty's side it was printing `mods: ctrl -> mods: ctrl`, since by
then only the key identity was still changing.

Rebased into two commits, the option and its tests.

One thing worth flagging: the branch is still on 9f5cc8a48 rather than current
master, because building master here now needs `slangc`, which I do not have.
Everything above is verified on that base. Happy to rebase forward if you would
rather, though I would not then be able to build what I am pushing.

kovidgoyal:
if you follow the instructions in
https://sw.kovidgoyal.net/kitty/build/

using dev.sh to setup your dev environment you will get slangc
automatically installed (at least you should hasnt been tested yet since
its so new).

kovidgoyal:
and regardless please do rebase on master as there are various conflicts that need fixing and I dont want to spend my time on that.

bjohas:
Follow-up bug fix in #10351, found while doing the rebase you asked for — it landed a few hours after this merged, so it never reached this branch.

On X11 `glfw_xkb_compile_keymap()` seeds its state from `xkb_x11_state_new_from_device()`, i.e. the server's physical modifier mask, bypassing `glfw_xkb_update_modifiers()` where the permutation is applied. A modifier held across a keymap reload is then reported as the key on the keycap until the next `XkbStateNotify`: under `remap_modifiers ctrl:super`, Ctrl+A reaches the child as a literal `0x01`. Layout switches cause reloads, as does any client asking X for a keysym it must bind to a spare keycode.

Issue #8147, opened 2024-12-21
make `cursor_trail` work on the same command prompt only


xuhdev:
**Is your feature request related to a problem? Please describe.**

Currently `cursor_trail` always shows cursor movement once the distance is beyond a certain threshold.

However, in most cases when the cursor makes a big movement, my eyes either know where
it will land or don't care. For example,

- After typing in a long command and pressing "Enter", I know it will show up in the next command line prompt.
- When a program outputs long segments of text, I don't care where the cursor is.

However, the unnecessary animation in these scenarios is distractive.

The moments I see cursor_trail is only needed when I move cursor within a prompt
buffer. For example, I may press `Alt-F` to jump a word, and the word has an
unknown length to my eyes until I look carefully.

I tried to adjust `cursor_trail_start_threshold` and `cursor_trail_decay`, but
they don't seem to be able to address the problem.

**Describe the solution you'd like**
Add a new option `cursor_trail_scenario` to specify when to enable cursor
movement.

wwqwq2313:
Hello, we tried to solve the issue.

This is what we did:

Added a new cursor_trail_scenario option to control when the cursor trail is shown. Updated the cursor trail logic to check this option before rendering.

You can review changes in this commit: https://github.com/wwqwq2313/kovidgoyal-kitty-8147/commit/3f6821ef45bf9104c621f082b41e3d6cdd29c120.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

kovidgoyal:
@jinhwanlazy please review.

jinhwanlazy:
While I understand what you're trying to achieve, I'm not sure if we can reliably implement this because the terminal doesn't know the semantics of the movement.

Technically we could detect whether a cursor movement was triggered by a control sequence (like a newline character) versus direct cursor positioning commands, the behavior would vary in many applications and I think it would look even more annoying.

xuhdev:
Sorry I didn't write the code. It's a feature request and it's a bot that wrote the changes...

kovidgoyal:
[...]

PR #10316, opened 2026-08-04
Fix cursor trail flashing during rapid command output (#8147)

EpicurianCoder:
I restricted the cursor trail animation to movement within the same command prompt. I modified screen.c to properly identify command output states (-2 in this case) by leveraging shell integration flags (prompt_settings.val). I made sure the state remains accurate even when the initial OUTPUT_START marker scrolls off the visible screen into the history buffer. These are all current features that tie into the current design seamlessly.

I also updated cursor_trail.c to suppress the trail opacity during these -2 states (and does so without breaking applications like Vim when opened) 

I added a Python unit test in screen_test.py to verify the state transitions during rapid output.

### DIFF
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -161,6 +161,12 @@ update_cursor_trail_corners(CursorTrail *ct, ndc_coords g, monotonic_t now, OSWi
 static void
 update_cursor_trail_opacity(CursorTrail *ct, Window *w, monotonic_t now)
 {
+    if (screen_cursor_at_a_shell_prompt(WD.screen) == -2)
+    {
+        ct->opacity = 0.0f;
+        return;
+    }
+
     const bool cursor_trail_always_visible = false;
     if (cursor_trail_always_visible)
     {
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -1,5 +1,6 @@
 #include <float.h>
 #include "state.h"
+#include "screen.h"
 
 #define WD w->render_data
 #define EDGE(axis, index) ct->cursor_edge_##axis[index]
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -5,43 +5,51 @@
 #define EDGE(axis, index) ct->cursor_edge_##axis[index]
 
 inline static float
-norm(float x, float y) {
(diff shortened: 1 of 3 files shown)


kovidgoyal:
There appear to be a whole bunch of autoformat based changes. please remove those and limit your PR to actual code changes only.

EpicurianCoder:
Okay, snap, my apologies, I didn't see my IDE had done this when I saved. Will disable, make some tweaks and redo shortly.

kitty/cursor_trail.c
```diff
@@ -1,55 +1,56 @@
...
+    case CURSOR_BEAM:
+        right = left + g.dx / WD.screen->cell_size.width * OPT(cursor_beam_thickness);
+        top = bottom + g.dy;
+        break;
+    case CURSOR_UNDERLINE:
+        right = left + g.dx;
+        top = bottom + g.dy / WD.screen->cell_size.height * OPT(cursor_underline_thickness);
+        break;
+    default:
+        break;
```
OBrutus, line 50: on a project like this, a 5,000-line reformat around a 15-line fix isn't just a style nit -> it's a signal that raises the maintainer's cost to trust everything else in the diff, because now they can't tell mechanical churn from intentional change without re-deriving it.

My comment is for keeping diffs minimal is itself a correctness practice here, not an aesthetic preference. It's how a solo maintainer keeps review tractable. That's a mentoring comment, and it lands better than "please revert formatting".

kitty/screen.c
```diff
@@ -2647,85 +3376,109 @@ screen_cursor_position(Screen *self, unsigned int line, unsigned int column) {
...
+int screen_cursor_at_a_shell_prompt(const Screen *self)
+{
+    if (self->cursor->y >= self->lines || self->linebuf != self->main_linebuf || !screen_is_cursor_visible(self))
+        return -1;
+    for (index_type y = self->cursor->y + 1; y-- > 0;)
+    {
+        switch (self->linebuf->line_attrs[y].prompt_kind)
+        {
+        case OUTPUT_START:
+            return -2;
```
OBrutus, line 3393: -1 and -2 are two different negatives meaning different things, and that a named enum or a separate query would be cleaner than a second magic negative.

OBrutus (changes requested):

PR #10337, opened 2026-08-09
shaders/slang: fix pipeline parsing and multi-pass module collisions

Safari77:
Fix two issues in custom shader pipeline parsing and IR compilation:

1. Fix `is_valid_named_texture` and `is_valid_slot` always returning False
   - `typing.get_args()` returns an empty tuple when called on `StrEnum` types (`NamedTexture`, `Slot`), causing `parse_pipeline_definition` to fail with "output_texture <name> does not appear in textures".
   - Switched checks to query `._value2member_map_` directly.

2. Fix ambiguous `fragment_main` symbol collision across pipeline passes
   - Importing multiple specialized Slang modules into wrapper files for multi-pass pipelines introduced duplicate `fragment_main` declarations into the same slot scope, causing `slangc` to throw E39999 (ambiguous function call).
   - Renamed `fragment_main` inside specialized module sources to a unique per-module identifier (`fragment_main_{modname}`) and updated wrapper call sites accordingly.

Without these fixes e.g. `output_texture persist` cannot be used.
Included a demo which persist the data between frames when mouse is clicked.
[northern_lights.pipeline.txt](https://github.com/user-attachments/files/30879131/northern_lights.pipeline.txt)
[northern_lights.slang.txt](https://github.com/user-attachments/files/30879132/northern_lights.slang.txt)
northern_lights

### DIFF
--- a/kitty/shaders/slang.py
+++ b/kitty/shaders/slang.py
@@ -20,7 +20,7 @@ from enum import StrEnum, auto
 from functools import lru_cache, partial
 from itertools import chain, product
 from types import MappingProxyType
-from typing import Any, Callable, Iterable, Iterator, Literal, NamedTuple, TypedDict, TypeGuard, get_args
+from typing import Any, Callable, Iterable, Iterator, Literal, NamedTuple, TypedDict, TypeGuard
 
 from kitty.constants import read_kitty_resource, shaders_dir, slangc
 from kitty.fast_data_types import (
@@ -1090,11 +1090,11 @@ class Slot(StrEnum):
 
 
 def is_valid_named_texture(x: str) -> TypeGuard[NamedTexture]:
-    return x in get_args(NamedTexture)
+    return x in NamedTexture._value2member_map_
 
 
 def is_valid_slot(x: str) -> TypeGuard[Slot]:
-    return x in get_args(Slot)
+    return x in Slot._value2member_map_
 
 
 VALID_VAR_TYPES: frozenset[str] = frozenset({'uint', 'int', 'float', 'double', 'bool'})
@@ -1271,7 +1271,7 @@ def parse_pipeline_definition(lines: Iterable[str], pipeline_name: str, pipeline
                 case 'viewport_size':
                     current_group['viewport_size'] = unit_float(parts[1]), unit_float(parts[1 if len(parts) < 3 else 2])
                 case 'output_texture':
-                    if not is_valid_named_texture(parts[1]):
+                    if not is_valid_named_texture(parts[1]) or parts[1] not in textures:
                         raise ValueError(f'output_texture {parts[1]} does not appear in textures')
                     current_group['output_texture'] = parts[1]
                 case 'shaders':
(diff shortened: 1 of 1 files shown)


kovidgoyal:
The second issue doesnt make sense to me fragment_main() is imported into a wrapper module only and not re exported from there so there should be no name collisions. I will look at it in more detail later.

Safari77:
I would get this error without the second fix.

```
[1,169] Failed to build custom shader for slot end with error: Failed to compile end.slang with command line:
slangc -warnings-as-errors all -lang slang -I /home/safari/.cache/kitty/shaders/c/lib -I /run/user/500/tmp/tmpk5q4pxyb -I /home/safari/.cache/kitty/shaders/c/slots -I /home/safari/.cache/kitty/shaders/c/lib -module-name end -o /home/safari/.cache/kitty/shaders/c/slots/end.slang-module -- -
and stderr:
error[E39999]: ambiguous call to 'fragment_main' with arguments of type (vector<float,4>, KittyTextures, KittyCustomShaderData, vector<float,4>, float)
 --> wrapper0.slang:7:25
  |
7 | ) { return fragment_main(inp, t, d, viewport, animation_progress); }
  |                         ^ ambiguous call to 'fragment_main' with arguments of type (vector<float,4>, KittyTextures, KittyCustomShaderData, vector<float,4>, float)
--'
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
error[E39999]: ambiguous call to 'fragment_main' with arguments of type (vector<float,4>, KittyTextures, KittyCustomShaderData, vector<float,4>, float)
 --> wrapper1.slang:7:25
  |
7 | ) { return fragment_main(inp, t, d, viewport, animation_progress); }
  |                         ^ ambiguous call to 'fragment_main' with arguments of type (vector<float,4>, KittyTextures, KittyCustomShaderData, vector<float,4>, float)
--'
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
error[E39999]: ambiguous call to 'fragment_main' with arguments of type (vector<float,4>, KittyTextures, KittyCustomShaderData, vector<float,4>, float)
 --> wrapper2.slang:7:25
  |
7 | ) { return fragment_main(inp, t, d, viewport, animation_progress); }
  |                         ^ ambiguous call to 'fragment_main' with arguments of type (vector<float,4>, KittyTextures, KittyCustomShaderData, vector<float,4>, float)
--'
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
note[E40011]: candidate: public func fragment_main(float4, KittyTextures, KittyCustomShaderData, float4, float) -> float4
```

kovidgoyal:
This is now fixed, but note that there were some API changes. signature of fragment_main() is changed. the group index is now availabline in KittyTextures so you dont need PASS variable. and also specialization constants should no longer be declared extern.

If you want, you can add northern lights as a builtin custom shader. That way if I do make any more API changes before the next release (unlikely but possible) it will get migrated automatically. Feel free to send a PR for that.

PR #10340, opened 2026-08-10
add northern lights custom shader demonstrating persistent textures

Safari77:
see top of `northern_lights.slang` for flags to disable star field or other features.

### DIFF
--- a/kitty/shaders/custom/northern_lights.pipeline
+++ b/kitty/shaders/custom/northern_lights.pipeline
@@ -1,0 +1,15 @@
+textures a persist
+
+startgroup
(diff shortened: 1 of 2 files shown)


Safari77:
if I remove `var int PASS n` from `northern_lights.pipeline`, why do I get
```bash
[0,358] Failed to build custom shader for slot end with error: cannot access local variable 'inc' where it is not associated with a value
```

kovidgoyal:
its a caching issue probably the cache invalidation logic is not too robust as yet. try clearing ~/.cache/kitty

PR #10348, opened 2026-08-12
Quote choose-files shell output

M-Hassan-Raza:
Quote selected paths before emitting shell output or pasting them into a prompt. This prevents spaces and shell metacharacters from changing the selected path.

Adds Go and Python regression tests.

### DIFF
--- a/kittens/choose_files/main.go
+++ b/kittens/choose_files/main.go
@@ -854,18 +854,15 @@ func (h *Handler) set_state_from_config(conf *Config, opts *Options) (err error)
 var default_cwd string
 var use_light_colors bool
 
-func quote_if_needed(x string) string {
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Its already quoting as far as I recall, isnt it?

M-Hassan-Raza:
> Its already quoting as far as I recall, isnt it?

It quotes whitespcae and $, but not other shell metacharacters. For example, shlex.Split("a;command") returns one token, so quote_if_needed() emits it unchanged and the shell interprets the semicolon. The same happens with globs, tilde expansion, redirects and backslashes. This switches to the existing shlex.Quote() helper and adds round-trip tests.

Please lmk if I'm missing smth here or you'd like a deeper explanation or a different approach

PR #10349, opened 2026-08-12
Report parallel test failures

M-Hassan-Raza:
Report failing subtests and nonzero worker exits instead of allowing false-green runs.

Keep Python tests in-process on macOS, where post-threading forked workers crash, and add regressions for both paths.

### DIFF
--- a/kitty_tests/main.py
+++ b/kitty_tests/main.py
@@ -298,10 +298,15 @@ class PipeTestResult(unittest.TestResult):
         super().addUnexpectedSuccess(test)
         self._send({'t': 'xpass', 'id': str(test), 'e': self._elapsed()})
 
+    def addSubTest(self, test: unittest.TestCase, subtest: unittest.TestCase, err: Any) -> None:
+        super().addSubTest(test, subtest, err)
+        if err is not None:
+            record_type = 'fail' if issubclass(err[0], test.failureException) else 'error'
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Why do you think the test launcher is initializing corefoundation? As far as I know it shouldnt be. And the tests are running fine in macOS in CI and on my mac. I have merged parts of your first commit as they made sense.

M-Hassan-Raza:
@kovidgoyal I based that suspicoin on the pre-fork all_fonts_map(True) call, which goes through the CoreText font path on macOS. For context, I reproduced it on an arm64 M5 Pro MacBook Pro running macOS 26.5.1 with the bundled Python 3.14.6: all seven workers exited with SIGSEGV, and the crash reports mentioned a multithreaded fork inside CoreFoundation.
Since it works in macOS CI and on your machine, this is likely specific to my OS/Python environment, and disabling parallelism for all macOS systems would be too broad. Thanks for merging the subtest and worker-exit fixes!

kovidgoyal:
I added a commit that uses exec() parapllel workers instead on macOS see
if it works for you.

M-Hassan-Raza:
> I added a commit that uses exec() parapllel workers instead on macOS see if it works for you.

Confirmed, this works on the affected machine. I tested 00a5ca54e twice with the exec-worker path: all 395 Python and 247 Go tests passed both times, with no worker errors, SIGSEGVs, or new crash reports. Thanks for the fix!

Issue #10344, opened 2026-08-11
[RFC] Custom shaders for fun and profit!


kovidgoyal:
## What is it?

Custom shaders let you attach GPU post-processing effects to your kitty terminal. They run after kitty finishes drawing everything else, so they can transform the final image any way you like — animated backgrounds, mouse effects, cursor trails, retro screen filters, window-focus highlights, you name it.

```
# kitty.conf
custom_shaders northern-lights
```

That's it. One line and your terminal is glowing with the Aurora Borealis. Several shaders ship with kitty out of the box; you can also write your own.

Custom shaders are zero overhead unless loaded. When they are enabled kitty takes a more resource intensive rendering path, rendering in layers, so you do pay a small resource cost for using them. Complexity of the actual custom shaders used also matters, of course. But even the most demanding of them, such as northern-lights only consumes ~ 3-5% of CPU and automatically turns off when the window is idle.

---

### Animated backgrounds

| Shader | |
|--------|--|
| `inside-the-matrix` — *See the bones of reality.* | <a href="https://download.calibre-ebook.com/videos/inside-the-matrix.webm" target="_blank">▶ watch</a> |
| `northern-lights` — *The ethereal Aurora Borealis.* | <a href="https://download.calibre-ebook.com/videos/northern-lights.webm" target="_blank">▶ watch</a> |
| `fireworks` — *Celebrate the sheer awesomeness of your terminal.* | <a href="https://download.calibre-ebook.com/videos/fireworks.webm" target="_blank">▶ watch</a> |
| `water` — *Pretend you are cool enough to code underwater.* | <a href="https://download.calibre-ebook.com/videos/water.webm" target="_blank">▶ watch</a> |

### Cursor trails

| Shader | |
|--------|--|
| `cursor-trail-blaze` — *Set your cursor on fire as it moves around.* | <a href="https://download.calibre-ebook.com/videos/cursor-trail-blaze.webm" target="_blank">▶ watch</a> |
| `cursor-trail-lightning` — *Make your cursor shoot lightning as it moves around.* | <a href="https://download.calibre-ebook.com/videos/cursor-trail-lightning.webm" target="_blank">▶ watch</a> |

### Mouse effects

| Shader | |
|--------|--|
| `pond-ripple` — *Clicking is like throwing stones in a pond.* | <a href="https://download.calibre-ebook.com/videos/pond-ripple.webm" target="_blank">▶ watch</a> |
[...]

PR #10373, opened 2026-08-18
Add configurable cursor trail particles

yumeminami:
## What changed

- Add an optional `cursor_trail_particles` pixie-dust effect based on the cursor trail.
- Keep particle state on the CPU and render small quads with the existing trail shader.
- Add controls for particle opacity, lifetime, density, speed, and curl.
- Preserve the existing `cursor_trail_start_threshold` behavior for the built-in trail.
- Add configuration parsing coverage and generated option bindings.

The feature is disabled by default. A minimal configuration is:

```conf
cursor_trail 1
cursor_trail_particles yes
```

The default particle tuning is intentionally visible while remaining short-lived; all behavior parameters can be adjusted independently.

## Why

The existing cursor trail makes large jumps easy to follow, but it offers only a continuous shape. Optional particles provide an additional visual cue for cursor movement while leaving current behavior unchanged for users who do not enable it.

## Validation

- `/opt/homebrew/bin/go run bypy/devenv.go build`
- `dependencies/darwin-arm64/python/Python.framework/Versions/Current/bin/python3 test.py --module options`
- `git diff --check`
- Full Go suite passed outside the restricted sandbox.
- Serial full Python suite: 378 passed, 21 skipped, with two unrelated local failures: completion data is missing `set-os-window-title`, and the SSH shell-integration test receives `TERM=dumb` instead of `kitty`.

### DIFF
--- a/kitty/cursor_trail.c
+++ b/kitty/cursor_trail.c
@@ -71,6 +71,85 @@ should_skip_cursor_trail_update(CursorTrail *ct, ndc_coords g, OSWindow *os_wind
     return false;
 }
 
+static uint32_t
+next_particle_random(CursorTrail *ct) {
+    if (!ct->particle_rng_inc) {
+        ct->particle_rng_state = 0x853C49E6748FEA9Bull;
+        ct->particle_rng_inc = (0xDA3E39CB94B95BDBull << 1u) | 1u;
+    }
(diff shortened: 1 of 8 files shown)


kovidgoyal:
This belongs as a custom shader. See
https://github.com/kovidgoyal/kitty/issues/10344

PR #10380, opened 2026-08-21
Fix clipping of scaled image placements at scroll margins

### LINKED ISSUE #10377: Scaled image placements are distorted instead of clipped when scrolling within margins
**Describe the bug**

The graphics protocol docs say that when page margins are defined, images scrolled against a margin "must be clipped". Kitty distorts scaled (`r=/c=`) placements instead of clipping them (observed on 0.48.2). Native-size placements clip correctly.

**To Reproduce**

Run the script below (observed on kitty 0.48.2). It shows an image with four color bands, scaled to one band per row, inside a scroll region, then scrolls the region. Expected: at the top margin, one full band is clipped off per scroll (red first), gone after 4 scrolls. 

```sh
#!/usr/bin/env bash
# 32x256px RGB image, four 64px color bands, displayed scaled into
# 4 columns x 4 rows (one band per row) inside a DECSTBM scroll
# region, then scrolled against the top margin.
set -u
esc=$'\033'

# red, yellow, green, blue bands; 2048 pixels each
bands() {
  printf '\377\073\060%.0s' $(seq 2048)
  printf '\377\314\000%.0s' $(seq 2048)
  printf '\064\307\131%.0s' $(seq 2048)
  printf '\012\204\377%.0s' $(seq 2048)
}
b64=$(bands | base64 | tr -d '\n')

printf '%s[2J%s[H' "$esc" "$esc"

# transmit in chunks
ctrl="q=2,a=t,f=24,s=32,v=256,i=1"
while [ -n "$b64" ]; do
  chunk=${b64:0:4096}
  b64=${b64:4096}
  [ -n "$b64" ] && m=1 || m=0
  printf '%s_G%s,m=%d;%s%s\\' "$esc" "$ctrl" "$m" "$chunk" "$esc"
  ctrl="q=2"
done

# text lines, then the image at rows 7-10
for row in 1 2 3 4 5 6 7 8 9 10; do
  printf '%s[%d;1Hpane text line %d' "$esc" "$row" "$row"
done
printf '%s[7;30H%s_Gq=2,a=p,i=1,p=1,c=4,r=4,C=1%s\\' "$esc" "$esc" "$esc"

# scroll region rows 3-10, scroll 9 times
printf '%s[3;10r%s[10;1H' "$esc" "$esc"
for s in 1 2 3 4 5 6 7 8 9; do
  printf 'scrolled line %d\n' "$s"
  sleep 0.5
done
printf '%s[r%s[12;1H' "$esc" "$esc"
```

**Screenshots**

https://github.com/user-attachments/assets/fe96026e-aaf0-4ba9-9b10-016b212c3752

liguangsheng:
Fixes #10377.

Scaled placements were clipped using the terminal cell height directly as a source-image pixel count. This caused too little source data to be removed, while the destination rectangle retained its original height, stretching the remaining image.

This change:

- converts each clipped terminal row to the corresponding number of source pixels
- renders explicitly sized placements using their clipped effective row count
- adds coverage for repeatedly clipping a scaled placement at a scroll margin

Tested with:

```text
./test.py --module graphics
```

### DIFF
--- a/kitty/graphics.c
+++ b/kitty/graphics.c
@@ -932,6 +932,17 @@ update_dest_rect(ImageRef *ref, uint32_t num_cols, uint32_t num_rows, CellPixelS
     ref->effective_num_cols = num_cols;
 }
 
+static float
+source_pixels_per_cell_y(const ImageRef *ref, CellPixelSize cell) {
+    if (ref->num_rows) {
+        return ref->src_height / MAX(1u, ref->effective_num_rows);
+    }
+    if (ref->num_cols) {
+        return cell.height * ref->src_width / (ref->num_cols * (float)cell.width);
+    }
+    return cell.height;
+}
+
 static ImageRef *
 create_ref(Image *img, ImageRef *clone_from) {
     ImageRef *ans = calloc(1, sizeof(ImageRef));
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Ah I just committed a fix myself. Feel free to review it and send another PR if there is something I missed.

Issue #10382, opened 2026-08-21
Pre-edit text is drawn in reverse video, and there is no way to change it


bjohas:
While using IBus in kitty I noticed the pre-edit ("undecided") text is rendered
in reverse video. Other applications on the same desktop underline it,
following the attributes IBus supplies in `IBusAttrList`. In running text the
inversion is considerably more distracting than an underline — it's a solid
block moving through the line as you type.

I can't find any way to change it. There is nothing for pre-edit styling in
`kitty/options/definition.py`, and the styling is a single hard-coded line in
`screen_draw_overlay_line()`:

```c
// kitty/screen.c
self->cursor->sgr.reverse ^= true;
```

That dates back to 5b1eaa741, the original "show the pre-edit text while the IME
is in progress" commit, so I assume it was a first-cut choice rather than a
considered one — but tell me if there's a reason for it I'm missing.

### The question behind this

That attribute list describes how to decorate the pre-edit run — underline,
foreground, background, each with a scope. It is visible in the very message
glfw parses; the docstring in `glfw/ibus_glfw.c` shows it:

```
variant  struct {
    string "IBusText"
    array [ ]
    string "ash "
    variant struct {
        string "IBusAttrList"     <-- attributes
        array [ ]
        array [ ]
    }
}
```

but `get_ibus_text_from_message()` extracts only the string and discards the
attribute list, so by the time `screen.c` draws it there is nothing left to
honour and a blanket style is the only option.

For CJK conversion that loses a distinction that carries meaning: engines
typically mark the segment currently being converted differently from the rest
of the pre-edit run. Collapsing all of it into one inverted block hides which
segment is active.

### What I'd propose

Honouring the attributes properly is the "right" fix but a much larger job, and
I don't want to assume you want that complexity for a feature this peripheral.
The cheap version would be an option:

```conf
ime_preedit_style reverse    # current behaviour, default
ime_preedit_style underline  # straight underline
ime_preedit_style curly      # etc.
```

`Cursor.sgr` already carries `uint8_t decoration` alongside `reverse`, and kitty
already renders all six underline styles, so the draw site becomes a lookup
instead of a toggle — no new rendering work.

Happy to implement whichever you prefer, or to drop it if you think the current
behaviour is right for a terminal specifically. I can see the argument that
underline collides with genuinely underlined output in a way it doesn't in a GUI
text field, and that pre-edit displacing real line content is worth making
unmistakable.

Two questions, then:

[...]

PR #10386, opened 2026-08-22
Style IME pre-edit text with italics and a dashed underline

bjohas:
Per your instruction on #10382: pre-edit text is now italic with a dashed
underline in the highlight color, rather than reverse video.

One thing to check, since your wording could be read either way: I took "color
being the highlight color" to mean the *underline's* colour, so it sets
`decoration_fg` and the text keeps its normal foreground. That matches the
`url_style`/`url_color` pair, where `url_color` is the decoration colour, and it
avoids drawing the text itself in `selection_background` (`#fffacd` by default),
which would be close to invisible on a light theme. If you meant the text
colour, it is a one-line change to `sgr.fg`.

The highlight color resolves through `colorprofile_to_color_with_fallback()`
from `highlight_bg`, falling back to `default_fg` when selection colours are
unset or special.

The previous code toggled `reverse` with XOR, so it restored itself. The three
new attributes are saved and put back explicitly, next to the wrap,
cursor-visibility and insert-replace modes already saved there.

Test added, asserting the rendered cells are italic, not reversed, decoration 5,
and `decoration_fg` packed as RGB with the selection background; plus that the
styling does not leak into the screen's own cursor. Verified against the built
binary:

```
cell 0: italic=True reverse=False decoration=5 decoration_fg=0xfffacd02 (type=2, rgb=0xfffacd)
```

Full suite passes here apart from two failures unrelated to this change and
present on master: `open_actions` argument splitting, and `glfw-wayland.so`,
which this machine cannot build (its wayland-protocols predates
`cursor-shape-v1`).

### DIFF
--- a/kitty/screen.c
+++ b/kitty/screen.c
@@ -4785,7 +4785,24 @@ screen_draw_overlay_line(Screen *self) {
     self->modes.mIRM = false;
     Cursor *orig_cursor = self->cursor;
     self->cursor = &(self->overlay_line.original_line.cursor);
-    self->cursor->sgr.reverse ^= true;
+    // Mark the pre-edit text as distinct from committed text: italic, with a
+    // dashed underline in the highlight color. Underline rather than reverse
+    // video matches what other terminals do (VTE and foot both underline), and
+    // dashed avoids colliding with the styles applications already use: curly
+    // for spell checking and straight for hyperlinks. Saved and restored around
+    // the draw, like the modes above.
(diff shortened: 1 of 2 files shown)


bjohas:
Thank you!

(Apologies for the formatting miss, wrong clang-format version locally. Will use ./autoformat in future.)

kovidgoyal:
No worries!

PR #10394, opened 2026-08-23
Revert "macos: explicitly enable modern window corners on macOS 26"

zzhaolei:
> This reverts commit b16221a1d3fd62db01cdc70df647b43168495310.

This reverts [#9890](https://github.com/kovidgoyal/kitty/pull/9890) because manually setting `CALayer.cornerCurve` is not the correct way to adopt the new macOS 26 window corners. AppKit should manage them automatically when kitty is built with the macOS 26 SDK.

It appears that the official releases do not use the macOS artifacts built by GitHub Actions. Are they built by a separate service? If so, its build SDK needs to be upgraded to the macOS 26 SDK for the new AppKit-managed window corners to take effect. Could it be upgraded?

Sorry for the unnecessary change and the extra maintenance work caused by the original PR.

### DIFF
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -3910,18 +3910,6 @@ apply_titlebar_color_settings(_GLFWwindow *window) {
 #undef tc
 }
 
-static void
-apply_window_corner_curve(_GLFWwindow *window) {
-    if (!window || !window->decorated) return;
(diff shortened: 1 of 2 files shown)


kovidgoyal:
See also: https://github.com/kovidgoyal/kitty/commit/a92348e460bc41cdbc21bddb8846ab9b9d4e7312

PR #10403, opened 2026-08-25
macOS: avoid stale mouse positions after focus changes

SuperGoodGame:
## Summary

- on macOS 26, refresh only kitty's internal hit-test coordinates immediately before valid mouse button dispatch and focused scroll dispatch
- use `glfwGetCursorPos()` directly, without calling `_glfwInputCursorPos()` or emitting a cursor callback/synthetic move
- add no extra query to ordinary cursor-movement callbacks; older macOS releases and non-Apple platforms are unchanged

## Problem

On macOS 26, AppKit can deliver an `NSEvent` whose `locationInWindow` is stale after an application focus change. kitty then keeps the old GLFW cursor position and can route clicks and scrolling to the wrong pane until the application is refocused.

This matches libsdl-org/SDL#15967, where the event position becomes stale while the global mouse position remains correct:
https://github.com/libsdl-org/SDL/issues/15967

I reproduced this on macOS 26.5.1 (25F80), Apple M5, with a three-pane splits layout. Debug input showed a cursor move on focus-in followed by button and scroll events that continued using the cached pane position.

## Implementation

The Cocoa implementation of `glfwGetCursorPos()` already reads `mouseLocationOutsideOfEventStream`, which remains current in the failing case. Immediately before discrete pointer-event routing, this patch copies that queried position into `OSWindow.mouse_x/y` only. It deliberately does not feed the value back through GLFW's cursor-input path, so it cannot generate a spurious kitty mouse-move event.

The existing `scroll_event()` path already queries the cursor for an unfocused OS window. The new callback-side refresh therefore applies only to focused scroll dispatch and does not duplicate that query.

## Testing

- `./dev.sh build` passes on current master
- full `./test.py`: 671/674 tests pass; the three remaining failures are environment-specific and unrelated to pointer input: one SSH fake login-shell path invokes `xcode-select`, and two Consolas selection cases expect bold/italic faces not installed on this system
- the equivalent narrow patch builds and launches as kitty 0.48.2 on the affected macOS 26.5.1 machine

No automated regression test is included because the failure depends on AppKit emitting a stale system event across a real application focus transition.

### DIFF
--- a/kitty/glfw.c
+++ b/kitty/glfw.c
@@ -636,6 +636,25 @@ cursor_enter_callback(GLFWwindow *w, int entered) {
     global_state.callback_os_window = NULL;
 }
 
+static bool
+refresh_mouse_position_for_hit_test(GLFWwindow *w, OSWindow *window) {
+#ifdef __APPLE__
+    if (__builtin_available(macOS 26.0, *)) {
+        // AppKit can report stale mouseMoved locations after application focus changes.
+        // Query the current position for discrete pointer events without emitting a move event.
+        double x, y;
+        glfwGetCursorPos(w, &x, &y);
+        window->mouse_x = x * window->viewport_x_ratio;
+        window->mouse_y = y * window->viewport_y_ratio;
+        return true;
+    }
+#else
+    (void)w;
+    (void)window;
+#endif
+    return false;
+}
+
 static void
 mouse_button_callback(GLFWwindow *w, int button, int action, int mods) {
     if (!set_callback_window(w)) return;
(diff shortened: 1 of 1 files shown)


kovidgoyal:
various concerns with this:
1) performance - querying the system every mouse move is inefficient
2) correctness - this can generate spurious mouse move events for kitty since you are calling glfwcursorinput in all the callbacks
3) neccessity - if this is an appkit bug then why not just wait for apple to fix it?

SuperGoodGame:
Thanks, those are fair concerns.

1/2: Agreed. The current patch is too broad. I can remove the `mouseMoved` workaround and avoid calling `_glfwInputCursorPos` from the button/scroll callbacks entirely. A narrower fix would refresh only kitty's hit-test coordinates via `glfwGetCursorPos()` immediately before button and scroll dispatch on macOS 26. The Cocoa implementation already obtains that value from `mouseLocationOutsideOfEventStream`. This would emit no cursor callback or synthetic move, and would add no work to ordinary mouse motion.

3: Waiting for Apple is certainly an option. My motivation for a workaround is that this is reproducible on the released macOS 26.5.1 and causes clicks and scrolling to be delivered to a different pane, with no reliable user workaround other than refocusing the application. SDL has independently observed the same AppKit behavior and found the out-of-event-stream/global query remains accurate. I would keep any workaround version-gated to macOS 26 and remove it once Apple fixes the underlying issue. If you would prefer not to carry an OS workaround at all, I understand.

Would the narrower hit-test-only refresh address the performance and correctness concerns? If so, I will update the PR accordingly.

SuperGoodGame:
Updated in `9dc4c0e` with the narrower approach.

The revised patch no longer touches `mouseMoved` or any Cocoa callback, and it never calls `_glfwInputCursorPos()`. It calls `glfwGetCursorPos()` only before valid mouse-button dispatch and focused scroll dispatch on macOS 26, then updates only kitty's `OSWindow.mouse_x/y` hit-test coordinates. Consequently it emits no cursor callback or synthetic mouse-move event and adds no work to ordinary mouse motion.

I also avoided an extra query for unfocused scrolling because the existing `scroll_event()` path already handles that case. Older macOS versions and non-Apple platforms retain the existing behavior.

The final diff is now limited to `kitty/glfw.c` (27 insertions, 4 deletions). `./dev.sh build` passes; 671/674 tests pass, with the same three machine-specific failures (the fake login-shell/Xcode lookup and two missing Consolas face cases).

PR #10412, opened 2026-08-27
Add an independent DnD copy mode

M-Hassan-Raza:
Local Copy drops use hard links when possible for performance. Add an opt-in independent mode for users who do not want destination writes to affect source files.

Move remains hard-link-first, and Copy keeps its current behavior by default.

### DIFF
--- a/kittens/dnd/drop.go
+++ b/kittens/dnd/drop.go
@@ -130,7 +130,8 @@ func uniqify_child_names(names []string, is_case_sensitive_filesystem bool) []st
 	return names
 }
 
-func do_local_copy(ctx context.Context, dest_dir *os.File, uri_list []string) (err error) {
+func do_local_copy(ctx context.Context, dest_dir *os.File, uri_list []string, action int, copy_mode string) (err error) {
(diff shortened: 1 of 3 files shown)


kovidgoyal:
IIRC I chose this deliberately for performance. in the case of move it should always use hard links. In the case of copy its debatable, perhaps an option controlling it.

M-Hassan-Raza:
> IIRC I chose this deliberately for performance. in the case of move it should always use hard links. In the case of copy its debatable, perhaps an option controlling it.

That makes sense. For Move, the hard-link-first path is a decent optimization since the source is removed only after a successful drop. 
Ig to mitigate the tradeoff making independent copies opt-in while preserving the current default seems reasonable. I can pass the selected action into the local-copy path, keep hard links for Move, and gate independent Copy behavior behind an option. 

Would that direction work for you?

kovidgoyal:
yes thats fine

M-Hassan-Raza:
> yes thats fine

Move and default Copy remain hard-link-first, with independent Copy now opt-in via `--copy-mode=independent` and covered across all action/mode combinations. Hopefully this is now in a mergeable state

Issue #8617, opened 2025-05-09
Feature Request: Rounded active window border


Parilia:
**Is your feature request related to a problem? Please describe.**
Not a problem per say, it just annoys me.

**Describe the solution you'd like**
Add the ability to have a togglable rounded corner on the acting and inactive window border. As in the one rendered in kitty not the actual kitty window.

A set one would be fine or the ability to set a specified roundness

**Additional context**
This border:
https://img.empire-rises.xyz/image/a6735a0e-941a-4586-beec-f4102918d1f7.png

kovidgoyal:
Not something I wish to spend time on, but a PR is welcome. Note that
drawing a rounded border means you cut off a part of the each corner
cell.

PR #10421, opened 2026-08-29
Add configurable rounded window borders

NightWatcher314:
## Summary

- add `window_border_radius` with px/pt units and a zero default
- render full pane borders with the existing rounded-rectangle shader while preserving rectangular edge hit regions
- leave minimal borders unchanged and support transparent backgrounds

## Testing

- `ruff check --fix`
- `make debug`
- `./test.py --module layout window_border_radius_geometry`
- `./test.py conf_parsing`
- release `kitty.app` build
- macOS visual smoke with 0.75 opacity and a split created with Cmd+Enter; 20 captured frames remained stable
- full `./test.py`: 689/690 passed; `test_font_selection` requires Source Code Pro, which is not installed on this host

Refs #8617

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -200,6 +200,8 @@ Detailed list of changes
 0.49.0 [future]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+- Add :opt:`window_border_radius` for rounded window borders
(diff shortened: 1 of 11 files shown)


NightWatcher314:
图片
here is the preview

PR #10422, opened 2026-08-29
Graphics protocol: Accept shm names without a leading slash

j-c-m:
POSIX portable shm names start with '/'. Darwin shm_open requires that form, glibc accepts both, so clients that omit the slash fail on macOS.


j-c-m:
So far I've only found mpv in non compliance, hopefully they patch.

j-c-m:
Mpv has merged the fix to be in compliance with the spec.

https://github.com/mpv-player/mpv/pull/18424

PR #10431, opened 2026-09-01
macOS: fulfill file promises lazily so pasted drop paths survive

### LINKED ISSUE #10430: macOS: dragged file-promise paths die within milliseconds (pasted path unresolvable)
## How I noticed

A Day of using kitty as my daily terminal: dragged a fresh screenshot thumbnail in to hand the path to an AI coding agent running in the shell (the pi coding-agent harness), and the agent's read failed with file-not-found. Assumed user error, tried again, sometimes it worked. What actually varied was only when the consumer read the path: reads inside the ~90 ms before promise fulfillment succeeded, everything after hit the already-deleted file. The 4/4 kernel traces later showed the mechanism was deterministic all along; the "sometimes" was pure read timing. This is the same consumer class as #10078 (AI coding assistants consuming dragged paths), where the encoding half was fixed; this is the deletion half.

## Summary

On macOS, dragging a floating screenshot thumbnail (or any file-promise drag) into kitty pastes a path that is dead within milliseconds: fulfilling the promise moves the provider's staging file away, and kitty then deletes its own received copy at end-of-drop. Every consumer of the pasted path gets ENOENT. This is the second landmine in the May 2026 file-promise code path: #10078 reported the percent-encoding half (fixed via #10054); this is the file-deletion half, which survives that fix.

## Environment

- kitty 0.48.2 (installed release) and current master (41095672e): identical behavior
- macOS 15.7.7 (24G720), Intel
- default config; the drag handling lives in glfw/cocoa_window.m and is config-independent

## To Reproduce

1. Cmd+Shift+4, select a region. The floating thumbnail appears.
2. Before it fades, drag it into a kitty window at a shell prompt.
3. A path like `/var/folders/.../TemporaryItems/NSIRD_screencaptureui_XXXX/Screenshot ....png` is pasted.
4. `cat` that path seconds later: `No such file or directory`.

Reproduced 4/4 times with kernel tracing. The same drops into native Terminal.app keep the file alive and readable every time.

## The juice: what five fs_usage traces actually caught

Recorded with `sudo fs_usage -w -f filesys | grep --line-buffered -Ei 'NSIRD|Screenshot|screencaptureui' | tee trace.txt`. The broken drops all show the same seven-event signature (example from a real run, kitty pid 541216):

```
19:06:57.473892  mkdir   NSIRD_screencaptureui_q91aPV      screencaptureui  (staging created)
19:06:57.485315  rename  NSIRD_screencaptureui_q91aPV/...  screencaptureui  (capture finalized)
19:06:57.744063  mkdir   NSIRD_kitty_fP8687                kitty            (promise dir)
19:06:57.835404  rename  NSIRD_screencaptureui_q91aPV/...  screencaptureui  (promise fulfillment MOVES the file into kitty's dir)
19:06:57.837571  open    NSIRD_kitty_fP8687/...            kitty
19:06:57.838013  unlink  NSIRD_kitty_fP8687/...            kitty            (2.6 ms after the move)
19:06:57.838091  rmdir   NSIRD_kitty_fP8687                kitty
```

The numbers across three traced drops:

```
trace   staging mkdir   kitty mkdir    promise move   kitty unlink   | KILL   total
2       16:41:55.187    16:41:55.506   16:41:55.593   16:41:55.597   | 4.3ms  410ms
3       19:06:57.473    19:06:57.744   19:06:57.835   19:06:57.838   | 2.6ms  364ms
4       19:11:00.520    19:11:00.840   19:11:00.919   19:11:00.921   | 2.3ms  402ms
```

Same pid creates the promise dir and unlinks its contents (kitty.541216 in all three). After every drop no usable copy exists anywhere:

- the pasted staging path is stale: the provider renames the file out of it when the promise is fulfilled
- kitty's received copy is unlinked by kitty at end-of-drop

Control trace, identical drag into Terminal.app: staging created, capture finalized, and then nothing. No fulfillment, no move, no unlink. The file stays readable for minutes. Two kernel events versus kitty's seven.

## Root cause (glfw/cocoa_window.m @ master)

1. `performDragOperation` publishes `text/uri-list` from the pasteboard NSURLs (the staging path), then immediately calls `receivePromisedFilesAtDestination` (line 1730 area). Fulfillment is a move, not a copy: the staging path just published is invalidated by kitty's own request.
2. `_glfwPlatformEndDrop` (line 1867) runs `free_drop_data`, whose `free_in_progress_drop_data` (line 1449) removes the whole promise temp dir at end-of-drop, destroying the only surviving copy.

## Fix direction

Fulfill promises lazily: keep the receivers at drop time and only call `receivePromisedFilesAtDestination` when a consumer actually requests file content (not the path-as-text uri-list). Path consumers are served from the pasteboard URLs, which stay valid because kitty never triggers the move. This mirrors Terminal.app's behavior and follows the same principle applied on #10166: do not transfer data before knowing the destination wants it. The existing `pending_requests` deferral machinery in `_glfwPlatformRequestDropData` already supports this; today it is bypassed because the uri-list is eagerly populated.

## A note on effort

#9316 was closed with "Not something I wish to spend time on, PR is welcome", so the time was spent here: five kernel traces, the timing table above, a Terminal.app control, root-cause analysis down to the source lines, a fix implemented and verified (after the patch the same drops produce zero NSIRD_kitty kernel events and the pasted paths stay readable minutes later), and a reproducible verification protocol. If reviewing is faster than re-deriving, the traces and the patch are ready to go.

saladin0x1:
Implements the fix proposed in #10430.

## Summary

`performDragOperation` published `text/uri-list` from the pasteboard URLs (the provider's staging path) and then called `receivePromisedFilesAtDestination` unconditionally at drop time. Fulfillment is a move, so the published path was invalidated by kitty's own request, and the received copy in the promise temp dir was removed at end-of-drop by `free_in_progress_drop_data`. Result: after every file-promise drag, no usable file existed anywhere and consumers of the pasted path got ENOENT.

## The change

Store the `NSFilePromiseReceiver`s at drop time (new `lazy_receivers` field) and fulfill them lazily in `_glfwPlatformRequestDropData`, on the first request for a content MIME:

- path-as-text consumers (the shell paste case) are served from the pasteboard URLs and never trigger fulfillment, so the provider keeps its staging file alive, matching Terminal.app's behavior
- content consumers (kitten dnd) still get fulfilled files: fulfillment runs on demand, results land in `path_map` as before, and the temp dir is still removed at end-of-drop after having actually served content
- the existing `pending_requests` deferral machinery handles the asynchronous window; it was previously bypassed because the uri-list was eagerly populated

Follows the same principle as the decision on #10166: do not transfer data before knowing the destination wants it.

## Verification

Before (stock 0.48.2, four traced drags) vs after (this patch, same drag):

| | before | after |
|---|---|---|
| kernel events on drop | 7 (mkdir, move, open, unlink, rmdir, ...) | 0 NSIRD_kitty events |
| received file lifetime | 2.3-4.3 ms | n/a (never received for path-only drops) |
| pasted path readable | only inside ~90 ms window | minutes later |
| signature | unique to kitty | identical to Terminal.app |

No test changes; kitty's DnD fake-event tests bypass the AppKit drag path this patch modifies.

Fixes #10430.

### DIFF
--- a/glfw/cocoa_platform.h
+++ b/glfw/cocoa_platform.h
@@ -121,6 +121,7 @@ typedef struct _GLFWDropData {
     struct {
         unsigned long long request_id;
         id temp_dir;
+        id lazy_receivers; // NSFilePromiseReceiver[] stored at drop time, fulfilled lazily on first content request
         id data_map; // map MIME to NSError or NSInputStream
         id path_map; // map MIME to NSError or NSInputStream or NSUrl
         id pending_requests;
--- a/glfw/cocoa_window.m
+++ b/glfw/cocoa_window.m
@@ -1453,6 +1453,7 @@ free_in_progress_drop_data(_GLFWwindow *window) {
         [[NSFileManager defaultManager] removeItemAtURL:d->in_progress_drop.temp_dir error:&error];
         [d->in_progress_drop.temp_dir release];
     }
+    if (d->in_progress_drop.lazy_receivers) [d->in_progress_drop.lazy_receivers release];
     if (d->in_progress_drop.data_map) [d->in_progress_drop.data_map release];
     if (d->in_progress_drop.path_map) [d->in_progress_drop.path_map release];
     if (d->in_progress_drop.pending_requests) [d->in_progress_drop.pending_requests release];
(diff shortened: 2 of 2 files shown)


kovidgoyal:
see #10430

saladin0x1:
TL;DR of the fix in #10431, for anyone finding this later.

Before (current code):

    You drag a screenshot in.
    Kitty says "gimme the file" — instantly.
      → the system MOVES the photo out of its home
        into kitty's pocket.
    Kitty tells the shell: "here's the address."
      → but that address was the photo's OLD home.
        Empty now.
    Kitty throws away its own copy ~3 ms later.

    Result: the address is a lie. The photo is gone.

After (the patch):

    You drag a screenshot in.
    Kitty says nothing. Takes nothing.
    Kitty tells the shell: "here's the address."
      → it's the photo's REAL home. Still there.
    The system keeps the photo alive for minutes,
    then cleans up on its own schedule.

    Only if a program says "give me the actual bytes"
    does kitty ask for the file — and then it cleans
    up after serving, exactly like today.

    Result: the address works. The photo is there when
    you look for it. Same as dragging into Terminal.

One line: stop grabbing a file nobody asked for, stop deleting it 3 ms later, and paste a path that actually resolves.

Issue #10404, opened 2026-08-25
Feature to add stretch/shrink split edges action


daninsky1:
This is to make easier to adjust split size.
The available solution is to use the fixed controls:
resize_window narrower
resize_window wider
resize_window taller
resize_window shorter

And that's fine.

The thing is that I want to be able to move the split, almost accordingly with the split orientation/focus that I am.
Like in the JetBrains IDE solutions with something like this:
stretch_window to_right
stretch_window to_left
stretch_window to_top
stretch_window to_bottom

Would be something like this in simple split configuration:
```
stretch_window to_right         | stretch_window to_left
┌────────────────┬────────────┐ | ┌────────────┬────────────────┐
│               >│            │ | │            │<               │
│               >│            │ | │            │<               │
│               >│            │ | │            │<               │
│               >│            │ | │            │<               │
│               >├────────────┤ | │            ├────────────────┤
│               >│            │ | │            │<               │
│               >│            │ | │            │<               │
│               >│            │ | │            │<               │
│               >│            │ | │            │<               │
└────────────────┴────────────┘ | └────────────┴────────────────┘
stretch_window to_top           | stretch_window to_bottom
┌──────────────┬──────────────┐ | ┌──────────────┬──────────────┐
│              │              │ | │              │              │
│              │              │ | │              │              │
│              │              │ | │              │              │
│              ├──────────────┤ | │              │              │
│              │⌃⌃⌃⌃⌃⌃⌃⌃⌃⌃⌃⌃⌃⌃│ | │              │⌄⌄⌄⌄⌄⌄⌄⌄⌄⌄⌄⌄⌄⌄│
│              │              │ | │              ├──────────────┤
│              │              │ | │              │              │
│              │              │ | │              │              │
│              │              │ | │              │              │
└──────────────┴──────────────┘ | └──────────────┴──────────────┘
```
So this creates a problem, what boundary should move when there are more then 1 surrounding the window. But I think kitty already solve this on resize_window 🤔.

kovidgoyal:
I dont understand what you are asking for. All four of your examples are
already covered by resize_window

Not to mention that these days you can resize windows by dragging the
borders between them.

daninsky1:
I'm sorry, my example wans't that precise.

I want to be able to move the split boundary in a absolute direction no matter which window is currently focused. Eg.:
With a vertical split I do:
ctrl+alt+right move the split to the right.

This should move the split boundary to the right no matter if I'm on the left or right window.

[...]

PR #10432, opened 2026-09-01
Add directional window border resizing in split layout

daninsky1:
Add four new values to 'resize_window' for moving a window border a fixed direction in the 'split' layout:
towards_left
towards_right
towards_top
towards_bottom.

I tried to keep the changes small and limited to the split layouts, because I'm not sure yet how the other layouts should behave with this action.

I've propagate additional boolean parameter to keep the implementation close to the existing code.

I've also added a documentation update and two tests

#10404

### DIFF
--- a/docs/layouts.rst
+++ b/docs/layouts.rst
@@ -312,7 +312,17 @@ action, for example::
    map ctrl+right resize_window wider
    map ctrl+up resize_window taller
    map ctrl+down resize_window shorter 3
-   # reset all windows in the tab to default sizes
+
+In the :ref:`Splits layout <splits_layout>`, you can also move a window border in a fixed
+direction, for example::
+
+   map ctrl+alt+left resize_window towards_left
+   map ctrl+alt+right resize_window towards_right
+   map ctrl+alt+up resize_window towards_top
+   map ctrl+alt+down resize_window towards_bottom
+
+To reset all windows in the tab to default sizes, use::
+
    map ctrl+home resize_window reset
 
 The :ac:`resize_window` action has a second optional argument to control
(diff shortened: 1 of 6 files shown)


kovidgoyal:
This needs to work with all layouts to be mergeable.

daninsky1:
We'll need to define an additional rule to make "towards_*" compatible with the resizing behavior of bias-based layouts.

I understand your concern now. I only knew about and considered the "splits" layout before, but it does become an issue with these layouts.

*the windows positioned between two borders/windows* is undefined for towards_*.

In the "splits" layout, this is solved because the pair-tree structure defines a single window border to move, and there is no bias-based distribution.
Windows at the edges are not a problem either, because the border to move is unambiguous and is compatible with the bias model.

Maybe thse possible behaviors, can solve the problem.

Alternate between the adjacent borders. This is not intended to mimic grow/shrink resizing, since that would not make sense for "towards_*". The goal would be to stay consistent with the existing resize action by moving only one split at a time and moving both of the posible borders similar to grow/shrink resizes.
Or, move only one adjacent border, probably the trailing border based on the window's position. This would be closer to the behavior of the splits layout, close, not equals!
Or, move both adjacent borders. This is effectively equivalent to moving the entire window, and in practice produces a similar result to consistently moving one side relative to the window.

kovidgoyal:
Not sure I understand your proposal. Alternating between opposite borders would basically mean moving the window not resizing it, unless I am misunderstanding. This would basically be the same as three? To me the scond proposal makes sense.

daninsky1:
Yes, the first and third are equivalent. Basically, they would move the window.

Alright, I'll try implementing the second one.

So, for windows between two splits, it will always move the trailing border based on the window position. This should work.

PR #10455, opened 2026-09-09
Shaped run cache, keep used byte counter vs walking count per put

j-c-m:
Another small tweak to the shaped run cache, keep a running byte counter to avoid the walking count on cache put. Although the walking count was noise next to the harfbuzz shaping, this is much better.


kovidgoyal:
merged, fixing the size accounting

j-c-m:
> merged, fixing the size accounting

Looks good, I was avoiding arena.h for scope, but I think it's the proper fix.

PR #10459, opened 2026-09-10
Fix idle sync to monitor - primarily macOS.

j-c-m:
I noticed this while testing the shaped run cache. Idle windows were still redrawing at vsync on macOS. It was barely noticeable on Linux/Wayland but on macOS the updateLayer/Damage loop kept redrawing the window at vsync pace even when idle, Linux was just possibly one extra frame draw. This PR now gates, and only request a frame if the PTY has fed any input. (A smaller shader fix was also required, probably a long standing unnoticed bug as it relied on vsync timing, instead of an explicit timer).

With sync_to_monitor yes, kitty requested a frame after every present (Wayland 04fd058e0, macOS abd4de731). On macOS a present can also make AppKit call updateLayer, which GLFW reports as window damage, so the window kept redrawing. After a swap, request the next frame only while the PTY is feeding.

It does not regress the FPS performance issue addressed in 04fd058e0, as it will will still request a new frame event after the render if any PTY input is read.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -261,6 +261,9 @@ Detailed list of changes
 
 - dnd kitten: Add an option to use file copies instead of hard links for copy drops (:pull:`10412`)
 
+- Fix :opt:`sync_to_monitor` presenting new frames while idle even when the
+  screen has not changed
+
 - Fix a malformed CSI escape sequence such as ``\e[?:`` corrupting the parser state so that subsequent SGR color codes are ignored (:iss:`10434`)
 
 - Graphics protocol: Fix a regression in 0.45.0 that caused the overwrite
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -264,6 +264,9 @@ Detailed list of changes
 - Fix :opt:`sync_to_monitor` presenting new frames while idle even when the
   screen has not changed
 
+- Fix always-active custom shaders such as ``water`` stopping when the window is
+  idle
+
 - Fix a malformed CSI escape sequence such as ``\e[?:`` corrupting the parser state so that subsequent SGR color codes are ignored (:iss:`10434`)
 
 - Graphics protocol: Fix a regression in 0.45.0 that caused the overwrite
--- a/kitty/child-monitor.c
+++ b/kitty/child-monitor.c
@@ -1018,7 +1018,8 @@ render_prepared_os_window(
     color_type active_window_bg,
     unsigned int num_visible_windows,
     bool all_windows_have_same_bg,
-    monotonic_t now) {
+    monotonic_t now,
+    bool input_read) {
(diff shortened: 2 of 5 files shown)


kovidgoyal:
Input being read is not the only source of needing to redraw, config reload, changing tab/layout/font size/scrolling etc all cause it as well.

j-c-m:
Those still trigger a redraw, just not an additional frame request afterwards unless there is pty feed.

kovidgoyal:
On Thu, Sep 10, 2026 at 07:03:19PM -0700, Jesse Miller wrote:
> j-c-m left a comment (kovidgoyal/kitty#10459)
>
> Those still trigger a redraw, just not an additional frame request afterwards unless there is pty feed.

I should probably read the code, but how would that work? Suppose there
is a scroll which triggers a redraw and there is no pty feed there would
be no frame request and therefore no actual draw

j-c-m:
Those are either still handled by the unchanged first call to request_frame_render in render_os_window, or for some things like a font or resize, kitty is already setting input_read to true in process_global_state, so if for some reason they don't get handled in the first request_frame_render, they'll be handled in the after swap  because input_read is still true.

kovidgoyal:
I added your second commit, but I dont see the redraw issue. Running
on macOS with the following patch

```diff
diff --git a/kitty/glfw.c b/kitty/glfw.c
index 5bf8d6347..d988c170d 100644
--- a/kitty/glfw.c
+++ b/kitty/glfw.c
@@ -460,6 +460,7 @@ dpi_change_callback(GLFWwindow *w, float x_scale UNUSED, float y_scale UNUSED) {

 static void
 refresh_callback(GLFWwindow *w) {
+    debug_rendering("refresh_callback()\n");
     if (!set_callback_window(w)) return;
     if (!global_state.callback_os_window->redraw_count) global_state.callback_os_window->redraw_count++;
     global_state.callback_os_window = NULL;
@@ -2921,6 +2922,7 @@ is_mouse_hidden(OSWindow *w) {

 void
 swap_window_buffers(OSWindow *os_window) {
+    debug_rendering("swap_window_buffers()\n");
     if (glfwAreSwapsAllowed(os_window->handle)) {
         glfwSwapBuffers(os_window->handle);
         os_window->keep_rendering_till_swap = 0;
```

and command line:

kitty --debug-rendering -o cursor_blink_interval=0

I see no continuous rendering (no prints of the swap_window_buffers())
function after the first 2-3.

This is with sync_to_monitor either enabled or disabled.

If I turn on cursor blinking then swaps happen in time with the cursor
animation. Similarly if a custom shader is running, then swaps happen in
time with that. But I dont see requesting frame becoming damage on
macOS.

Closing but feel free to discuss further

j-c-m:
I could no longer reproduce it after restarting on macOS. I'll watch if it ever comes back.

PR #10468, opened 2026-09-14
macOS: Preserve long IME commits outside key handling

xiaobai050:
Cocoa input methods can submit an entire result through `insertText:replacementRange:` outside key handling. These commits currently pass through the 512-byte key-event buffer, so text longer than 511 UTF-8 bytes is silently discarded. For example, committing 171 three-byte Chinese characters clears the preedit but delivers no text.

Dispatch event-loop commits directly using `GLFW_IME_COMMIT_TEXT`. Keep preedit clearing, control-character filtering, and per-key accumulation as before. The existing length limit during `keyDown` and `flagsChanged` is outside this fix's scope.

Validation on macOS 26.5.1, ARM64:

- A standalone local probe, kept outside this PR, passes all 24 cases on the fixed Cocoa module. Nine oversized-commit cases fail against a separately built, unmodified upstream module. Cases cover UTF-8 byte boundaries, long Chinese and emoji input, attributed strings, preedit state, successive commits, and existing key-handler behavior. The probe captures the native GLFW keyboard callback; it does not automate a physical input method or verify delivery to a terminal child.
- All 24 local cases also pass with the Cocoa backend and probe built using AddressSanitizer and UndefinedBehaviorSanitizer, with leak detection disabled and undefined behavior set to halt execution.
- After removing the test integration, `./test.py --module check_build` passes (9 passed, 1 skipped). Ruff lint, Python formatting on the restored test module, and `git diff --check` pass.
- Prior full-suite comparison found the same two font-selection assertion failures on the fixed and unmodified upstream builds.


kovidgoyal:
Remove the tests, a test that depends on source code and clang is not acceptable.

xiaobai050:
Removed the tests and their documentation in 7cbea2f. The PR now contains only the IME fix and changelog entry. I have kept the probe outside the repository for local validation.

PR #10473, opened 2026-09-14
Custom shaders: Add configurable inactive-pane tinting with on-demand redraws

xiaobai050:
Extend `focus-highlight`, the shader used by `dim-inactive-windows`, to blend inactive panes toward a configurable color. A documented gray-tint pipeline covers explicit red/green diff backgrounds, text and graphics together, while optional bounds handling preserves the tab bar and the active pane's padding. Alpha and terminal color queries are preserved. Existing pipelines retain the default black dimming and bounds behavior.

The same pipeline can use `animation_step 0` to redraw on content or shader events without periodic frames. Stop events and duration limits still repaint when a group ends, including when another static group remains active. Also clear the compiled-program cache when disabling a shader, so disabling and re-enabling the same pipeline works. These changes are separated into three commits within this PR.

Validation:

- Three focused Slang tests pass, including compiling the documented tint pipeline and the disable/re-enable regression. Ruff and formatting checks pass for the changed Python, C and Slang files.
- The local macOS Dev build passes 12 framebuffer scenarios covering truecolor/ANSI/256-color backgrounds, reverse video, text, images, padding, focus/tab changes, OS focus loss, transparency, overlays, a single pane and unchanged default dimming. The PR's focused Python tests reuse that build's native module.
- With the test window focused, the periodic control rendered 28 frames during the idle observation while the static pipeline rendered only the one requested screenshot frame. Additional real-window checks pass for static duration expiry, animated duration expiry and a static stop event, including a repaint before the next screenshot.

[Fork CI passed all eight jobs](https://github.com/xiaobai050/kitty/actions/runs/34857327273) for this exact head (`18023ea`), including Linux packaging, type checks, static binaries and macOS Brew packaging/benchmarks. The earlier upstream Linux package failure was an HTTP 504 while downloading the test font archive.


xiaobai050:
The red `Linux package` check is due to a dependency download failure: downloading `NerdFontsSymbolsOnly.tar.xz` returned HTTP 504 (Gateway Time-out) during dependency setup, before the package build or tests ran ([failed job](https://github.com/kovidgoyal/kitty/actions/runs/34857355375/job/104020246445)). This is a transient download issue, not a build/test failure caused by this change; re-running the failed job should be sufficient.

All eight CI jobs, including `Linux package`, passed in the fork for the exact same PR head, `18023eaafa05e4be979cfb579638f12ca68c553d`: https://github.com/xiaobai050/kitty/actions/runs/34857327273. This was already noted in the PR description; adding it here as well for visibility.

PR #10481, opened 2026-09-15
Cursor-trail enhancements: motion blur and antialiasing

drone1:
## Summary
I love the cursor trail feature (shoutout @jinhwanlazy!) and have been using it for years, but thought it might benefit from some optional motion blur/antialiasing.

Specifically, this PR adds temporal supersampling motion blur, which simulates a shutter opening and closing like a real camera. More technically, it integrates the moving cursor geometry over a simulated camera exposure interval.

Before &mdash; notice aliasing
image

After &mdash; aliasing + motion blur
image

The shader handles block, beam, and underline cursors in both focused and unfocused windows.

## Try it

Enable cursor trails and the bundled shader with:
```bash
cursor_trail 1
custom_shaders cursor-trail-default
```

Without the custom shader, Kitty continues to use its existing built-in cursor trail renderer. The enhanced path is a normal bundled custom shader and is only compiled and run when selected.

## Shader options

The rendering controls below are compile-time constants in `kitty/shaders/custom/cursor-trail-default.slang`. They can be edited there or overridden with `var` directives in a custom pipeline; they are not Kitty configuration options.

```bash
MOTION_BLUR true             # integrate the trail over time
ANTIALIASING true             # integrate pixel coverage over subpixels
MOTION_BLUR_SAMPLES 64        # temporal samples per frame (1..256)
ANTIALIASING_SAMPLES 16       # spatial samples per frame (1..256)
CONNECTED false               # false: original disconnected trail; true: reach target
MIN_OPACITY 0.2               # opacity floor for the blurred trail
TARGET_BLEND_START 0.1        # start radial whitening at this distance fraction
TARGET_BLEND_END 0.02         # reach full white at this distance fraction
```

The existing `cursor_trail_decay` and `cursor_trail_start_threshold` Kitty options still control animation timing and when a trail starts.

### Connected vs. Disconnected Trail
I also added a mode that _keeps the trail connected_ to the target cursor.

Disconnected (default):
image

Connected (new):
image

[The examples above use the shader defaults for temporal and spatial supersampling.]

Set `CONNECTED` to `true` in the shader or pipeline to select connected mode.

## Testing

For an in-tree build, after installing `slangc` (or setting `SLANGC` to its path), I run:

```bash
env -u PYTHONHOME -u PYTHONPATH \
  PYTHONPATH="$PWD" \
  LD_LIBRARY_PATH="/usr/lib/x86_64-linux-gnu:/usr/lib:$PWD/dependencies/linux-amd64/lib" \
  SLANGC=slangc \
  ./kitty/launcher/kitty \
    --config NONE \
    -o cursor_trail=1 \
    -o custom_shaders=cursor-trail-default \
    -o shell_integration=no-cursor \
    -o cursor_shape=block \
    /bin/zsh
```

Then I exercise large, frequent cursor jumps with:

```bash
kitty_tests/cursor_trail_test.zsh 250 0.06 block
kitty_tests/cursor_trail_test.zsh 250 0.06 beam
kitty_tests/cursor_trail_test.zsh 250 0.06 underline
```

The custom-shader demo video is generated by `kitty/shaders/custom/demo.py cursor-trail-default`.


kovidgoyal:
Some comments:

1) I suggest simply making these options in the shader not kitty.conf
In fact you can use the existing cursor-trail-default.slang having
options that control the rendering. Document this briefly in the documentation
of the cursor_trail option, and in more detail in the shader itself.

2) remove the screenshot matrix and code to generate it. You can instead
create a demo video similar and add it to the demo videos in the custom
shaders doc page. A video is both more compelling and takes up less
space. If you are on linux you can use the existing script to generate
the video at custom/demo.py otherwise you will have to shoot it
manually. You can upload it here and I will take care of linking it up.

